PR #3, opened 2019-02-26
Add integration tests task for CASSANDRA-15031

clohfink:


.circleci/config.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  build:
+    docker:
+      # specify the version you desire here
+      - image: circleci/openjdk:8-jdk
+
+    working_directory: ~/repo
+
+    environment:
+      # Customize the JVM maximum heap limit
```
dineshjoshi: Will we need to adjust these in the future? If not, we can remove it.

.circleci/config.yml
```diff
@@ -0,0 +1,40 @@
...
+      - save_cache:
+          paths:
+            - ~/.gradle
+          key: v1-dependencies-{{ checksum "build.gradle" }}
+
+      # make sure it builds with build steps like swagger docs and dist
+      - run: gradle build
+
+      # run tests!
+      - run: gradle check
```
dineshjoshi: Is this going to use the gradle wrapper that is checked in?

build.gradle
```diff
@@ -2,6 +2,8 @@ plugins {
     id 'java'
     id 'application'
     id 'idea'
+    id 'findbugs'
```
dineshjoshi: While you're at it, can you also add the jacoco support?

src/main/java/org/apache/cassandra/sidecar/CQLSession.java
```diff
@@ -0,0 +1,112 @@
...
+import com.google.inject.Singleton;
+
+/**
+ * Represents a connection to Cassandra cluster. Currently supports returning the local connection only as
+ * defined in the Configuration.
+ */
+@Singleton
+public class CQLSession
+{
+    private static final Logger logger = LoggerFactory.getLogger(CassandraSidecarDaemon.class);
```
dineshjoshi: Wrong logger name. This should be `CQLSession.class`

src/main/java/org/apache/cassandra/sidecar/CQLSession.java
```diff
@@ -0,0 +1,112 @@
...
+                                 // tests can create a lot of these Cluster objects, to avoid creating HWTs and
+                                 // event thread pools for each we have the override
+                                 .withNettyOptions(nettyOptions)
+                                 .build();
+                localSession = cluster.connect();
+            }
+        }
+        catch (Exception e)
+        {
+            logger.debug("Failed to reach Cassandra", e);
```
dineshjoshi: Shouldn't this be at error level?

src/main/java/org/apache/cassandra/sidecar/routes/HealthService.java
```diff
@@ -18,55 +18,83 @@
...
             logger.error("Error while performing health check", e);
         }
     }
 
-    synchronized public void stop()
+    private synchronized void maybeRegisterHostListener()
+    {
+        if (!registered)
+        {
+            if (session != null && session.getLocalCql().isPresent())
```
dineshjoshi: Using optional instead of a null check doesn't add any value. See: 
https://stackoverflow.com/a/40801034

build.gradle
```diff
@@ -80,6 +82,7 @@ dependencies {
     testCompile group: 'org.cassandraunit', name: 'cassandra-unit-shaded', version: '3.3.0.2'
     testCompile 'com.datastax.cassandra:cassandra-driver-core:3.6+:tests'
     testCompile 'org.apache.commons:commons-exec:1.3+'
+    testCompile group: 'com.datastax.oss.simulacron', name: 'simulacron-driver-3x', version: '0.8.7'
```
vinaykumarchella: IIRC, `simulacron` does not support JMX interaction that Cassandra provides (specifically MBeans), 
which is imperative as part of sidecar implementation as we end up dealing with it a lot.  I am fine with
`simulacron` for now as we don't have any JMX interaction yet in the sidecar, and we can think about JMX interaction when we get there.

src/main/java/org/apache/cassandra/sidecar/CQLSession.java
clohfink: its actually an expected scenario with a NoHostAvailableException, so ill just break that one out as a debug and an error for other exceptions

build.gradle
clohfink: simulacron is not Cassandra, its just a CQL server thats super light weight and programatic responses. Its also something we can fire up 100 instances with like 20mb of ram. For JMX we will need to use something else yes, but im just testing the CQL connectivity and driver logic

checkstyle.xml
```diff
@@ -0,0 +1,393 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vinaykumarchella: Just curious: Are these configurations handcrafted? If they are, should we consider using any predefined formats? e.g., `google-java-format` etc.,

src/test/java/org/apache/cassandra/sidecar/HealthServiceIntegrationTest.java
```diff
@@ -0,0 +1,259 @@
+/*
```
vinaykumarchella: nit: Should place all integration tests under `src/integration`? or different package inside `test` itself

checkstyle.xml
clohfink: they are hand crafted, cassandra has a unique code style requirement: https://cassandra.apache.org/doc/latest/development/code_style.html

if we could use googles style id rather just run the style auto formatter on git commit so we just didnt have to think about it anymore but since we have to match a kinda 1-off thing this is best option i could find to maintain it.

PR #4, opened 2019-05-10
Virtual tables API endpoints for sidecar for CASSANDRA-15124

clohfink:


api.yaml
```diff
@@ -15,6 +15,130 @@ tags:
     description: Execute, coordinate, or schedule operations
 
 paths:
+  /api/v1/threadpools:
```
dineshjoshi: Is there a way we could auto-generate this spec. In the past I've used Django REST Framework which autogenerated the docs and schemas - https://www.django-rest-framework.org/topics/documenting-your-api/ . Quick search turned up this: https://github.com/outofcoffee/vertx-oas The reason we should consider doing this is, unless we're very diligent about keeping schemas in sync with the actual routes, the spec gets stale and goes out of sync from the APIs. Alternatively, we should maybe have a test that will check whether the schema is in conformance with the APIs. WDYT?

src/main/java/org/apache/cassandra/sidecar/CQLSession.java
```diff
@@ -73,19 +98,66 @@ public synchronized Session getLocalCql()
...
+                                                 .withoutMetrics()
+                                                 // tests can create a lot of these Cluster objects, to avoid creating
+                                                 // HWTs and event thread pools for each we have the override
+                                                 .withNettyOptions(nettyOptions);
+
+                if (useSSL)
+                {
+                    try
+                    {
+                        if (trustStorePath != null && trustStorePassword != null)
```
dineshjoshi: I think this might be required for future tests as well. Perhaps consider adding this into a utility class?

src/main/java/org/apache/cassandra/sidecar/MainModule.java
```diff
@@ -96,6 +97,12 @@ public Router vertxRouter(Vertx vertx, HealthService healthService)
 
         // API paths
         router.route().path("/api/v1/__health").handler(healthService::handleHealth);
+
+        router.route().path("/api/v1/compactions").handler(infoService::handleCompactions);
```
dineshjoshi: Perhaps consider changing the path to `/api/v1/info/compactions`? That way all actions related to the `InfoService` are under the same path.

src/main/java/org/apache/cassandra/sidecar/VirtualTables.java
```diff
@@ -0,0 +1,342 @@
...
+                                new DefaultPropertyMapper()
+                                    .setNamingStrategy(
+                                        new DefaultNamingStrategy(
+                                            NamingConventions.LOWER_CAMEL_CASE, NamingConventions.LOWER_SNAKE_CASE))
+                                    .setPropertyAccessStrategy(PropertyAccessStrategy.FIELDS))
+                            .build();
+    }
+
+    @Nullable
+    MappingManager manger()
```
dineshjoshi: Typo? Did you mean `manager()`?

src/main/java/org/apache/cassandra/sidecar/VirtualTables.java
```diff
@@ -0,0 +1,342 @@
...
+        {
+            return blockedTasks;
+        }
+
+        public long getBlockedTasksAllTime()
+        {
+            return blockedTasksAllTime;
+        }
+
+        public JsonObject json()
```
dineshjoshi: Do we need to customize this?

src/main/java/org/apache/cassandra/sidecar/routes/HealthCheck.java
```diff
@@ -64,7 +64,7 @@ private boolean check()
                 return result;
             }
         }
-        catch (NoHostAvailableException nha)
+        catch (NoHostAvailableException | ConnectionException nha)
         {
             logger.trace("NoHostAvailableException in HealthCheck - Cassandra Down");
```
dineshjoshi: This message needs to change. Could you also add the exception to the logger so it prints the stacktrace in the logs?

src/main/java/org/apache/cassandra/sidecar/routes/HealthCheck.java
```diff
@@ -64,7 +64,7 @@ private boolean check()
                 return result;
             }
         }
-        catch (NoHostAvailableException nha)
+        catch (NoHostAvailableException | ConnectionException nha)
```
dineshjoshi: Nit: Consider renaming this variable to `e` or something.

dineshjoshi (changes requested):
Overall looks good. I just have some minor suggestions.

src/main/java/org/apache/cassandra/sidecar/MainModule.java
```diff
@@ -83,7 +84,7 @@ public HttpServer vertxServer(Vertx vertx, Configuration conf, Router router)
 
     @Provides
     @Singleton
-    public Router vertxRouter(Vertx vertx, HealthService healthService)
+    public Router vertxRouter(Vertx vertx, HealthService healthService, InfoService infoService)
```
tolbertam: Capturing this from jira:

> When start sidecar would be nice to log something indicating how to get to the docs endpoint.

Would still be nice to have this as isn't clear how/where to hit the swagger endpoint and would be a nice UX thing  to log a URL for first time users.

api.yaml
tolbertam: 👍 could be really nice to generate documentation from code using vertex-oas or something based on swagger-core.  One benefit of doing it by hand is you can come up with nice examples like what is done here, although maybe that is something you can express from code as well somehow.

src/main/java/org/apache/cassandra/sidecar/MainModule.java
tolbertam: On another note I wonder if we should get rid of `docs/index.html` and promote swagger up to docs, as it seems more useful than redoc and includes the examples.

tolbertam (commented):
Don't have much to add since my last review.  Had just one comment, but 👍 otherwise.

api.yaml
```diff
@@ -15,6 +15,130 @@ tags:
...
+        '503':
+          description: Resource not available due to version conflict or connectivity to cluster
+          content:
+            application/json:
+              schema:
+                type: object
+                items:
+                  $ref: '#/components/schemas/Error'
+
+  /api/v1/settings:
```
nvharikrishna: "v1" indicates major api version, but mentioning it here leaves the possibility of of having multiple major versions of an endpoint where others don't have. Can we use server object to mention the version like shown here: https://swagger.io/specification/#serverObject ?

PR #6, opened 2020-02-28
CASSANDRA-15610 Upgraded gradle dependencies, some of which are blockers for full jav…

rustyrazorblade:
…a 11 support.

* Upgrading from findbugs (JDK 8 only) to Spotbugs.  FindBugs was abandoned
  years ago and will not be updated.
* Pinned the transitive Guava dependency to version 28-2.re.
* Upgraded the DataStax driver to version 3.8
* Upgraded Gradle to version 6.2.1

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,19 @@
+
+buildscript {
+    dependencies {
+        // findBugs needs a newer version of Guava in the buildscript.
+        // otherwise it throws an exception
+        classpath "com.google.guava:guava:28.2-jre"
+    }
+}
+
 plugins {
     id 'java'
     id 'application'
     id 'idea'
     id 'checkstyle'
     id 'jacoco'
-    id 'findbugs'
+    id "com.github.spotbugs" version "3.0.0"
     id 'org.hidetake.swagger.generator' version '2.16.0'
 }
 
@@ -73,6 +82,7 @@ dependencies {
     compile 'org.slf4j:slf4j-api:1.7.25'
     compile 'ch.qos.logback:logback-core:1.2.3'
     compile 'ch.qos.logback:logback-classic:1.2.3'
+
     compile 'com.datastax.cassandra:cassandra-driver-core:3.6+'
     compile group: 'com.google.inject', name: 'guice', version: '4.2.2'
     compile group: 'org.apache.commons', name: 'commons-configuration2', version: '2.4'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -73,10 +85,15 @@ dependencies {
     compile 'org.slf4j:slf4j-api:1.7.25'
     compile 'ch.qos.logback:logback-core:1.2.3'
     compile 'ch.qos.logback:logback-classic:1.2.3'
-    compile 'com.datastax.cassandra:cassandra-driver-core:3.6+'
+
+    compile 'com.datastax.cassandra:cassandra-driver-core:3.8+'
     compile group: 'com.google.inject', name: 'guice', version: '4.2.2'
     compile group: 'org.apache.commons', name: 'commons-configuration2', version: '2.4'
 
+    compile(group: 'com.google.guava', name: 'guava', version: '28.2-jre') {
```
dineshjoshi: We don't need this. The guava version update in the buildscript is sufficient.

build.gradle
```diff
@@ -73,10 +85,15 @@ dependencies {
     compile 'org.slf4j:slf4j-api:1.7.25'
     compile 'ch.qos.logback:logback-core:1.2.3'
     compile 'ch.qos.logback:logback-classic:1.2.3'
-    compile 'com.datastax.cassandra:cassandra-driver-core:3.6+'
+
+    compile 'com.datastax.cassandra:cassandra-driver-core:3.8+'
```
dineshjoshi: Simulacron has some issues with 3.8 driver. Can we stay with the 3.6? The integration tests failed with 3.8 but downgrading to 3.6 worked.

rustyrazorblade: Yeah, we can roll this back.

build.gradle
rustyrazorblade: We still need it.  Guava isn't JDK 11 compatible till version 27, and the buildscript block only affects the build tooling, not the artifacts.  I've verified this locally.

rustyrazorblade: For reference on the Guava side: https://github.com/google/guava/issues/3272

rustyrazorblade: Here's the buildscan of the dependency tree if I remove the explicit Guava block:

image

when I build it with the forced guava version:

image

dineshjoshi: I understand. I would like to take a conservative approach in upgrading guava. Here's my concern -

```
+--- com.datastax.cassandra:cassandra-driver-core:3.6+ -> 3.6.0
|    +--- io.netty:netty-handler:4.0.56.Final -> 4.1.30.Final (*)
|    +--- com.google.guava:guava:19.0 -> 28.2-jre
|    |    +--- com.google.guava:failureaccess:1.0.1
|    |    +--- com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava
```

This might actually break the driver. So, ideally we should go to a driver that supports Guava 28. Right now v3.8 of the driver seems to be breaking the integration tests so we cannot upgrade that. Let's address this as a follow up ticket?

rustyrazorblade: Fair enough, I'll update my branch.

PR #11, opened 2020-04-21
Jon/cassandrasc 14/packaging

rustyrazorblade:


rustyrazorblade:
Just making sure packaging works, and we're able to install from them.  We could definitely stage them down the line, that's a great idea.

PR #13, opened 2020-04-29
Resteasy swagger integration

dineshjoshi:


build.gradle
```diff
@@ -222,6 +215,15 @@ applicationDistribution.from("LICENSE.txt") {
     into ""
 }
 
+resolve {
```
rustyrazorblade, line 218: Mind adding a comment?  I had to look up what plugin this was for.  I suggest:

`Resolves project openAPI specification and saves the result in JSON`

dineshjoshi: great point. updated.

PR #14, opened 2020-05-13
Jon/23/multiple cassandra versions

rustyrazorblade:
CASSANDRASC-23


.circleci/config.yml
```diff
@@ -100,10 +100,12 @@ jobs:
...
       - install_java:
           version: adoptopenjdk-11-hotspot
-
+      - restore_cache:
+          keys:
+            - tarballs
       - run: sudo update-java-alternatives -s adoptopenjdk-11-hotspot-amd64 && java -version
-
-      - run: ./gradlew -i clean build --stacktrace
+      - run: mkdir ~/tarballs
```
michaelsembwever, line 107: please can this `mkdir` be put inside gradle?

In hope that CI configurations are as thin as possible, so it is as simple as possible to repeat these steps outside of CircleCI, whether that be on locally or in https://ci-cassandra.apache.org/job/cassandra%7Esidecar/ 

(trivial here i know, just hoping that we'll see better practices and examples with a new project than what we have with that main project…)

rustyrazorblade: Good suggestion, I'll move it in.

cassandra-integration-tests/src/test/java/org/apache/cassandra/sidecar/common/StatusTest.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class StatusTest
+{
+    private static final Logger logger = LoggerFactory.getLogger(StatusTest.class);
+
+    @BeforeEach
+    void setupData(CassandraTestContext context)
+    {
+        logger.info("setup {}", context.container);
+
```
dineshjoshi, line 41: Nit: Extra newline.

cassandra-integration-tests/src/test/java/org/apache/cassandra/sidecar/common/testing/CassandraPod.java
```diff
@@ -0,0 +1,360 @@
...
+        Boolean started = false;
+        String response = "";
+
+        for (int i = 0; i < maxTime; i++)
+        {
+            // we sleep in the beginning because the pod will never be ready right away
+            // sometimes K8 seems to hang in CI as well, so this might be enough to let the pod start
+            try
+            {
+                Thread.sleep(1000);
```
dineshjoshi, line 177: Potentially log a debug statement right before sleeping?

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
```diff
@@ -0,0 +1,197 @@
...
+    {
+        this.cqlSession = cqlSession;
+        this.versionProvider = provider;
+        this.refreshRate = refreshRate;
+    }
+
+    public synchronized void start()
+    {
+        logger.info("Starting health check");
+        // TODO: maybe unhardcode
```
dineshjoshi, line 77: Do you want to address this before merging?

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraVersion.java
```diff
@@ -0,0 +1,159 @@
...
+
+    private static final Pattern pattern = Pattern.compile(VERSION_REGEXP);
+    private static final Pattern SNAPSHOT = Pattern.compile("-SNAPSHOT");
+
+    public final int major;
+    public final int minor;
+    public final int patch;
+
+    // not going to use these for equality checks but we might need to know what it is at some point
+    private final String extra;
```
dineshjoshi, line 52: What is extra? I presume strings like alpha, beta, etc? If so 4.0.0-alpha4 and 4.0.0-alpha5 are technically not the same releases.

common/src/test/java/org/apache/cassandra/sidecar/common/CassandraVersionTest.java
```diff
@@ -0,0 +1,114 @@
...
+package org.apache.cassandra.sidecar.common;
+
+import org.junit.jupiter.api.Test;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+
+/**
+ * Tests that CassandraVersion comparisons work correctly
+ */
+public class CassandraVersionTest
```
dineshjoshi, line 28: Need to add negative test cases for `CassandraVersion::create(String v)`. You should also add tests to validate that the result of `CassandraVersion::toString` is parseable by `CassandraVersion::create(String v)`.

scripts/check-kube.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
dineshjoshi, line 1: Empty script?

dineshjoshi (changes requested):
Looks good to me overall. Just minor nits and clarifications.

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
rustyrazorblade: Ah yes, I originally had the time hard coded, removed it, then forgot to remove the comment.

cassandra-integration-tests/src/test/java/org/apache/cassandra/sidecar/common/testing/CassandraPod.java
rustyrazorblade: Will do

dineshjoshi (approved):

PR #17, opened 2020-10-20
Fixing a style issue

tharanga:


yifan-c:
Thank you @tharanga for the patch! It looks like the java file in question no longer exist. The project has added checkstyle and run it as part of build to guard the inconsistent code style. So I believe the issue was fixed. 

That being said, looking forward to your other contributions! 

I am closing the PR as it no longer applies.

PR #20, opened 2021-09-17
Add stream SSTable API to stream SSTable components

sarankk:
We want to create an API for streaming SSTable components using Vertx’s zero copy sendFile API


src/main/java/com/google/common/util/concurrent/SidecarRateLimiter.java
```diff
@@ -0,0 +1,45 @@
+package com.google.common.util.concurrent;
+
+/**
+ * Custom guava Rate Limiter, uses SmoothBursty Ratelimiter
+ */
+public class SidecarRateLimiter
```
yifan-c, line 7: It simple wraps a RateLimiter and delegate all calls to it. The class seems unnecessary. Can you elaborate?

src/main/java/com/google/common/util/concurrent/SidecarRateLimiter.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public void setRate(final double permitsPerSecond)
+    {
+        this.rateLimiter.setRate(permitsPerSecond);
+    }
+
+    public double getRate()
+    {
+        return this.rateLimiter.getRate();
+    }
```
yifan-c: Rate is only set via `Configuration`, the setter provided is not used and unnecessary. 
Same for the getter.

src/main/java/org/apache/cassandra/sidecar/Configuration.java
```diff
@@ -255,10 +291,17 @@ public Builder setSslEnabled(boolean enabled)
...
+            return this;
+        }
+
         public Configuration build()
         {
-            return new Configuration(cassandraHost, cassandraPort, host, port, healthCheckFrequencyMillis, isSslEnabled,
-                                     keyStorePath, keyStorePassword, trustStorePath, trustStorePassword);
+            return new Configuration(cassandraHost, cassandraPort, cassandraDataDirs, host, port,
+                    healthCheckFrequencyMillis, isSslEnabled, keyStorePath, keyStorePassword, trustStorePath,
+                    trustStorePassword, rateLimitStreamRequestsPerSecond);
```
yifan-c:
```suggestion
            return new Configuration(cassandraHost, cassandraPort, cassandraDataDirs, host, port,
                                     healthCheckFrequencyMillis, isSslEnabled, keyStorePath, keyStorePassword, trustStorePath,
                                     trustStorePassword, rateLimitStreamRequestsPerSecond);
```

src/main/java/org/apache/cassandra/sidecar/exceptions/RangeException.java
```diff
@@ -0,0 +1,22 @@
+package org.apache.cassandra.sidecar.exceptions;
+
+/**
+ * Custom exception
+ */
+public class RangeException extends RuntimeException
```
yifan-c, line 6: Constructors like `public RangeException()` and `public RangeException(Throwable cause)` are not used.

src/main/java/org/apache/cassandra/sidecar/models/HttpResponse.java
```diff
@@ -0,0 +1,88 @@
...
+    {
+        response.setStatusCode(HttpResponseStatus.INTERNAL_SERVER_ERROR.code()).setStatusMessage(msg).end();
+    }
+
+    public void sendFile(File file)
+    {
+        response.putHeader(HttpHeaderNames.CONTENT_TYPE, HttpHeaderValues.APPLICATION_OCTET_STREAM)
+                .putHeader(HttpHeaderNames.CONTENT_LENGTH, Long.toString(file.length()))
+                .sendFile(file.getAbsolutePath());
+    }
```
yifan-c, line 71: Those 2 methods are not used.

src/main/java/org/apache/cassandra/sidecar/models/Range.java
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        return new Range(Math.max(start, range.start()), Math.min(end, range.end()));
+    }
+
+    /**
+     * Accepted string formats "start-end", both ends of the range required to be parsed
+     * Sample accepted formats "1-2", "232-2355"
+     */
+    public static Range parseAbsolute(@NotNull String range)
```
yifan-c: How about making `parseAbsolute` a private static function? so that there is no need to check if wrong method was used. 
Providing a single `parse` function makes it easier to use too.

src/main/java/org/apache/cassandra/sidecar/models/Range.java
```diff
@@ -0,0 +1,181 @@
...
+
+import org.apache.cassandra.sidecar.exceptions.RangeException;
+
+/**
+ * Accepted Range formats are start-end, start-, -suffix_length
+ * start-end (start = start index of the range, end = end index of the range, both inclusive)
+ * start- (start = start index of the range, end = end of file)
+ * -suffix-length (Requested length from end of file)
+ */
+public class Range
```
yifan-c, line 18: Parse the string representation of a range seems clumsy. How about defining header fields dedicated for START, END, and SUFFIX-LENGTH? So there is no need to extract fields using regular expression.

src/main/java/org/apache/cassandra/sidecar/utils/CachedFilePathBuilder.java
```diff
@@ -0,0 +1,219 @@
...
+            }
+        }
+    }
+
+    /**
+     * Enum to hold key length values
+     */
+    public enum KeyLength
+    {
+        ONE, TWO, THREE, FOUR
```
yifan-c: nit: it is only used to represent 1, 2, 3, 4... declaring enum types seem overly complicated.

src/main/java/org/apache/cassandra/sidecar/utils/FileStreamer.java
```diff
@@ -0,0 +1,110 @@
...
+ * General handler for serving files
+ */
+@Singleton
+public class FileStreamer
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(FileStreamer.class);
+    private static final ScheduledExecutorService SCHEDULER = Executors.newScheduledThreadPool(1,
+            new ThreadFactoryBuilder().setNameFormat("acquirePermit").setDaemon(true).build());
+    private static final Duration DELAY = Duration.ofSeconds(5);
+    private static final Duration TIMEOUT = Duration.ofSeconds(10);
```
yifan-c: Maybe make them configurable.

src/test/java/org/apache/cassandra/sidecar/AbstractHealthServiceTest.java
```diff
@@ -129,9 +130,9 @@ private WebClientOptions getWebClientOptions()
         return options;
     }
 
-    @DisplayName("Should return HTTP 503 Failure when check=False")
+    @DisplayName("Should return HTTP 200 OK when check=False")
```
yifan-c: Wrong display name. "200 but NOT_OK"
But.. if the changes to the health endpoint is not necessary for the patch. Please just revert.

src/main/java/org/apache/cassandra/sidecar/routes/HealthService.java
```diff
@@ -53,18 +52,13 @@ public HealthService(CassandraAdapterDelegate cassandra)
     }
 
     @Operation(summary = "Health Check for Cassandra's status",
-    description = "Returns HTTP 200 if Cassandra is available, 503 otherwise",
-    responses = {
-    @ApiResponse(responseCode = "200", description = "Cassandra is available"),
-    @ApiResponse(responseCode = "503", description = "Cassandra is not available")
-    })
+    description = "Returns HTTP 200 if Sidecar is available")
```
yifan-c: The changes in this file do not seem to be related with the patch?

yifan-c (changes requested):

src/main/java/org/apache/cassandra/sidecar/models/Range.java
dineshjoshi: The [range header is a HTTP standard](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range). Its best to stick to standards as standard HTTP clients can be used to interact with this API.

src/main/java/com/google/common/util/concurrent/SidecarRateLimiter.java
sarankk: @yifan-c we wanted to use `this.rateLimiter.queryEarliestAvailable(nowMicros);` method to get time to wait, this method is package protected hence using this wrapper class

src/main/java/org/apache/cassandra/sidecar/exceptions/RangeException.java
sarankk: Removed them

src/main/java/org/apache/cassandra/sidecar/models/HttpResponse.java
sarankk: We might have to set `InternalServerError` in other handlers and likewise `sendFile` method might be directly called if range of file to be streamed is not known

src/main/java/org/apache/cassandra/sidecar/routes/HealthService.java
sarankk: @yifan-c We wanted to avoid having sidecar's health response dependent on Cassandra's health status. Since the changes needed were minimal included these along with this patch

src/main/java/org/apache/cassandra/sidecar/utils/CachedFilePathBuilder.java
sarankk: Mainly added to improve readability of this switch statement https://github.com/apache/cassandra-sidecar/blob/0083e0d62dd0a9aa63fa977ffeea6d7609f387c3/src/main/java/org/apache/cassandra/sidecar/utils/CachedFilePathBuilder.java#L76

yifan-c: Numbers are not best for readability. Please declare the enum types that are more descriptive. 
For example, `ONE` represents `JUST_KEYSPACE`, `TWO` means `KEYSPACE_TABLE`, `THREE` is `KEYSPACE_TABLE_SNAPSHOT`, so on so forth.

src/main/java/com/google/common/util/concurrent/SidecarRateLimiter.java
yifan-c: OK. Can you add a java doc to clarify?

yifan-c (approved):

dineshjoshi (approved):

PR #21, opened 2021-09-22
Decouple health response code from cassandra instance's health status

sarankk:


src/main/java/org/apache/cassandra/sidecar/routes/HealthService.java
```diff
@@ -53,18 +52,13 @@ public HealthService(CassandraAdapterDelegate cassandra)
     }
 
     @Operation(summary = "Health Check for Cassandra's status",
-    description = "Returns HTTP 200 if Cassandra is available, 503 otherwise",
-    responses = {
-    @ApiResponse(responseCode = "200", description = "Cassandra is available"),
-    @ApiResponse(responseCode = "503", description = "Cassandra is not available")
-    })
+    description = "Returns HTTP 200 if Sidecar is available")
```
yifan-c, line 46: How about this?
```suggestion
    description = "Returns HTTP 200 if Sidecar is available. The availability status of Cassandra is indicated in the response body.")
```

yifan-c (commented):
just a nit. Thanks for the patch.

yifan-c (approved):

src/main/java/org/apache/cassandra/sidecar/routes/HealthService.java
```diff
@@ -53,18 +52,13 @@ public HealthService(CassandraAdapterDelegate cassandra)
...
+    description = "Returns HTTP 200 if Sidecar is available. The availability status of Cassandra is indicated in the response body.")
     @Produces(MediaType.APPLICATION_JSON)
     @GET
     public Response doGet()
     {
         Boolean up = cassandra.isUp();
-        int status = up ? HttpResponseStatus.OK.code() : HttpResponseStatus.SERVICE_UNAVAILABLE.code();
-        return Response.status(status).entity(Json.encode(ImmutableMap.of("status", up ?
-                                                                                    "OK" : "NOT_OK"))).build();
+        return Response.status(HttpResponseStatus.OK.code()).entity(Json.encode(ImmutableMap.of("status",
```
dineshjoshi: Instead of having one end point returning details of two services, I propose that we have two health check endpoints one for each service. So introduce two health endpoints. One for each service. That makes it easier to integrate it with Load Balancers as they generally rely only on the endpoint's HTTP response code and not the contents of the response. Two endpoints I propose are -

```
sidecar - /api/v1/__health
cassandra - /api/v1/cassandra/__health
```

The convention should be the same - return HTTP 200 OK if the service is up, otherwise HTTP 503.

dineshjoshi (changes requested):

PR #28, opened 2022-04-27
CASSANDRASC-36: Support for ErrorHandler in Sidecar

frankgh:
This commit adds a default `ErrorHandler` to the `Route#failureHandler()`. The default implementation
for Sidecar is the `JsonErrorHandler` implementation which is also provided as part of this commit.

The `JsonErrorHandler` class serializes `HttpException`s and `REQUEST_TIMEOUT`s to a JSON representation
that provides additional information to the response.

Additionally, we allow for custom implementations of the `ErrorHandler` to be injected (via module
overrides). This allows downstream projects to provide custom implementations of the `ErrorHandler` to
fit the specific needs of the project.


src/main/java/org/apache/cassandra/sidecar/utils/JsonErrorHandler.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Handles failures in Sidecar, and provides a detailed JSON response payload for {@link HttpException HttpExceptions}
+ * and {@code REQUEST_TIMEOUT} errors.
+ */
+public class JsonErrorHandler implements ErrorHandler
+{
+    /**
+     * {@inheritDoc}
+     */
```
yifan-c: This seems not to be necessary. Not expecting to generate javaDoc for this project.

src/main/java/org/apache/cassandra/sidecar/utils/JsonErrorHandler.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    public void handle(RoutingContext ctx)
+    {
+        Throwable t = ctx.failure();
+
+        if (t instanceof HttpException)
```
yifan-c: Why not format all failures in json??

src/main/java/org/apache/cassandra/sidecar/utils/JsonErrorHandler.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param statusCode the HTTP status code for the response
+     * @param payload    the JSON payload for the response
+     */
+    private void writeResponse(RoutingContext ctx, int statusCode, JsonObject payload)
+    {
+        HttpServerResponse response = ctx.response();
+        if (!response.ended() && !response.closed())
+        {
+            response.setStatusCode(statusCode)
+                    .putHeader(HttpHeaders.CONTENT_TYPE.toString(), "application/json")
```
yifan-c: The content type of the error response is always json. It can conflict with the `Accept` header provided by the client. See https://www.rfc-editor.org/rfc/rfc7231#section-5.3.2
Meanwhile, it looks like the `io.vertx.ext.web.handler.impl.ErrorHandlerImpl` has the logic to encode the response in json and other formats. For example, 

```java
//  register the out-of-box error handler for routes.
    .failureHandler(ErrorHandler.create(vertx))

//  client request with the Accept header
    .putHeader("Accept", "application/json")

//  response body content
//  {"error":{"code":400,"message":"Bad Request"}}
```

yifan-c (commented):
Mostly questions for the initial review. It seems to me that `JsonErrorHandler` is not necessary.

frankgh:
Discussed offline with @yifan-c and we agreed that `JsonErrorHandler` is not necessary. We can default to the provided `ErrorHandlerImpl`

yifan-c (approved):

src/test/java/org/apache/cassandra/sidecar/LoggerHandlerInjectionTest.java
```diff
@@ -34,55 +36,75 @@
...
         context.awaitCompletion(5, TimeUnit.SECONDS);
     }
 
+    @AfterEach
+    void tearDown() throws InterruptedException
+    {
+        final CountDownLatch closeLatch = new CountDownLatch(1);
+        server.close(res -> closeLatch.countDown());
+        vertx.close();
+        if (closeLatch.await(60, TimeUnit.SECONDS)) logger.info("Close event received before timeout.");
```
sarankk: Minor nit: missing if else brackets

sarankk (approved):
+1 LGTM

src/test/java/org/apache/cassandra/sidecar/LoggerHandlerInjectionTest.java
frankgh: thanks!

src/test/java/org/apache/cassandra/sidecar/LoggerHandlerInjectionTest.java
```diff
@@ -34,55 +36,78 @@
...
+            logger.error("Close event timed out.");
+        }
+    }
+
     @DisplayName("Should log at error level when the request fails with a 500 code")
     @Test
     public void testInjectedLoggerHandlerLogsAtErrorLevel(VertxTestContext testContext)
     {
-        helper("/500-route", testContext, 500, "Internal Server Error");
+        helper("/500-route", testContext, 500,
```
dineshjoshi: Please don't use magic numbers.

dineshjoshi (changes requested):

dineshjoshi (approved):

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/2e233ec579e4d2a23021116027d75e776e7ad9ec

PR #30, opened 2022-06-28
CASSANDRASC-38 Add endpoint to list snapshot files

frankgh:
This commit adds two new endpoints to allow listing snapshot files.

The first endpoint takes a snapshot name as a path parameter, and searches for
all the snapshots matching the provided name. The result lists all the snapshot
files for all matching snapshots. Additionally, secondary index files can be
included in the response by providing the query param
`includeSecondaryIndexFiles=true`.

```
/api/v1/snapshots/:snapshot
```

The second endpoint takes a keyspace, table name, and snapshot name and searches
for a unique snapshot matching the provided snapshot name in the given keyspace
and table name. The results lists the snapshot files matching the given keyspace,
table name, and snapshot name. Similarly to the first endpoint, secondary index
files can be included in the response by providing the query param
`includeSecondaryIndexFiles=true`.

```
/api/v1/keyspace/:keyspace/table/:table/snapshots/:snapshot
```

### DIFF
--- a/src/test/java/org/apache/cassandra/sidecar/snapshots/AbstractSnapshotPathBuilderTest.java
+++ b/src/test/java/org/apache/cassandra/sidecar/snapshots/AbstractSnapshotPathBuilderTest.java
@@ -1,9 +1,28 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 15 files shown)


src/test/java/org/apache/cassandra/sidecar/snapshots/AbstractSnapshotPathBuilderTest.java
```diff
@@ -109,149 +129,16 @@ void setup() throws IOException
...
     {
-        assertThat(dataDir0.delete());
-        assertThat(dataDir1.delete());
+        dataDir0.delete();
+        dataDir1.delete();
     }
 
     abstract SnapshotPathBuilder initialize(Vertx vertx, InstancesConfig instancesConfig);
 
-    @Test
```
frankgh, line 121: these are really unit tests, moved them to `StreamSSTableComponentRequestTest`

PR #32, opened 2022-07-09
CASSANDRAC-39: Allow Cassandra input validation to be configurable

frankgh:
It is sometimes desirable to allow configuring the Cassandra input validation in case
we want to further restrict the existing validations. In this commit, we introduce the
ability to inject the ValidationConfiguration. A default YamlValidationConfiguration is
provided and the sidecar.yaml is updated to reflect the existing default validation.

### DIFF
--- a/src/main/java/org/apache/cassandra/sidecar/MainModule.java
+++ b/src/main/java/org/apache/cassandra/sidecar/MainModule.java
@@ -27,6 +27,7 @@ import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Set;
+import java.util.function.UnaryOperator;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.SidecarRateLimiter;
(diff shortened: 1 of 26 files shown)


src/main/java/org/apache/cassandra/sidecar/MainModule.java
```diff
@@ -166,90 +198,84 @@ public Router vertxRouter(Vertx vertx,
...
+        = yamlConf.subset(CASSANDRA_INPUT_VALIDATION);
+        Set<String> forbiddenKeyspaces
+        = new HashSet<>(validationConfiguration.getCollection(String.class, CASSANDRA_FORBIDDEN_KEYSPACES, null));
+        String allowedPatternForDirectory
+        = validationConfiguration.get(String.class,
+                                                                        CASSANDRA_ALLOWED_CHARS_FOR_DIRECTORY);
+        String allowedPatternForComponentName
+        = validationConfiguration.get(String.class, CASSANDRA_ALLOWED_CHARS_FOR_COMPONENT_NAME);
+        String allowedPatternForRestrictedComponentName
+        = validationConfiguration.get(String.class, CASSANDRA_ALLOWED_CHARS_FOR_RESTRICTED_COMPONENT_NAME);
```
yifan-c: The format could be improved... 
And `new HashSet<>(validationConfiguration.getCollection(String.class, CASSANDRA_FORBIDDEN_KEYSPACES, null)` will throw NPE if the forbidden keyspaces is not defined. The collection returned with default value `null` (not the `null` passed in the line. Then hash set new with `null` throws. 

Proposed change
```suggestion
        var validation = yamlConf.subset(CASSANDRA_INPUT_VALIDATION);
        Set<String> forbiddenKeyspaces = new HashSet<>(validation.getList(String.class,
                                                                          CASSANDRA_FORBIDDEN_KEYSPACES,
                                                                          Collections.emptyList()));
        UnaryOperator<String> readString = key -> validation.get(String.class, key);
        String allowedPatternForDirectory = readString.apply(CASSANDRA_ALLOWED_CHARS_FOR_DIRECTORY);
        String allowedPatternForComponentName = readString.apply(CASSANDRA_ALLOWED_CHARS_FOR_COMPONENT_NAME);
        String allowedPatternForRestrictedComponentName = readString.apply(
            CASSANDRA_ALLOWED_CHARS_FOR_RESTRICTED_COMPONENT_NAME);
```

frankgh: Thanks for the feedback, `var` is not available in Java 8. I have pushed a new commit

yifan-c (approved):

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/a7a2c29e990acb2363eb7a15cc4b970ebdc04753

PR #33, opened 2022-07-13
CASSANDRASC-40 Fix search in list snapshot endpoint

frankgh:
This commit fixes test setup in SnapshotUtils. Because of the incorrect test setup
the execution is providing incorrect results. For example, assume the following path

/cassandra-test/data/ks/tbl/snapshots/test-snapshot

The test was configuring data directories as ["/cassandra-test/data"], but in a real
execution data directories is provided as ["/cassandra-test"]. This is causing the
endpoint to return incorrect values in the JSON payload.

Additionally, the response was providing the port for Cassandra and not the Sidecar
port.

### DIFF
--- a/src/main/java/org/apache/cassandra/sidecar/routes/ListSnapshotFilesHandler.java
+++ b/src/main/java/org/apache/cassandra/sidecar/routes/ListSnapshotFilesHandler.java
@@ -35,8 +35,7 @@ import io.vertx.core.http.HttpServerRequest;
 import io.vertx.core.net.SocketAddress;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.handler.HttpException;
-import org.apache.cassandra.sidecar.cluster.InstancesConfig;
-import org.apache.cassandra.sidecar.cluster.instance.InstanceMetadata;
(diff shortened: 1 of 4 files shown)


src/main/java/org/apache/cassandra/sidecar/routes/ListSnapshotFilesHandler.java
```diff
@@ -130,20 +121,18 @@ private ListSnapshotFilesResponse buildResponse(String host,
...
-        InstanceMetadata instanceMetadata = instancesConfig.instanceFromHost(host);
-        int sidecarPort = instanceMetadata.port();
-        Path dataDirPath = Paths.get(instanceMetadata.dataDirs().get(DATA_DIR_INDEX));
         ListSnapshotFilesResponse response = new ListSnapshotFilesResponse();
         String snapshotName = request.getSnapshotName();
+        int sidecarPort = configuration.getPort();
 
         for (Pair<String, FileProps> file : fileList)
         {
-            Path pathFromDataDir = dataDirPath.relativize(Paths.get(file.getLeft()));
```
frankgh, line 141: we don't rely on relative paths here for the fix. We use the indexes counting from the end of the path to get the values instead.

PR #34, opened 2022-07-13
CASSANDRASC-40 Fix search in list snapshot endpoint

frankgh:
This commit fixes test setup in SnapshotUtils. Because of the incorrect test setup
the execution is providing incorrect results. For example, assume the following path

/cassandra-test/data/ks/tbl/snapshots/test-snapshot

The test was configuring data directories as ["/cassandra-test/data"], but in a real
execution data directories is provided as ["/cassandra-test"]. This is causing the
endpoint to return incorrect values in the JSON payload.

Additionally, the response was providing the port for Cassandra and not the Sidecar
port.

### DIFF
--- a/common/src/main/java/org/apache/cassandra/sidecar/common/data/ListSnapshotFilesResponse.java
+++ b/common/src/main/java/org/apache/cassandra/sidecar/common/data/ListSnapshotFilesResponse.java
@@ -52,14 +52,14 @@ public class ListSnapshotFilesResponse
      */
     public static class FileInfo
     {
-        public final long size;
-        public final String host;
-        public final int port;
-        public final int dataDirIndex;
-        public final String snapshotName;
-        public final String keySpaceName;
-        public final String tableName;
-        public final String fileName;
+        private final long size;
+        private final String host;
+        private final int port;
+        private final int dataDirIndex;
+        private final String snapshotName;
+        private final String keySpaceName;
+        private final String tableName;
+        private final String fileName;
 
         public FileInfo(@JsonProperty("size") long size,
                         @JsonProperty("host") String host,
@@ -83,6 +83,70 @@ public class ListSnapshotFilesResponse
         public String ssTableComponentPath()
         {
             return Paths.get(keySpaceName, tableName, fileName).toString();
+        }
+
+        /**
+         * @return the size of the file
+         */
+        public long getSize()
+        {
+            return size;
+        }
+
+        /**
+         * @return the host where the file is hosted
+         */
+        public String getHost()
+        {
+            return host;
+        }
+
+        /**
+         * @return the port of the service hosting the file
+         */
+        public int getPort()
+        {
+            return port;
+        }
+
+        /**
+         * @return the index to the data directory
+         */
+        public int getDataDirIndex()
+        {
+            return dataDirIndex;
+        }
+
+        /**
+         * @return the name of the snapshot
+         */
+        public String getSnapshotName()
+        {
+            return snapshotName;
+        }
+
+        /**
+         * @return the name of the keyspace this file belongs to
+         */
+        public String getKeySpaceName()
+        {
+            return keySpaceName;
+        }
+
+        /**
+         * @return the name of the table this file belongs to
+         */
+        public String getTableName()
+        {
+            return tableName;
+        }
+
+        /**
+         * @return the name of the file
+         */
+        public String getFileName()
+        {
+            return fileName;
         }
 
         public boolean equals(Object o)
--- a/common/src/main/java/org/apache/cassandra/sidecar/common/data/ListSnapshotFilesResponse.java
+++ b/common/src/main/java/org/apache/cassandra/sidecar/common/data/ListSnapshotFilesResponse.java
@@ -23,11 +23,15 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 
+import com.google.common.annotations.Beta;
+
 import com.fasterxml.jackson.annotation.JsonProperty;
 
 /**
- * A class representing a response for the {@link ListSnapshotFilesRequest}
+ * A class representing a response for the {@link ListSnapshotFilesRequest}.
+ * This class is expected to evolve and has been mark with the {@link Beta} annotation.
  */
+@Beta
 public class ListSnapshotFilesResponse
 {
     private final List<FileInfo> snapshotFilesInfo;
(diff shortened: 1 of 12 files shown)


src/main/java/org/apache/cassandra/sidecar/routes/ListSnapshotFilesHandler.java
```diff
@@ -130,20 +121,18 @@ private ListSnapshotFilesResponse buildResponse(String host,
                                                     ListSnapshotFilesRequest request,
                                                     List<Pair<String, FileProps>> fileList)
     {
-        InstanceMetadata instanceMetadata = instancesConfig.instanceFromHost(host);
-        int sidecarPort = instanceMetadata.port();
-        Path dataDirPath = Paths.get(instanceMetadata.dataDirs().get(DATA_DIR_INDEX));
         ListSnapshotFilesResponse response = new ListSnapshotFilesResponse();
         String snapshotName = request.getSnapshotName();
+        int sidecarPort = configuration.getPort();
```
yifan-c, line 127: 👍

src/main/java/org/apache/cassandra/sidecar/routes/ListSnapshotFilesHandler.java
```diff
@@ -130,20 +121,18 @@ private ListSnapshotFilesResponse buildResponse(String host,
...
         for (Pair<String, FileProps> file : fileList)
         {
-            Path pathFromDataDir = dataDirPath.relativize(Paths.get(file.getLeft()));
+            Path path = Paths.get(file.getLeft());
 
             String keyspace = request.getKeyspace();
             // table name might include a dash (-) with the table UUID so we always use it as part of the response
-            String tableName = pathFromDataDir.getName(TABLE_NAME_SUBPATH_INDEX).toString();
-            String fileName = pathFromDataDir.getName(FILE_NAME_SUBPATH_INDEX).toString();
+            String tableName = path.getName(path.getNameCount() - TABLE_NAME_SUBPATH_INDEX_FROM_END).toString();
```
yifan-c: Counting from the end... does it work if includeSecondaryIndexFiles? It would add another level

frankgh: good point. This will not work with secondary index files. I will address

src/main/java/org/apache/cassandra/sidecar/routes/ListSnapshotFilesHandler.java
```diff
@@ -146,6 +160,12 @@ private ListSnapshotFilesResponse buildResponse(String host,
         return response;
     }
 
+    private boolean isSecondaryIndexFile(Path path)
+    {
+        return path.getNameCount() >= 4 &&
+               path.getName(path.getNameCount() - 4).endsWith(SNAPSHOTS_DIR_NAME);
```
yifan-c: I think it can still fail in some cases. 

For example, `/cassandra/snapshot/ks/tbl/sstable-files`. The data directory is `/cassandra/snapshot`. 

Since there is no scheme for the directory names, finding the root and counting from root can be more reliable.

frankgh: I thought about that, but it's unlikely that this will occur since the table appends the UUID to the directory name. In any case, I am thinking we can also validate for the `keyspace`, by changing the signature to this:
```java
    private boolean isSecondaryIndexFile(Path path, String keyspace)
    {
        return path.getNameCount() >= 6 &&
               path.getName(path.getNameCount() - 6).endsWith(keyspace) &&
               path.getName(path.getNameCount() - 4).endsWith(SNAPSHOTS_DIR_NAME);
    }
```
What do you think?

frankgh: discussed offline and we decided it's best to have `SnapshotPathBuilder#listSnapshotDirectory` return a List of `SnapshotFile` objects instead of returning a list of `Pair` objects.

common/src/main/java/org/apache/cassandra/sidecar/common/data/ListSnapshotFilesResponse.java
```diff
@@ -52,14 +52,14 @@ public List<FileInfo> getSnapshotFilesInfo()
...
-        public final String tableName;
-        public final String fileName;
+        private final long size;
+        private final String host;
+        private final int port;
+        private final int dataDirIndex;
+        private final String snapshotName;
+        private final String keySpaceName;
+        private final String tableName;
+        private final String fileName;
```
yifan-c: The change in this file does not seem to be necessary. Also, according to https://cassandra.apache.org/_/development/code_style.html#boilerplate, we prefer `public final` over getters.

yifan-c:
Regarding the refactor, it still has the path extraction code, which I think is a dup and can be avoided. Please take a look at the alternative https://github.com/yifan-c/cassandra-sidecar/commit/9ce5c6dbc2522fe4f43f07efbbb141b57e19a9cc

frankgh:
> Regarding the refactor, it still has the path extraction code, which I think is a dup and can be avoided. Please take a look at the alternative [yifan-c@9ce5c6d](https://github.com/yifan-c/cassandra-sidecar/commit/9ce5c6dbc2522fe4f43f07efbbb141b57e19a9cc)

Thanks for that, I spent some time exploring that implementation, but I ended up with more complexity in the code. In the end, I decided to step back for a minute and explore other options. I decided against the proposed approach because:
- the code becomes more complex
- some existing classes that depend on the existing API would need to be refactored
- the approach above was doing something very specific for the `listSnapshotDirectory` API

In the end, I realized we have the `snapshotDirectory` available to us in the `ListSnapshotFilesHandler`. That directory contains all the information required for us to build the response.

I ended up with a more simplified code, and moved the logic to the `SnapshotDirectory` class.

Please let me know what are your thoughts on this approach. I think it's actually looking much cleaner now, and it fixes a few assumptions as well, such as the data directory index in the response.

I have also marked the `ListSnapshotFilesResponse` class as `@Beta`. I suspect that we do not need that many elements in the response payload. The response is used in conjunction with the `StreamSSTableComponentHandler`, the handler does not need that information, so I think we should reconsider what fields are required in the response payload. But maybe let's address that as a follow up?

yifan-c (approved):

frankgh:
closed via 07deb07d2e57a32c67643385403980a9a3c3fc95

PR #35, opened 2022-08-05
CASSANDRASC-41: Refactor range and add toString

yifan-c:


src/main/java/org/apache/cassandra/sidecar/models/Range.java
```diff
@@ -31,132 +31,116 @@
...
-    public long end()
-    {
-        return this.end;
+        if (left == -1L && right == -1L) // matching "bytes=-"
+        {
+            throw invalidRangeHeaderException(rangeHeader);
+        }
+        else if (left == -1L) // matching "bytes=-1"
+        {
+            long len = Math.min(right, fileSize); // correct the range if it exceeds file size.
```
yifan-c, line 86: This is a change of behavior. I feel handling an over-sized range gracefully helps to reduce the number of request/retry. Client will need to (and _should_) read data according to content-length header.

src/main/java/org/apache/cassandra/sidecar/models/Range.java
```diff
@@ -31,132 +31,133 @@
...
+                                            "Supported Range formats are bytes=<start>-<end>, bytes=<start>-, bytes=-<suffix-length>");
     }
 
-    public boolean isValidHttpRange()
+    // An initialized range is always valid; invalid params fail range initialization.
+    private Range(final long start, final long end)
     {
-        return start >= 0 && end >= start && length > 0;
+        Preconditions.checkArgument(start >= 0, "Range start can not be negative");
+        Preconditions.checkArgument(end >= start, "Range does not satisfy boundary requirements");
```
frankgh, line 126: these two need to be of type `RangeException` instead of `IllegalArgumentException`

src/main/java/org/apache/cassandra/sidecar/models/Range.java
```diff
@@ -31,132 +31,133 @@
...
+    public static Range parseHeader(final String header, final long fileSize)
     {
-        this.start = start;
-        this.end = end;
-        this.length = length(start, end);
+        if (header == null)
+        {
+            return new Range(0, fileSize - 1);
+        }
+        if (!header.startsWith(RANGE_UNIT + "="))
```
frankgh: is this check required? It seems redundant since we will do the regex check in `Range#parse`

frankgh (commented):
discussed offline about adding some unit tests

src/main/java/org/apache/cassandra/sidecar/models/Range.java
frankgh: disregard this one

frankgh (approved):
+1 Looks good to me!

src/main/java/org/apache/cassandra/sidecar/models/Range.java
yifan-c: No. I will remove it.

PR #36, opened 2022-08-30
CASSANDRASC-42: Keyspaces API endpoint

frankgh:


src/main/java/org/apache/cassandra/sidecar/MainModule.java
```diff
@@ -194,6 +197,19 @@ public Router vertxRouter(Vertx vertx,
...
+        router.get(API_V1_VERSION + keyspacesRoute)
+              .handler(keyspacesHandler);
+
+        // handles:
+        // - /api/v1/keyspace/:keyspace and
+        // - /api/v1/keyspace/:keyspace/table/:table
+        router.route()
+              .method(HttpMethod.HEAD)
+              .method(HttpMethod.GET)
+              .pathRegex("\\/api\\/v1\\/keyspace\\/(?<keyspace>[^\\/]+)(\\/table\\/(?<table>[^\\/]+))?")
```
frankgh: I'm not too happy about having the regex. Maybe we should just use two strings instead:
```suggestion
              .path(API_V1_VERSION + "/keyspace/:keyspace");

        router.route()
              .method(HttpMethod.HEAD)
              .method(HttpMethod.GET)
              .path(API_V1_VERSION + "/keyspace/:keyspace/table/:table");
```

src/main/java/org/apache/cassandra/sidecar/models/Range.java
```diff
@@ -116,7 +116,8 @@ private static long parseLong(String valStr, String rangeHeader)
     private static IllegalArgumentException invalidRangeHeaderException(String rangeHeader)
     {
         return new IllegalArgumentException("Invalid range header: " + rangeHeader + ". " +
-                                            "Supported Range formats are bytes=<start>-<end>, bytes=<start>-, bytes=-<suffix-length>");
+                                            "Supported Range formats are bytes=<start>-<end>, " +
```
frankgh, line 119: `./gradlew build` was not happy about long lines

src/test/java/org/apache/cassandra/sidecar/RangeTest.java
```diff
@@ -71,7 +71,8 @@ public void testInvalidRangeFormat()
         {
             Range.parseHeader(rangeHeader, Long.MAX_VALUE);
         });
-        String msg = "Invalid range header: bytes=2344--3432. Supported Range formats are bytes=<start>-<end>, bytes=<start>-, bytes=-<suffix-length>";
+        String msg = "Invalid range header: bytes=2344--3432. Supported Range formats are bytes=<start>-<end>, " +
```
frankgh, line 74: unrelated changes in this file, but `./gradlew build was complaining`

src/test/java/org/apache/cassandra/sidecar/routes/StreamSSTableComponentHandlerTest.java
```diff
@@ -315,9 +316,10 @@ void testSuffixRangeExceeds(VertxTestContext context)
                 .send(context.succeeding(response -> context.verify(() ->
                 {
                     assertThat(response.statusCode()).isEqualTo(OK.code());
+                    AbstractStringAssert<?> as =
                     assertThat(response.getHeader(HttpHeaderNames.CONTENT_LENGTH.toString()))
-                        .isEqualTo("4")
-                        .describedAs("Server should shrink the range to the file length");
+                        .describedAs("Server should shrink the range to the file length")
```
frankgh, line 280: `describeAs` is annotated with `org.assertj.core.util.CheckReturnValue` which is documented as `Findbugs handles any annotation with name "CheckReturnValue" in return value check.` . This was failing the build with warning `RV_RETURN_VALUE_IGNORED: Method ignores return value`

src/main/java/org/apache/cassandra/sidecar/MainModule.java
yifan-c: +1 to the brain relief.

src/main/java/org/apache/cassandra/sidecar/routes/KeyspacesHandler.java
```diff
@@ -0,0 +1,227 @@
...
+    {
+        KeyspaceRequest requestParams = extractParamsOrThrow(context);
+        String host = getHost(context);
+        SocketAddress remoteAddress = context.request().remoteAddress();
+        InstanceMetadata instanceMeta = instancesConfig.instanceFromHost(host);
+        LOGGER.debug("KeyspacesHandler received request: {} from: {}. Instance: {}",
+                     requestParams, remoteAddress, host);
+
+        getMetadata(instanceMeta)
+        .onFailure(throwable ->
```
yifan-c: nit: indent the chained invocations.

src/main/java/org/apache/cassandra/sidecar/routes/KeyspacesHandler.java
```diff
@@ -0,0 +1,227 @@
...
+        {
```
yifan-c, line 110: for lambda, we can have the `{` right after `->`, rather than the style in the `onSuccess` block, which waste the heading spaces for each line.

src/main/java/org/apache/cassandra/sidecar/routes/KeyspacesHandler.java
```diff
@@ -0,0 +1,227 @@
...
+                                           "Unable to reach Cassandra service",
+                                           throwable));
+        })
+        .onSuccess(metadata ->
+                   {
+                       KeyspaceMetadata ksMetadata;
+                       if (metadata == null)
+                       {
+                           LOGGER.error("Failed to obtain keyspace metadata for request '{}'", requestParams);
+                           context.fail(HttpResponseStatus.INTERNAL_SERVER_ERROR.code());
```
yifan-c: nit: return early and move `ksMetadata` declaration after the check.

src/main/java/org/apache/cassandra/sidecar/routes/KeyspacesHandler.java
```diff
@@ -0,0 +1,227 @@
...
+
+        getMetadata(instanceMeta)
+        .onFailure(throwable ->
+        {
+            LOGGER.error("Failed to obtain keyspace metadata for request '{}'", requestParams, throwable);
+            context.fail(new HttpException(HttpResponseStatus.SERVICE_UNAVAILABLE.code(),
+                                           "Unable to reach Cassandra service",
+                                           throwable));
+        })
+        .onSuccess(metadata ->
```
yifan-c: How about move it into a method? For a lambda, it is quite long.

src/main/java/org/apache/cassandra/sidecar/routes/KeyspacesHandler.java
```diff
@@ -0,0 +1,227 @@
...
+                                                                   requestParams.getKeyspace());
+                               context.fail(new HttpException(HttpResponseStatus.NOT_FOUND.code(), errorMessage));
+                           }
+                           else
+                           {
+                               // keyspace / table request
+                               buildTableResponse(context, tableMetadata);
+                           }
+                       }
+                   });
```
yifan-c: A bit more verbose alternative is the following. It separate the value assignment apart from conditional statement to make it easier to read. 

```java
void handleWithInstanceMetadata(context, metadata)
{
  if (metadata == null)
  {
    // set request as failed and return
  }

  if (has no keyspace param)
  {
    // list keyspaces and return
  }

  // retrive keyspace metadata

  if (keyspace metadata does not exist)
  {
    // set request as failed and return
  }
  
  if (has no table param)
  {
    // get keyspace and return
  }

  // retrieve table metadata

  if (table metadata does not exist)
  {
    // fail and return
  }

  // get table and return
}
```

src/main/java/org/apache/cassandra/sidecar/routes/KeyspacesHandler.java
```diff
@@ -0,0 +1,227 @@
...
+                       KeyspaceMetadata ksMetadata;
+                       if (metadata == null)
+                       {
+                           LOGGER.error("Failed to obtain keyspace metadata for request '{}'", requestParams);
+                           context.fail(HttpResponseStatus.INTERNAL_SERVER_ERROR.code());
+                       }
+                       else if (requestParams.getKeyspace() == null)
+                       {
+                           // keyspaces request
+                           buildKeyspacesResponse(context, metadata);
```
yifan-c: let's call it `listKeyspaces`

src/main/java/org/apache/cassandra/sidecar/routes/KeyspacesHandler.java
```diff
@@ -0,0 +1,227 @@
...
+                       {
+                           // keyspace does not exist
+                           String errorMessage = String.format("Keyspace '%s' does not exist.",
+                                                               requestParams.getKeyspace());
+                           context.fail(new HttpException(HttpResponseStatus.NOT_FOUND.code(), errorMessage));
+                       }
+                       else if (requestParams.getTableName() == null)
+                       {
+                           // keyspace request
+                           buildKeyspaceResponse(context, ksMetadata);
```
yifan-c: let's call it `getKeyspace` or `getTables`. Both have the same meaning.

So that it is much easier to distinguish between `buildKeyspacesResponse` and `buildKeyspaceResponse`

src/main/java/org/apache/cassandra/sidecar/routes/KeyspacesHandler.java
```diff
@@ -0,0 +1,227 @@
...
+                           {
+                               String errorMessage = String.format("Table '%s' does not exist in the '%s' keyspace",
+                                                                   requestParams.getTableName(),
+                                                                   requestParams.getKeyspace());
+                               context.fail(new HttpException(HttpResponseStatus.NOT_FOUND.code(), errorMessage));
+                           }
+                           else
+                           {
+                               // keyspace / table request
+                               buildTableResponse(context, tableMetadata);
```
yifan-c: and.. `getTable` to be consistent.

yifan-c (commented):
1st pass. I have not looked at the tests yet.

src/main/java/org/apache/cassandra/sidecar/routes/KeyspacesHandler.java
frankgh: +1. I'm updating `checkstyle.xml` to avoid linter errors

frankgh:
Closing in favor of Schema API. See https://issues.apache.org/jira/browse/CASSANDRASC-42 for additional details

PR #38, opened 2022-10-10
CASSANDRASC-44 Refactor health check to use vertx timer

frankgh:
Vertx API offers a periodic timer that integrates with it's internal thead pooling mechanism. In this commit, we utilize vertx's periodic timer in favor of using a `Executors.newSingleThreadScheduledExecutor()` on each delegate.

Another benefit of this approach is that if the cluster topology changes, i.e. node replacement, cluster expansion / shrink, then the health checks will be performed against the actual nodes in the cluster, assuming we receive an updated view of the cluster when invoking the `Configuration#getInstancesConfig()#instances()` method. We no longer need to worry about decommissioning the single thread executors running on each delegate.

### DIFF
--- a/common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
+++ b/common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
@@ -47,7 +47,7 @@ public class CassandraAdapterDelegate implements ICassandraAdapter, Host.StateLi
 {
     private final CQLSession cqlSession;
     private final CassandraVersionProvider versionProvider;
-    private Session session;
+    private volatile Session session;
     private SimpleCassandraVersion currentVersion;
     private ICassandraAdapter adapter;
     private volatile boolean isUp = false;
--- a/common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
+++ b/common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
@@ -19,10 +19,6 @@
 package org.apache.cassandra.sidecar.common;
 
 import java.util.List;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
```diff
@@ -127,7 +95,7 @@ public synchronized void checkSession()
      * Should be called on initial connect as well as when a server comes back since it might be from an upgrade
      * synchronized so we don't flood the DB with version requests
      *
-     * If the healthcheck determines we've changed versions, it should load the proper adapter
+     * <p>If the healthcheck determines we've changed versions, it should load the proper adapter
```
yifan-c: nit: either remove the `<p>` tag or close it with `</p>`. I do not see the advantage of adding the html tag, so lean to removing.

src/main/java/org/apache/cassandra/sidecar/CassandraSidecarDaemon.java
```diff
@@ -97,6 +101,20 @@ private void validate()
...
+     * The health check is executed in a blocking thread to prevent the event-loop threads from blocking.
+     *
+     * @param timerId the ID of the periodic timer
+     */
+    private void healthCheck(Long timerId)
+    {
+        config.getInstancesConfig()
+              .instances()
+              .forEach(instanceMetadata ->
+                       vertx.executeBlocking(promise -> instanceMetadata.delegate().healthCheck()));
```
yifan-c, line 115: If my understanding is correct. 
In this setup, if the prior check takes a longer time that is greater than the interval, there could be 2 threads that runs `healthCheck` concurrently. Due to the check is `synchronized`, there will be one thread being blocked. 
In the prior setup, no additional thread is to be blocked. 

I do not mean that the prior one is better.

One step back, the health check is something we want to perform every X seconds. If a prolonged check just finished, the next one should not be immediately performed, which is the problem for both setups. 

Can we add a `shouldCheck(timeInMillis)` method and skip a check if it has just checked very recently?

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
frankgh: this is a javadoc standard I believe. You need to open new paragraphs with the paragraph (`<p>`) tag. I will close the tag

src/main/java/org/apache/cassandra/sidecar/CassandraSidecarDaemon.java
frankgh: discussed offline, and we decided the approach in the PR is sufficient. We explored another alternative where we schedule the next health check after all of the checks complete. And another alternative where each delegate will ensure that the check was performed at least x milliseconds after the last completed run

dineshjoshi (approved):
LGTM +1

PR #39, opened 2022-10-11
CASSANDRASC-45: Delegate methods to the RateLimiter

frankgh:
Sidecar offers a `SidecarRateLimiter` class that internally uses the `com.google.common.util.concurrent.RateLimiter`. In this commit, we  expose public methods of the `RateLimiter` class using the delegate pattern. These methods will allow us to tweak the settings of the `RateLimiter` that are available to us

### DIFF
--- a/src/main/java/com/google/common/util/concurrent/SidecarRateLimiter.java
+++ b/src/main/java/com/google/common/util/concurrent/SidecarRateLimiter.java
@@ -18,6 +18,10 @@
 
 package com.google.common.util.concurrent;
 
+import java.util.concurrent.TimeUnit;
+
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+
 /**
  * Wrapper class over guava Rate Limiter, uses SmoothBursty Ratelimiter. This class mainly exists to expose
  * package protected method queryEarliestAvailable of guava RateLimiter
@@ -36,6 +40,8 @@ public class SidecarRateLimiter
         return new SidecarRateLimiter(permitsPerSecond);
     }
 
+    // Delegated methods
+
     /**
      * Returns earliest time permits will become available
      */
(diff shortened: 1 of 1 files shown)


yifan-c (approved):
LGTM. 

Wondering if there is a non-blocking rate limiter library alternative?

src/main/java/com/google/common/util/concurrent/SidecarRateLimiter.java
```diff
@@ -51,4 +57,73 @@ public boolean tryAcquire()
...
     }
+
+    /**
+     * Updates the stable rate of the internal {@code RateLimiter}, that is, the {@code permitsPerSecond}
+     * argument provided in the factory method that constructed the {@code RateLimiter}.
+     *
+     * @param permitsPerSecond the new stable rate of this {@code RateLimiter}
+     * @throws IllegalArgumentException if {@code permitsPerSecond} is negative or zero
+     */
+    public void setRate(double permitsPerSecond)
```
dineshjoshi: `setRate(double permitsPerSecond) -> rate(double permitsPerSecond)`

src/main/java/com/google/common/util/concurrent/SidecarRateLimiter.java
```diff
@@ -51,4 +57,73 @@ public boolean tryAcquire()
...
+    {
+        this.rateLimiter.setRate(permitsPerSecond);
+    }
+
+    public void doSetRate(double permitsPerSecond, long nowMicros)
+    {
+        rateLimiter.doSetRate(permitsPerSecond, nowMicros);
+    }
+
+    public double getRate()
```
dineshjoshi: `double getRate() -> double rate()`

dineshjoshi (changes requested):
minor changes to naming

frankgh:
> Wondering if there is a non-blocking rate limiter library alternative?

No idea, but it'd be a good idea to look into that. I will take this up as some research

dineshjoshi (approved):

PR #40, opened 2022-10-14
CASSANDRASC-46: Migrate minikube to testcontainers for integration tests

frankgh:
The existing Cassandra Sidecar integration testing suite uses Minikube to provision a Cassandra service and it performs tests against the running database. This database runs on Minikube.

A new alternative is to use [testcontainers](https://www.testcontainers.org), which requires Docker to run tests. The concept is similar, but the benefit is that *testcontainers* is well integrated into the java ecosystem and it works well with junit5. By replacing Minikube with `testcontainers` we can simplify the setup process and reduce the complexity for running integration tests.

Additionally, `testcontainers` is supported as part of the testing infrastructure inside [CircleCI](https://www.testcontainers.org/supported_docker_environment/continuous_integration/circle_ci/). Currently, our Minikube tests are broken both locally and in CircleCI. Moving to `testcontainers` would also unblock us for further development.


cassandra-integration-tests/src/test/java/org/apache/cassandra/sidecar/common/testing/TestVersionSupplier.java
```diff
@@ -28,17 +28,17 @@
...
- * Ideally, we'd probably have concurrent runs of the test infrastructure each running tests against one specific
+ * <p>Ideally, we'd probably have concurrent runs of the test infrastructure each running tests against one specific
  * version of C*, but we don't need that yet given we only have one version.
  */
 public class TestVersionSupplier
 {
     Stream<TestVersion> getTestVersions()
     {
-        return Stream.of(new TestVersion("4.0.0", new Cassandra40Factory(), "cassandra40"));
+        return Stream.of(new TestVersion("4.0.0", new Cassandra40Factory(), "cassandra:4.0.6"));
```
yifan-c: What about reading the image tag from env? The tags can become unavailable when new images are released. For example, 4.0.5 no longer exist.

dineshjoshi:
LGTM +1

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/05c0bbe29f75d678596af09abb0c68d15e93f7ba

PR #41, opened 2022-11-02
CASSANDRASC-47: Introduce JMX foundation in Sidecar

frankgh:
In this commit, we introduce the JMX foundation in Sidecar to enable the ability to communicate with the Cassandra process. This commit adds new configuration parameters to configure the a `org.apache.cassandra.sidecar.common.JmxClient`. This client is available as part of the Cassandra delegate.

A new interface is introduced and exposed through the `org.apache.cassandra.sidecar.common.ICassandraAdapter`. The new interface `StorageOperations` is intended to interface with the Cassandra StorageService. This commit provides an example implementation of the `takeSnapshot` method, which is also found in the Cassandra code base. This should allow us to interact with the Cassandra process to create snapshots.

A fix is required in the `CassandraSidecarDaemon` class, where the `healthCheck` runs the first time only after the configured health check frequency (millis) has passed. This causes issues in the unit tests as well as the actual execution of the service, as the `adapter` will be `null` until the first health check is performed. To fix this issue, we perform a health check right after the server successfully starts up.

Additional integration tests are added for testing the JMX integration with Cassandra. In the test, we spin up a new Cassandra container using `testcontainers` and we perform validation against the StorageService in Cassandra.

Co-authored-by: Doug Rohrer <drohrer@apple.com>


cassandra-integration-tests/src/testFixtures/java/org/apache/cassandra/sidecar/common/containers/CassandraContainer.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Constructs a new {@link CassandraContainer} with the {@link #JMX_PORT JMX port} exposed
+     *
+     * @param dockerImageName the name of the docker image to use for the container
+     */
+    public CassandraContainer(String dockerImageName)
+    {
+        super(dockerImageName);
+//        addExposedPort(JMX_PORT);
```
yifan-c: nit: remove the commented line if it no longer needed.

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/Cassandra40StorageOperations.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.apache.cassandra.sidecar.common.JmxClient;
+import org.apache.cassandra.sidecar.common.StorageOperations;
+
+/**
+ * An implementation of the {@link StorageOperations} that interfaces with Cassandra 4.0
+ */
+public class Cassandra40StorageOperations implements StorageOperations
+{
+    private static final String SS_OBJ_NAME = "org.apache.cassandra.db:type=StorageService";
```
yifan-c: nit: no need to abbreviate, use `STORAGE_SERVICE_OBJ_NAME`

common/src/main/java/org/apache/cassandra/sidecar/common/ICassandraAdapter.java
```diff
@@ -28,6 +28,14 @@
  */
 public interface ICassandraAdapter
 {
+    /**
+     * @return a list of {@link NodeStatus} for the Cassandra cluster
+     */
```
yifan-c: TBH, I do not know why the return type is a list. The `ICassandraAdapter` is to interact with the local Cassandra node that pairs with the sidecar. Why returning the list of node status? Sidecar does not support bulk operation yet, and it is not even implemented. 
Let not add the comment for the method. The comment might be no accurate. We might want to refactor the method later since it does not make too much sense.

common/src/main/java/org/apache/cassandra/sidecar/common/JmxClient.java
```diff
@@ -0,0 +1,241 @@
...
+     *
+     * @param clientClass the management interface that the MBean exports, which will
+     *                    also be implemented by the returned proxy
+     * @param remoteName  the name of the MBean within {@code connection} to forward to
+     * @param func        the remote function that will be executed through the proxy
+     * @param <TClient>   the type of the proxy client
+     * @param <TResult>   the type of the result returned by the proxy
+     * @return the result of the remote function after being executed via the proxy
+     */
+    public <TClient, TResult> TResult apply(Class<TClient> clientClass,
```
yifan-c: Generic types named in the form that looks like class names are misleading and confusing. Can you just do `C` for client and `R` for result?

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
```diff
@@ -174,6 +176,13 @@ public List<NodeStatus> getStatus()
         return adapter.getStatus();
     }
 
+    @Override
+    public StorageOperations storageOperations()
+    {
+        checkSession();
```
yifan-c: nit: remove the check. JMX operation does not require session to be established, although there is no harm to check here.

common/src/main/java/org/apache/cassandra/sidecar/common/StorageOperations.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package org.apache.cassandra.sidecar.common;
+
+import java.util.Map;
+
+/**
+ * An interface that defines interactions with the storage system in Cassandra
+ */
+public interface StorageOperations
```
yifan-c, line 29: nit: should it be called `JmxOperations`, which interacts with different components in cassandra (not limited to storage) via JMX, or `StorageJmxOperations` if want to limit for storage service.

yifan-c (commented):
glanced through for the first pass.

common/src/main/java/org/apache/cassandra/sidecar/common/StorageOperations.java
frankgh: I implemented this following the comment in this class: https://github.com/apache/cassandra-sidecar/blob/trunk/common/src/main/java/org/apache/cassandra/sidecar/common/ICassandraAdapter.java#L26 . This makes sense to me as well, it seems we should have some interfaces that expose different functionality. In this case, it so happens that the Storage ops are done via JMX, but this could change at some point. I would prefer to keep this name for the class.

frankgh: Thinking more about this, I think I'm mixing the interface that sidecar exposes to the endpoints, and the interface used for the Storage Service in Cassandra. This specific `StorageOperations` is an interface for the REST layer, and this doesn't have to be a mirror of the JMX interface that Cassandra exposes. So I will go ahead and refactor this code to separate the two interfaces.

frankgh: I pushed code to address this comment. Let me know what you think @yifan-c

yifan-c: Sounds good to me

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/Cassandra40StorageOperations.java
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void takeSnapshot(@NotNull String tag, @NotNull String keyspace, @NotNull String table,
+                             @Nullable Map<String, String> options)
+    {
+        requireNonNull(tag, "snapshot tag must be non-null");
+        requireNonNull(keyspace, "keyspace for the  must be non-null");
+        requireNonNull(table, "table must be non-null");
```
yifan-c, line 57: 👍 to actually enforce the validation for `@NotNull`

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/StorageJmxOperations.java
```diff
@@ -0,0 +1,18 @@
+package org.apache.cassandra.sidecar.cassandra40;
+
+import java.util.Map;
+
+/**
+ * An interface that pulls a method from the Cassandra Storage Service Proxy
+ */
+interface StorageJmxOperations
```
yifan-c: Why it is in `cassandra40`? Do you expect it to be different in a different cassandra release?

yifan-c (approved):
LGTM

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/StorageJmxOperations.java
frankgh: yeah, the interface evolves with versions, so I expect this to change in 41 and most likely the interface evolved and it looks different in 30, 31 and so on

common/src/main/java/org/apache/cassandra/sidecar/common/JmxClient.java
```diff
@@ -0,0 +1,237 @@
...
+    /**
+     * Calls a remote function through the JMX bean proxy
+     *
+     * @param clientClass the management interface that the MBean exports, which will
+     *                    also be implemented by the returned proxy
+     * @param remoteName  the name of the MBean within {@code connection} to forward to
+     * @param caller      the consumer to be called through the proxy
+     * @param <C>         the type of the proxy client
+     */
+    public <C> void call(Class<C> clientClass, String remoteName, Consumer<C> caller)
```
yifan-c: It is a bit counter-intuitive that `call` has no return value. I am thinking of `Callable`.

Similarly, `client.apply` reads not well. Semantically, `apply` reminds me of function evaluation. But `JmxClient` is not a function, instead it is a client to invoke/call a remote procedure.

How about remove both method and make `getProxy` public? This one the chained method call also looks fluent, and it no longer take a lambda, which can grow complex and look messy. For example,
```java
        jmxClient.getProxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME)
                 .takeSnapshot(tag, options, keyspace + "." + table);
```
Remember to call `checkConnection()` in `getProxy`, so it is identical to `apply` and `call`.

frankgh: trying out how this would look like, and I agree that the call looks fluent. Now the question is whether exposing the proxy is okay or not. I think since we expose it to internal classes, it should be fine. The proxy will be consumed directly by the interfacing class between the endpoint and Cassandra, so this would limit the exposure of the proxy.

dineshjoshi (approved):
+1

frankgh:
Merged via 02cc6548f291528e9749a51d103463f9552f4b4e

PR #42, opened 2022-11-10
CASSANDRASC-48: Adding new endpoint that gives information about the release version & partitioner name of a node

jyothsnakonisa:
Adds new APIs

GET /api/v1/cassandra/status
  - Returns the NodeStatus with release version & partitioner info of the node receiving the request GET /api/v1/cassandra/status?instanceId={id}
  - Returs the NodeStatus with release version & partitioner info of a given instanceId
  
  https://app.circleci.com/pipelines/github/jyothsnakonisa/cassandra-sidecar/2/workflows/d24ab1b4-5149-4ae5-80d6-7c5d8c3dc5cb


src/main/java/org/apache/cassandra/sidecar/routes/CassandraHealthService.java
```diff
@@ -68,16 +71,35 @@ public CassandraHealthService(InstanceMetadataFetcher metadataFetcher)
...
+    }
+
+    @Operation(summary = "Cassandra's node status",
+    description = "Returns HTTP 200 if Cassandra is available, 503 otherwise",
+    responses = {
+    @ApiResponse(responseCode = "200", description = "Cassandra status is available"),
+    @ApiResponse(responseCode = "503", description = "Cassandra status is not available")
+    })
+    @GET
+    @Path("/v1/cassandra/status")
```
frankgh: does settings make more sense?
```suggestion
    @Path("/v1/cassandra/settings")
```

src/main/java/org/apache/cassandra/sidecar/routes/CassandraHealthService.java
```diff
@@ -68,16 +71,35 @@ public CassandraHealthService(InstanceMetadataFetcher metadataFetcher)
     public Response getCassandraHealth(@Context HttpServerRequest req,
                                        @QueryParam(AbstractHandler.INSTANCE_ID) Integer instanceId)
     {
-        CassandraAdapterDelegate cassandra;
-        if (instanceId != null)
+        CassandraAdapterDelegate cassandra = getCassandraAdapterDelegate(req.host(), instanceId);
+        return getHealthResponse(cassandra);
+    }
+
+    @Operation(summary = "Cassandra's node status",
```
frankgh: the name of the class `CassandraHealthService` doesn't make sense anymore if we add another endpoint here. We should either rename the class, or add a new class for the new endpoint.

src/main/java/org/apache/cassandra/sidecar/routes/CassandraHealthService.java
```diff
@@ -68,16 +71,35 @@ public CassandraHealthService(InstanceMetadataFetcher metadataFetcher)
...
+    {
+        NodeStatus nodeStatus;
+        try
         {
-            cassandra = metadataFetcher.getDelegate(instanceId);
+            CassandraAdapterDelegate cassandra = getCassandraAdapterDelegate(req.host(), instanceId);
+            nodeStatus = cassandra.getStatus();
         }
-        else
+        catch (IllegalArgumentException e)
```
frankgh, line 74: I don't know if this try/catch block is necessary, I am not able to pin point exactly where the `IllegalArgumentException` would come from

frankgh (commented):
I left a few comments

src/main/java/org/apache/cassandra/sidecar/routes/CassandraHealthService.java
jyothsnakonisa: @frankgh If the instanceId is incorrect then an IllegalArgumentException is being thrown. You can refer the test case `validRequestWithInvalidInstanceId `. Without this, we would throw internal server error to the client, which I think is incorrect.

frankgh: should we instead register an exception mapper to handle `IllegalArgumentException` errors?

```java
public class IllegalArgumentExceptionMapper implements ExceptionMapper<IllegalArgumentException>
{
    @Override
    public Response toResponse(IllegalArgumentException exception)
    {
        return Response.status(Response.Status.BAD_REQUEST)
                       ...
    }
}
```

and then inside `org.apache.cassandra.sidecar.MainModule#configureServices()` we can register it
```
resteasyDeployment.getProviderFactory().registerProvider(IllegalArgumentExceptionMapper.class);
```

yifan-c: Should all `IllegalArgumentException`s be mapped to 400? It might not hold true for all.

common/src/main/java/org/apache/cassandra/sidecar/common/NodeStatus.java
```diff
@@ -19,10 +19,39 @@
...
+
+    /**
+     * Constructs a new {@link NodeStatus} object with the Cassandra node's release version and partitioner
+     * information.
+     *
+     * @param releaseVersion the release version of the Cassandra node
+     * @param partitioner    the partitioner used by the Cassandra node
+     */
+    public NodeStatus(@JsonProperty("releaseVersion") String releaseVersion,
+                      @JsonProperty("partitioner") String partitioner)
```
nvharikrishna: Neither releaseVersion nor partitoner says node is up or down. Why this class is called as NodeStatus?

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+
+package org.apache.cassandra.sidecar.common;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Holds information about the Status of a node
```
frankgh:
```suggestion
 * Holds information about specific node settings
```

src/main/java/org/apache/cassandra/sidecar/routes/CassandraSettingsService.java
```diff
@@ -0,0 +1,66 @@
+package org.apache.cassandra.sidecar.routes;
```
frankgh, line 19: can we add the license header to this file?

src/test/java/org/apache/cassandra/sidecar/routes/CassandraSettingsServiceTest.java
```diff
@@ -0,0 +1,118 @@
+package org.apache.cassandra.sidecar.routes;
```
frankgh, line 18: can we add the license header to this file as well

frankgh (approved):
Looks good to me. I think the only pending on this PR from my point of view is some license headers missing in the new files

src/main/java/org/apache/cassandra/sidecar/routes/CassandraSettingsService.java
jyothsnakonisa: Added the license information.

src/test/java/org/apache/cassandra/sidecar/routes/CassandraSettingsServiceTest.java
jyothsnakonisa: added license information.

common/src/main/java/org/apache/cassandra/sidecar/common/ICassandraAdapter.java
```diff
@@ -28,6 +26,6 @@
  */
 public interface ICassandraAdapter
 {
-    List<NodeStatus> getStatus();
+    NodeSettings getStatus();
```
nvharikrishna: Now this method won't become getSettings() ?

Probably, this could be something like this:

```
boolean getStatus();
NodeSettings getSettings();
```

src/main/java/org/apache/cassandra/sidecar/routes/CassandraSettingsService.java
```diff
@@ -0,0 +1,84 @@
...
+            CassandraAdapterDelegate cassandra = metadataFetcher.getDelegate(req.host(), instanceId);
+            nodeSettings = cassandra.getStatus();
+        }
+        catch (IllegalArgumentException e)
+        {
+            return Response.status(HttpResponseStatus.BAD_REQUEST.code()).entity(e.getMessage()).build();
+        }
+        int statusCode = nodeSettings != null
+                ? HttpResponseStatus.OK.code()
+                : HttpResponseStatus.SERVICE_UNAVAILABLE.code();
```
nvharikrishna, line 80: :+1:

src/main/java/org/apache/cassandra/sidecar/routes/CassandraSettingsService.java
```diff
@@ -0,0 +1,84 @@
...
+    @Operation(summary = "Cassandra's node settings",
+            description = "Returns HTTP 200 if Cassandra is available,400 on incorrect instanceId, 503 otherwise",
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "Cassandra status is available"),
+                    @ApiResponse(responseCode = "400", description = "Incorrect instanceId"),
+                    @ApiResponse(responseCode = "503", description = "Cassandra status is not available")
+            })
+    @GET
+    @Path("/v1/cassandra/settings")
+    public Response status(@Context HttpServerRequest req,
```
nvharikrishna: Shall we rename this method too (settings instead of status)?

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
```diff
@@ -50,7 +50,7 @@ public class CassandraAdapterDelegate implements ICassandraAdapter, Host.StateLi
     private volatile Session session;
     private SimpleCassandraVersion currentVersion;
     private ICassandraAdapter adapter;
-    private volatile boolean isUp = false;
+    private volatile NodeSettings status = null;
```
nvharikrishna: Can the status be renamed to settings or nodeSettings? Doesn't it make sense to keep both of them (isUP and nodeSettings)?

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
```diff
@@ -133,45 +133,46 @@ private void healthCheckInternal()
...
         }
         catch (NoHostAvailableException e)
         {
             logger.error("Unexpected error connecting to Cassandra instance.", e);
             // The cassandra node is down.
             // Unregister the host listener and nullify the session in order to get a new object.
-            isUp = false;
+            status = null;
             maybeUnregisterHostListener(activeSession);
             session = null;
```
nvharikrishna, line 169: On a different note, doesn't it make sense to close the session in this case?

nvharikrishna (commented):
Few minor comments.

common/src/main/java/org/apache/cassandra/sidecar/common/ICassandraAdapter.java
frankgh: or just `settings()` following Cassandra conventions

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
jyothsnakonisa: @nvharikrishna I think we want to keep the session open as long as the node is running, @frankgh  correct me if I am wrong.

yifan-c (approved):
LGTM. 
The indentations in some lines are off. I can re-indent on commit.

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
nvharikrishna: At this line existing session is assigned with null (throwing away the existing session). Thought it is better to close before assigning it to null.

nvharikrishna:
The comment I have is not related to the context of this PR. Don't want to block you on that. LGTM from my side.

nvharikrishna (approved):

dineshjoshi (approved):
+1 once you address @nvharikrishna's comment.

jyothsnakonisa:
@dineshjoshi addressed @nvharikrishna's comment on session closing.

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
```diff
@@ -133,45 +133,47 @@ private void healthCheckInternal()
...
         }
         catch (NoHostAvailableException e)
         {
             logger.error("Unexpected error connecting to Cassandra instance.", e);
             // The cassandra node is down.
             // Unregister the host listener and nullify the session in order to get a new object.
-            isUp = false;
+            nodeSettings = null;
             maybeUnregisterHostListener(activeSession);
+            session.close();
```
frankgh, line 168: we probably need to wrap this in a try/catch block , errors coming from the session should be safely ignored. And we probably want `activeSession` here instead of `session`

yifan-c:
It is already committed. Closing the PR

PR #43, opened 2022-12-01
CASSANDRASC-49: Set the copyright profile on generating idea project

yifan-c:
Automatically add the ApacheV2 license header to new files


frankgh (commented):
does it makes sense to add validation to the linter as well for the expected header?

PR #45, opened 2023-05-09
CEP-28: Implement Bulk API endpoints and introduce the Sidecar Client to Support Bulk Analytics

frankgh:
This PR implements the remaining endpoints needed to perform Bulk Analytics operations that allow
reading and writing data from Cassandra in Bulk. The new endpoints include:

- Endpoint to create snapshots
- Endpoint to clear a snapshot
- Endpoint to upload SSTable components
- Endpoint to clean up uploads for SSTable components
- Endpoint to import SSTable components
- Endpoint to retrieve gossip info
- Endpoint to retrieve the time skew for the server
- Endpoint to retrieve the ring information

Sidecar Client

Introduces the fully featured sidecar client to access Cassandra Sidecar endpoints.
It offers support for retries and Sidecar instance selection policies. The client
project itself is technology-agnostic, but we provide a vertx implementation for
the `HttpClient`. The Sidecar vertx-client can be published as a shaded-jar to be
consumed by clients where the dependencies can cause issues, especially in environments
where the dependencies are not always controlled by the consumers (for example Spark).

Co-authored-by: Saranya Krishnakumar <saranya_k@apple.com>
Co-authored-by: Yifan Cai <ycai@apache.org>
Co-authored-by: Francisco Guerrero <frank.guerrero@gmail.com>
Co-authored-by: Doug Rohrer <drohrer@apple.com>


cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicas.java
```diff
@@ -0,0 +1,417 @@
...
+
+    private boolean isSubsetOf(TokenRangeReplicas other)
+    {
+        if (this.partitioner != other.partitioner)
+            throw new IllegalStateException("Token ranges being compared do not have the same partitioner");
+
+        BigInteger maxValue = this.partitioner.maxToken;
+        if (this.start.compareTo(other.start) >= 0)
+        {
+            // TODO:
```
dineshjoshi: TODO?

vertx-client/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Configuration status="ERROR">
+    <Properties>
+        <Property name="LOG_PATTERN">%d{yyyy-MM-dd'T'HH:mm:ss.SSSZ} %p %m%n</Property>
+        <Property name="APP_LOG_ROOT">c:/temp</Property>
```
dineshjoshi: c:/temp?

src/test/resources/sidecar_with_single_multiple_instances.yaml
```diff
@@ -38,6 +38,8 @@ sidecar:
       - stream_requests_per_sec: 5000
       - delay_sec: 5
       - timeout_sec: 10
+  - allowable_time_skew_in_minutes: 60
```
dineshjoshi, line 41: Is 60 minutes a reasonable default?

src/test/java/org/apache/cassandra/sidecar/routes/sstableuploads/BaseUploadsHandlerTest.java
```diff
@@ -0,0 +1,204 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Common functionality for the SSTable Uploads {@link SSTableUploadHandler}, {@link SSTableImportHandler},
+ * and {@link SSTableCleanupHandler} tests.
+ */
+class BaseUploadsHandlerTest
+{
+    protected final Logger logger = LoggerFactory.getLogger(this.getClass());
```
dineshjoshi, line 70: logger -> LOGGER

dineshjoshi (changes requested):
Thanks @frankgh. Minor nits only in comments.

I have also taken the time to put my feedback in [this](https://github.com/dineshjoshi/cassandra-sidecar/commits/CEP-28-bulk-apis) branch. Please cherry pick the commits. Thanks.

src/test/java/org/apache/cassandra/sidecar/routes/sstableuploads/BaseUploadsHandlerTest.java
yifan-c: The `logger` is not `static`.

frankgh: yeah, classes that extend this base class can use the logger.

cassandra40/build.gradle
```diff
@@ -12,7 +13,34 @@ repositories {
...
+    testImplementation "org.junit.jupiter:junit-jupiter-params:${project.junitVersion}"
+    testImplementation "org.assertj:assertj-core:3.24.2"
+    testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:${project.junitVersion}"
+
+    testImplementation('org.mockito:mockito-core:4.10.0')
+    testImplementation('org.mockito:mockito-inline:4.10.0')
+
+}
+
+publishing {
```
yifan-c, line 37: nit: maybe declare it in the `allprojects` block of the root project.

src/test/resources/sidecar_with_single_multiple_instances.yaml
frankgh: this is configurable. Do you have a different number in mind to suggest?

frankgh:
@dineshjoshi thanks for the review. I've applied your patch and addressed your comments. Let me know if there is any additional feedback

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/RingProvider.java
```diff
@@ -0,0 +1,211 @@
...
+
+    private static String formatOwns(boolean showEffectiveOwnership, DecimalFormat ownsFormat, Float owns)
+    {
+        if (showEffectiveOwnership && owns != null)
+            return ownsFormat.format(owns);
+        return UNKNOWN_SHORT;
+    }
+
+    /**
+     * Data class to get status of endppints
```
yifan-c: nit
```suggestion
     * Data class to get status of endpoints
```

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/StorageJmxOperations.java
```diff
@@ -1,12 +1,103 @@
...
+     * @return a map of tokens to endpoints in ascending order
+     */
+    Map<String, String> getTokenToEndpointWithPortMap();
+
+    /**
+     * Effective ownership is % of the data each node owns given the keyspace
+     * we calculate the percentage using replication factor.
+     * If Keyspace == null, this method will try to verify if all the keyspaces
+     * in the cluster have the same replication strategies and if yes then we will
+     * use the first else a empty Map is returned.
```
yifan-c: nit
```suggestion
     * use the first else an empty Map is returned.
```

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicas.java
```diff
@@ -0,0 +1,417 @@
...
+     * since all subsequent ranges have to be subsets.
+     *
+     * @param other the range we are currently processing to check if "this" intersects it
+     * @return true if "this" range intersects the other
+     */
+    private boolean intersects(TokenRangeReplicas other)
+    {
+        return (other.end.compareTo(partitioner.maxToken) == 0 && this.start.compareTo(other.start) > 0) ||
+               this.start.compareTo(other.end) < 0;
+    }
```
yifan-c: I think this implementation is wrong. You want to compare `this.end.compareTo(other.start) < 0;`. Before the comparison, it needs to sort the 2 ranges by `start`, and make sure to compare left range with right range.
~Beside that, the first condition that checks with the maxToken of the partitioner seems not necessary. It only wants to compare 2 ranges.~
(Edit: looks like wrap around is considered. In that case, either this or that range needs to unwrap to check intersection.)

The below test fails 

```java
    @Test
    void testInterests()
    {
        TokenRangeReplicas range1 = new TokenRangeReplicas(BigInteger.valueOf(1), BigInteger.valueOf(10), Partitioner.Murmur3, new HashSet<>());
        TokenRangeReplicas range2 = new TokenRangeReplicas(BigInteger.valueOf(9), BigInteger.valueOf(12), Partitioner.Murmur3, new HashSet<>());
        assertThat(range1.intersects(range2)).isTrue();
        assertThat(range2.intersects(range1)).isTrue();

        TokenRangeReplicas range3 = new TokenRangeReplicas(BigInteger.valueOf(1), BigInteger.valueOf(10), Partitioner.Murmur3, new HashSet<>());
        TokenRangeReplicas range4 = new TokenRangeReplicas(BigInteger.valueOf(11), BigInteger.valueOf(20), Partitioner.Murmur3, new HashSet<>());
        assertThat(range3.intersects(range4)).isFalse();
        assertThat(range4.intersects(range3)).isFalse();
    }
```

yifan-c (changes requested):
Feel free to ignore the nits.

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicas.java
```diff
@@ -0,0 +1,415 @@
...
+
+import com.google.common.base.Objects;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.jetbrains.annotations.NotNull;
+
+
+/**
+ * Representation of a token range and the corresponding mapping to replica-set hosts
```
yifan-c: add a sentence to explain which end of the range is exclusive. I would assume `(start, end]`

cassandra40/build.gradle
frankgh: Will skip for now. I think we should create a JIRA to upgrade gradle and do this as part of that effort. What do you think?

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicas.java
```diff
@@ -0,0 +1,415 @@
...
+    {
+        return Objects.hashCode(start, end, partitioner);
+    }
+
+    private boolean isWrapAround()
+    {
+        return start.compareTo(end) >= 0;
+    }
+
+    private boolean isSubsetOf(TokenRangeReplicas other)
```
yifan-c: Does it check if this range is entirely covered by the other range? In other word, the `other` range contains `this` range. 

If so, I think the implementation is wrong. See test below
```java
    @Test
    void testSubsetRelationship()
    {
        // wraps around
        TokenRangeReplicas range1 = new TokenRangeReplicas(BigInteger.valueOf(100), BigInteger.valueOf(-100), Partitioner.Murmur3, new HashSet<>());
        TokenRangeReplicas range2 = new TokenRangeReplicas(BigInteger.valueOf(100), Partitioner.Murmur3.maxToken, Partitioner.Murmur3, new HashSet<>());
        assertThat(range1.isSubsetOf(range2)).isFalse();
        assertThat(range2.isSubsetOf(range1)).isTrue();
    }
```

nit: the name could be more succinct if it is called "contains". The implementation needs to be adjusted if being renamed.

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicas.java
```diff
@@ -0,0 +1,415 @@
...
+    /**
+     * For subset ranges, this is used to determine if a range is larger than the other by comparing start-end lengths
+     * If both ranges end at the min, we compare starting points to determine the result.
+     * When the left range is the only one ending at min, it is always the larger one since all subsequent ranges
+     * in the sorted range list have to be smaller.
+     *
+     * @param other the next range in the range list to compare
+     * @return true if "this" range is larger than the other
+     */
+    private boolean isLarger(TokenRangeReplicas other)
```
yifan-c: 1. Please write unit tests for this method. 
2. What qualifies a range to be larger? I am still confused after reading the java doc. The first sentence says it compares the lengths, which makes sense. But the later sentences seem to indicate that it also depends on the position of the ranges.

checkstyle.xml
```diff
@@ -139,6 +139,7 @@ page at http://checkstyle.sourceforge.net/config.html -->
 
         <module name="JavadocType">
             <property name="scope" value="protected" />
+            <property name="allowUnknownTags" value="true" />
```
yifan-c, line 142: 👍

client/src/main/java/org/apache/cassandra/sidecar/client/RequestContext.java
```diff
@@ -0,0 +1,455 @@
...
+/**
+ * The context for a given request that include the {@link InstanceSelectionPolicy}, the {@link RetryPolicy}, and
+ * the {@link Request}
+ */
+public class RequestContext
+{
+    protected static final SchemaRequest FULL_SCHEMA_REQUEST = new SchemaRequest();
+    protected static final TimeSkewRequest TIME_SKEW_REQUEST = new TimeSkewRequest();
+    protected static final NodeSettingsRequest NODE_SETTINGS_REQUEST = new NodeSettingsRequest();
+    protected static final RingRequest RING_REQUEST = new RingRequest();
```
yifan-c, line 54: nit: the request objects, as well as the request builder methods below, do not feel should part of `RequestContext`, which supplies the context for a request. 
But I think it is fine to leave them as is for now.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarConfig.java
```diff
@@ -0,0 +1,125 @@
...
+     */
+    public static final class Builder
+    {
+        private int maxRetries = DEFAULT_MAX_RETRIES;
+        private long retryDelayMillis = DEFAULT_RETRY_DELAY_MILLIS;
+        private long maxRetryDelayMillis = DEFAULT_MAX_RETRY_DELAY_MILLIS;
+
+        public Builder()
+        {
+        }
```
yifan-c: It can be removed.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarInstanceImpl.java
```diff
@@ -0,0 +1,105 @@
...
+    public String hostname()
+    {
+        return hostname;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+
```
yifan-c: remove the empty line.

yifan-c (changes requested):

client/src/main/java/org/apache/cassandra/sidecar/client/exception/UnexpectedStatusCodeException.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class UnexpectedStatusCodeException extends RuntimeException
+{
+    /**
+     * Constructs a new {@link UnexpectedStatusCodeException} with the given {@code statusCode} and {@code body}
+     *
+     * @param statusCode the response status code
+     * @param body       the body of the response
+     */
+    public UnexpectedStatusCodeException(int statusCode, String body)
```
yifan-c, line 32: Is it safe to assume the body is always `String`?

client/src/main/java/org/apache/cassandra/sidecar/client/request/Request.java
```diff
@@ -0,0 +1,95 @@
...
+    {
+        return null;
+    }
+
+    /**
+     * @return the HTTP method to be used for the request
+     */
+    public HttpMethod method()
+    {
+        return HttpMethod.GET;
```
yifan-c: nit: it would read more clear to not supply a default HttpMethod, and enforce the subclasses to define it.

client/src/main/java/org/apache/cassandra/sidecar/client/retry/BasicRetryPolicy.java
```diff
@@ -0,0 +1,297 @@
...
+import org.apache.cassandra.sidecar.client.exception.ResourceNotFoundException;
+import org.apache.cassandra.sidecar.client.request.Request;
+
+/**
+ * A basic {@link RetryPolicy} supporting standard status codes
+ */
+public class BasicRetryPolicy extends RetryPolicy
+{
+    protected static final String RETRY_AFTER = "Retry-After";
+    // Use -1 to retry indefinitely
```
yifan-c: nit: define a constants

client/src/main/java/org/apache/cassandra/sidecar/client/retry/ExponentialBackoffRetryPolicy.java
```diff
@@ -0,0 +1,90 @@
...
+     *
+     * @param attempts the number of attempts already performed for this request
+     * @return the number of milliseconds to wait before attempting the next request
+     */
+    @Override
+    protected long retryDelayMillis(int attempts)
+    {
+        long exponentialValue = (long) Math.pow(2, attempts - 1);
+        // do not multiply times retryDelayMillis, if we've reached the Long.MAX_VALUE already to avoid overflows
+        long retryDelay = exponentialValue == Long.MAX_VALUE ? exponentialValue : exponentialValue * retryDelayMillis;
```
yifan-c: `exponentialValue * retryDelayMillis` could overflow and result in a negative value. 
But it will never happen in practice, no human could possibly wait for this long time.

client/src/main/java/org/apache/cassandra/sidecar/client/selection/OrderedInstanceSelectionPolicy.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Returns an iterator of {@link SidecarInstance instances}.
+     *
+     * @return an iterator of {@link SidecarInstance instances}
+     */
+    @NotNull
+    public Iterator<SidecarInstance> iterator()
+    {
+        return instancesProvider.instances().iterator();
```
yifan-c, line 53: The order depends in the value returned from `instancesProvider.instances()`, which is implementation dependent. For example, the provider could shuffle the instances and return. In this example, the `OrderedInstanceSelectionPolicy` is no different from `RandomInstanceSelectionPolicy`. 

nit: It might want to sort the instances, so the order is guaranteed.

yifan-c (commented):
current progress: ~1/4

cassandra40/build.gradle
yifan-c: As said, feel free to ignore the nits.

client/src/main/java/org/apache/cassandra/sidecar/client/retry/ExponentialBackoffRetryPolicy.java
frankgh: this check here is so it doesn't overflow. Math.pow will return Long.MAX_VALUE if it will overflow.

client/src/main/java/org/apache/cassandra/sidecar/client/selection/OrderedInstanceSelectionPolicy.java
frankgh: good point. I guess this really means that we won't randomly sort the instances provided

client/src/main/java/org/apache/cassandra/sidecar/client/exception/UnexpectedStatusCodeException.java
frankgh: good point. I think this can be a binary file that we stream

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,165 @@
...
+                                       partitioner,
+                                       keyspace);
+
+        return new TokenRangeReplicasResponse(
+                writeReplicas,
+                mappingsToUnwrappedReplicaSet(rangeToEndpointMappings, hostToDatacenter, partitioner));
+    }
+
+    private List<TokenRangeReplicasResponse.ReplicaInfo>
+    writeReplicasFromPendingRanges(Map<List<String>, List<String>> readReplicaMappings,
```
yifan-c: It is confusing to call it read replicas. We can use the same term, natural, as used in Cassandra code base.
```suggestion
    writeReplicasFromPendingRanges(Map<List<String>, List<String>> naturalReplicaMappings,
```

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,165 @@
...
+                                   Map<List<String>, List<String>> pendingRangeMappings,
+                                   Map<String, String> hostToDatacenter,
+                                   Partitioner partitioner,
+                                   String keyspace)
+    {
+        LOGGER.debug("Pending token ranges for keyspace={}, pendingRangeMappings={}", keyspace, pendingRangeMappings);
+        // Merge primary and pending replicas to generate candidates for write-replicas
```
yifan-c:
```suggestion
        // Merge natural and pending range replicas to generate candidates for write-replicas
```

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,165 @@
...
+                                                  pendingRangeMappings.entrySet().stream())
+                                                  .map(entry ->
+                                                       new TokenRangeReplicas(
+                                                               new BigInteger(entry.getKey().get(0)),
+                                                               new BigInteger(entry.getKey().get(1)),
+                                                               partitioner,
+                                                               new HashSet<>(entry.getValue())))
+                                                  .collect(Collectors.toList());
+
+        // Candidate write-replica mappings are normalized by removing overlapping ranges
```
yifan-c: It took me some time to finally realize what the `normalize` and `deoverlap` methods do. I think adding the following comments should help future readers. 

```suggestion
        // Candidate write-replica mappings (merged from natural and pending ranges) are normalized by consolidate overlapping ranges
        // For an overlapping range that is included in both natural and pending ranges, say R_natural and R_pending 
        // (where R_natural == R_pending), the replicas of both R_natural and R_pending should receive writes. 
        // Therefore, the write-replicas of such range is the union of both replica sets. 
        // The procedure (esp. normalize method) implements the consolidation process.
```

yifan-c (commented):
added a few suggestions for `TokenRangeReplicaProvider`

client/src/main/java/org/apache/cassandra/sidecar/client/selection/OrderedInstanceSelectionPolicy.java
yifan-c: yeah. It feels off semantically. But it is likely not an issue to worry about pragmatically.

client/src/main/java/org/apache/cassandra/sidecar/client/retry/ExponentialBackoffRetryPolicy.java
yifan-c: It will. Say `exponentialValue == Long.MAX_VALUE / 2` and `retryDelayMillis == 3`, and boom!
```shell
jshell> exponentialValue * retryDelayMillis
$10 ==> -4611686018427387907
```

client/src/main/java/org/apache/cassandra/sidecar/client/selection/OrderedInstanceSelectionPolicy.java
frankgh: what I can probably do for now is write a more descriptive javadoc. Let me know if that works out

yifan-c: It is fine to leave it as-is. The original comment is a nit.

client/src/test/java/org/apache/cassandra/sidecar/client/retry/BasicRetryPolicyTest.java
```diff
@@ -0,0 +1,376 @@
...
+        defaultBasicRetryPolicy.onResponse(future, mockRequest, mockResponse, null, 1, false,
+                                           (attempts, retryDelayMillis) -> fail("Should never retry"));
+        future.join();
+        assertThat(future.isDone()).isTrue();
+        assertThat(future.get()).isSameAs(mockResponse);
+    }
+
+    @ParameterizedTest(name = "{index} => canRetryOnADifferentHost={0}")
+    @ValueSource(booleans = { true, false })
+    void testCompletesWithNotFoundStatusCode(boolean canRetryOnADifferentHost)
```
yifan-c, line 111: 👍

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicas.java
frankgh: adding this test and providing a fixed implementation

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicas.java
frankgh: Adding this test and providing a fixed implementation

client/src/main/java/org/apache/cassandra/sidecar/client/retry/ExponentialBackoffRetryPolicy.java
frankgh: ah yeah, that will overflow for sure. I will make it more robust

client/src/test/java/org/apache/cassandra/sidecar/client/retry/BasicRetryPolicyTest.java
```diff
@@ -0,0 +1,376 @@
...
+
+        retryPolicy.onResponse(future, mockRequest, mockResponse, null, 1, false,
+                               (attempts, retryDelayMillis) -> fail("Should not retry"));
+        future.join();
+        assertThat(future.isDone()).isTrue();
+        assertThat(future.get()).isSameAs(mockResponse);
+        assertThat(retryActionCalls.get()).isEqualTo(maxRetries);
+    }
+
+    private static Stream<Arguments> clientStatusCodeArguments()
```
yifan-c, line 208: nit: All `4XX` are client error. The stream returned is not the complete set. The original implementation only returns  the established client error codes, which is a subset. That being said, it is rarely a good idea to have custom status code. I only suggest to cover all 4XX code from the perspective of testing, because all of them technically represent client errors, regardless of the predefined or the custom. 

```suggestion
    private static Stream<Arguments> clientStatusCodeArguments()
    {
        return IntStream.range(400, 500).boxed().map(Arguments::of);
    }
```

client/src/test/java/org/apache/cassandra/sidecar/client/retry/BasicRetryPolicyTest.java
```diff
@@ -0,0 +1,376 @@
...
+    @ParameterizedTest(name = "{index} => statusCode={0}")
+    @MethodSource("clientStatusCodeArguments")
+    void testRetriesWithClientError(int statusCode)
+    {
+        when(mockResponse.statusCode()).thenReturn(statusCode);
+        testWithRetries(mockRequest, mockResponse, null, 2, 200, 200, false, false);
+        testWithRetries(mockRequest, mockResponse, null, 2, 200, 0, true, false);
+    }
+
+    private static Stream<Arguments> serverStatusCodeArguments()
```
yifan-c, line 225: same for serve error, `5xx`

client/src/test/java/org/apache/cassandra/sidecar/client/retry/ExponentialBackoffRetryPolicyTest.java
```diff
@@ -0,0 +1,139 @@
...
+        mockRequest = mock(Request.class);
+        mockResponse = mock(HttpResponse.class);
+        when(mockRequest.requestURI()).thenReturn("/api/uri");
+    }
+
+    @Test
+    void testOverflows()
+    {
+        long delay = new ExponentialBackoffRetryPolicy(1, 1, 0)
+                     .retryDelayMillis(66); // 2 ^ 63 + 1 overflows
```
yifan-c: It should be `64` instead of `66`. `Long.MAX_VALUE == 2 ^ 63`. So to trigger the current overflow check (which is not complete), the input attempts should be 63 + 1 = 64.

The reason why the current check is not complete can be demoed in this test, which fails.

```java
        delay = new ExponentialBackoffRetryPolicy(1, 8, 0)
                .retryDelayMillis(63); // 2 ^ 62 does not overflow, but 2 ^ 62 * 8 overflows. It should cap the value at MAX_VALUE.
        assertThat(delay).isEqualTo(Long.MAX_VALUE);
```

client/src/test/java/org/apache/cassandra/sidecar/client/retry/RunnableOnStatusCodeRetryPolicyTest.java
```diff
@@ -0,0 +1,101 @@
...
+import static org.assertj.core.api.Fail.fail;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Unit tests for the {@link RunnableOnStatusCodeRetryPolicy} class
+ */
+class RunnableOnStatusCodeRetryPolicyTest
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(RunnableOnStatusCodeRetryPolicyTest.class);
```
yifan-c: not used.

client/src/test/java/org/apache/cassandra/sidecar/client/selection/OrderedInstanceSelectionPolicyTest.java
```diff
@@ -0,0 +1,79 @@
...
+    @Test
+    public void testIterator()
+    {
+
+        SidecarInstancesProvider provider = new SimpleSidecarInstancesProvider(mockInstanceList);
+        InstanceSelectionPolicy instanceSelectionPolicy = new OrderedInstanceSelectionPolicy(provider);
+        Iterator<SidecarInstance> iterator = instanceSelectionPolicy.iterator();
+
+        assertThat(iterator.hasNext()).isTrue();
+        assertThat(iterator.hasNext()).isTrue().as("Expected to be true");
```
yifan-c: nit: duplication

client/src/test/java/org/apache/cassandra/sidecar/client/selection/RandomInstanceSelectionPolicyTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Test
+    public void testIterator()
+    {
+        SidecarInstancesProvider provider = new SimpleSidecarInstancesProvider(mockInstanceList);
+        InstanceSelectionPolicy instanceSelectionPolicy = new RandomInstanceSelectionPolicy(provider);
+        Iterator<SidecarInstance> iterator = instanceSelectionPolicy.iterator();
+
+        assertThat(iterator.hasNext()).isTrue();
+        assertThat(iterator.hasNext()).isTrue().as("Expected to be true");
```
yifan-c: nit: duplication

client/src/test/java/org/apache/cassandra/sidecar/client/selection/SingleInstanceSelectionPolicyTest.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Test
+    void testIterator()
+    {
+        InstanceSelectionPolicy instanceSelectionPolicy = new SingleInstanceSelectionPolicy(mockSidecarInstance);
+        Iterator<SidecarInstance> iterator = instanceSelectionPolicy.iterator();
+
+        assertThat(iterator.hasNext()).isTrue();
+        assertThat(iterator.hasNext()).isTrue().as("Expected to be true");
```
yifan-c, line 49: nit: duplication

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicaProvider.java
frankgh: yeah, makes sense, it always felt odd to call it read replicas.

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/request/GossipInfoRequestTestParameters.java
```diff
@@ -0,0 +1,72 @@
...
+               "\"releaseVersion\":\"4.0.8\",\"rpcAddress\":\"172.17.0.2\"," +
+               "\"hostId\":\"33cae238-8203-41c1-880f-8cdf98ee6720\",\"nativeAddressAndPort\":\"172.17.0.2:9042\"," +
+               "\"load\":\"89082.0\",\"sstableVersions\":\"big-nb\",\"tokens\":\"<hidden>\",\"rpcReady\":\"true\"," +
+               "\"status\":\"NORMAL,-9223372036854775808\",\"dc\":\"LO\",\"netVersion\":\"12\",\"statusWithPort\":" +
+               "\"NORMAL,-9223372036854775808\"},\"/127.0.0.2:7000\":{\"generation\":\"1679961465\"," +
+               "\"schema\":\"4994b214-7a05-35ea-a11c-d092a554dd5d\",\"rack\":\"101000201\",\"heartbeat\":\"2108\"," +
+               "\"releaseVersion\":\"4.0.8\",\"hostId\":\"dba02656-ea8c-4a1d-8011-cbc0dab5f411\"," +
+               "\"nativeAddressAndPort\":\"127.0.0.2:9042\",\"load\":\"89084.0\",\"sstableVersions\":\"big-nb\"," +
+               "\"tokens\":\"<hidden>\",\"rpcReady\":\"true\",\"dc\":\"LO\",\"netVersion\":\"12\"," +
+               "\"statusWithPort\":\"NORMAL,-3074457345618258603\"}}[root@b77ac3582e25 /]";
```
yifan-c: `[root@b77ac3582e25 /]`?? 

Btw, the nit on preparing the response body is to build a map and serialize into json. This way, the code reads more structured, instead of a large block of string. Feel free to ignore the nit though.

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/request/RingRequestForKeyspaceTestParameters.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.cassandra.sidecar.common.data.RingResponse;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Unit tests for client requests accessing the ring endpoint for a given keyspace
+ */
+public class RingRequestForKeyspaceTestParameters implements RequestTestParameters<RingResponse>
+{
+    public static final String KEYSPACE = "tutorialspoint";
```
yifan-c: hmm... it seems to be the name of a website?

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
```diff
@@ -145,47 +124,73 @@ private void healthCheckInternal()
...
             {
-                currentVersion = newVersion;
-                adapter = versionProvider.getCassandra(nodeSettings.releaseVersion()).create(cqlSession, jmxClient);
-                logger.info("Cassandra version change detected. New adapter loaded: {}", adapter);
+                // update the nodeSettings cache.
+                SimpleCassandraVersion previousVersion = currentVersion;
+                currentVersion = SimpleCassandraVersion.create(releaseVersion);
+                adapter = versionProvider.cassandra(releaseVersion)
+                                         .create(cqlSessionProvider, jmxClient);
+                nodeSettings = newNodeSettings;
```
yifan-c, line 137: There is a potential race that set `nodeSettings` to an unexpected value. But since it runs periodically, it should be corrected in the following runs. So.. I think it is acceptable.

common/src/main/java/org/apache/cassandra/sidecar/common/data/GossipInfoResponse.java
```diff
@@ -0,0 +1,292 @@
...
+     * <p>Note: When adding a new field, make sure there is a pairing access method defined in {@link GossipInfo}.
+     */
+    protected enum GossipField
+    {
+        GENERATION,
+        HEARTBEAT,
+
+        // Below are copied from org.apache.cassandra.gms.ApplicationState
+        // Note: that all padding fields are not included.
+        @Deprecated STATUS, //Deprecated and unsued in 4.0, stop publishing in 5.0, reclaim in 6.0
```
yifan-c:
```suggestion
        @Deprecated STATUS, //Deprecated and unused in 4.0, stop publishing in 5.0, reclaim in 6.0
```

common/src/main/java/org/apache/cassandra/sidecar/common/data/GossipInfoResponse.java
```diff
@@ -0,0 +1,292 @@
...
+         * Then, put the new key and value.
+         *
+         * @param snakeCaseKey the key to convert
+         * @param value        the value for the gossip info entry
+         */
+        public void camelizeKeyAndPut(String snakeCaseKey, String value)
+        {
+            String key = toLowerCamelCase(snakeCaseKey.toUpperCase());
+            super.put(key, value);
+        }
```
yifan-c: How do you make sure the input `String snakeCaseKey` is indeed in snake case. I would suggest to convert the value into enum `GossipField` first by looking up using `valueOf()`, and relocate the `toLowerCamelCase()` to the enum, which simplifies the implementation a bit, since validation for null or empty is no longer required. 

```java
    // in protected enum GossipField
        String toLowerCamelCase()
        {
            String upperUnderscoreCase = name();

            StringBuilder sb = new StringBuilder(upperUnderscoreCase.length());
            sb.append(Character.toLowerCase(upperUnderscoreCase.charAt(0)));

            int length = upperUnderscoreCase.length();
            for (int i = 1; i < length; i++)
            {
                if (upperUnderscoreCase.charAt(i) == '_' && i + 1 < length)
                {
                    i++;
                    sb.append(upperUnderscoreCase.charAt(i));
                }
                else
                {
                    sb.append(Character.toLowerCase(upperUnderscoreCase.charAt(i)));
                }
            }

            return sb.toString();
        }
```

yifan-c (commented):
Submit the comments so far. Current progress at ~2/5.

client/src/test/java/org/apache/cassandra/sidecar/client/selection/SingleInstanceSelectionPolicyTest.java
frankgh: I'm testing idempotency of hasNext here, so I will add a comment to make it clear

client/src/test/java/org/apache/cassandra/sidecar/client/selection/RandomInstanceSelectionPolicyTest.java
frankgh: testing idempotency of hasNext. I will add a comment to make it clear

client/src/test/java/org/apache/cassandra/sidecar/client/selection/OrderedInstanceSelectionPolicyTest.java
frankgh: testing idempotency of hasNext. I will add a comment to make it clear

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/request/RingRequestForKeyspaceTestParameters.java
frankgh: it should just be testkeyspace or something like that. I will update it

client/src/main/java/org/apache/cassandra/sidecar/client/exception/UnexpectedStatusCodeException.java
frankgh: this constructor is called from 
```
    UnexpectedStatusCodeException unexpectedStatusCode(HttpResponse response)
    {
        return new UnexpectedStatusCodeException(response.statusCode(), response.contentAsString());
    }
```
and `response.contentAsString()` is a best effort attempt to convert the raw response to a UTF-8 String,

client/src/test/java/org/apache/cassandra/sidecar/client/retry/ExponentialBackoffRetryPolicyTest.java
frankgh: updated comment, and added a new test case

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/request/GossipInfoRequestTestParameters.java
frankgh: removed the added string, but I agree with you about making the test prep more robust

frankgh: for now, I will keep the clean string.

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
frankgh: this shouldn't be an issue because this method is guaranteed to only run by a single thread

common/src/test/java/org/apache/cassandra/sidecar/common/SimpleCassandraVersionProviderTest.java
```diff
@@ -78,7 +78,7 @@ void ensureOutOfOrderInsertionWorks()
                           .add(new V41())
                           .add(new V30()).build();
 
-        ICassandraFactory cassandra = provider.getCassandra(SimpleCassandraVersion.create("4.0.0"));
+        ICassandraFactory cassandra = provider.cassandra(SimpleCassandraVersion.create("4.0.0"));
         assertThat(cassandra).hasSameClassAs(new V40());
```
yifan-c, line 82: ~To test `ensureOutOfOrderInsertionWorks`, should it get V30 from version `SimpleCassandraVersion.create("3.0.0")`?~

~After I made the change, it actually fails, meaning the insertion order has to be in the ascending order of the versions. Either updating the test case or the implementation of builder is needed.~

It is unrelated with this patch, please ignore.

src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
```diff
@@ -50,14 +48,9 @@
...
-     * @return {@link CQLSession} for connecting with instance via CQL
+     * @return staging directory for the uploads of the cassandra instance
      */
-    CQLSession session();
-
-    /**
-     * @return {@link JmxClient} for connecting with the instance via JMX
-     */
-    JmxClient jmxClient();
+    String uploadsStagingDir();
```
yifan-c: Can you call it `stagingDir`? It is just a directory for the staging data, not necessary to distinguish between scenarios like upload or others.

src/test/resources/sidecar_with_single_multiple_instances.yaml
frankgh: discussed offline. we decided to leave the value as is for now.

src/main/java/org/apache/cassandra/sidecar/concurrent/ConcurrencyLimiter.java
```diff
@@ -0,0 +1,112 @@
...
+ *         // to make sure that permit is always released
+ *         context.addEndHandler(v -&gt; limiter.releasePermit());
+ *         // .. some expensive operation
+ *     }
+ * </pre>
+ */
+public class ConcurrencyLimiter
+{
+    @VisibleForTesting
+    final AtomicInteger permits = new AtomicInteger(0);
```
yifan-c: instead of exposing the mutable `AtomicInteger`, can you add a method to return the current `permits: int` and make the filed private instead? 

```java
    public int acquiredPermits()
    {
        return permits.get();
    }
```

src/main/java/org/apache/cassandra/sidecar/data/SchemaRequest.java
```diff
@@ -15,24 +15,34 @@
...
+package org.apache.cassandra.sidecar.data;
 
 import org.jetbrains.annotations.Nullable;
 
 /**
  * Holder class for the {@link org.apache.cassandra.sidecar.routes.SchemaHandler}
  * request parameters
  */
-public class SchemaRequest extends QualifiedTableName
+public class SchemaRequest
```
yifan-c, line 26: Should we rename this group of classes to Payload in this patch? For this one, `SchemaRequestPayload`

src/main/java/org/apache/cassandra/sidecar/routes/FileStreamHandler.java
```diff
@@ -24,54 +24,60 @@
...
 /**
  * Handler for sending out files.
  */
-public class FileStreamHandler implements Handler<RoutingContext>
+public class FileStreamHandler extends AbstractHandler<String>
 {
-    private static final Logger LOGGER = LoggerFactory.getLogger(FileStreamHandler.class);
     public static final String FILE_PATH_CONTEXT_KEY = "fileToTransfer";
-
+    private static final Logger LOGGER = LoggerFactory.getLogger(FileStreamHandler.class);
```
yifan-c: logger defined in base class already

src/main/java/org/apache/cassandra/sidecar/routes/SnapshotsHandler.java
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected SnapshotRequest extractParamsOrThrow(final RoutingContext context)
+    {
+        boolean includeSecondaryIndexFiles =
+        "true".equalsIgnoreCase(context.request().getParam(INCLUDE_SECONDARY_INDEX_FILES, "false"));
```
yifan-c: nit
```suggestion
        boolean includeSecondaryIndexFiles = Boolean.parseBoolean(context.request().getParam(INCLUDE_SECONDARY_INDEX_FILES, "false"));
```

src/main/java/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableCleanupHandler.java
```diff
@@ -0,0 +1,110 @@
...
+                               HttpServerRequest httpRequest,
+                               String host,
+                               SocketAddress remoteAddress,
+                               String uploadId)
+    {
+        if (context.request().method() != HttpMethod.DELETE)
+        {
+            context.fail(wrapHttpException(HttpResponseStatus.BAD_REQUEST, "Unknown API requested"));
+            return;
+        }
```
yifan-c: Why would the request router route such request? It is only reregistered for `delete`

src/main/java/org/apache/cassandra/sidecar/utils/BaseFileSystem.java
```diff
@@ -0,0 +1,125 @@
...
+
+import io.vertx.core.Future;
+import io.vertx.core.file.FileProps;
+import io.vertx.core.file.FileSystem;
+import org.apache.cassandra.sidecar.cluster.InstancesConfig;
+import org.apache.cassandra.sidecar.common.utils.CassandraInputValidator;
+import org.apache.cassandra.sidecar.concurrent.ExecutorPools;
+
+/**
+ * Provides functionality for filesytem operations
```
yifan-c:
```suggestion
 * Provides functionality for filesystem operations
```

src/main/java/org/apache/cassandra/sidecar/utils/CacheFactory.java
```diff
@@ -0,0 +1,101 @@
...
+        return Caffeine.newBuilder()
+                       .ticker(ticker)
+                       .executor(MoreExecutors.directExecutor())
+                       .expireAfterAccess(expireAfterAccessDuration)
+                       .maximumSize(maximumSize)
+                       .recordStats()
+                       .removalListener((RemovalListener<SSTableImporter.ImportOptions, Future<Void>>)
+                                        (options, result, cause) -> {
+                                            LOGGER.debug("Removed entry '{}' with options '{}' from SSTable Import " +
+                                                         "Cache  and cause {}", result, options, cause);
```
yifan-c:
```suggestion
                                                         "Cache and cause {}", result, options, cause);
```

src/main/java/org/apache/cassandra/sidecar/utils/MD5ChecksumVerifier.java
```diff
@@ -0,0 +1,101 @@
...
+            digest = MessageDigest.getInstance("MD5");
+        }
+        catch (NoSuchAlgorithmException e)
+        {
+            return Future.failedFuture(e);
+        }
+
+        Promise<String> result = Promise.promise();
+        file.pause()
+            .setReadBufferSize(64 * 1024)
```
yifan-c: Declare a constant for the buffer size?

src/main/java/org/apache/cassandra/sidecar/utils/RequestUtils.java
```diff
@@ -18,30 +18,29 @@
...
-            // remove brackets from ipv6 addresses
-            return host.startsWith("[") ? host.substring(1, host.length() - 1) : host;
-        }
-        return address;
+        String value = request.getParam(headerName);
+        if ("true".equalsIgnoreCase(value))
+            return true;
+        if ("false".equalsIgnoreCase(value))
+            return false;
+        return defaultValue;
```
yifan-c, line 44: Use `Boolean.parseBoolean()` instead?

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicas.java
frankgh: Added some unit tests. I also added javadocs clarifying that this method assumes that the ranges are normalized and unwrapped, i.e. `this` comes before `other` AND there's no wrapping around the min token

client/src/main/java/org/apache/cassandra/sidecar/client/exception/UnexpectedStatusCodeException.java
frankgh: will leave for now as it's an unexpected case, what do you think?

client/src/main/java/org/apache/cassandra/sidecar/client/RequestContext.java
frankgh: maybe we can have it as a follow up

src/main/java/org/apache/cassandra/sidecar/utils/SSTableImporter.java
```diff
@@ -0,0 +1,587 @@
...
+         */
+        public void unlock()
+        {
+            isQueueInUse.set(false);
+        }
+
+        public boolean removeIf(Predicate<? super Pair<Promise<Void>, ImportOptions>> filter)
+        {
+            return super.removeIf(filter);
+        }
```
yifan-c: Remove it?

src/test/integration/org/apache/cassandra/sidecar/IntegrationTestBase.java
```diff
@@ -0,0 +1,131 @@
...
+import io.vertx.ext.web.client.WebClient;
+import io.vertx.junit5.VertxTestContext;
+import org.apache.cassandra.sidecar.common.data.QualifiedTableName;
+import org.apache.cassandra.sidecar.common.testing.CassandraTestContext;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Base class for integration test.
+ * Start a docker container of cassandra at the begining of each test, and
```
yifan-c:
```suggestion
 * Start a docker container of cassandra at the beginning of each test, and
```

vertx-client-shaded/build.gradle
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+}
+
+/**
+ * A Transformer which updates the Netty JAR META-INF/ resources to accurately
+ * reference shaded class names.
+ */
+@CacheableTransformer
+class NettyResourceTransformer implements Transformer {
```
yifan-c, line 90: MAGIC!

yifan-c (commented):
Last batch done

src/main/java/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableCleanupHandler.java
frankgh: oh interesting. that should not be there.

src/main/java/org/apache/cassandra/sidecar/utils/RequestUtils.java
frankgh: it's not quite the same. This method takes only true or false, case insensitively. parseBoolean compares to true only.

vertx-client-shaded/build.gradle
frankgh: this is just renaming netty directories in the fat jar :)

src/main/java/org/apache/cassandra/sidecar/data/SchemaRequest.java
frankgh: this is not really the payload, these are path params.

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
yifan-c: yep

frankgh:
Thanks for the review @yifan-c . I've pushed a few commits and I hope I've addressed your comments fully. Let me know if there is anything additional to address

yifan-c (commented):
LGTM. I will take a second pass of the token range replica changes tomorrow.

src/main/java/org/apache/cassandra/sidecar/utils/RequestUtils.java
yifan-c: I thought it is equivalent to 
```java
        return Boolean.parseBoolean(request.getParam(headerName, String.valueOf(defaultValue)));
```

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicas.java
```diff
@@ -0,0 +1,515 @@
...
+        validateMatchingPartitioner(other);
+
+        BigInteger maxValue = this.partitioner.maxToken;
+        if (this.start.compareTo(other.start) == 0)
+        {
+            if (this.end.equals(maxValue)) return 1;
+            else if (other.end.equals(maxValue)) return -1;
+            else return this.end.compareTo(other.end);
+        }
+        else return this.start.compareTo(other.start);
```
yifan-c: It looks wrong to me..

Say R1 (1, MAX] and R2 (1, MAX], apparently R1 == R2, but according to this implement `R1.compareTo(R2) == 1`

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicas.java
```diff
@@ -0,0 +1,515 @@
...
+     */
+    @Override
+    public boolean equals(Object o)
+    {
+        if (!(o instanceof TokenRangeReplicas))
+        {
+            return false;
+        }
+        TokenRangeReplicas that = (TokenRangeReplicas) o;
+        return (this.start.equals(that.start) && this.end.equals(that.end) && this.partitioner == that.partitioner);
```
yifan-c: Use `Objects.equals` instead..

cassandra40/src/main/java/org/apache/cassandra/sidecar/cassandra40/TokenRangeReplicas.java
```diff
@@ -0,0 +1,515 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public int hashCode()
+    {
+        return Objects.hashCode(start, end, partitioner);
```
yifan-c: `Objects.hash`

dineshjoshi (approved):
+1

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/38cdacb2e7418e2aefbcffb1754dcd324c46028d

PR #46, opened 2023-05-19
CASSANDRASC-50: Deprecate the sidecar cassandra health endpoint conta…

frankgh:
…ining instance segment

This commit deprecates the Cassandra Health endpoint containing the instance segment in the path. This endpoint is currently unused and it is replaced by the health endpoint with the `instanceId` query string parameter. Since the `instanceId` is optional we move the path param (mandatory) to the query param (optional).

This commit also moves the CassandraHealthService from jax RS to a vertx Handler. It also moves the HealthService to a inlined handler, simplifying the service.


dineshjoshi (approved):
+1

src/main/java/org/apache/cassandra/sidecar/cluster/InstancesConfig.java
```diff
@@ -37,14 +37,16 @@
      *
      * @param id instance's id
      * @return instance meta information
+     * @throws IllegalArgumentException when the instance with {@code id} does not exist
      */
-    InstanceMetadata instanceFromId(final int id);
+    InstanceMetadata instanceFromId(int id) throws IllegalArgumentException;
```
yifan-c: Does `NoSuchElementException` fit better?

yifan-c (approved):
LGTM with 1 nit

src/main/java/org/apache/cassandra/sidecar/cluster/InstancesConfig.java
frankgh: yeah, I think this makes more sense. I will update it

PR #47, opened 2023-05-31
CASSANDRASC-52 Sidecar Returns Own Version in Node Settings

5:
As of now, Sidecar's node settings endpoint returns Cassandra version and partitioner only.
This patch adds Sidecar's own build version into the node settings object for reporting purposes.


common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
```diff
@@ -16,33 +16,74 @@
...
-    private final String releaseVersion;
+    private static final Logger LOGGER = LoggerFactory.getLogger(NodeSettings.class);
+
     private final String partitioner;
+    private final String releaseVersion;
+    private final String sidecarVersion;
+
+    private static String getSidecarVersion()
+    {
+        try (InputStream version = NodeSettings.class.getResourceAsStream("/sidecar.version"))
```
frankgh: we should read this once only. Otherwise it will be an expensive operation

5: Done.

build.gradle
```diff
@@ -174,6 +174,12 @@ dependencies {
     integrationTestImplementation("org.testcontainers:junit-jupiter:${project.testcontainersVersion}")
 }
 
+jar {
+    doFirst {
+        new File("$projectDir/src/main/resources/sidecar.version").text = version
+    }
+}
+
```
yifan-c: Does it create a version file with the content? 
What happens if the file already exist? 
Does it make sense to checkin the version file?

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
```diff
@@ -16,33 +16,75 @@
...
+     * release version, and Sidecar version information
+     *
      * @param partitioner    the partitioner used by the Cassandra node
+     * @param releaseVersion the release version of the Cassandra node
+     * @param sidecarVersion the version of the Sidecar on the Cassandra node
      */
-    public NodeSettings(@JsonProperty("releaseVersion") String releaseVersion,
-                        @JsonProperty("partitioner") String partitioner)
+    public NodeSettings(@JsonProperty("partitioner")    String partitioner,
+                        @JsonProperty("releaseVersion") String releaseVersion,
```
yifan-c: What is the reason of switching the order? It causes other refactorings in the same PR.

yifan-c (commented):
have a few questions.

build.gradle
5: It does, silently overwriting the file if it exists.
I don't think we should check the file in — having it change automatically may be distracting/confusing, and changing it manually will be easy to forget.

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
5: I wanted to keep two version numbers (of Cassandra and of Sidecar) next to each other in method signatures.

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
```diff
@@ -16,33 +16,75 @@
...
+
+    private static String getSidecarVersion()
+    {
+        try (InputStream version = NodeSettings.class.getResourceAsStream("/sidecar.version"))
+        {
+            return IOUtils.toString(version, StandardCharsets.UTF_8);
+        }
+        catch (Exception exception)
+        {
+            LOGGER.error("Failed to ex Sidecar version", exception);
```
frankgh: not sure what you meant to say here, is it parse?
```suggestion
            LOGGER.error("Failed to parse Sidecar version", exception);
```

5: Fixed, that was a typo, thanks.

common/build.gradle
```diff
@@ -23,6 +23,7 @@ test {
 }
 
 dependencies {
+    implementation(group: 'commons-io', name: 'commons-io', version: '2.12.0')
```
frankgh: common should have a minimal amount of dependencies. Let's not add this dependency just for reading the input stream

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
yifan-c: I do not think the order is fixed even if you change the order in the param list, unless `@JsonPropertyOrder` is declared. 

It kind of reminds me of the clarity of the payload. See the example below. It feels unclear. Because the fields in node settings response payload were only about the cassandra node below. Now, we have a field about sidecar, that is sitting next to the cassandra fields at the _same_ level. 
We certainly cannot change the schema of the existing fields. I guess you do not want to add an extra endpoint just for sidecar. 
What do you think about adding a sidecar json object, that includes the sidecar version and other related settings to be added in the future? 

```json
{
  "partitioner": "Murmur3Partitioner",
  "releaseVersion": "4.0",
  "sidecarVersion": "1.0"
}
```

The alternative,
```java
{
  "partitioner": "Murmur3Partitioner",
  "releaseVersion": "4.0",
  "sidecar": {
    "version": "1.0"
  }
}
```

build.gradle
5: Added a comment explaining the current behavior.

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
5: Honestly, this feels like a bit of an overkill to me, as there's a chance there will be no other Sidecar-related values…
But I can introduce another class just for this one attribute, if you insist.

That being said, I don't think changing the existing schema is completely out of the question.
This code was published very recently and is probably not actively used by anyone, so we could try renaming `releaseVersion` into `cassandraVersion`…

common/build.gradle
5: Replaced call to `IOUtils.toString()` with a more verbose JDK-based implementation.

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
```diff
@@ -16,33 +16,82 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(NodeSettings.class);
+    private static final String SIDECAR_VERSION = getSidecarVersion();
+
     private final String partitioner;
+    private final String releaseVersion;
+    private final String sidecarVersion;
+
+    private static String getSidecarVersion()
+    {
+        try (InputStream input = NodeSettings.class.getResourceAsStream("/sidecar.version");
```
frankgh: maybe define the string as a constant?

5: I don't think having a constant here will help, as it cannot be referenced from the Gradle build file… Local variable, maybe?

5: Added a local final variable.

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
5: Changed the Sidecar version field type as per our conversation, let me know whether this is what you had in mind.

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
```diff
@@ -16,33 +16,80 @@
...
-     * information.
+     * Constructs a new {@link NodeSettings} object with the Cassandra node's release version
+     * and partitioner information, uses Sidecar version from the currently loaded binary
      *
      * @param releaseVersion the release version of the Cassandra node
      * @param partitioner    the partitioner used by the Cassandra node
      */
+    public NodeSettings(String releaseVersion, String partitioner)
+    {
+        this(releaseVersion, partitioner, Collections.singletonMap(VERSION, SIDECAR_VERSION));
```
frankgh: a tiny object creation optimization is to declare the map as a static field. That way we no longer need to declare `SIDECAR_VERSION`. The downside of this approach is that the code will change if we add any other field to the map in the future where the new value in the map is dynamic.

frankgh (commented):
The PR looks good to me. Validated locally

```
➜  cassandra-sidecar git:(r/pr47) ✗ curl -s localhost2:9043/api/v1/cassandra/settings  | jq
{
  "releaseVersion": "4.0.9-SNAPSHOT",
  "partitioner": "org.apache.cassandra.dht.Murmur3Partitioner",
  "sidecar": {
    "version": "unknown"
  }
}
```

frankgh:
added the step to run and now I get the version :
```
➜  cassandra-sidecar git:(r/pr47) ✗ curl -s localhost2:9043/api/v1/cassandra/settings  | jq
{
  "releaseVersion": "4.0.9-SNAPSHOT",
  "partitioner": "org.apache.cassandra.dht.Murmur3Partitioner",
  "sidecar": {
    "version": "1.0-SNAPSHOT"
  }
}
```

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
5: Done.

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
```diff
@@ -16,33 +16,80 @@
...
+import org.slf4j.LoggerFactory;
 
 /**
  * Holds information about the specific node settings
  */
 public class NodeSettings
 {
+    private static final Logger LOGGER = LoggerFactory.getLogger(NodeSettings.class);
+    private static final String VERSION = "version";
+    private static final Map<String, String> SIDECAR = Collections.singletonMap(VERSION, getSidecarVersion());
```
dineshjoshi: I don't recommend opening a file during static initialization of variables. While this seems like a good idea, it should be avoided. I would prefer reading this explicitly during server initialization and populating the node settings (perhaps injecting it via Guice?).

dineshjoshi (changes requested):

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
5: In this specific case it was reading from an internal resource, and swallowing all possible exceptions, but I agree that this can lead to issues that are difficult to identify. Added a separate commit that uses Guice to retrieve current Sidecar version.

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
```diff
@@ -57,23 +82,41 @@ public String partitioner()
...
+    public String sidecarVersion()
+    {
+        return sidecar.get(VERSION);
+    }
+
     /**
      * {@inheritDoc}
      */
-    public boolean equals(Object o)
+    public boolean equals(Object other)
```
yifan-c, line 94: Please do not add those unrelated changes (e.g. unnecessary variable renaming) for *future* contributions. (We can keep this one, since it is changed already)

src/main/java/org/apache/cassandra/sidecar/MainModule.java
```diff
@@ -304,4 +314,28 @@ public ChecksumVerifier checksumVerifier(Vertx vertx)
     {
         return new MD5ChecksumVerifier(vertx.fileSystem());
     }
+
+    @Provides
+    @Singleton
+    @Named("SidecarVersion")
+    private static String sidecarVersion()
+    {
+        final String resource = "/sidecar.version";
```
yifan-c: remove `final`

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
```diff
@@ -57,23 +82,41 @@ public String partitioner()
...
 
+    @JsonProperty("sidecar")
+    public Map<String, String> sidecar()
+    {
+        return sidecar;
+    }
+
+    public String sidecarVersion()
+    {
+        return sidecar.get(VERSION);
```
yifan-c: can you do a null check for `sidecar`? It is prone to NPE.

src/main/java/org/apache/cassandra/sidecar/MainModule.java
```diff
@@ -304,4 +314,28 @@ public ChecksumVerifier checksumVerifier(Vertx vertx)
     {
         return new MD5ChecksumVerifier(vertx.fileSystem());
     }
+
+    @Provides
+    @Singleton
+    @Named("SidecarVersion")
+    private static String sidecarVersion()
```
yifan-c: nit: drop `static`. I do not think it is required. The scope is singleton already.

Does the annotation work with `private` methods? It feel counterintuitive semantically, even if it works. Can you change it to `public`?

src/main/java/org/apache/cassandra/sidecar/MainModule.java
```diff
@@ -304,4 +314,28 @@ public ChecksumVerifier checksumVerifier(Vertx vertx)
...
+                output.write(buffer, 0, length);
+            }
+            return output.toString(StandardCharsets.UTF_8.name());
+        }
+        catch (Exception exception)
+        {
+            LOGGER.error("Failed to retrieve Sidecar version", exception);
+        }
+        return "unknown";
+    }
```
yifan-c, line 325: There is no test for the method (reading from the resource stream). Can you add a test?

yifan-c (changes requested):

src/main/java/org/apache/cassandra/sidecar/MainModule.java
5: Removed.

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
5: Acknowledged.

src/main/java/org/apache/cassandra/sidecar/MainModule.java
5: Missed that when copy-pasting, fixed.

src/main/java/org/apache/cassandra/sidecar/MainModule.java
5: This is a tricky one — since the version number is embedded into the JAR itself, I don't see what to compare it with. I could read it out from resources in the test code as well, but then I'll be just testing two copies of the same code against each other, which may not be helpful.

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
5: Done.

src/main/java/org/apache/cassandra/sidecar/MainModule.java
frankgh: It feels to me that we should have a `SidecarVersionProvider` class instead, that should lazily initialize the version when requested, as opposed to doing it upfront in the `MainModule` with a `@Named` annotation for the string

5: I actually did that first, but then simplified the code to a singleton String.
Provider would definitely make sense if there was a way for Sidecar version to change at the runtime.
Currently it is built into the binary, so I don't see the need for a separate class.

src/main/java/org/apache/cassandra/sidecar/MainModule.java
frankgh: Well, the reason I think value makes more sense is because it allows you to potentially retry reading from the file if retrieving the version fails during `MainModule` initialization.

src/main/java/org/apache/cassandra/sidecar/MainModule.java
```diff
@@ -304,4 +314,28 @@ public ChecksumVerifier checksumVerifier(Vertx vertx)
...
+            int length;
+            while ((length = input.read(buffer)) >= 0)
+            {
+                output.write(buffer, 0, length);
+            }
+            return output.toString(StandardCharsets.UTF_8.name());
+        }
+        catch (Exception exception)
+        {
+            LOGGER.error("Failed to retrieve Sidecar version", exception);
```
frankgh:
```suggestion
            LOGGER.error("Failed to retrieve Sidecar version from resource={}", resource, exception);
```

5: Done.

src/main/java/org/apache/cassandra/sidecar/MainModule.java
yifan-c: Regarding the test, you can add a *test* resource called `sidecar.version`, and write a testing like this.

```java
    @Test
    public void testReadSidecarVersion()
    {
        String version = injector.getInstance(Key.get(String.class, Names.named("SidecarVersion")));
        Assertions.assertThat(version).isNotNull().isEqualTo("1.0.0");
    }
```

5: Done, thank you.

src/test/java/org/apache/cassandra/sidecar/MainModuleTest.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+public class MainModuleTest {
+    @Test
+    public void testSidecarVersion()
+    {
+        Injector injector = Guice.createInjector(Modules.override(new MainModule()).with(new TestModule()));
+
+        String sidecarVersion = injector.getInstance(Key.get(String.class, Names.named("SidecarVersion")));
+
+        assertEquals("1.0-TEST", sidecarVersion);
```
yifan-c, line 43: Thanks for adding the test, but it does not run successfully. 
This is the result I get
```
org.opentest4j.AssertionFailedError: expected: <1.0-TEST> but was: <1.0-SNAPSHOT>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1142)
	at org.apache.cassandra.sidecar.MainModuleTest.testSidecarVersion(MainModuleTest.java:23)
```

yifan-c: What happens is that `:jar` task runs before `:test` task. So the test actually reads from the snapshot jar built.

5: Added the missing test resource file.

yifan-c (commented):
In general, the patch looks good.
There are minor issues with build still. 
Can you run `./gradlew clean build` and fix the errors?

src/main/java/org/apache/cassandra/sidecar/MainModule.java
```diff
@@ -311,4 +320,28 @@ public ChecksumVerifier checksumVerifier(Vertx vertx)
...
+            int length;
+            while ((length = input.read(buffer)) >= 0)
+            {
+                output.write(buffer, 0, length);
+            }
+            return output.toString(StandardCharsets.UTF_8.name());
+        }
+        catch (Exception exception)
+        {
+            LOGGER.error("Failed to retrieve Sidecar version from resource {}", resource, exception);
```
dineshjoshi: What's the consequence of not having the correct sidecar version? IMO, this should be fatal. If sidecar version is expected and is not available, we should bail on the start.

@frankgh @yifan-c any opinions?

dineshjoshi (changes requested):

src/main/java/org/apache/cassandra/sidecar/MainModule.java
5: To explain my motivation: if we don't have a resource with version build into Sidecar, then something went wrong with the artifact build process, but it doesn't prevent us from starting up and serving requests, so just log an error, and return "`unknown`", which only appears on the usage dashboard.

frankgh: Well, we will definitely not have the version for development, unit testing, or some gradle targets. Which means that we will need to handle those cases if we are going to go with that behavior.

As Yuriy mentioned above, if the file is missing it means there was some issue during the build process. I think we should not worry much about that case. I'm leaning more towards not failing the startup for two reasons. The first reason is that we should/can rely on the build process to correctly generate the version file. The second reason is that we will need to handle the cases where it's fine to not have the version (i.e unit tests, gradlew run, etc)

5: I just pushed an additional commit implementing what we all just discussed.
Please take another look.

src/main/java/org/apache/cassandra/sidecar/utils/SidecarVersionProvider.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Retrieves, caches, and provides build version of this Sidecar binary
+ */
+public class SidecarVersionProvider
+{
+    private final String sidecarVersion;
+
+    public SidecarVersionProvider(String resource)
+    {
+        try (InputStream input = NodeSettings.class.getResourceAsStream(resource);
```
yifan-c: Import  `NodeSettings` does not seem to be necessary.

5: Thanks, replaced with a call to `getClass()`.

dineshjoshi (approved):
+1

PR #51, opened 2023-06-15
CASSANDRASC-55: Extract the in-jvm dtest template for use in other projects.

JeetKunDoug:
This also required moving some additional classes to `common` so they could be used in the new test-framework project.


JeetKunDoug:
Clean CI build:
https://app.circleci.com/pipelines/github/JeetKunDoug/cassandra-sidecar/41/workflows/d2759a94-4443-410c-b012-69a81605f06b

JeetKunDoug:
CI running at https://app.circleci.com/pipelines/github/JeetKunDoug/cassandra-sidecar/43/workflows/cab30162-fcbb-4b6c-91d0-d8ff95a4815c (updated after fixing the license file issue)

src/test/integration/org/apache/cassandra/testing/CassandraIntegrationTest.java
```diff
@@ -94,4 +96,32 @@
...
+     */
+    boolean buildCluster() default true;
+
+    /**
+     * If the integration test does not need to be run on each version of Cassandra, set this to false
+     *      and it will be run only on the first version specified.
+     * @return true if the test should be run on all tested versions of Cassandra,
+     *         false if it should be run on the first version.
+     */
+    boolean versionDependent() default true;
```
yifan-c, line 126: it makes a lot of sense!

build.gradle
```diff
@@ -20,6 +20,7 @@
 
 import java.nio.file.Files
 import java.nio.file.Paths
+import java.nio.file.Files
```
frankgh, line 25: odd, this is imported already in line 21

common/src/main/java/org/apache/cassandra/sidecar/cluster/InstancesConfigImpl.java
```diff
@@ -75,7 +77,18 @@ public InstanceMetadata instanceFromHost(String host) throws NoSuchElementExcept
...
         {
-            throw new NoSuchElementException("Instance with host address " + host + " not found");
+            try
+            {
+                instanceMetadata = hostToInstanceMetadata.get(InetAddress.getByName(host).getHostAddress());
+            }
+            catch (UnknownHostException e)
+            {
+                throw new RuntimeException(e);
+            }
```
frankgh: We should use the DnsResolver here. Which means you'll need to pass it to the constructor, or we change the signature of the `instanceFromHost` method to take the DnsResolver. It seems like all the calling points for `instanceFromHost` should have access to the DnsResolver, so we could change the signature
```suggestion
            try
            {
                instanceMetadata = hostToInstanceMetadata.get(dnsResolver.resolve(host));
            }
            catch (UnknownHostException e)
            {
                throw new RuntimeException("Unable to resolve instance with host address " + host, e);
            }
```

build.gradle
JeetKunDoug: Fixed.

src/test/integration/org/apache/cassandra/testing/ConfigurableCassandraTestContext.java
```diff
@@ -0,0 +1,72 @@
...
+        this.builder = builder;
+    }
+
+    public UpgradeableCluster configureCluster(Consumer<UpgradeableCluster.Builder> configurator) throws IOException
+    {
+        if (cluster != null)
+        {
+            throw new IllegalStateException(BUILT_CLUSTER_CANNOT_BE_CONFIGURED_ERROR);
+        }
+        configurator.accept(builder);
```
frankgh, line 51: 👍

src/test/integration/org/apache/cassandra/sidecar/testing/CassandraSidecarTestContext.java
```diff
@@ -87,8 +122,13 @@ private InstancesConfig buildInstancesConfig(CassandraVersionProvider versionPro
...
             assertThat(dataDirParentPath).isNotNull()
-                                         .exists();
-            String uploadsStagingDirectory = dataDirParentPath.resolve("staging").toFile().getAbsolutePath();
+                      .exists();
+            Path stagingPath = dataDirParentPath.resolve("staging");
+            String uploadsStagingDirectory = stagingPath.toFile().getAbsolutePath();
+            if (!Files.exists(stagingPath))
+            {
+                Files.createDirectory(stagingPath);
+            }
```
frankgh: similar to `mkdir -p ...`
```suggestion
            Files.createDirectories(stagingPath);
```

frankgh (commented):
Looks good to me. +1 (nb) from me . Thanks for this work

JeetKunDoug:
Latest CI: https://app.circleci.com/pipelines/github/JeetKunDoug/cassandra-sidecar/48/workflows/536c4ad3-1471-4e5c-bf74-858fd9a6cf94

adapters/base/build.gradle
```diff
@@ -59,7 +60,7 @@ publishing {
         maven(MavenPublication) {
             from components.java
             groupId project.group
-            artifactId "${archivesBaseName}"
+            artifactId "adapters-base"
```
frankgh, line 63: how about `base-adapter`?

build.gradle
```diff
@@ -72,11 +77,22 @@ allprojects {
...
+    codeCheckTasks.dependsOn(tasks.withType(RatTask))
+    codeCheckTasks.dependsOn(tasks.withType(SpotBugsTask))
+
+    tasks.withType(Test) {
+        shouldRunAfter(codeCheckTasks)
+        shouldRunAfter(tasks.withType(Checkstyle))
+        shouldRunAfter(tasks.withType(RatTask))
+        shouldRunAfter(tasks.withType(SpotBugsTask))
+    }
+
```
yifan-c, line 95: So, you want to achieve the effect of `check` with `test`? If you run `./gradlew check`, it should run test and the checks specified.

.gitignore
```diff
@@ -77,7 +77,7 @@ src/dist/*
 *.logdir_IS_UNDEFINED
 
 # Sidecar version - generated by build.gradle
-src/main/resources/sidecar.version
+common/src/main/resources/sidecar.version
```
yifan-c: nit: it could be just `sidecar.version`

common/src/main/java/org/apache/cassandra/sidecar/cluster/InstancesConfigImpl.java
```diff
@@ -75,7 +79,18 @@ public InstanceMetadata instanceFromHost(String host) throws NoSuchElementExcept
...
         if (instanceMetadata == null)
         {
-            throw new NoSuchElementException("Instance with host address " + host + " not found");
+            try
+            {
+                instanceMetadata = hostToInstanceMetadata.get(dnsResolver.resolve(host));
+            }
+            catch (UnknownHostException e)
+            {
+                throw new RuntimeException(e);
```
yifan-c: nit: Should it throw `NoSuchElementException` in this case too? An unresolved host is very much the same as the instance metadata cannot be found. From API consumer's perspective, it either looks up the instance metadata, or the value does not exist.

adapters/base/build.gradle
JeetKunDoug: `adapters-base` matches the directory structure `adapters/base` but I suppose `base-adapter` rolls off the tongue a bit more easily - @yifan-c any opinions here (I'll leave it as is until we come to consensus but don't have a strong opinion either way)

yifan-c: Just providing a data point
https://mvnrepository.com/search?q=base
All the artifacts in the first page has `-base` instead of `base-`

src/test/integration/org/apache/cassandra/sidecar/testing/CassandraSidecarTestContext.java
```diff
@@ -142,6 +182,8 @@ public String toString()
     public void close()
     {
         instancesConfig.instances().forEach(instance -> instance.delegate().close());
+        // TODO: Do we need to close sessions?
+//        sessionProviders.
```
yifan-c: It is to be resolved?

adapters/base/build.gradle
frankgh: in that case I +1 -base :) . Thanks for the data point

build.gradle
frankgh: I think it is described in line 59:
> // Force checkstyle, rat, and spotBugs to run before test tasks for faster feedback

Just tried running `./gradlew test` locally (by removing license header from a couple of java files), and it doesn't modify the test behavior. 

`./gradlew check` fails fast before running tests.

common/src/main/java/org/apache/cassandra/sidecar/cluster/InstancesConfigImpl.java
JeetKunDoug: Done

common/src/main/java/org/apache/cassandra/sidecar/cluster/InstancesConfigImpl.java
JeetKunDoug: So I'd like to be able to include the details of the UnknownHostException, but `NoSuchElementException` doesn't have a constructor that takes a cause as an argument... for now, will just include some additional text in the exception's description.

src/test/integration/org/apache/cassandra/sidecar/testing/CassandraSidecarTestContext.java
JeetKunDoug: Yup - thanks for catching it... was a simple one-liner I just hadn't remember to get back to.

JeetKunDoug: To be a bit more specific, I'm fairly certain that shutting down the Sidecar server itself will eventually close these, but in case somehow the sidecar server doesn't, it makes sense to close them explicitly here.

common/src/main/java/org/apache/cassandra/sidecar/cluster/InstancesConfigImpl.java
```diff
@@ -85,10 +85,12 @@ public InstanceMetadata instanceFromHost(String host) throws NoSuchElementExcept
...
             {
-                throw new NoSuchElementException("Instance with host address "
-                                                 + host +
-                                                 " not found, and an error occurred when " +
-                                                 "attempting to get its IP address.");
+                NoSuchElementException error = new NoSuchElementException("Instance with host address "
+                                                      + host +
+                                                      " not found, and an error occurred when " +
+                                                      "attempting to resolve its IP address.");
+                error.initCause(e);
```
yifan-c, line 92: 👍

PR #53, opened 2023-06-21
CASSANDRASC-57 Remove RESTEasy

frankgh:
This commit removes RESTEasy and associated dependencies from the project. Vertx's handler has been now preferred and there are no more endpoints on RESTEasy in this project.


src/main/java/org/apache/cassandra/sidecar/routes/CassandraHealthService.java
```diff
@@ -1,104 +0,0 @@
...
-import org.apache.cassandra.sidecar.common.CassandraAdapterDelegate;
-import org.apache.cassandra.sidecar.utils.InstanceMetadataFetcher;
-
-/**
- * Provides a simple REST endpoint to determine if a node is available
- */
-@Singleton
-@Path("/api")
-@Produces(MediaType.APPLICATION_JSON)
-public class CassandraHealthService
```
frankgh, line 48: unfortunately, this was missed in CASSANDRASC-50. It should have been removed there

PR #54, opened 2023-06-22
CASSANDRASC-58: Support retries in Sidecar Client on Invalid Checksum

frankgh:
In rare occasions an SSTable upload will receive a corrupted SSTable. Bit flips are expected to occur occassionally while transmitting SSTables from client to server.

This commit adds support for retries in Sidecar Client when a checksum mismatch is encountered during SSTable upload. Allowing for clients to retry

### DIFF
--- a/common/src/main/java/org/apache/cassandra/sidecar/common/http/SidecarHttpResponseStatus.java
+++ b/common/src/main/java/org/apache/cassandra/sidecar/common/http/SidecarHttpResponseStatus.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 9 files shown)


common/src/main/java/org/apache/cassandra/sidecar/common/http/SidecarHttpResponseStatus.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package org.apache.cassandra.sidecar.common.http;
+
+import io.netty.handler.codec.http.HttpResponseStatus;
+
+/**
+ * Non-standard HTTP response status codes used in Sidecar
+ */
+public class SidecarHttpResponseStatus extends HttpResponseStatus
```
yifan-c, line 26: 👍

yifan-c (approved):
lgtm

PR #55, opened 2023-06-23
CASSANDRASC-59: Expose JMX host and port from JMXClient

frankgh:
In this commit, we expose the JMX host and port from JMXClient to make it available to implementations of the `ICassandraAdapter`.

This information can be valuable as different implementations and integrations need to have this information available during the application execution.


frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/a4ca352a1720744a2424c64c498f31b2460924e8

PR #56, opened 2023-06-27
CASSANDRASC-61: Add Release Audit Tool (RAT) plugin to Sidecar

frankgh:
This commit adds the Release Audit Tool (RAT) plugin to `build.gradle` which adds a new build task `rat`. This new task will make sure that the license headers are present in source files during the `check` task.

To run the RAT plugin, you can run:

```
./gradlew rat
```

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -4,3 +4,11 @@ Copyright 2023- The Apache Software Foundation
 This product includes software developed by The Apache Software
 Foundation (http://www.apache.org/).
 
+
+Eclipse Public License - v 1.0,GNU Lesser General Public License
+
+Logback Classic Module (1.2.3)
+logback-classic module
+
+Logback Core Module (1.2.3)
+logback-core module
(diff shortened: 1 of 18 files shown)


dineshjoshi (approved):
+1

NOTICE.txt
```diff
@@ -4,3 +4,11 @@ Copyright 2023- The Apache Software Foundation
 This product includes software developed by The Apache Software
 Foundation (http://www.apache.org/).
 
+
```
michaelsembwever: The format here is typically

product/project name
copyright
This project includes … 


More info on this at https://infra.apache.org/licensing-howto.html

frankgh: @michaelsembwever I think we don't need to add anything to NOTICE.txt after reviewing the link you sent, as well as what we do for Apache Cassandra. In fact, Cassandra has logback dependencies, but since they are not _bundled_, we don't need to add it to the notice.

From Cassandra's `build.xml`:
```
          <dependency groupId="ch.qos.logback" artifactId="logback-core" version="1.2.9"/>
          <dependency groupId="ch.qos.logback" artifactId="logback-classic" version="1.2.9"/>
```

michaelsembwever (approved):
+1

PR #57, opened 2023-06-27
CASSANDRASC-62: Fix the uncaught exception in client when processing response

yifan-c:


frankgh (approved):
+1. I couldn't find any other places where this might be an issue in the client

dineshjoshi:
+1

PR #58, opened 2023-06-28
CASSANDRASC-60 Adds endpoint to serve read/write replica-sets by toke…

arjunashok:
…n-ranges

### Update 08/18
- Updated endpoint response to include replica states
- Updated algorithm to resolve range intersections
- Added in-jvm dtest based integration tests for the sidecar endpoint

### Changes

This commit introduces a new sidecar endpoint (`token-range-replicas`) which exposes cluster topology information by keyspace, namely token ranges and endpoint mappings (including natural and pending ranges). 

The response consists of token ranges mapped to read and write replica-sets grouped by datacenter .  Write replica-sets will include mappings for pending ranges to nodes in transient states such as joining or leaving.

https://app.circleci.com/pipelines/github/arjunashok/cassandra-sidecar


frankgh (commented):
PR looks good in general, but would it make sense to add the endpoint to the client so it can be consumed by third-party libraries?

arjunashok:
> PR looks good in general, but would it make sense to add the endpoint to the client so it can be consumed by third-party libraries?

Yep, added to client in the next commit.

README.md
```diff
@@ -49,6 +49,19 @@ The build script supports two parameters:
...
+For multi-node in-jvm dtests, network aliases will need to be setup for each Cassandra node. The tests assume each node's 
+ip address is 127.0.0.x, where x is the node id. 
+
+For example if you populated your cluster with 3 nodes, create interfaces for 127.0.0.2 and 127.0.0.3 (the first node of course uses 127.0.0.1).
+
+### macOS network aliases
+To get up and running, create a temporary alias for every node except the first:
+
+```
+sudo ifconfig lo0 alias 127.0.0.2
```
frankgh: I think we have up to 10 for the existing tests, should we just tell devs to do this:
```suggestion
 for i in {2..10}; do sudo ifconfig lo0 alias "127.0.0.${i}"; done
```

src/test/integration/org/apache/cassandra/testing/TestVersionSupplier.java
```diff
@@ -35,7 +35,7 @@ public class TestVersionSupplier
     Stream<TestVersion> testVersions()
     {
         // By default, we test 2 versions that will exercise oldest and newest supported versions
-        String versions = System.getProperty("cassandra.sidecar.versions_to_test", "4.0,5.0");
+        String versions = System.getProperty("cassandra.sidecar.versions_to_test", "5.0");
```
arjunashok: Currently validating tests with both versions enabled. Will revert this change

src/test/integration/org/apache/cassandra/testing/CassandraTestTemplate.java
```diff
@@ -148,34 +149,21 @@ private BeforeEachCallback beforeEach()
...
-                                      {
-                                          config.with(Feature.GOSSIP);
-                                      }
-                                      if (annotation.network())
-                                      {
-                                          config.with(Feature.NETWORK);
-                                      }
-                                  });
+                UpgradeableCluster.build(originalNodeCount)
+//                                  .withDynamicPortAllocation(true) // to allow parallel test runs
```
arjunashok: Temporarily disabled parallel runs as it was not playing well with in-jvm dtest based integration tests

src/test/integration/org/apache/cassandra/sidecar/routes/TokenRangeIntegrationMovingTest.java
```diff
@@ -0,0 +1,232 @@
...
+                                                      }
+        );
+
+        runMovingTestScenario(context,
+                              BBHelperMovingNode.TRANSIENT_STATE_START,
+                              BBHelperMovingNode.TRANSIENT_STATE_END,
+                              cluster);
+    }
+
+    // TODO: Multiple replica-safe node movements in same DC, different DCs
```
arjunashok: Will remove and add more coverage for multiple movements

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
```diff
@@ -0,0 +1,472 @@
...
+            }
+            rangeConsumer.accept(next);
+            // when next is subset of current, we keep tracking current
+            current = (current.contains(next)) ? current : next;
+            next = null;
+        }
+        return next;
+    }
+
+    // TODO: Verify why we need all replicas from queue
```
arjunashok, line 432: Resolved. Will remove

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/Partitioner.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.math.BigInteger;
+
+/**
+ * Represents types of Partitioners supported and the corresponding starting token values
+ */
+public enum Partitioner
+{
+    Murmur3(BigInteger.valueOf(Long.MIN_VALUE), BigInteger.valueOf(Long.MAX_VALUE)),
+    Random(BigInteger.ONE.negate(), BigInteger.valueOf(2).pow(127).subtract(BigInteger.ONE));
```
frankgh: so looking at the Cassandra code, for Murmur3 the maximum is Long.MAX_VALUE (https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/dht/Murmur3Partitioner.java#L51). but for Random in Cassandra the maximum is `2^127` (https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/dht/RandomPartitioner.java#L53) , however here we declare it as `(2^127) - 1`. Any reason we subtract one for random here?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,249 @@
...
+    public TokenRangeReplicaProvider(JmxClient jmxClient)
+    {
+        this.jmxClient = jmxClient;
+    }
+
+    public TokenRangeReplicasResponse tokenRangeReplicas(String keyspace, Partitioner partitioner)
+    {
+        Objects.requireNonNull(keyspace, "keyspace must be non-null");
+
+        StorageJmxOperations storage = jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME);
```
frankgh: Refer to `RingProvider`, we need to provide an initialization method in case an extension class wishes to reuse the code with a custom StorageJmxOperation interface (i.e. Cassandra 3.11 implementation)
```suggestion
        StorageJmxOperations storage = initializeStorageOps();
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,249 @@
...
+
+        StateWithReplacement state = new StateWithReplacement(joiningNodes, leavingNodes, movingNodes, gossipInfo);
+
+        return replicaSet.stream()
+                         .collect(Collectors.toMap(Function.identity(), state::of));
+    }
+
+    private String getRawGossipInfo()
+    {
+        return jmxClient.proxy(ClusterMembershipJmxOperations.class, FAILURE_DETECTOR_OBJ_NAME)
```
frankgh, line 149: Similar comment to above, also make sure the initialization methods are protected so they can be overridden 
```suggestion
        return initializeClusterMembershipOps()
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,249 @@
...
+                                                                                         rep.end().toString(),
+                                                                                         replicasByDc);
+                                   })
+                                   .collect(Collectors.toList());
+    }
+
+    private Map<String, String> groupHostsByDatacenter(Set<String> replicaSet)
+    {
+        EndpointSnitchJmxOperations endpointSnitchInfo = jmxClient.proxy(EndpointSnitchJmxOperations.class,
+                                                                         ENDPOINT_SNITCH_INFO_OBJ_NAME);
```
frankgh: Similar comment
```suggestion
        EndpointSnitchJmxOperations endpointSnitchInfo = initializeEndpointProxy();
```

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -162,6 +163,23 @@ public CompletableFuture<TimeSkewResponse> timeSkew(List<? extends SidecarInstan
...
+     * @return a completable future of the token-range replicas
+     */
+    public CompletableFuture<TokenRangeReplicasResponse> tokenRangeReplicas(List<? extends SidecarInstance> instances,
+                                                                            String keyspace)
+    {
+        SidecarInstancesProvider instancesProvider = new SimpleSidecarInstancesProvider(instances);
+        InstanceSelectionPolicy instanceSelectionPolicy = new RandomInstanceSelectionPolicy(instancesProvider);
+        return executeRequestAsync(requestBuilder()
+                                   .instanceSelectionPolicy(instanceSelectionPolicy)
+                                   .tokenRangeReplicasRequest(keyspace).build());
```
frankgh: minor nit
```suggestion
                                   .tokenRangeReplicasRequest(keyspace)
                                   .build());
```

common/src/main/java/org/apache/cassandra/sidecar/common/CassandraAdapterDelegate.java
```diff
@@ -240,21 +240,20 @@ public boolean isUp()
...
-        nodeSettings = null;
         try
         {
             jmxClient.close();
         }
         catch (IOException e)
         {
             // Can't throw unchecked exceptions here, so wrap and rethrow
-            throw new RuntimeException(e);
+            LOGGER.warn("Unable to close JMX client", e);
```
frankgh, line 255: update comment above, since we are not re-throwing the exception anymore

common/src/main/java/org/apache/cassandra/sidecar/common/data/GossipInfoResponse.java
```diff
@@ -53,6 +53,33 @@
  */
 public class GossipInfoResponse extends HashMap<String, GossipInfoResponse.GossipInfo>
 {
+    /**
+     * Overrides the {@link #get(Object)} method. The gossip info keys usually start with the format
+     * {@code /ip:port}. Some clients may be unaware of the preceding {@code slash}, and lookups can
```
frankgh, line 58: this bug is exhibited in analytics library as seen here: https://github.com/apache/cassandra-analytics/blob/trunk/cassandra-analytics-core/src/main/java/org/apache/cassandra/spark/bulkwriter/CassandraClusterInfo.java#L523

common/src/main/java/org/apache/cassandra/sidecar/common/data/TokenRangeReplicasResponse.java
```diff
@@ -0,0 +1,157 @@
...
+     * @return returns replica to it's state information mapping
+     */
+    @JsonProperty("replicaState")
+    public Map<String, String> replicaState()
+    {
+        return replicaState;
+    }
+
+    /**
+     * @return returns the {@link ReplicaInfo} instances representing write replicas for each token range
```
frankgh:
```suggestion
     * @return the {@link ReplicaInfo} instances representing write replicas for each token range
```

common/src/main/java/org/apache/cassandra/sidecar/common/data/TokenRangeReplicasResponse.java
```diff
@@ -0,0 +1,157 @@
...
+                                      @JsonProperty("writeReplicas") List<ReplicaInfo> writeReplicas,
+                                      @JsonProperty("readReplicas") List<ReplicaInfo> readReplicas)
+    {
+        this.replicaState = replicaState;
+        this.writeReplicas = writeReplicas;
+        this.readReplicas = readReplicas;
+    }
+
+    /**
+     * @return returns replica to it's state information mapping
```
frankgh:
```suggestion
     * @return the replica to state information mapping
```

common/src/main/java/org/apache/cassandra/sidecar/common/data/TokenRangeReplicasResponse.java
```diff
@@ -0,0 +1,157 @@
...
+     * @return returns the {@link ReplicaInfo} instances representing write replicas for each token range
+     */
+    @JsonProperty("writeReplicas")
+    public List<ReplicaInfo> writeReplicas()
+    {
+        return writeReplicas;
+    }
+
+    /**
+     * @return returns the {@link ReplicaInfo} instances representing read replicas for each token range
```
frankgh:
```suggestion
     * @return the {@link ReplicaInfo} instances representing read replicas for each token range
```

src/main/java/org/apache/cassandra/sidecar/routes/TokenRangeReplicaMapHandler.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    protected void processFailure(Throwable cause, RoutingContext context, String host, SocketAddress remoteAddress,
+                                  TokenRangeReplicasRequest request)
+    {
+        if (cause instanceof AssertionError &&
+            StringUtils.contains(cause.getMessage(), "Unknown keyspace"))
+        {
+            context.fail(new HttpException(HttpResponseStatus.NOT_FOUND.code(), cause.getMessage()));
```
frankgh: nit, since we have these helper methods:
```suggestion
            context.fail(HttpExceptions.wrapHttpException(HttpResponseStatus.NOT_FOUND, cause.getMessage()));
```

src/main/java/org/apache/cassandra/sidecar/routes/TokenRangeReplicaMapHandler.java
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        executorPools.service().executeBlocking(promise -> {
+            try
+            {
+                context.json(storageOperations.tokenRangeReplicas(request.keyspace(), metadata.getPartitioner()));
+            }
+            catch (UnknownHostException e)
+            {
+                processFailure(e, context, host, remoteAddress, request);
```
frankgh: it looks like we never throw the `UnknownHostException` exception, we are wrapping it in `RuntimeException`, should we catch that instead here?

frankgh (commented):
Added some more comments

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/Partitioner.java
yifan-c: Second to this. The value range of random partition token should be the same as in Cassandra.

Besides that, can you add a brief description to clarify the range's ends. The minimum token is not inclusive, and the maximum token is. In math notion, it is `(MIN, MAX]`

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
yifan-c: Not an AI, but maybe refactor the code in a follow-up, so that `jmxClient` is not available in the scope to prevent calling `jmxClient.proxy` directly.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,249 @@
...
+        List<TokenRangeReplicas> replicas = Stream.concat(
+                                                  naturalReplicaMappings.entrySet().stream(),
+                                                  pendingRangeMappings.entrySet().stream())
+                                                  .map(entry -> TokenRangeReplicas.generateTokenRangeReplicas(
+                                                  new BigInteger(entry.getKey().get(0)),
+                                                  new BigInteger(entry.getKey().get(1)),
+                                                  partitioner,
+                                                  new HashSet<>(entry.getValue())))
+                                                  .flatMap(Collection::stream)
+                                                  .collect(Collectors.toList());
```
yifan-c: nit: add some indentation to make it easier to read

```suggestion
        List<TokenRangeReplicas> replicas = Stream.concat(naturalReplicaMappings.entrySet().stream(), 
                                                          pendingRangeMappings.entrySet().stream())
                                                  .map(entry -> TokenRangeReplicas.generateTokenRangeReplicas(
                                                      new BigInteger(entry.getKey().get(0)),
                                                      new BigInteger(entry.getKey().get(1)),
                                                      partitioner,
                                                      new HashSet<>(entry.getValue())))
                                                  .flatMap(Collection::stream)
                                                  .collect(Collectors.toList());
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,249 @@
...
+                                  Map<String, String> hostToDatacenter,
+                                  Partitioner partitioner)
+    {
+        return replicasByTokenRange.entrySet().stream()
+                                   .map(entry -> TokenRangeReplicas.generateTokenRangeReplicas(
+                                   new BigInteger(entry.getKey().get(0)),
+                                   new BigInteger(entry.getKey().get(1)),
+                                   partitioner,
+                                   new HashSet<>(entry.getValue())))
+                                   .flatMap(Collection::stream)
```
yifan-c: nit: readability
```suggestion
        return replicasByTokenRange.entrySet().stream()
                                   .map(entry -> TokenRangeReplicas.generateTokenRangeReplicas(
                                       new BigInteger(entry.getKey().get(0)),
                                       new BigInteger(entry.getKey().get(1)),
                                       partitioner,
                                       new HashSet<>(entry.getValue())))
                                   .flatMap(Collection::stream)
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,249 @@
...
+     * do this we analyze gossip info and set 'Replacing' state for node replacing a dead node.
+     * {@link StateWithReplacement} is used to set replacing state for a node.
+     *
+     * <p>We are adding this state for token range replica provider endpoint. To send out replicas for a
+     * range along with state of replicas including replacing state.
+     */
+    static class StateWithReplacement extends RingProvider.State
+    {
+        private static final String STATE_REPLACING = "Replacing";
+        private final Set<String> joiningNodes;
```
yifan-c: Why does it declare another copy of `joiningNodes`? It is available from the base class (but need to make it protected or add getter).

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
JeetKunDoug: Same (make things protected) goes for any members that need to be accessible by the derived class (or provide a protected getter if you feel that's necessary, but I think it's mostly the `private final JmxClient jmxClient`, which would be OK to be protected).

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    public TokenRangeReplicasResponse tokenRangeReplicas(String keyspace, Partitioner partitioner)
+    {
+        Objects.requireNonNull(keyspace, "keyspace must be non-null");
+
+        StorageJmxOperations storage = jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME);
+
+        // Retrieve map of primary token ranges to endpoints that describe the ring topology
+        Map<List<String>, List<String>> rangeToEndpointMappings = storage.getRangeToEndpointWithPortMap(keyspace);
```
JeetKunDoug: General comment about this line of code + most of the rest of this class - It's _really hard_ to figure out what's what in this class with everything being maps of native types - I realize that we can't change the return types of the JMX endpoints, but can we please try to transform things as soon as possible after we get them to something more meaningful types (Datacenter, Range, ReplicaInfo, Token, some of which I realize we don't have yet)? It would make all of this logic significantly easier to understand, and prevent bugs in the future.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
JeetKunDoug: The indentation/formatting here is horrible to follow - I realize this is the default the project uses, but is there a way we can improve this, as it's really quite difficult to understand what's at what level of the concat/map/etc. here...
Alternately, break this down into smaller chunks and assign each to an intermediate variable with a good name.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,249 @@
...
+                                       Map<String, List<String>> replicasByDc =
+                                       replicasByDataCenter(hostToDatacenter, rep.replicaSet());
+                                       return new TokenRangeReplicasResponse.ReplicaInfo(rep.start().toString(),
+                                                                                         rep.end().toString(),
+                                                                                         replicasByDc);
+                                   })
+                                   .collect(Collectors.toList());
+    }
+
+    private Map<String, String> groupHostsByDatacenter(Set<String> replicaSet)
```
JeetKunDoug: This method is misnamed - it doesn't group anything, it just enriches the host with a datacenter - maybe just `buildHostToDatacenterMapping`?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,249 @@
...
+        @Override
+        String of(String endpoint)
+        {
+            if (joiningNodes.contains(endpoint))
+            {
+                GossipInfoResponse.GossipInfo gossipInfoEntry = gossipInfo.get(endpoint);
+
+                if (gossipInfoEntry != null)
+                {
+                    LOGGER.info("Found gossipInfoEntry={}", gossipInfoEntry);
```
JeetKunDoug: NIT: Debug?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
```diff
@@ -0,0 +1,472 @@
...
+    }
+
+    /**
+     * For subset ranges, this is used to determine if a range is larger than the other by comparing start-end lengths
+     * If both ranges end at the min, we compare starting points to determine the result.
+     * When the left range is the only one ending at min, it is always the larger one since all subsequent ranges
+     * in the sorted range list have to be smaller.
+     * <p>
+     * This method assumes that the ranges are normalized and unwrapped, i.e.
+     * 'this' comes before 'other' AND there's no wrapping around the min token
```
JeetKunDoug: This assumption should be checked with a precondition, or we should write the method to not care about if `this` comes before `other`.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
yifan-c: nit: should we only print the endpoint and status? Those 2 values should be enough to reason about the runtime here.

JeetKunDoug (changes requested):
Partial review but wanted to get out what I could now - will continue later.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
```diff
@@ -0,0 +1,472 @@
...
+        return Objects.hash(start, end, partitioner);
+    }
+
+    private void validateRangesForComparison(@NotNull TokenRangeReplicas other)
+    {
+        if (this.partitioner != other.partitioner)
+            throw new IllegalStateException("Token ranges being compared do not have the same partitioner");
+    }
+
+    protected boolean contains(TokenRangeReplicas other)
```
yifan-c: The `protected` methods are not for extension, since there is no subclass. Please not use `protected`.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,249 @@
...
+     * We want to identity a joining node, to replace a dead node, differently from a newly joining node. To
+     * do this we analyze gossip info and set 'Replacing' state for node replacing a dead node.
+     * {@link StateWithReplacement} is used to set replacing state for a node.
+     *
+     * <p>We are adding this state for token range replica provider endpoint. To send out replicas for a
+     * range along with state of replicas including replacing state.
+     */
+    static class StateWithReplacement extends RingProvider.State
+    {
+        private static final String STATE_REPLACING = "Replacing";
```
yifan-c: How about create an enum for the state? It helps in 2 ways, strong typed and comprehensive matching.

adapters/base/src/test/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicasTest.java
```diff
@@ -0,0 +1,847 @@
...
+    public void simpleTest()
+    {
+        List<TokenRangeReplicas> simpleList = createSimpleTokenRangeReplicaList();
+        LOGGER.info("Input:" + simpleList);
+        List<TokenRangeReplicas> rangeList = TokenRangeReplicas.normalize(simpleList);
+        LOGGER.info("Result:" + rangeList);
+        assertThat(hasOverlaps(rangeList)).isFalse();
+    }
+
+    // TODO: Validate unwrapping separately
```
yifan-c: Not sure if it is addressed by `wrappedMultiOverlapTest` already.

build.gradle
```diff
@@ -326,6 +326,8 @@ tasks.register("integrationTest", Test) {
     useJUnitPlatform() {
         includeTags "integrationTest"
     }
+// Uncomment below to run unit tests in parallel
+//    maxParallelForks = Runtime.runtime.availableProcessors() * 2
```
yifan-c: Just curious, why it is commented out?

common/build.gradle
```diff
@@ -41,6 +41,8 @@ repositories {
 
 test {
     useJUnitPlatform()
+// Uncomment below to run unit tests in parallel
+//    maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
```
yifan-c: Another curiosity, why the `maxParallelForks` calculations are different in different build files?

common/src/main/java/org/apache/cassandra/sidecar/common/CQLSessionProvider.java
```diff
@@ -126,8 +126,15 @@ public synchronized Session localCql()
...
-    public synchronized void close()
+    public Session close()
     {
+        Session localSession;
+        synchronized (this)
+        {
+            localSession = this.localSession;
+            this.localSession = null;
+        }
+
```
yifan-c, line 137: 👍

common/src/main/java/org/apache/cassandra/sidecar/common/JmxClient.java
```diff
@@ -259,12 +259,16 @@ private Map<String, Object> buildJmxEnv()
     }
 
     @Override
-    public synchronized void close() throws IOException
+    public void close() throws IOException
     {
-        JMXConnector connector = jmxConnector;
-        if (connector != null)
+        JMXConnector connector;
+        synchronized (this)
```
yifan-c, line 329: Can you make `connect()` (where `jmxConnector` is update) synchronized? It is synchronized in its only callsite. It won't hurt performance when synchronizing `connnect()`, but more error-proof.

common/src/main/java/org/apache/cassandra/sidecar/common/JmxClient.java
```diff
@@ -259,12 +259,16 @@ private Map<String, Object> buildJmxEnv()
...
         {
+            connector = jmxConnector;
             jmxConnector = null;
```
yifan-c, line 332: Reset `connected` to false too?

src/main/java/org/apache/cassandra/sidecar/routes/TokenRangeReplicaMapHandler.java
yifan-c: right. The method in the interface is declared to throw `UnknownHostException`, but the implementation does not. It is wrapped in `RuntimeException`. Can you update the interface method to remove the throws? 
The `RuntimeException` does not need to be handled separately, it reaches to `onFailure`, which calls `processFailure`, if throws.

src/test/integration/org/apache/cassandra/sidecar/routes/BaseTokenRangeIntegrationTest.java
```diff
@@ -0,0 +1,795 @@
...
+        return expectedRanges;
+    }
+
+    protected Set<String> instancesFromReplicaSet(List<TokenRangeReplicasResponse.ReplicaInfo> replicas)
+    {
+        return replicas.stream()
+                       .map(r -> r.replicasByDatacenter().values())
+                       .flatMap(Collection::stream)
+                       .flatMap(Collection::stream)
+                       .collect(Collectors.toSet());
```
yifan-c: nit: one less line if you do this 
```suggestion
        return replicas.stream()
                       .flatMap(r -> r.replicasByDatacenter().values().stream())
                       .flatMap(Collection::stream)
                       .collect(Collectors.toSet());
```

src/test/integration/org/apache/cassandra/sidecar/routes/BaseTokenRangeIntegrationTest.java
```diff
@@ -0,0 +1,795 @@
...
+            Uninterruptibles.awaitUninterruptibly(TRANSIENT_STATE_END);
+            orig.call();
+        }
+    }
+
+    /**
+     * ByteBuddy helper for multiple leaving nodes
+     */
+    @Shared
+    public static class BBHelperMultipleLeavingNodes
```
yifan-c: Each BBHelper class is used in a single test class, and their implementations (install) highly depends on the test class, e.g. nodeNumber assumption. Therefore, it makes it hard to share the implementation with other test classes. How about move it to the test class where it is actually used?

src/test/integration/org/apache/cassandra/sidecar/routes/MultiDcTokenSupplier.java
```diff
@@ -0,0 +1,64 @@
...
+
+import org.apache.cassandra.distributed.api.TokenSupplier;
+
+/**
+ * Static factory holder that provides a multi-DC token supplier
+ */
+public class MultiDcTokenSupplier
+{
+
+    static TokenSupplier evenlyDistributedTokens(int numNodes, int numDcs, int numTokens)
```
yifan-c: I guess you mean this
```suggestion
    static TokenSupplier evenlyDistributedTokens(int numNodesPerDc, int numDcs, int numTokensPerNode)
```

Then, what does the result of `numTokensPerNode * numDcs` mean?

src/test/integration/org/apache/cassandra/sidecar/routes/MultiDcTokenSupplier.java
```diff
@@ -0,0 +1,64 @@
...
+    {
+        long totalTokens = (long) numNodes * numDcs * numTokens;
+        BigInteger increment = BigInteger.valueOf((Long.MAX_VALUE / totalTokens) * 4);
```
yifan-c: Where is `4` coming from?

src/test/resources/logback-test.xml
```diff
@@ -20,7 +20,7 @@
 
   <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
     <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
-      <level>DEBUG</level>
+      <level>INFO</level>
```
yifan-c: I think `DEBUG` is more suitable for tests.

yifan-c (commented):
Thanks for the patch! I left questions and suggestions.

build.gradle
JeetKunDoug: It requires a Cassandra fix (https://issues.apache.org/jira/browse/CASSANDRA-18722) that isn't yet committed.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: As you have called out, the native types are the return types of the JMX endpoints. I am transforming this into a holder `TokenRangeReplicas` further in the flow. Will look into moving it ahead.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: Will make the change similar to how RingHandler does this for consistency

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: Will update. This method was grouping by DC is a previous iteration which I did not rename when modified.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: Always in favor of using enums for grouped constants such as this case. Will update.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: Switching to debug log(per the above comment), as we don't really need this logged at info level.

build.gradle
arjunashok: Parallel runs, when enabled do not seem to play well with in-jvm dtests. Since this was introduced in one of the prior commits in this PR, I will remove these commented lines for now. We can re-introduce it as an optimization.

src/main/java/org/apache/cassandra/sidecar/routes/TokenRangeReplicaMapHandler.java
arjunashok: Will update.

src/test/integration/org/apache/cassandra/sidecar/routes/MultiDcTokenSupplier.java
arjunashok: So, we perform `numNodes * numDcs * numTokensPerNode` to compute the total no. tokens across all DCs and nodes

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
arjunashok: Makes sense. Seems like we're no longer using this method with the new token-range intersection resolution algorithm. Will remove.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
```diff
@@ -0,0 +1,472 @@
...
+     * @return true if "this" range intersects the other
+     */
+    protected boolean intersects(TokenRangeReplicas other)
+    {
+        if (this.compareTo(other) > 0)
+            throw new IllegalStateException(
+            String.format("Token ranges - (this:%s other:%s) are not ordered", this, other));
+
+        return this.end.compareTo(other.start) > 0 && this.start.compareTo(other.end) < 0; // Start exclusive (DONE)
+    }
```
JeetKunDoug: Rather then throwing if this is < other, how about just reversing the comparison in the return statement:

```suggestion
    protected boolean intersects(TokenRangeReplicas other)
    {
        boolean inOrder = this.compareTo(other) <= 0;
        TokenRangeReplicas first = inOrder ? this : other;
        TokenRangeReplicas last = inOrder ? other : this;

        return first.end.compareTo(last.start) > 0 && first.start.compareTo(last.end) < 0; // Start exclusive (DONE)
    }
```

common/build.gradle
arjunashok: This was introduced by @frankgh in a prior commit to my branch, so I will let him respond. Leaving it out of this commit for now.

common/src/main/java/org/apache/cassandra/sidecar/common/JmxClient.java
arjunashok: Good catch. This would need to be reset to false to prevent false-positives from `checkConnection`. Will add.

src/test/integration/org/apache/cassandra/sidecar/routes/MultiDcTokenSupplier.java
arjunashok: This was hard-coded to account for the no. DCs. Will update to be as folllows. `2` is as-is from the existing `TokenSupplier` (to account for the token-space being from Long.MIN - Long.MAX).

`(Long.MAX_VALUE / totalTokens)* 2 * numDCs`

We broaden the increment space by numDcs as the tokens assigned to neighboring nodes across different DCs are consecutive (i.e. delta 1).

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: Agreed. Will update.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
arjunashok: Yep, not needed. Will switch to package-private for tests

adapters/base/src/test/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicasTest.java
arjunashok: Yep this is addressed. Will remove

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
yifan-c: ~The formatting was not done yet~

GH page displayed stale code.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,263 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    @NotNull
+    private static Map<String, List<String>> replicasByDataCenter(Map<String, String> hostToDatacenter,
+                                                                  Collection<String> replicas)
+    {
+        return replicas.stream().collect(Collectors.groupingBy(hostToDatacenter::get,
+                                                               Collectors.filtering(replicas::contains, toList())));
```
yifan-c: The `adapter` subproject declare source compatibility of 1.8, but `Collectors.filtering` is available after java 9.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,263 @@
...
+                                                                       new BigInteger(entry.getKey().get(1)),
+                                                                       partitioner,
+                                                                       new HashSet<>(entry.getValue())))
+                              .flatMap(Collection::stream)
+                              .collect(toList());
+    }
+
+
+    private List<ReplicaInfo> readReplicasFromReplicaMapping(List<TokenRangeReplicas> naturalTokenRangeReplicas,
+                                                             Map<String, String> hostToDatacenter)
```
yifan-c, line 163: unused variable.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,263 @@
...
+    protected String getRawGossipInfo()
+    {
+        return jmxClient.proxy(ClusterMembershipJmxOperations.class, FAILURE_DETECTOR_OBJ_NAME)
+                        .getAllEndpointStatesWithPort();
+    }
+
+    private List<ReplicaInfo> writeReplicasFromPendingRanges(List<TokenRangeReplicas> tokenRangeReplicaSet,
+                                                             Map<String, String> hostToDatacenter)
+    {
+//        Map<String, String> hostToDatacenter = buildHostToDatacenterMapping(tokenRangeReplicaSet);
```
yifan-c: rm this?

yifan-c (commented):
Thanks for addressing the comments. Just spot a few oversights.

src/test/integration/org/apache/cassandra/testing/CassandraTestTemplate.java
frankgh: we should be able to leverage this feature now. https://issues.apache.org/jira/browse/CASSANDRA-18722 has been merged

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/NodeInfo.java
```diff
@@ -0,0 +1,69 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.adapters.base;
+
+/**
+ * Holder class for Node related
+ */
+public class NodeInfo
```
frankgh, line 26: I am +0 on this change, there's nothing that we leverage from the enum effectively

src/test/resources/logback-test.xml
```diff
@@ -20,7 +20,7 @@
 
   <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
     <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
-      <level>INFO</level>
+      <level>DEBUG</level>
```
frankgh: I think INFO is fine, the tests already print way too much output

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/NodeInfo.java
arjunashok: Like I said earlier, I am in favor of enums (not necessarily in this holder class) for improved readability and maintainability of related constants, which in this case, relates to node states and statuses. 

This is how Cassandra manages these constants as well
https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/service/StorageService.java#L423

@frankgh - Do you see any reasons to not go down this route?

src/test/resources/logback-test.xml
arjunashok: I don't have a strong opinion on this, can revert since these are tests.

@yifan-c - Does this make sense, since you recommended this change?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/Partitioner.java
arjunashok: Updated

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/NodeInfo.java
frankgh: I just don't see the benefit, I'm not against or for it, I'm neutral. I would not use an enum if we are just providing constants. This is what Cassandra does for ring: https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/tools/nodetool/Ring.java#L172

src/test/integration/org/apache/cassandra/testing/CassandraTestTemplate.java
arjunashok: Thanks. will update and enable parallel runs

common/build.gradle
arjunashok: @frankgh - Can you comment on @yifan-c 's question, since this was introduced in your change?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: Addressed

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: Addressed

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: Addressed

src/test/resources/logback-test.xml
yifan-c: Typically `Debug` is used when running test. What are the other scenarios of using debug? It is (almost) never considered in production. So.. if we do not really enable debug level, what is the point of having those debug message in the code :D
I think the way too many log messages is a different issue. Some useless chatty logger can be disabled or set to INFO. 
We need to put more thoughts. For now, I am fine with "INFO" if Francisco feels strong about it.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/NodeInfo.java
JeetKunDoug: The `nodetool/Ring.java` is producing _human readable_ output though, and otherwise I think Arjun's `Mode` example in C* seems more appropriate. I can imagine other usage of the enum in future work potentially, but more importantly it provides a  way to group all of the things in one place. You could, I suppose, just have a bunch of `public static final String` members on an otherwise static class just to have another container for the constant strings...

We could, however, do something like this:
```java
public enum NodeState
{
    JOINING,
    LEAVING,
    MOVING,
    NORMAL,
    REPLACING;

    private final String displayName;

    NodeState()
    {
        String firstChar = String.valueOf(name().charAt(0)).toLowerCase(Locale.ROOT);
        displayName = name().toLowerCase().replaceFirst(firstChar, firstChar.toUpperCase(Locale.ROOT));
    }

    public String displayName() {
        return displayName;
    }
}
```
And then you can use `displayName` where you need the human-readable version.

This gets rid of the hard-coding of string names, but maybe is otherwise more complicated - you could also just have a helper method somewhere that does the enum -> Title case conversion.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -0,0 +1,269 @@
...
+
+    public TokenRangeReplicasResponse tokenRangeReplicas(String keyspace, Partitioner partitioner)
+    {
+        Objects.requireNonNull(keyspace, "keyspace must be non-null");
+
+        StorageJmxOperations storage = initializeStorageOps();
+
+        // Retrieve map of primary token ranges to endpoints that describe the ring topology
+        Map<List<String>, List<String>> naturalReplicaMappings = storage.getRangeToEndpointWithPortMap(keyspace);
+        LOGGER.debug("Natural token range mappingsfor keyspace={}, pendingRangeMappings={}",
```
JeetKunDoug: NIT: missing space in comment
```suggestion
        LOGGER.debug("Natural token range mappings for keyspace={}, pendingRangeMappings={}",
```

adapters/base/src/test/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicasTest.java
```diff
@@ -0,0 +1,824 @@
...
+        }
+        return false;
+    }
+
+    private boolean checkContains(List<TokenRangeReplicas> resultList, TokenRangeReplicas expected)
+    {
+        return resultList.stream()
+                         .map(TokenRangeReplicas::toString)
+                         .anyMatch(r -> r.equals(expected.toString()));
+    }
```
JeetKunDoug: NIT: This is a personal preference of mine, but I like to see the _tests_ first in a test class, and helper methods at the end - these are details that I generally don't need to see until I'm trying to debug a failure or if the method doesn't have a good enough name (in which case, we should rename it).

adapters/base/src/test/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicasTest.java
```diff
@@ -0,0 +1,824 @@
...
+
+    // non-overlapping ranges
+    @Test
+    public void simpleTest()
+    {
+        List<TokenRangeReplicas> simpleList = createSimpleTokenRangeReplicaList();
+        LOGGER.info("Input:" + simpleList);
+        List<TokenRangeReplicas> rangeList = TokenRangeReplicas.normalize(simpleList);
```
JeetKunDoug, line 48: Given all this test does is assert that there are no overlaps in the final list, we should at least generate an input list that has some overlaps, or is there some other thing we could validate here, like the input list and output list are equal (because there should have been no change?) I think this works/passes, and would be a better assertion than just there were no overlaps (because there were no overlaps in the input), although maybe you do both?

JeetKunDoug (commented):
Just submitting comments so far - will submit individual comments ongoing so you can respond without waiting for a complete review.

src/test/resources/logback-test.xml
frankgh: yeah, let's keep info for now until we make logging more significant for tests

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
```diff
@@ -255,6 +256,31 @@ public void testTimeSkewFromReplicaSet() throws Exception
...
+                                            "{\"datacenter1\":[\"127.0.0.1:7000\"]}}],\"readReplicas\":" +
+                                            "[{\"start\":\"-9223372036854775808\",\"end\":\"9223372036854775807\"," +
+                                            "\"replicasByDatacenter\":{\"datacenter1\":[\"127.0.0.1:7000\"]}}]}";
+        MockResponse response = new MockResponse().setResponseCode(OK.code()).setBody(tokenRangeReplicasAsString);
+        enqueue(response);
+
+        TokenRangeReplicasResponse result = client.tokenRangeReplicas(instances.subList(1, 2), keyspace)
+                                                  .get(30, TimeUnit.SECONDS);
+        assertThat(result).isNotNull();
+        assertThat(result.writeReplicas()).hasSize(1);
```
JeetKunDoug, line 347: As we've discussed in other places - can we validate that the data the mock server responded with was actually what we got? It's good to make sure there aren't any JSON serialization issues by actually validating everything here rather than just the shape of the response.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
arjunashok: Apart from here there are other operations (eg. `processIntersectingRanges`) that rely on the ranges being processed to be sorted and breaks if the prerequisite is not met. Although that should "never" happen since we are explicitly sorting and unwrapping upfront. 

Like you mentioned earlier, I'm leaning towards performing the assertions on the assumptions up-front, if for any reason the ranges are not what we expect them to be.

arjunashok: Another option would be to perform the assertion within `processIntersectingRanges` instead and do what you're suggesting in the `intersects()` method. That way, we only fail in the event there are intersections (and the ranges are out of order).

adapters/base/src/test/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicasTest.java
arjunashok: Yep agree, I'd blame IntelliJ's insertion ordering :)  Addressed

adapters/base/src/test/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicasTest.java
arjunashok: Yep, will add validation for input/output lists match.

The intention behind the current check is that `normalize` does not do anything bad when there are no overlaps as the simplest case of the test, with other test cases covering the overlapping/sub-set etc. scenarios.

src/test/integration/org/apache/cassandra/sidecar/routes/BaseTokenRangeIntegrationTest.java
```diff
@@ -0,0 +1,257 @@
...
+                                                                                        new BigInteger(r.end())))
+                                                             .collect(Collectors.toList());
+
+        List<Range<BigInteger>> readRanges = readReplicaSet.stream()
+                                                           .map(r -> Range.openClosed(new BigInteger(r.start()),
+                                                                                      new BigInteger(r.end())))
+                                                           .collect(Collectors.toList());
+
+
+        assertThat(writeRanges.size()).isEqualTo(writeReplicaSet.size());
```
JeetKunDoug: NIT: The size comparison isn't necessary given `containsExactlyElementsOf`"verifies that actual contains exactly the elements of the given iterable and nothing else in the same order."

src/test/integration/org/apache/cassandra/sidecar/routes/BaseTokenRangeIntegrationTest.java
```diff
@@ -0,0 +1,257 @@
...
+                       .flatMap(r -> r.replicasByDatacenter().values().stream())
+                       .flatMap(Collection::stream)
+                       .collect(Collectors.toSet());
+    }
+
+    protected void validateWriteReplicaMappings(List<TokenRangeReplicasResponse.ReplicaInfo> writeReplicas,
+                                              Map<String, Map<Range<BigInteger>, List<String>>> expectedRangeMapping)
+    {
+        CassandraIntegrationTest annotation = sidecarTestContext.cassandraTestContext().annotation;
+        assertThat(writeReplicas).hasSize(expectedRangeMapping.get("datacenter1").size());
```
JeetKunDoug: Having trouble wrapping my head around why writeReplicas would always be the size of only the first dc here - can you clarify (maybe add a comment) as to why this is the case?

README.md
JeetKunDoug: Note: There are tests in the multi-dc scenarios that go up to .12, so please increase the end of the range to at least 12 (maybe 20?).

src/test/integration/org/apache/cassandra/sidecar/routes/TokenRangeIntegrationMultiDCBasicTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+import io.vertx.junit5.VertxExtension;
+import io.vertx.junit5.VertxTestContext;
+import org.apache.cassandra.sidecar.common.data.TokenRangeReplicasResponse;
+import org.apache.cassandra.testing.CassandraIntegrationTest;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test the token range replica mapping endpoint with cassandra container.
```
JeetKunDoug: NIT: `with cassandra container` isn't true any more - we should replace this comment everywhere it appears with `with the in-jvm dtest framework`

src/test/integration/org/apache/cassandra/sidecar/routes/TokenRangeIntegrationLeavingTest.java
```diff
@@ -0,0 +1,786 @@
...
+                                UpgradeableCluster cluster,
+                                Map<String, Map<Range<BigInteger>, List<String>>> expectedRangeMappings)
+    throws Exception
+    {
+        try
+        {
+            CassandraIntegrationTest annotation = sidecarTestContext.cassandraTestContext().annotation;
+            Set<String> dcReplication;
+
+            if (annotation.numDcs() > 1)
```
JeetKunDoug: This logic is repeated a bunch of times - maybe move to the integration test base?

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
arjunashok: Addressed

src/test/integration/org/apache/cassandra/testing/TestVersionSupplier.java
```diff
@@ -35,7 +35,7 @@ public class TestVersionSupplier
     Stream<TestVersion> testVersions()
     {
         // By default, we test 2 versions that will exercise oldest and newest supported versions
-        String versions = System.getProperty("cassandra.sidecar.versions_to_test", "4.0,5.0");
+        String versions = System.getProperty("cassandra.sidecar.versions_to_test", "5.1");
```
JeetKunDoug: Before we merge, please make this "4.0,5.1" or "4.1,5.1" so we don't lose 4.X testing

src/test/integration/org/apache/cassandra/sidecar/routes/BaseTokenRangeIntegrationTest.java
arjunashok: This validates the number of ranges match the expected no. ranges. Will add comment

src/test/integration/org/apache/cassandra/sidecar/routes/BaseTokenRangeIntegrationTest.java
arjunashok: yep, redundant and not required. will remove

README.md
arjunashok: Addressed

arjunashok:
Changes were merged into trunk. Closing

PR #59, opened 2023-06-29
CASSANDRASC-63: Support credential rotation in JmxClient

frankgh:
JMX credentials in a Cassandra instance can be rotated on a cadence, on every bounce, or by some other means. In those cases, the `JmxClient` will no longer be able to connect to the instance completely losing the ability to talk to that instance.

In this commit, we allow the `JmxClient` to support credential changes to be continue to talk to the Cassandra instance uninterrupted without any potential downtime to the Sidecar service.

### DIFF
--- a/common/src/main/java/org/apache/cassandra/sidecar/common/JmxClient.java
+++ b/common/src/main/java/org/apache/cassandra/sidecar/common/JmxClient.java
@@ -25,6 +25,9 @@ import java.rmi.server.RMIClientSocketFactory;
 import java.rmi.server.RMISocketFactory;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
+import java.util.function.BooleanSupplier;
+import java.util.function.Supplier;
 import javax.management.JMX;
 import javax.management.MBeanServerConnection;
 import javax.management.MalformedObjectNameException;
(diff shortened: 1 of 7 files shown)


common/src/main/java/org/apache/cassandra/sidecar/common/JmxClient.java
```diff
@@ -158,6 +162,14 @@ private void connect()
...
                                        iox);
         }
+        catch (SecurityException securityException)
+        {
+            connected = false;
+            String errorMessage = securityException.getMessage() != null
+                                  ? securityException.getMessage()
+                                  : "JMX Authentication failed";
+            throw new JmxAuthenticationException(errorMessage, securityException);
+        }
```
yifan-c, line 172: Theoretically, it also needs to catch `RuntimeException` since it _could_ be thrown from the suppliers.

frankgh: I think the desired behavior is to let `RuntimeException`s bubble up and be handled up the stack. We actually throw a `RuntimeException` on line 162, and that will bubble up

yifan-c: Talked offline. The thrown runtime exception needs to be wrapped in a dedicated exception type to avoid being caught and handled unexpectedly.

PR #60, opened 2023-07-11
CASSANDRASC-64: File descriptor is not being closed on MD5 checksum

frankgh:
This commit fixes an issue where the file descriptors are not being closed during the MD5 checksum during SSTable upload. This can potentially cause the JVM to run out of available file descriptors during execution, and it would force an operator to restart the Sidecar process to return to a working state.

In this commit, we ensure the file is closed after the checksum is calculated.

Additionally, this commit fixes a rare ConcurrentModificationException encountered in `SSTableImporter` whree the `importQueuePerHost` does not use a thread-safe map.

### DIFF
--- a/src/main/java/org/apache/cassandra/sidecar/utils/HttpExceptions.java
+++ b/src/main/java/org/apache/cassandra/sidecar/utils/HttpExceptions.java
@@ -54,7 +54,7 @@ public class HttpExceptions
      */
     public static HttpException wrapHttpException(HttpResponseStatus status, Throwable cause)
     {
-        return wrapHttpException(status, cause != null ? cause.getMessage() : null, cause);
+        return wrapHttpException(status, null, cause);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/main/java/org/apache/cassandra/sidecar/utils/HttpExceptions.java
```diff
@@ -54,7 +54,7 @@ public static HttpException cassandraServiceUnavailable()
      */
     public static HttpException wrapHttpException(HttpResponseStatus status, Throwable cause)
     {
-        return wrapHttpException(status, null, cause);
+        return wrapHttpException(status, cause != null ? cause.getMessage() : null, cause);
```
yifan-c: I would against passing the cause message as the response payload. The message is internal to server and probably does not make sense to the client, or even not desired to be exposed. 
In the scenario where payload is needed, it should set the response message payload explicitly.

yifan-c (approved):

src/main/java/org/apache/cassandra/sidecar/utils/HttpExceptions.java
frankgh: makes sense, I have addressed this feedback. Thanks

PR #61, opened 2023-07-13
Enrich RetriesExhaustedException to have more information for better visibility

yifan-c:


client/src/main/java/org/apache/cassandra/sidecar/client/HttpResponse.java
```diff
@@ -62,4 +62,9 @@ default String contentAsString()
      * @return the headers for the response
      */
     Map<String, List<String>> headers();
+
+    /**
+     * @return the server that returns the response
+     */
+    SidecarInstance respondingServer();
```
frankgh: Would it make more sense to call this instance instead?
```suggestion
    SidecarInstance sidecarInstance();
```

client/src/main/java/org/apache/cassandra/sidecar/client/RequestExecutor.java
```diff
@@ -142,62 +160,6 @@ public void close() throws Exception
...
-        if (iterator.hasNext())
-        {
-            executeWithRetries(future, iterator, iterator.next(), context, attempt);
-        }
-        else
-        {
-            // exhausted retries on all available hosts
-            future.completeExceptionally(new RetriesExhaustedException(attempt, context.request(), throwable));
-        }
-    }
```
yifan-c, line 170: reviewer note: I found this method being redundant. It is only called when `iterator.hasNext()` is true, since the condition has been checked on the only call-site. The else branch is never hit. Then, the method has the same effect as its overload. Therefore, I am removing this method and the similar one for stream, for code simplicity.

client/src/main/java/org/apache/cassandra/sidecar/client/RequestExecutor.java
```diff
@@ -305,26 +267,21 @@ private void applyRetryPolicy(CompletableFuture<HttpResponse> future,
...
-                                request, throwable);
-                }
-                schedule(delay, () -> executeWithRetries(future, iterator, sidecarInstance, context, nextAttempt));
+                logger.warn("Retrying request on {} instance after {}ms. " +
+                            "Failed on instance={}, attempt={}, statusCode={}",
+                            nextInstance == sidecarInstance ? "same" : "next", delay,
+                            nextInstance, attempt, statusCode, throwable);
             }
+            SidecarInstance instance = nextInstance;
+            schedule(delay, () -> executeWithRetries(future, iterator, instance, context, nextAttempt));
```
yifan-c: reviewer note: Same effect as the original implementation, but simplified.

client/src/main/java/org/apache/cassandra/sidecar/client/retry/RetryPolicy.java
```diff
@@ -55,31 +54,6 @@ public abstract void onResponse(CompletableFuture<HttpResponse> responseFuture,
...
-     * @param attempts  the number of attempts for the request
-     * @param request   the HTTP request
-     * @param throwable the underlying exception
-     * @return a {@link RetriesExhaustedException} with the number of {@code attempts} performed before the retries
-     */
-    RetriesExhaustedException retriesExhausted(int attempts, Request request, Throwable throwable)
-    {
-        return new RetriesExhaustedException(attempts, request, throwable);
-    }
-
```
yifan-c, line 82: reviewer note: The exception creation methods are more helpful to be publicly accessible. They should not be limited to the RetryPolicy and its subclasses. Additionally, it is a lot of javadoc copy-and-paste by having creation method in the constructors and here.

client/src/main/java/org/apache/cassandra/sidecar/client/exception/RetriesExhaustedException.java
```diff
@@ -28,24 +29,52 @@ public class RetriesExhaustedException extends RuntimeException
...
+
+    /**
+     * Constructs an exception with the number of {@code attempts} performed for the request.
+     *
+     * @param attempts      the number of attempts performed for the request
+     * @param request       the HTTP request
+     * @param lastResponse  the last failed HTTP response
+     * @param throwable     the underlying exception
+     */
+    protected RetriesExhaustedException(int attempts,
```
yifan-c, line 67: review note: There is no need to create multiple constructor overloads.

client/src/main/java/org/apache/cassandra/sidecar/client/HttpResponseImpl.java
```diff
@@ -32,17 +32,22 @@ public class HttpResponseImpl implements HttpResponse
     private final String statusMessage;
     private final byte[] raw;
     private final Map<String, List<String>> headers;
+    private final SidecarInstance respondingServer;
```
frankgh: NIT: instance for consistency with the `SidecarInstance` class name
```suggestion
    private final SidecarInstance sidecarInstance;
```

client/src/main/java/org/apache/cassandra/sidecar/client/RequestExecutor.java
```diff
@@ -101,10 +100,20 @@ public <T> T executeRequest(RequestContext context, long timeout, TimeUnit unit)
      */
     public <T> CompletableFuture<T> executeRequestAsync(RequestContext context)
     {
+        Iterator<SidecarInstance> iterator = context.instanceSelectionPolicy().iterator();
+        SidecarInstance instance = iterator.next();
```
frankgh: This seems wrong. `iterator.next` if implemented according to spec, will throw a `NoSuchElementException` if the iteration has no more elements. We must call `hasNext` before calling `next`

client/src/main/java/org/apache/cassandra/sidecar/client/RequestExecutor.java
```diff
@@ -122,9 +131,18 @@ public void streamRequest(RequestContext context, StreamConsumer streamConsumer)
     {
         Objects.requireNonNull(streamConsumer, "streamConsumer must be non-null");
         Iterator<SidecarInstance> iterator = context.instanceSelectionPolicy().iterator();
-
+        SidecarInstance instance = iterator.next();
```
frankgh: Again , we shouldn't call `next` without checking `hasNext`

client/src/main/java/org/apache/cassandra/sidecar/client/HttpResponseImpl.java
```diff
@@ -32,17 +32,22 @@ public class HttpResponseImpl implements HttpResponse
...
 
     /**
      * Constructs a response object with the provided values
      *
      * @param statusCode    the status code of the response
      * @param statusMessage the status message of the response
      * @param headers       the headers from the response
+     * @param server        the server that returns the response
```
frankgh, line 43: this javadoc seems to reference an inexistent param, and `./gradlew javadocs` will probably fail
```suggestion
     * @param server        the server that returns the response
```

client/src/main/java/org/apache/cassandra/sidecar/client/RequestExecutor.java
```diff
@@ -101,10 +100,20 @@ public <T> T executeRequest(RequestContext context, long timeout, TimeUnit unit)
...
+        Iterator<SidecarInstance> iterator = context.instanceSelectionPolicy().iterator();
+        SidecarInstance instance = iterator.next();
         CompletableFuture<T> resultFuture = new CompletableFuture<>();
+        if (instance == null)
+        {
+            resultFuture.completeExceptionally(new IllegalStateException("InstanceSelectionPolicy " +
+                                                                         context.instanceSelectionPolicy()
+                                                                                .getClass()
+                                                                                .getSimpleName() +
+                                                                         " selects 0 instance"));
```
frankgh:
```suggestion
                                                                         " selects 0 instances"));
```

client/src/main/java/org/apache/cassandra/sidecar/client/RequestExecutor.java
frankgh: 👍

client/src/main/java/org/apache/cassandra/sidecar/client/exception/RetriesExhaustedException.java
```diff
@@ -28,24 +29,52 @@ public class RetriesExhaustedException extends RuntimeException
...
+     * @param request       the HTTP request
+     * @param lastResponse  the last failed HTTP response
      */
-    public RetriesExhaustedException(int attempts, Request request)
+    public static RetriesExhaustedException of(int attempts,
+                                               Request request,
+                                               HttpResponse lastResponse)
     {
-        this(attempts, request, null);
+        return of(attempts, request, lastResponse, null);
```
frankgh, line 40: call the ctor directly?

client/src/main/java/org/apache/cassandra/sidecar/client/RequestExecutor.java
```diff
@@ -305,26 +267,21 @@ private void applyRetryPolicy(CompletableFuture<HttpResponse> future,
...
-                {
-                    logger.warn("Retrying request on same instance after {}ms. Failed on instance={}, " +
-                                "attempt={}, statusCode={}, request={}", delay, sidecarInstance, attempt, statusCode,
-                                request, throwable);
-                }
-                schedule(delay, () -> executeWithRetries(future, iterator, sidecarInstance, context, nextAttempt));
+                logger.warn("Retrying request on {} instance after {}ms. " +
+                            "Failed on instance={}, attempt={}, statusCode={}",
+                            nextInstance == sidecarInstance ? "same" : "next", delay,
+                            nextInstance, attempt, statusCode, throwable);
```
frankgh: Wouldn't the failed instance be `sidecarInstance` instead?
```suggestion
                            sidecarInstance, attempt, statusCode, throwable);
```

client/src/main/java/org/apache/cassandra/sidecar/client/RequestExecutor.java
```diff
@@ -305,26 +267,21 @@ private void applyRetryPolicy(CompletableFuture<HttpResponse> future,
         context.retryPolicy()
                .onResponse(future, request, response, throwable, attempt, retryOnNewHost, (nextAttempt, delay) -> {
             String statusCode = response != null ? String.valueOf(response.statusCode()) : "<Not Available>";
+            SidecarInstance nextInstance = sidecarInstance;
```
frankgh: This can be simplified a bit more:
```suggestion
                        SidecarInstance nextInstance = iterator.hasNext()
                                           ? iterator.next()
                                           : sidecarInstance;

            if (response == null || response.statusCode() != HttpResponseStatus.ACCEPTED.code())
            {
                logger.warn("Retrying stream on {} instance after {}ms. " +
                            "Failed on instance={}, attempt={}, statusCode={}",
                            nextInstance == sidecarInstance ? "same" : "next", delay,
                            sidecarInstance, attempt, statusCode, throwable);
            }
            schedule(delay, () -> streamWithRetries(future, consumer, iterator, nextInstance, context, nextAttempt));
```

frankgh (commented):
I've added some comments. It looks like tests are failing in CI. In general, looks good, but I do have some concerns about the iterator.

client/src/main/java/org/apache/cassandra/sidecar/client/HttpResponse.java
yifan-c: Yeah. For consistency, `sidecarInstance` fits better and I will update. 
I called it "server" because the meaning of instance is vague.

client/src/main/java/org/apache/cassandra/sidecar/client/RequestExecutor.java
yifan-c: good call.

client/src/main/java/org/apache/cassandra/sidecar/client/exception/RetriesExhaustedException.java
yifan-c: I feel those static creation methods make the code look nicer by eliminating the need to explicitly pass null.

client/src/main/java/org/apache/cassandra/sidecar/client/RequestExecutor.java
yifan-c: good catch

client/src/main/java/org/apache/cassandra/sidecar/client/exception/RetriesExhaustedException.java
frankgh: I mean, it's one additional call in the stack, but it might get optimized by the jvm.

src/test/integration/org/apache/cassandra/sidecar/routes/SnapshotsHandler2IntegrationTest.java
```diff
@@ -0,0 +1,143 @@
...
+import org.apache.cassandra.sidecar.IntegrationTestBase;
+import org.apache.cassandra.sidecar.common.data.QualifiedTableName;
+import org.apache.cassandra.sidecar.common.testing.CassandraIntegrationTest;
+import org.apache.cassandra.sidecar.common.testing.CassandraTestContext;
+
+import static io.netty.handler.codec.http.HttpResponseStatus.OK;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@ExtendWith(VertxExtension.class)
+class SnapshotsHandler2IntegrationTest extends IntegrationTestBase
```
frankgh: I'm not sure I understand this change

yifan-c: The commit is **not** to be merged. 
I was just testing with CI. There is memory leak in the in-jvm dtest. The integration test gets killed when there are too many test cases and OOM'd

frankgh (approved):
+1, looks good. Thanks for addressing the comments.

PR #62, opened 2023-07-18
CASSANDRASC-66 Fix failing unit tests in Apache CI

frankgh:
Currently, tests running under {{org.apache.cassandra.sidecar.HealthServiceSslTest}} [are failing](https://ci-cassandra.apache.org/job/cassandra~sidecar/42/) when running inside ASF's CI. Logs are showing that some resources (keystore and truststore) are not found. This is causing the tests to fail.

In this commit, we read the resource from the stream, which is guaranteed to exist as long as the resource exists and the resource name is correct, then we write the resource to a temporary directory, and use the file name to set the keystore and truststore path as part of the configuration options.


common/src/testFixtures/java/org/apache/cassandra/sidecar/common/ResourceUtils.java
```diff
@@ -0,0 +1,74 @@
...
+     * @param classLoader     the class loader for the resource
+     * @param destinationPath the destination path to write the file
+     * @param resourceName    the name of the resource to be loaded
+     * @return the {@link Path} to the created resource
+     */
+    public static Path writeResourceToTempDir(ClassLoader classLoader, Path destinationPath, String resourceName)
+    {
+        try
+        {
+            Path certFilePath = destinationPath.resolve(resourceName);
```
yifan-c: `cert`?

common/src/testFixtures/java/org/apache/cassandra/sidecar/common/ResourceUtils.java
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Writes a resource with {@code resourceName} loaded from the {@link ClassLoader classLoader} into the
+     * {@code destinationPath}
+     *
+     * @param classLoader     the class loader for the resource
+     * @param destinationPath the destination path to write the file
+     * @param resourceName    the name of the resource to be loaded
+     * @return the {@link Path} to the created resource
+     */
+    public static Path writeResourceToTempDir(ClassLoader classLoader, Path destinationPath, String resourceName)
```
yifan-c: The method name says "toTempDir", but it takes "destinationPath", which might not be located in a tempDir.

common/src/testFixtures/java/org/apache/cassandra/sidecar/common/ResourceUtils.java
```diff
@@ -0,0 +1,74 @@
...
+    {
+        try
+        {
+            Path certFilePath = destinationPath.resolve(resourceName);
+
+            // ensure parent directory is created
+            Files.createDirectories(certFilePath.getParent());
+
+            try (InputStream inputStream = classLoader.getResourceAsStream(resourceName);
+                 OutputStream outputStream = Files.newOutputStream(certFilePath.toFile().toPath()))
```
yifan-c: `path.toFile().toPath()`..

scripts/build-shaded-dtest-jar-local.sh
```diff
@@ -25,33 +25,33 @@ SCRIPT_DIR=$( dirname -- "$( readlink -f -- "$0"; )"; )
...
 # Install the version that will be shaded
-mvn install:install-file               \
-   -Dfile=./build/dtest-${CASSANDRA_VERSION}.jar \
-   -DgroupId=org.apache.cassandra      \
-   -DartifactId=${DTEST_ARTIFACT_ID} \
-   -Dversion=${CASSANDRA_VERSION}-${GIT_HASH}         \
-   -Dpackaging=jar                     \
-   -DgeneratePom=true                  \
-   -DlocalRepositoryPath=${REPO_DIR}
+"${SCRIPT_DIR}/mvnw" install:install-file                            \
```
yifan-c, line 38: should we consider windows development and run `mvnw.cmd`, or assume no such use case and discard the cmd file?

scripts/build-shaded-dtest-jar-local.sh
```diff
@@ -25,33 +25,33 @@ SCRIPT_DIR=$( dirname -- "$( readlink -f -- "$0"; )"; )
...
-    -Drelocation.prefix=shaded-${GIT_HASH} \
-    -nsu -U
+"${SCRIPT_DIR}/mvnw" --file "${SCRIPT_DIR}/relocate-dtest-dependencies.pom" package     \
+                     -Drevision="${CASSANDRA_VERSION}"                                  \
+                     -DskipTests                                                        \
+                     -Ddtest.version="${CASSANDRA_VERSION}-${GIT_HASH}"                 \
+                     -Dmaven.repo.local="${REPO_DIR}"                                   \
+                     -DoutputFilePath="${DTEST_JAR_DIR}/dtest-${CASSANDRA_VERSION}.jar" \
+                     -Drelocation.prefix="shaded-${GIT_HASH}"                           \
+                     --no-snapshot-updates --update-snapshots
```
yifan-c, line 55: 👍  on using the full option name

common/src/testFixtures/java/org/apache/cassandra/sidecar/common/ResourceUtils.java
frankgh: does `writeResourceToPath` makes more sense?

common/src/testFixtures/java/org/apache/cassandra/sidecar/common/ResourceUtils.java
frankgh: good catch!

scripts/build-shaded-dtest-jar-local.sh
frankgh: didn't Cassandra drop support for windows? I would not support sidecar for windows if the parent project doesn't support it

yifan-c: good call. Then we should remove the cmd file. And there is gradle.bat file could be removed too.

frankgh: done

PR #63, opened 2023-07-21
CASSANDRASC-67 Fix relocation of native libraries for vertx-client-sh…

frankgh:
…aded

OpenSSL is currently unavailable in the vertx-client-shaded library. This is due to the incorrect relocation of the netty native libraries under META-INF/native/libnetty.

In this commit we fix the relocation of the native libraries and we ensure that the relocation is properly configured by adding tests that guarantee OpenSSL loads correctly with the correct relocation.


scripts/get-hostname.sh
```diff
@@ -0,0 +1,30 @@
...
+
+if [[ $# -eq 0 ]] ; then
+    echo 'A URI must be provided'
+    exit 1
+fi
+
+hostname=$(echo "${1}" | sed -E 's/^\s*.*:\/\///g')
+hostname=${hostname%%/*}
+hostname=${hostname%%:*}
+hostname=$(echo "${hostname}" | cut -d@ -f2-)
```
yifan-c: It would be helpful to add a comment on what those lines do.

scripts/build-dtest-jars.sh
```diff
@@ -27,7 +27,8 @@ mkdir -p "${BUILD_DIR}"
 
 # host key verification
 mkdir -p ~/.ssh
-ssh-keyscan github.com >> ~/.ssh/known_hosts
+REPO_HOST=$("${SCRIPT_DIR}/get-hostname.sh" "${REPO}")
```
yifan-c: nit: instead of eval the script, it could source the script if the functionality is defined as a function.

vertx-client-shaded/build.gradle
```diff
@@ -55,8 +72,7 @@ shadowJar {
     relocate 'com.fasterxml.jackson.databind', 'o.a.c.sidecar.client.shaded.com.fasterxml.jackson.databind'
     relocate 'io.netty', 'o.a.c.sidecar.client.shaded.io.netty'
     relocate 'io.vertx', 'o.a.c.sidecar.client.shaded.io.vertx'
-    relocate 'META-INF/native/libnetty', 'META-INF/native/sidecar_client_netty_shaded_netty'
-    relocate 'META-INF/native/netty', 'META-INF/native/io_sidecar_client_netty_shaded_netty'
+    relocate 'META-INF/native/libnetty', 'META-INF/native/libo_a_c_sidecar_client_shaded_netty'
     relocate 'META-INF/versions/11/io/vertx', 'META-INF/versions/11/o/a/c/sidecar/client/shaded/io/vertx'
```
yifan-c, line 76: line#75 is using `_` to delimit, but line#76 is using `/` for `o/a/c`. Can you update them to use the same pattern?

vertx-client/build.gradle
```diff
@@ -56,7 +56,11 @@ dependencies {
     api(group: 'io.vertx', name: 'vertx-web-client', version: '4.4.1') {
         exclude group: 'com.fasterxml.jackson.core', module: 'jackson-core'
     }
-    implementation(group: 'io.netty', name: 'netty-tcnative-boringssl-static', version: '2.0.44.Final')  // for openSSL
+    implementation(group: 'io.netty', name: 'netty-tcnative-boringssl-static', version: '2.0.61.Final')  // for openSSL
+    implementation(group: 'io.netty', name: 'netty-tcnative-boringssl-static', version: '2.0.61.Final', classifier: 'linux-x86_64')  // for openSSL
+    implementation(group: 'io.netty', name: 'netty-tcnative-boringssl-static', version: '2.0.61.Final', classifier: 'osx-aarch_64')  // for openSSL
+    implementation(group: 'io.netty', name: 'netty-tcnative-boringssl-static', version: '2.0.61.Final', classifier: 'linux-aarch_64')  // for openSSL
+    implementation(group: 'io.netty', name: 'netty-tcnative-boringssl-static', version: '2.0.61.Final', classifier: 'osx-x86_64')  // for openSSL
```
yifan-c: Hmm... I think the dependency at line#59 already includes all variants. It either downloads all variants or download with a classifier.

frankgh: actually, behavior changed after version `2.0.48.Final`. https://github.com/netty/netty-tcnative/issues/716

vertx-client-shaded/build.gradle
frankgh: those two lines are not doing quite the same. In line 75 we are renaming all the files inside the `META-INF/native` directory that start with name `libnetty`. And we end up with the following names.

```
jar -tf ~/.m2/repository/org/apache/cassandra/sidecar/vertx-client-all/1.0.0-local/vertx-client-all-1.0.0-local.jar | grep netty_tcn

META-INF/native/libo_a_c_sidecar_client_shaded_netty_tcnative_linux_x86_64.so
META-INF/native/libo_a_c_sidecar_client_shaded_netty_tcnative_osx_aarch_64.jnilib
META-INF/native/libo_a_c_sidecar_client_shaded_netty_tcnative_linux_aarch_64.so
META-INF/native/libo_a_c_sidecar_client_shaded_netty_tcnative_osx_x86_64.jnilib
```

This is what netty expects to be in the classpath to be able to load the native library.

For line 76, we are changing the expected directory structure for vertx, where the original is `io/vertx`, it now expects the directory structure to match the package name of the shaded version. The result is the following:

```
 ls -R 11
o

11/o:
a

11/o/a:
c

11/o/a/c:
sidecar

11/o/a/c/sidecar:
client

11/o/a/c/sidecar/client:
shaded

11/o/a/c/sidecar/client/shaded:
io

11/o/a/c/sidecar/client/shaded/io:
vertx

11/o/a/c/sidecar/client/shaded/io/vertx:
core

11/o/a/c/sidecar/client/shaded/io/vertx/core:
DeploymentOptions.class impl

11/o/a/c/sidecar/client/shaded/io/vertx/core/impl:
LoaderManager.class
```

scripts/functions.sh
```diff
@@ -0,0 +1,42 @@
...
+  hostname=${hostname%%/*}
+  # removes the first encounter of a colon : until the end of the string i.e git@github.com:apache -> git@github.com
+  hostname=${hostname%%:*}
+  # removes the username portion of an URI i.e. git@github.com -> github.com
+  hostname=$(echo "${hostname}" | cut -d@ -f2-)
+
+  echo "${hostname}"
+}
+
+test_get_hostname() {
```
yifan-c, line 38: 👍 (except my terminal tab exit when running the test, which is expected :))

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/ccda6798caae7eb7ef11ac9ee248f2e80720a6e2

PR #64, opened 2023-08-11
CASSANDRASC-68 Add interface to capture bytes streamed and written metrics

sarankk:


src/main/java/org/apache/cassandra/sidecar/MainModule.java
```diff
@@ -304,4 +305,11 @@ public SidecarVersionProvider sidecarVersionProvider()
     {
         return new SidecarVersionProvider("/sidecar.version");
     }
+
+    @Provides
+    @Singleton
+    public MetricsPublisher metricsPublisher()
+    {
+        return MetricsPublisher.DummyMetricsPublisher.INSTANCE;
```
yifan-c: I'd expect to see "dummy" classes in test, but not in production code. The publisher does not nothing, can you rename it to `NoOpMetricsPublisher`? It describes what the publisher does.

src/main/java/org/apache/cassandra/sidecar/metrics/MetricsPublisher.java
```diff
@@ -0,0 +1,51 @@
...
+package org.apache.cassandra.sidecar.metrics;
+
+/**
+ * Interface used for publishing metrics across endpoints in Sidecar.
+ */
+public interface MetricsPublisher
+{
+    void captureCount(Metric metric);
+
+    void captureAggregation(Metric metric, long value);
```
yifan-c: `captureAggregation` is invoked in both `SSTableUploadHandler` and `FileStreamer`. How does an actual publisher distinguish the metrics?

yifan-c (changes requested):
The patch requires a second thought. The current design doesn't seem to be working. 
Generally, you have 2 design choices. 
1. In the publisher interface, you declare all the metrics to publish as methods. For example, the publisher has those method, `bytesUploaded(long bytesCount)`, `bytesStreamed(long bytesCount)`. Compiler makes sure that in the actual implementation, it has the metrics publishing implemented comprehensively. The drawback of this approach is the lack of flexibility. You have to modify the interface when adding new metrics. 
2. You declare the metric data type (not the current enum!), which contains the name, value, type, etc. The publisher now can be abstract. It only needs to have the publish method. Much more flexible than option 1. I'd prefer to this option. Please checkout https://metrics.dropwizard.io/

src/main/java/org/apache/cassandra/sidecar/metrics/MetricsPublisher.java
sarankk: Discussed this offline and clarified, we distinguish them by the metric's name

yifan-c:
Unrelated with the patch content, can you include the jira ticket number in the PR title? It helps to link between the jira and pr.

src/main/java/org/apache/cassandra/sidecar/utils/FileStreamer.java
```diff
@@ -123,6 +128,8 @@ private void acquireAndSend(HttpResponse response,
                                {
                                    LOGGER.debug("Streamed file {} successfully to client {}. Instance: {}", filename,
                                                 response.remoteAddress(), response.host());
+                                   metricRegistry.register(MetricRegistry.name(FileStreamer.class, "bytes_streamed"),
+                                                     (Gauge<Long>) () -> fileLength);
```
yifan-c: It does not work as expected. 
The code here tries to register the Gauge metric whenever a file streams successfully and only the first attempt succeeds. It then, only publishes the `fileLength` of that file and once. 

It should not use `Gauge` type. Instead, it should be `Histograms` to not loose any update.

frankgh (approved):
+1 looks good to me

src/main/java/org/apache/cassandra/sidecar/MainModule.java
```diff
@@ -84,6 +85,7 @@ public Vertx vertx()
     {
         return Vertx.vertx(new VertxOptions().setMetricsOptions(new DropwizardMetricsOptions()
                                                                 .setEnabled(true)
+                                                                .setRegistryName("cassandra-sidecar-metrics-registry")
```
yifan-c: This line is not needed. I will remove it on committing.

PR #66, opened 2023-08-22
[CASSANDRASC-70] Added Client Methods for Obtaining Sidecar and Cassandra Health

5:
Methods for obtaining Sidecar and Cassandra health need to be added to the client.
Both of these methods will be used for start-up validation by the Analytics library.

### DIFF
--- a/client/src/main/java/org/apache/cassandra/sidecar/client/RequestContext.java
+++ b/client/src/main/java/org/apache/cassandra/sidecar/client/RequestContext.java
@@ -22,6 +22,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.apache.cassandra.sidecar.client.request.CassandraHealthRequest;
 import org.apache.cassandra.sidecar.client.request.CleanSSTableUploadSessionRequest;
 import org.apache.cassandra.sidecar.client.request.ClearSnapshotRequest;
 import org.apache.cassandra.sidecar.client.request.CreateSnapshotRequest;
@@ -33,6 +34,7 @@ import org.apache.cassandra.sidecar.client.request.Request;
 import org.apache.cassandra.sidecar.client.request.RingRequest;
 import org.apache.cassandra.sidecar.client.request.SSTableComponentRequest;
 import org.apache.cassandra.sidecar.client.request.SchemaRequest;
+import org.apache.cassandra.sidecar.client.request.SidecarHealthRequest;
 import org.apache.cassandra.sidecar.client.request.TimeSkewRequest;
 import org.apache.cassandra.sidecar.client.request.UploadSSTableRequest;
 import org.apache.cassandra.sidecar.client.retry.ExponentialBackoffRetryPolicy;
@@ -48,6 +50,8 @@ import org.apache.cassandra.sidecar.common.utils.HttpRange;
  */
 public class RequestContext
 {
+    protected static final SidecarHealthRequest SIDECAR_HEALTH_REQUEST = new SidecarHealthRequest();
+    protected static final CassandraHealthRequest CASSANDRA_HEALTH_REQUEST = new CassandraHealthRequest();
     protected static final SchemaRequest FULL_SCHEMA_REQUEST = new SchemaRequest();
     protected static final TimeSkewRequest TIME_SKEW_REQUEST = new TimeSkewRequest();
     protected static final NodeSettingsRequest NODE_SETTINGS_REQUEST = new NodeSettingsRequest();
@@ -171,6 +175,28 @@ public class RequestContext
         {
             this.retryPolicy = retryPolicy;
             return this;
+        }
+
+        /**
+         * Sets the {@code request} to be a {@link SidecarHealthRequest}
+         * and returns a reference to this Builder enabling method chaining
+         *
+         * @return a reference to this Builder
+         */
+        public Builder sidecarHealthRequest()
+        {
+            return request(SIDECAR_HEALTH_REQUEST);
+        }
+
+        /**
+         * Sets the {@code request} to be a {@link CassandraHealthRequest}
+         * and returns a reference to this Builder enabling method chaining
+         *
+         * @return a reference to this Builder
+         */
+        public Builder cassandraHealthRequest()
+        {
+            return request(CASSANDRA_HEALTH_REQUEST);
         }
 
         /**
--- a/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
+++ b/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
@@ -28,12 +28,14 @@ import org.slf4j.LoggerFactory;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import org.apache.cassandra.sidecar.client.request.ImportSSTableRequest;
 import org.apache.cassandra.sidecar.client.retry.IgnoreConflictRetryPolicy;
+import org.apache.cassandra.sidecar.client.retry.OncePerInstanceRetryPolicy;
 import org.apache.cassandra.sidecar.client.retry.RetryPolicy;
 import org.apache.cassandra.sidecar.client.retry.RunnableOnStatusCodeRetryPolicy;
 import org.apache.cassandra.sidecar.client.selection.InstanceSelectionPolicy;
 import org.apache.cassandra.sidecar.client.selection.RandomInstanceSelectionPolicy;
 import org.apache.cassandra.sidecar.common.NodeSettings;
 import org.apache.cassandra.sidecar.common.data.GossipInfoResponse;
+import org.apache.cassandra.sidecar.common.data.HealthResponse;
 import org.apache.cassandra.sidecar.common.data.ListSnapshotFilesResponse;
 import org.apache.cassandra.sidecar.common.data.RingResponse;
 import org.apache.cassandra.sidecar.common.data.SSTableImportResponse;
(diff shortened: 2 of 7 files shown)


client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -68,6 +69,26 @@ public SidecarClient(SidecarInstancesProvider instancesProvider,
...
     }
 
+    /**
+     * Executes the Sidecar health request using the default retry policy and configured selection policy
+     *
+     * @return a completable future of the Sidecar health response
+     */
+    public CompletableFuture<HealthResponse> sidecarHealth()
+    {
+        return executor.executeRequestAsync(requestBuilder().sidecarHealthRequest().build());
```
frankgh: I think for these calls we need a custom retry policy. We probably want to limit the number or retries, I think we also need to support sidecar health checks on a specific instance, and a list of instances

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -68,6 +69,26 @@ public SidecarClient(SidecarInstancesProvider instancesProvider,
...
+    }
+
+    /**
+     * Executes the Cassandra health request using the default retry policy and configured selection policy
+     *
+     * @return a completable future of the Cassandra health response
+     */
+    public CompletableFuture<HealthResponse> cassandraHealth()
+    {
+        return executor.executeRequestAsync(requestBuilder().cassandraHealthRequest().build());
```
frankgh: same comment as above

common/src/main/java/org/apache/cassandra/sidecar/common/data/HealthResponse.java
```diff
@@ -0,0 +1,90 @@
...
+    public boolean isOk()
+    {
+        return status.equals("OK");
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public int hashCode() {
```
frankgh: formatting?

common/src/main/java/org/apache/cassandra/sidecar/common/data/HealthResponse.java
```diff
@@ -0,0 +1,90 @@
...
+        return status.hashCode();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean equals(Object other) {
```
frankgh: formatting is off here, can we move this bracket to the next line?

frankgh (commented):
Looks good, but I think we need to have a custom retry policy

common/src/main/java/org/apache/cassandra/sidecar/common/data/HealthResponse.java
5: Done.

common/src/main/java/org/apache/cassandra/sidecar/common/data/HealthResponse.java
5: Done.

common/src/main/java/org/apache/cassandra/sidecar/common/data/HealthResponse.java
```diff
@@ -0,0 +1,93 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.common.data;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.cassandra.sidecar.client.request.CassandraHealthRequest;
+import org.apache.cassandra.sidecar.client.request.SidecarHealthRequest;
```
yifan-c: You cannot import code from `client` subproject into `commons`. The dependency is other way around

client/src/main/java/org/apache/cassandra/sidecar/client/RequestContext.java
```diff
@@ -173,6 +177,26 @@ public Builder retryPolicy(RetryPolicy retryPolicy)
             return this;
         }
 
+        /**
+         * Sets the {@code request} to be a {@link SidecarHealthRequest} and returns a reference to this Builder enabling method chaining
```
yifan-c: break the line into 2

client/src/main/java/org/apache/cassandra/sidecar/client/RequestContext.java
```diff
@@ -173,6 +177,26 @@ public Builder retryPolicy(RetryPolicy retryPolicy)
...
+         *
+         * @return a reference to this Builder
+         */
+        public Builder sidecarHealthRequest()
+        {
+            return request(SIDECAR_HEALTH_REQUEST);
+        }
+
+        /**
+         * Sets the {@code request} to be a {@link CassandraHealthRequest} and returns a reference to this Builder enabling method chaining
```
yifan-c: break the line into 2

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
yifan-c: Rather than assume a retry policy and hardcoded in the client, what do you think about this?

```java
    public CompletableFuture<HealthResponse> sidecarHealth(Consumer<RequestContext.Builder> requestOverrides)
    {
        RequestContext.Builder builder = requestBuilder().sidecarHealthRequest();
        requestOverrides.accept(builder);
        return executor.executeRequestAsync(builder.build());
    }
```

5: As we discussed, changed this to perform no retries.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
5: Ditto.

common/src/main/java/org/apache/cassandra/sidecar/common/data/HealthResponse.java
5: Removed the imports, they were only used in the JavaDoc.

client/src/main/java/org/apache/cassandra/sidecar/client/RequestContext.java
5: Done.

client/src/main/java/org/apache/cassandra/sidecar/client/RequestContext.java
5: Done.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -68,6 +70,32 @@ public SidecarClient(SidecarInstancesProvider instancesProvider,
...
+    /**
+     * Executes the Sidecar health request using the configured selection policy and with no retries
+     *
+     * @return a completable future of the Sidecar health response
+     */
+    public CompletableFuture<HealthResponse> sidecarHealth()
+    {
+        return executor.executeRequestAsync(requestBuilder()
+                .sidecarHealthRequest()
+                .retryPolicy(new NoRetryPolicy())
```
frankgh: this is not quite what we want. The `NoRetryPolicy` will only check a single instance and skip the rest. Ideally, we want to check if any instance is available. I think a `BasicRetryPolicy` with some sensible defaults makes sense.

frankgh (commented):
can we also add some tests for these new endpoints?

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
5: Replaced with `new BasicRetryPolicy(1)`, but it looks to me like the behavior is exactly same…
Also, added unit tests for newly added methods.

5: Should I use the `ExponentialBackoffRetryPolicy` instead?

5: Added a new `OncePerInstanceRetryPolicy` as per Francisco's suggestion.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
5: Added a new `OncePerInstanceRetryPolicy` as per Francisco's suggestion.

client/src/main/java/org/apache/cassandra/sidecar/client/retry/OncePerInstanceRetryPolicy.java
```diff
@@ -0,0 +1,52 @@
...
+package org.apache.cassandra.sidecar.client.retry;
+
+import java.util.concurrent.CompletableFuture;
+
+import io.netty.handler.codec.http.HttpResponseStatus;
+import org.apache.cassandra.sidecar.client.HttpResponse;
+import org.apache.cassandra.sidecar.client.exception.RetriesExhaustedException;
+import org.apache.cassandra.sidecar.client.request.Request;
+
+public class OncePerInstanceRetryPolicy extends RetryPolicy
```
frankgh, line 32: can we add a javadoc here?

frankgh (approved):
+1 looks good to me

client/src/main/java/org/apache/cassandra/sidecar/client/retry/OncePerInstanceRetryPolicy.java
5: Done.

PR #67, opened 2023-08-25
CASSANDRASC-71 Allow configuring permissions for uploaded SSTables

frankgh:
This commit introduces a new configuration for SSTable uploads (`file_permissions`) which allows
an operator to configure the desired file permissions used for files that are uploaded via SSTable
upload.

### DIFF
--- a/src/main/dist/conf/sidecar.yaml
+++ b/src/main/dist/conf/sidecar.yaml
@@ -75,6 +75,7 @@ sidecar:
   sstable_upload:
     concurrent_upload_limit: 80
     min_free_space_percent: 10
+    # file_permissions: "rw-r--r--" # when not specified, the default file permissions are owner read & write, group & others read
   allowable_time_skew_in_minutes: 60
   sstable_import:
     poll_interval_millis: 100
--- a/src/main/java/org/apache/cassandra/sidecar/config/SSTableUploadConfiguration.java
+++ b/src/main/java/org/apache/cassandra/sidecar/config/SSTableUploadConfiguration.java
@@ -18,6 +18,8 @@
 
 package org.apache.cassandra.sidecar.config;
 
+import java.nio.file.attribute.PosixFilePermission;
+
 /**
  * Configuration for SSTable component uploads on this service
  */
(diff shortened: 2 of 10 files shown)


src/main/dist/conf/sidecar.yaml
```diff
@@ -75,6 +75,7 @@ sidecar:
   sstable_upload:
     concurrent_upload_limit: 80
     min_free_space_percent: 10
+    # file_permissions: # when not specified, the default permissions are "rw-r--r--"
```
yifan-c: Rephrase the comment so the part before `#` is a valid form, it provides the guidance on the acceptable format of the value. 
```suggestion
    # file_permissions: "rw-r--r--" # when not specified, the default file permissions is owner read & write, others & group read
```

src/main/java/org/apache/cassandra/sidecar/config/impl/SSTableUploadConfigurationImpl.java
```diff
@@ -78,4 +110,20 @@ public float minimumSpacePercentageRequired()
...
+
+    /**
+     * Returns the String representation of a set of posix file permissions used during an SSTable file upload.
+     * When an SSTable file is created the specified permissions will be used to create the file.
+     * For example, the String {@code rw-r--r--} represents the set of permissions
+     * {@link PosixFilePermission#OWNER_READ}, {@link PosixFilePermission#OWNER_WRITE},
+     * {@link PosixFilePermission#GROUP_READ}, and {@link PosixFilePermission#OTHERS_READ}.
+     *
+     * @return the String representation of a set of posix file permissions used during an SSTable file upload
+     */
```
yifan-c: nit: can we simply have the following
```
    /**
     * {@inheritDoc}
     */
```

I know that the other methods in the class have the copy-pasted java docs already.. would appreciate less code, including repeated docs.

frankgh: good point, IntelliJ did the copy/paste work for me already :), but it's more work for the reviewer. I will follow your recommendation. Thanks

frankgh:
Closed via 6b650d956ed9ca57b99065cdc3c2c81d2ef0c2d1

PR #68, opened 2023-09-02
CASSANDRASC-72 [CircleCI] Split unit tests and integration tests

frankgh:
As the number of integration tests continues to grow, it is desirable to split unit tests and integration tests. It is also desirable to parallelize integration tests, each integration test should be its own job. The goals of this improvement are:

- Fail fast on checkstyle or minor errors
- Speed up test runtime by running integration tests in parallel
- Isolate failing tests to specific combinations of Cassandra

In this commit, we run unit tests individually for both java 8 and java 11. We split integration tests into its own jobs, split per java version and cassandra version combination.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,7 @@ jobs:
   # Runs java 8 tests on a docker image
   unit_java8:
     docker:
-     - image: circleci/openjdk:8-jdk-stretch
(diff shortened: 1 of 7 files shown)


frankgh:
Closed via f8605b3c3dfdfdc5ef1b8327f4fd657efca8f9b7

PR #69, opened 2023-09-18
CASSANDRASC-73 Updates token-ranges endpoint to return additional ins…

arjunashok:
…tance metadata

### Changes
The sidecar `token-range-replicas` endpoint response is updated to include a `ReplicaMetadata` section which includes the following attrbutes. This replaces the `replicaState` attribute which included a mapping of the replica IP address/port to the `State` of the instance.

ReplicaMetadata attributes:
- State - State of the instance. eg. `Normal`, `Joining`, `Leaving`, `Replacing`
- Status - Status of the instance. eg. `Up`, `Down`
- Name - Hostname/FQDN of the instance resolved from the IP address in the below field
- Address - IP address of the instance
- Port - Port as supplied by the replica-set from C* JMX endpoints
- Datacenter

CI: https://app.circleci.com/pipelines/github/arjunashok/cassandra-sidecar/63/workflows/6440e2ae-0e1e-4b02-b4f8-32d3491f560d

### DIFF
--- a/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
+++ b/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
@@ -49,24 +49,16 @@ public class CassandraStorageOperations implements StorageOperations
     protected final TokenRangeReplicaProvider tokenRangeReplicaProvider;
 
     /**
-     * Creates a new instance with the provided {@link JmxClient}
+     * Creates a new instance with the provided {@link JmxClient} and {@link DnsResolver}
      *
      * @param jmxClient the JMX client used to communicate with the Cassandra instance
+     * @param dnsResolver the DNS resolver used to lookup replicas
      */
     public CassandraStorageOperations(JmxClient jmxClient, DnsResolver dnsResolver)
     {
-        this(jmxClient,
-             new RingProvider(jmxClient, dnsResolver),
-             new TokenRangeReplicaProvider(jmxClient, dnsResolver));
-    }
-
-    public CassandraStorageOperations(JmxClient jmxClient,
-                                      RingProvider ringProvider,
-                                      TokenRangeReplicaProvider tokenRangeProvider)
-    {
         this.jmxClient = jmxClient;
-        this.ringProvider = ringProvider;
-        this.tokenRangeReplicaProvider = tokenRangeProvider;
+        this.ringProvider = new RingProvider(jmxClient, dnsResolver);
+        this.tokenRangeReplicaProvider =  new TokenRangeReplicaProvider(jmxClient, dnsResolver);
     }
 
     /**
--- a/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
+++ b/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
@@ -55,14 +55,18 @@ public class CassandraStorageOperations implements StorageOperations
      */
     public CassandraStorageOperations(JmxClient jmxClient, DnsResolver dnsResolver)
     {
-        this(jmxClient, new RingProvider(jmxClient, dnsResolver));
+        this(jmxClient,
+             new RingProvider(jmxClient, dnsResolver),
+             new TokenRangeReplicaProvider(jmxClient, dnsResolver));
     }
 
-    public CassandraStorageOperations(JmxClient jmxClient, RingProvider ringProvider)
+    public CassandraStorageOperations(JmxClient jmxClient,
+                                      RingProvider ringProvider,
+                                      TokenRangeReplicaProvider tokenRangeProvider)
     {
         this.jmxClient = jmxClient;
         this.ringProvider = ringProvider;
-        this.tokenRangeReplicaProvider = new TokenRangeReplicaProvider(jmxClient);
+        this.tokenRangeReplicaProvider = tokenRangeProvider;
     }
 
     /**
--- a/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
+++ b/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
@@ -30,6 +30,7 @@ import java.util.Objects;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import com.google.common.net.HostAndPort;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -146,9 +147,10 @@ public class TokenRangeReplicaProvider
                          .map(replica -> {
                              try
                              {
+                                 HostAndPort hap = HostAndPort.fromString(replica);
                                  return new ReplicaMetadata(state.of(replica),
                                                             status.of(replica),
-                                                            dnsResolver.reverseResolve(replica),
+                                                            dnsResolver.reverseResolve(hap.getHost()),
                                                             replica,
                                                             hostToDatacenter.get(replica));
                              }
--- a/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
+++ b/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
@@ -151,7 +151,8 @@ public class TokenRangeReplicaProvider
                                  return new ReplicaMetadata(state.of(replica),
                                                             status.of(replica),
                                                             dnsResolver.reverseResolve(hap.getHost()),
-                                                            replica,
+                                                            hap.getHost(),
+                                                            hap.getPort(),
                                                             hostToDatacenter.get(replica));
                              }
                              catch (UnknownHostException e)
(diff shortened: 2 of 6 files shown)


common/src/main/java/org/apache/cassandra/sidecar/common/data/TokenRangeReplicasResponse.java
```diff
@@ -154,4 +155,74 @@ public String toString()
...
+        public String status()
+        {
+            return status;
+        }
+
+        /**
+         * @return FQDN of the node
+         */
+        @JsonProperty("name")
+        public String name()
```
frankgh: this should just be called `fqdn`, see https://github.com/apache/cassandra-sidecar/blob/trunk/common/src/main/java/org/apache/cassandra/sidecar/common/data/RingEntry.java#L171

arjunashok: Addressed

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -115,22 +121,51 @@ private List<TokenRangeReplicas> transformRangeMappings(Map<List<String>, List<S
...
 
         StateWithReplacement state = new StateWithReplacement(joiningNodes, leavingNodes, movingNodes, gossipInfo);
+        RingProvider.Status status = new RingProvider.Status(liveNodes, deadNodes);
 
         return replicaSet.stream()
                          .map(TokenRangeReplicas::replicaSet)
                          .flatMap(Collection::stream)
                          .distinct()
-                         .collect(Collectors.toMap(Function.identity(), state::of));
+                         .map(r -> new ReplicaMetadata(state.of(r),
```
frankgh: reading this is confusing, can we give `r` a more descriptive name?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -115,22 +121,51 @@ private List<TokenRangeReplicas> transformRangeMappings(Map<List<String>, List<S
...
+                                                       status.of(r),
+                                                       resolveReplica(r),
+                                                       r,
+                                                       hostToDatacenter.get(r)))
+                         .collect(Collectors.toList());
+
+    }
+
+    private String resolveReplica(String r)
```
frankgh: use replica here?
```suggestion
    private String resolveReplica(String replica)
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -115,22 +121,51 @@ private List<TokenRangeReplicas> transformRangeMappings(Map<List<String>, List<S
...
+    {
+        String hostName;
+        try
+        {
+            hostName = dnsResolver.reverseResolve(r.split(":")[0]);
```
frankgh: We have some class that will get the host and port for use, just use the class. Splitting and getting an index is prone to errors. I believe the library is HostAndPort
```suggestion
            hostName = dnsResolver.reverseResolve(r.split(":")[0]);
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -115,22 +121,51 @@ private List<TokenRangeReplicas> transformRangeMappings(Map<List<String>, List<S
...
+        }
+        catch (Exception e)
+        {
+            // Swallow exception to not fail-fast on DNS resolution failure
+            LOGGER.warn("Failed to resolve hostname for instance {}: {}", r, e);
```
frankgh: I think we want to print the exception with the stacktrace here
```suggestion
            LOGGER.warn("Failed to resolve hostname for instance={}", r, e);
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -115,22 +121,51 @@ private List<TokenRangeReplicas> transformRangeMappings(Map<List<String>, List<S
...
+            hostName = "";
```
frankgh: why empty? maybe use the unresolved IP?

frankgh (commented):
I've added some comments

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: I explicitly left this empty to denote the resolution failure, so the client can choose to handle it differently in this event. For instance, it can retry instead of making a sidecar request with the IP, which would fail to verify the client identity with TLS.

frankgh: that sounds like a bad idea. Why don't we mimic what `RingProvider` is doing instead. If there are host resolution errors, we should fail the request, client will see a 500, clients should leverage retries in that case.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: Addressed

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: Addressed

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
arjunashok: From offline discussion, the resolver will not fail on resolution failures. Decided to allow exception to be thrown on failures from the resolver. Addressed

yifan-c:
The patch looks good to me! Waiting for the new CI run to finish.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -55,14 +55,18 @@ public class CassandraStorageOperations implements StorageOperations
...
     public CassandraStorageOperations(JmxClient jmxClient, DnsResolver dnsResolver)
     {
-        this(jmxClient, new RingProvider(jmxClient, dnsResolver));
+        this(jmxClient,
+             new RingProvider(jmxClient, dnsResolver),
+             new TokenRangeReplicaProvider(jmxClient, dnsResolver));
     }
 
-    public CassandraStorageOperations(JmxClient jmxClient, RingProvider ringProvider)
+    public CassandraStorageOperations(JmxClient jmxClient,
```
frankgh: I wonder why we have this constructor. It only seems to be used from the call in line 58. Should we remove it?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -115,22 +122,45 @@ private List<TokenRangeReplicas> transformRangeMappings(Map<List<String>, List<S
...
                          .distinct()
-                         .collect(Collectors.toMap(Function.identity(), state::of));
+                         .map(replica -> {
+                             try
+                             {
+                                 HostAndPort hap = HostAndPort.fromString(replica);
+                                 return new ReplicaMetadata(state.of(replica),
+                                                            status.of(replica),
+                                                            dnsResolver.reverseResolve(hap.getHost()),
+                                                            replica,
```
frankgh: since we already have it split, let's add an address field and a port field. 
```suggestion
                                                            hap.getHost(),
                                                            hap.getPort(),
```

frankgh (approved):

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -49,20 +49,16 @@ public class CassandraStorageOperations implements StorageOperations
...
-        this(jmxClient, new RingProvider(jmxClient, dnsResolver));
-    }
-
-    public CassandraStorageOperations(JmxClient jmxClient, RingProvider ringProvider)
     {
         this.jmxClient = jmxClient;
-        this.ringProvider = ringProvider;
-        this.tokenRangeReplicaProvider = new TokenRangeReplicaProvider(jmxClient);
+        this.ringProvider = new RingProvider(jmxClient, dnsResolver);
+        this.tokenRangeReplicaProvider =  new TokenRangeReplicaProvider(jmxClient, dnsResolver);
```
frankgh, line 61: minor cosmetic issues, can you reformat the files touched in this PR?
```suggestion
        this.tokenRangeReplicaProvider = new TokenRangeReplicaProvider(jmxClient, dnsResolver);
```

yifan-c:
Closing the PR as it is merged already.

PR #70, opened 2023-09-21
CASSANDRASC-74 Stream sstable components API fails on secondary index…

frankgh:
… files

In this commit, we fix streaming secondary index SSTable component files. We add tests to validate that the index files can be streamed. We also add compatibility for older clients that don't have the fix.

[CASSANDRASC-74](https://issues.apache.org/jira/browse/CASSANDRASC-74)


src/main/java/org/apache/cassandra/sidecar/routes/StreamSSTableComponentHandler.java
```diff
@@ -64,43 +65,32 @@ public void handleInternal(RoutingContext context,
                                SocketAddress remoteAddress,
                                StreamSSTableComponentRequest request)
     {
-        validate(request)
```
frankgh, line 67: Removed validation since the validation is already happening at the `snapshotPathBuilder.build` step

common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -55,6 +57,8 @@ public final class ApiEndpointsV1
     public static final String SNAPSHOTS_ROUTE = API_V1 + PER_KEYSPACE + PER_TABLE + PER_SNAPSHOT;
     // Replaces DEPRECATED_COMPONENTS_ROUTE
     public static final String COMPONENTS_ROUTE = SNAPSHOTS_ROUTE + PER_COMPONENT;
+    public static final String COMPONENTS_WITH_INDEX_ROUTE_SUPPORT = SNAPSHOTS_ROUTE + PER_INDEX
```
yifan-c: To be more clear, the name should indicate that it is for secondary index. The name `index` has been used else where in the other classes too. Please update them. 

Suggestions: `COMPONENTS_WITH_SI_ROUTE_SUPPORT` or `COMPONENTS_WITH_SECONDARY_INDEX_ROUTE_SUPPORT`

I think using `SI` is fine, as it is a common term in Cassandra world.

common/src/main/java/org/apache/cassandra/sidecar/common/data/SSTableComponent.java
```diff
@@ -35,8 +39,21 @@ public class SSTableComponent
...
+    }
+
+    /**
+     * Constructor for the holder class
+     *
+     * @param qualifiedTableName the qualified table name in Cassandra
+     * @param indexName          the name of the index for the SSTable component
+     * @param componentName      the name of the SSTable component
+     */
+    public SSTableComponent(QualifiedTableName qualifiedTableName, @Nullable String indexName, String componentName)
```
yifan-c: Is it possible to just modify the existing constructor? Let's try to not add more code.

src/main/java/org/apache/cassandra/sidecar/data/StreamSSTableComponentRequest.java
```diff
@@ -43,24 +45,74 @@ public class StreamSSTableComponentRequest extends SSTableComponent
...
+     *
+     * @param qualifiedTableName the qualified table name in Cassandra
+     * @param snapshotName       the name of the snapshot
+     * @param componentName      the name of the SSTable component
+     */
+    public StreamSSTableComponentRequest(QualifiedTableName qualifiedTableName,
+                                         String snapshotName,
+                                         String componentName)
+    {
+        this(qualifiedTableName, snapshotName, null, componentName);
```
yifan-c: Similar to the above comment. How about just modify the constructor? The class is for server internal usage, not a public API.

src/main/java/org/apache/cassandra/sidecar/data/StreamSSTableComponentRequest.java
```diff
@@ -43,24 +45,74 @@ public class StreamSSTableComponentRequest extends SSTableComponent
...
+        String snapshotName = context.pathParam("snapshot");
+        String indexName = context.pathParam("index");
+        String componentName = context.pathParam("component");
+
+        if (indexName == null)
+        {
+            return new StreamSSTableComponentRequest(qualifiedTableName,
+                                                     snapshotName,
+                                                     componentName);
+        }
```
yifan-c: The nullness check seems unnecessary. The `StreamSSTableComponentRequest` allows taking null value.

src/main/java/org/apache/cassandra/sidecar/snapshots/SnapshotPathBuilder.java
```diff
@@ -283,6 +287,13 @@ protected void validate(StreamSSTableComponentRequest request)
         validator.validateTableName(request.tableName());
         validator.validateSnapshotName(request.snapshotName());
         // Only allow .db and TOC.txt components here
+        String indexName = request.indexName();
```
yifan-c: secondaryIndexName or siName.

Not to be confused with primary index, which is an sstable component.

src/main/java/org/apache/cassandra/sidecar/data/StreamSSTableComponentRequest.java
```diff
@@ -43,24 +45,54 @@ public class StreamSSTableComponentRequest extends SSTableComponent
...
+        this(new QualifiedTableName(keyspace, tableName, true), snapshotName, null, componentName);
+    }
+
+    /**
+     * Constructor for the holder class
+     *
+     * @param keyspace      the keyspace in Cassandra
+     * @param tableName     the table name in Cassandra
+     * @param snapshotName  the name of the snapshot
+     * @param indexName     the name of the index for the SSTable component
```
yifan-c: secondaryIndexName?

frankgh:
closed via https://github.com/apache/cassandra-sidecar/commit/c7b170cf2e7764159a3d9cf4ca0abc6db1659e51

PR #71, opened 2023-09-22
CASSANDRASC-75: Shade Jackson completely to prevent incompatibility issues

JeetKunDoug:
Before, we we shading almost all of Jackson but left the annotations unshaded/relocated. However, this causes problems in older Spark environments where the annotations in the class path don’t quite match what the other shaded parts of Jackson expect (missing classes being the most serious issue). This can cause the library to fail on certain Spark versions. This commit removes the exclusion of the annotations project from shading in the vertx-client-shaded project.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,13 +80,13 @@ jobs:
      - image: circleci/openjdk:8-jdk-stretch
     environment:
       INTEGRATION_MAX_PARALLEL_FORKS: 1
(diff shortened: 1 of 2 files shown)


frankgh (approved):
Looks good to me. Can you please link the JIRA to this PR?

JeetKunDoug:
https://issues.apache.org/jira/projects/CASSANDRASC/issues/CASSANDRASC-75

yifan-c:
Closing the PR as it is merged.

PR #72, opened 2023-10-16
CASSANDRASC-77 Upgrade to vertx 4.4.6

frankgh:
Vertx 4.4.6 brings two features that we integrate into Sidecar.

1. Hot reloading of SSL certificates. This allows a running cluster to reload certificates without having to restart the service.

2. Traffic shaping options. This allows to introduce protections for the service. It allows configuring ingress/egress limits.

Additionally, this patch introduces the SidecarServerEvents messaging. It leverages vertx's EventBus to publish and consume messages when server starts, server stops, on CQL connection ready, or CQL disconnection, and when all CQL connections are ready.

Patch for [CASSANDRASC-77](https://issues.apache.org/jira/browse/CASSANDRASC-77)


common/src/main/java/org/apache/cassandra/sidecar/common/JmxClient.java
```diff
@@ -65,94 +64,34 @@ public class JmxClient implements NotificationListener, Closeable
...
-    {
-        this(jmxServiceURL, () -> role, () -> password, () -> false);
-    }
-
-    public JmxClient(String host,
-                     int port,
-                     Supplier<String> roleSupplier,
-                     Supplier<String> passwordSupplier,
-                     BooleanSupplier enableSslSupplier)
+    private JmxClient(Builder builder)
```
frankgh, line 76: decided to go with the builder here, because the number of constructors was getting crazy

spotbugs-exclude.xml
```diff
@@ -48,7 +48,7 @@
 
     <!-- Ignore GC call in CassandraTestTemplate as we want to GC after cluster shutdown -->
     <Match>
-        <Class name="org.apache.cassandra.sidecar.testing.CassandraTestTemplate" />
+        <Class name="org.apache.cassandra.sidecar.test.CassandraTestTemplate" />
```
frankgh, line 52: when using `testing` IntelliJ was not able to locate this class, which caused a lot of issues in IDE. So I decided to move the package from testing -> test. Now everything is good it seems

src/main/java/org/apache/cassandra/sidecar/CassandraSidecarDaemon.java
```diff
@@ -18,177 +18,75 @@
...
-    private final Vertx vertx;
-    private final HttpServer server;
-    private final SidecarConfiguration config;
-    private final InstancesConfig instancesConfig;
-    private final ExecutorPools executorPools;
-
-    private long healthCheckTimerId;
-
-    @Inject
-    public CassandraSidecarDaemon(Vertx vertx,
```
frankgh, line 60: this code is now captured under server/Server.java

src/main/java/org/apache/cassandra/sidecar/config/yaml/ServiceConfigurationImpl.java
```diff
@@ -88,117 +107,45 @@ public class ServiceConfigurationImpl implements ServiceConfiguration
...
+        this(builder());
     }
 
-    public ServiceConfigurationImpl(String host)
+    /**
+     * Constructs a new {@link ServiceConfigurationImpl} with the configured {@link Builder}
+     *
+     * @param builder the builder object
+     */
+    protected ServiceConfigurationImpl(Builder builder)
```
frankgh, line 129: this is another case of exploding number of constructors. So I decided to simplify with a builder

src/main/java/org/apache/cassandra/sidecar/routes/GossipInfoHandler.java
```diff
@@ -24,11 +24,11 @@
 import io.vertx.core.http.HttpServerRequest;
 import io.vertx.core.net.SocketAddress;
 import io.vertx.ext.web.RoutingContext;
-import org.apache.cassandra.sidecar.common.CassandraAdapterDelegate;
 import org.apache.cassandra.sidecar.common.ClusterMembershipOperations;
 import org.apache.cassandra.sidecar.common.data.GossipInfoResponse;
 import org.apache.cassandra.sidecar.common.utils.GossipInfoParser;
 import org.apache.cassandra.sidecar.concurrent.ExecutorPools;
+import org.apache.cassandra.sidecar.utils.CassandraAdapterDelegate;
```
frankgh: I moved some classes away from common (where they didn't make sense), and put them in server.

src/main/java/org/apache/cassandra/sidecar/snapshots/SnapshotPathBuilder.java
```diff
@@ -140,79 +139,77 @@ public Future<List<SnapshotFile>> listSnapshotDirectory(String snapshotDirectory
...
-                                                    // aggregate the results and return the full list
-                                                    snapshotList.addAll(idxPropList);
-                                                    promise.complete(snapshotList);
-                                                });
-                             });
+              List<Future<FileProps>> futures = list.stream()
+                                                    .map(fs::props)
+                                                    .collect(Collectors.toList());
+
+              Future.all(futures)
```
frankgh, line 146: CompositeFuture was deprecated. Using Future now which now supports typed futures

src/main/dist/conf/sidecar.yaml
```diff
@@ -68,10 +68,20 @@ sidecar:
...
+  accept_backlog: 1024
+  server_verticle_instances: 1
   throttle:
     stream_requests_per_sec: 5000
     delay_sec: 5
     timeout_sec: 10
+  traffic_shaping:
+    inbound_global_bandwidth_bps: 0               # 0 implies unthrottled, the inbound bandwidth in bytes per second
+    outbound_global_bandwidth_bps: 0              # 0 implies unthrottled, the outbound bandwidth in bytes per second
+    peak_outbound_global_bandwidth_bps: 419430400 # 400 Mbps, the peak outbound bandwidth in bytes per second
```
yifan-c: `Mbps` means megabits per seconds. 
The value here is indeed `MBps`.

It is too easy to be confused with the abbreviation, i.e. upper B and lower b. 

I would just put this for the comment 

"the peak outbound bandwidth in bytes per second. The default is 400 mebibytes per second"

src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
```diff
@@ -20,7 +20,8 @@
 
 import java.util.List;
 
-import org.apache.cassandra.sidecar.common.CassandraAdapterDelegate;
+import org.apache.cassandra.sidecar.utils.CassandraAdapterDelegate;
```
yifan-c: `CassandraAdapterDelegate` does not seem to fit in the `utils` package. In the existing packages, `cluster` would be a better fit.

src/main/java/org/apache/cassandra/sidecar/concurrent/ExecutorPools.java
```diff
@@ -119,9 +120,45 @@ public long setPeriodic(long delay, Handler<Long> handler)
         public long setPeriodic(long delay, Handler<Long> handler, boolean ordered)
         {
             return vertx.setPeriodic(delay,
-                                     id -> workerExecutor.executeBlocking(promise -> {
+                                     id -> workerExecutor.executeBlocking(() -> {
                                          handler.handle(id);
-                                         promise.complete();
+                                         return id;
+                                     }, ordered));
```
yifan-c: instead of scheduling. It can just call the other `setPeriodic` overloads.

```
        public long setPeriodic(long delay, Handler<Long> handler, boolean ordered)
        {
            return setPeriodic(delay, delay, handler, ordered);
        }
```

src/main/java/org/apache/cassandra/sidecar/concurrent/ExecutorPools.java
```diff
@@ -119,9 +120,45 @@ public long setPeriodic(long delay, Handler<Long> handler)
...
+         *                     then they will be no ordering guarantees
+         * @return the unique ID of the timer
+         */
+        public long setPeriodic(long initialDelay, long delay, Handler<Long> handler, boolean ordered)
+        {
+            return vertx.setPeriodic(initialDelay,
+                                     delay,
+                                     id -> workerExecutor.executeBlocking(() -> {
+                                         handler.handle(id);
+                                         return id;
```
yifan-c: Although the new library suggests to use the new API that takes `Callable`, one difference I noticed in the `WorkerExecutorImpl` is that the implementation with `Callable` does not check whether Worker executor is closed, meaning the callable is scheduled to the closing pool regardless. 

It could be a concern.

I would revert and use the deprecated API.

src/main/java/org/apache/cassandra/sidecar/config/ServiceConfiguration.java
```diff
@@ -33,11 +36,33 @@ public interface ServiceConfiguration
...
 
+    /**
+     * Returns a list of ports where the Sidecar process will bind and listen for connections. Defaults to
+     * the configured {@link #port()}
+     *
+     * @return a list of ports where Sidecar will listen
+     */
+    default List<Integer> ports()
+    {
+        return Collections.singletonList(port());
```
yifan-c: The matrix of network interface array and port array is confusing. 
For example, for interface array [i1, i2] and port array [8000, 9000], does it listens to [i1:8000, i1:9000, i2:8000, i2:9000] or [i1:8000, i2:9000]?

For better API clarity, I would just return a list of `SocketAddress`.

src/main/java/org/apache/cassandra/sidecar/config/SslConfiguration.java
```diff
@@ -28,6 +28,29 @@ public interface SslConfiguration
      */
     boolean enabled();
 
+    /**
+     * Returns {@code true} if the OpenSSL engine should be used, {@code false} otherwise. The OpenSSL engine
+     * will only be enabled if the native libraries for OpenSSL have been loaded correctly.
+     *
+     * @return {@code true} if the OpenSSL engine should be used, {@code false} otherwise
```
yifan-c: nit: 
- duplicated sentences. 
- Since configuring `true` does not guarantee OpenSSL is picked, it is only an intent or preference. How about renaming the config to `preferOpenSSL`?

src/main/java/org/apache/cassandra/sidecar/config/TrafficShapingConfiguration.java
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Configuration for the global traffic shaping options. These TCP server options enable configuration of
+ * bandwidth limiting. Both inbound and outbound bandwidth can be limited through these options.
+ */
+public interface TrafficShapingConfiguration
+{
+    /**
+     * Default inbound bandwidth limit in bytes/sec = 0 (0 implies unthrottled)
+     */
+    long DEFAULT_INBOUND_GLOBAL_BANDWIDTH_LIMIT = 0;
```
yifan-c: Why defining the constants in the interface? How about moving them to `TrafficShapingConfigurationImpl`?

src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -241,11 +237,27 @@ public Router vertxRouter(Vertx vertx,
...
+    @Provides
+    @Singleton
+    public PeriodicTaskExecutor periodicTaskExecutor(Vertx vertx,
+                                                     ExecutorPools executorPools,
+                                                     HealthCheckPeriodicTask healthCheckTask)
+    {
+        EventBus eventBus = vertx.eventBus();
+        PeriodicTaskExecutor executor = new PeriodicTaskExecutor(executorPools);
+        eventBus.localConsumer(ON_SERVER_START, message -> executor.schedule(healthCheckTask));
+        eventBus.localConsumer(ON_SERVER_STOP, message -> executor.unschedule(healthCheckTask));
```
yifan-c: nit: the event bus message registration for health check fits better in `HealthCheckPeriodicTask` constructor. 

The method here should just return an executor

src/main/java/org/apache/cassandra/sidecar/server/SidecarServerEvents.java
```diff
@@ -0,0 +1,131 @@
...
+     * The {@link io.vertx.core.eventbus.EventBus} address where server stop/shutdown events will be published.
+     * Server stop events will be published whenever Sidecar is stopping or shutting down.
+     */
+    public static final String ON_SERVER_STOP = SidecarServerEvents.class.getName() + ".ON_SERVER_STOP";
+
+    /**
+     * The {@link io.vertx.core.eventbus.EventBus} address where events will be published when a CQL connection for
+     * a given instance has been established. The instance identifier will be passed as part of the message.
+     */
+    public static final String ON_CASSANDRA_CQL_READY = SidecarServerEvents.class.getName() + ".ON_CASSANDRA_CQL_READY";
```
yifan-c: Instead of hardcode the names, you can declare the enum instead. It store all the name in the list that can be accessed using the enum's ordinal. 

```java
    enum Event
    {
        ON_SERVER_START;

        private static List<String> eventNames = Arrays.stream(values())
                                                       .map(e -> e.getDeclaringClass().getName() + '.' + e.name())

        String eventName()
        {
            return eventNames.get(ordinal());
        }
    }
```

src/main/java/org/apache/cassandra/sidecar/tasks/HealthCheckPeriodicTask.java
```diff
@@ -0,0 +1,90 @@
...
+                               instanceMetadata.delegate().healthCheck();
+                               promise.complete();
+                           }
+                           catch (Throwable cause)
+                           {
+                               promise.fail(cause);
+                               LOGGER.error("Unable to complete health check on instance={}",
+                                            instanceMetadata.id(), cause);
+                           }
+                       }));
```
yifan-c: nit: submit with `unordered == true`.

src/main/dist/conf/sidecar.yaml
frankgh: yeah! thanks

src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
frankgh: yeah, agreed

src/main/java/org/apache/cassandra/sidecar/concurrent/ExecutorPools.java
frankgh: ok, I think it's a valid concern. I will double check, but it looks like the exception now should be `RejectedExecutionException`. I assume the intention was to maintain the pool API expectations instead of throwing an `IllegalStateException`

src/main/java/org/apache/cassandra/sidecar/server/SidecarServerEvents.java
frankgh: makes sense, refactored

src/main/java/org/apache/cassandra/sidecar/config/TrafficShapingConfiguration.java
frankgh: yeah, I guess these are specific to yaml

src/main/java/org/apache/cassandra/sidecar/config/ServiceConfiguration.java
frankgh: makes sense, I will modify the API

yifan-c (approved):
+1

dineshjoshi (approved):
+1

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/8045f8eedad4510ecbf66bb739d464eec741aced

PR #73, opened 2023-10-19
CASSANDRASC-78 Fix token-ranges endpoint to handle gossip-info respon…

arjunashok:
…ses without 'status'

### Details
This is a fix to look for the host status in ‘Status’ and ‘StatusWithPort’ attributes in GossipInfo response  to determine the ‘state’ of the node.

Currently, we only check for ‘status’ which can be missing from gossipInfo in some cases, which will result in a replacement node status to be reported as `Joining` instead of `Replacing`.

### Testing
- Added related unit tests to reproduce and validate fix

### DIFF
--- a/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
+++ b/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
@@ -285,7 +285,9 @@ public class TokenRangeReplicaProvider
                 {
                     LOGGER.debug("Found gossipInfoEntry={}", gossipInfoEntry);
                     String hostStatus = gossipInfoEntry.status();
-                    if (hostStatus != null && hostStatus.startsWith("BOOT_REPLACE,"))
+                    String hostStatusWithPort = gossipInfoEntry.statusWithPort();
+                    if ((hostStatus != null && hostStatus.startsWith("BOOT_REPLACE,")) ||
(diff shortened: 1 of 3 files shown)


frankgh (approved):
Good find! +1

yifan-c (approved):

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicaProvider.java
```diff
@@ -285,7 +285,9 @@ String of(String endpoint)
                 {
                     LOGGER.debug("Found gossipInfoEntry={}", gossipInfoEntry);
                     String hostStatus = gossipInfoEntry.status();
-                    if (hostStatus != null && hostStatus.startsWith("BOOT_REPLACE,"))
+                    String hostStatusWithPort = gossipInfoEntry.statusWithPort();
+                    if ((hostStatus != null && hostStatus.startsWith("BOOT_REPLACE,")) ||
+                        (hostStatusWithPort != null && hostStatusWithPort.startsWith("BOOT_REPLACE,")))
```
yifan-c, line 290: nit: it might be worth to add a warning when `STATUS` field is missing when cluster runs in 4.x, since it marks a potential gossip bug. It is not the sidecar's responsibility to monitor it, so I am fine without it too. I would defer to your decision.

arjunashok: Oops, seeing this late. I agree with you that we could do without the log line, since it really does not affect the sidecar or this APIs behavior.

PR #74, opened 2023-10-20
CASSANDRASC-79: Sidecar should be able to load metadata even if the local instance is unavailable

JeetKunDoug:
- Add SidecarLoadBalancingPolicy to connect to additional nodes beyond local
- Enable Gossip by default (the client really needs it on)
- Retry keyspace creation if it fails
- Use shared executor/hashedwheeltimer netty options for all tests.
  This restores the previous behavior from when the NettyOptions parameter
  was first added, although we're not sure exactly why it's helpful, or if
  it is at all.
- Also store test results as artifacts
- Fix multi-threaded access issues with synchronized for now. Also fix an issue where the host couldn't be found if it wasn't using 9042 for the port.
- Prevent spurious reconnects of ignored nodes on onUp events
- Use `QueryOptions#setReprepareOnUp(false)` to prevent an onUp event from reconnecting to a node we don't want a connection with.
- Integration test for the load balancing polilcy + Adds fixes for issues found while testing.
- Use `ClusterUtils.awaitRingStatus` for retries when waiting for a node to shut down


src/test/integration/org/apache/cassandra/testing/AbstractCassandraTestContext.java
```diff
@@ -67,10 +69,24 @@ public void close()
             {
                 cluster.close();
             }
-            catch (ShutdownException shutdownException)
+            // ShutdownException may be thrown from a different classloader, and therefore the standard
+            // `catch (ShutdownException)` won't always work - compare the canonical names in stead.
```
frankgh:
```suggestion
            // `catch (ShutdownException)` won't always work - compare the canonical names instead.
```

common/src/main/java/org/apache/cassandra/sidecar/common/ICassandraAdapter.java
```diff
@@ -35,6 +40,15 @@ public interface ICassandraAdapter
 
     NodeSettings nodeSettings();
 
+    default ResultSet executeLocal(String query)
```
frankgh, line 52: can we please add javadocs for all the new methods in the interface?

common/src/main/java/org/apache/cassandra/sidecar/common/ICassandraFactory.java
```diff
@@ -26,9 +28,10 @@ public interface ICassandraFactory
     /**
      * Creates a new {@link ICassandraAdapter} with the provided {@link CQLSessionProvider} and {@link JmxClient}
      *
-     * @param session the session to the Cassandra database
-     * @param client  the JMX client to connect to the Cassandra database
+     * @param session             the session to the Cassandra database
+     * @param client              the JMX client to connect to the Cassandra database
+     * @param broadcastRpcAddress the native transport address and port of the instance
```
frankgh: should we call this parameter `localInstanceNativeAddress` instead?

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -332,6 +361,11 @@ protected void markAsDownAndMaybeNotify()
     @Nullable
     private <T> T fromAdapter(Function<ICassandraAdapter, T> getter)
     {
+        if (this.adapter == null)
```
frankgh: I am a strong -1 on this change. This basically defeats the purpose of the health checks, which periodically checks for connection state. This is a way to prevent from a thundering herd situation. The client should retry on 503 (Service Unavailable) which is returned when we encounter this situation.

src/main/java/org/apache/cassandra/sidecar/config/DriverConfiguration.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * The driver configuration to use when connecting to Cassandra
+ */
+public interface DriverConfiguration
+{
+    @JsonProperty("contact_points")
+    List<InetSocketAddress> contactPoints();
```
frankgh, line 34: can we please add javadocs here as well?

src/main/java/org/apache/cassandra/sidecar/config/DriverConfiguration.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * The driver configuration to use when connecting to Cassandra
+ */
+public interface DriverConfiguration
+{
+    @JsonProperty("contact_points")
```
frankgh: Annotations under the implementation instead?
```suggestion
```

src/main/java/org/apache/cassandra/sidecar/config/DriverConfiguration.java
```diff
@@ -0,0 +1,39 @@
...
+    List<InetSocketAddress> contactPoints();
+
+    @JsonProperty("num_connections")
```
frankgh: Annotations under the implementation instead?
```suggestion
```

src/main/java/org/apache/cassandra/sidecar/config/DriverConfiguration.java
```diff
@@ -0,0 +1,39 @@
...
+    int numConnections();
+
+    @JsonProperty("local_dc")
```
frankgh: Annotations under the implementation instead?
```suggestion
```

src/main/java/org/apache/cassandra/sidecar/config/yaml/SidecarConfigurationImpl.java
```diff
@@ -304,6 +330,8 @@ public Builder cassandraInputValidationConfiguration(CassandraInputValidationCon
             return update(b -> b.cassandraInputValidationConfiguration = configuration);
         }
 
+
```
frankgh: Nit: extra line break
```suggestion
```

src/main/java/org/apache/cassandra/sidecar/config/yaml/DriverConfigurationImpl.java
```diff
@@ -0,0 +1,54 @@
...
+    @JsonProperty("contact_points")
+    public List<InetSocketAddress> contactPoints()
+    {
+        return contactPoints;
+    }
+
+    @JsonProperty("num_connections")
+    public int numConnections()
+    {
+        return numConnections;
```
frankgh, line 46: Should we do validation of the input? i.e. do not allow 0 or negative numbers? or what is the expectation to handle those values?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
```diff
@@ -42,16 +49,18 @@ public class CassandraAdapter implements ICassandraAdapter
...
     protected final JmxClient jmxClient;
-    private final CQLSessionProvider session;
+    private final CQLSessionProvider cqlSessionProvider;
     private final String sidecarVersion;
+    private final InetSocketAddress broadcastRpcAddress;
 
-    public CassandraAdapter(DnsResolver dnsResolver, JmxClient jmxClient, CQLSessionProvider session,
-                            String sidecarVersion)
+    public CassandraAdapter(DnsResolver dnsResolver, JmxClient jmxClient, CQLSessionProvider cqlSessionProvider,
+                            String sidecarVersion, InetSocketAddress broadcastRpcAddress)
```
frankgh: I feel we need a better name for the `broadcastRpcAddress` variable

common/src/main/java/com/datastax/driver/core/DriverExtensions.java
```diff
@@ -0,0 +1,67 @@
...
+     * @param cluster The cluster object
+     * @param host    the host to which reconnect attempts will be made
+     */
+    public static void startPeriodicReconnectionAttempt(Cluster cluster, Host host)
+    {
+        cluster.manager.startPeriodicReconnectionAttempt(host, false);
+    }
+
+    /**
+     * Gets a Host instance from metadata based on the native
```
frankgh: incomplete sentence maybe? based on the native socket address?

common/src/main/java/com/datastax/driver/core/DriverExtensions.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+package com.datastax.driver.core;
+
+import java.net.InetSocketAddress;
+
+/**
+ * A collection of methods that require access to package-private members in the datastax driver.
+ */
+
```
frankgh: nit : remove extra empty line

src/main/java/org/apache/cassandra/sidecar/config/yaml/SidecarConfigurationImpl.java
```diff
@@ -292,6 +306,18 @@ public Builder healthCheckConfiguration(HealthCheckConfiguration healthCheckConf
             return update(b -> b.healthCheckConfiguration = healthCheckConfiguration);
         }
 
+        /**
+         * Sets the {@code driverConfiguration} and returns a reference to this Builder enabling
+         * method chaining.
+         *
+         * @param driverConfiguration the {@code cassandraInputValidationConfiguration} to set
```
frankgh:
```suggestion
         * @param driverConfiguration the {@code driverConfiguration} to set
```

src/main/java/org/apache/cassandra/sidecar/config/yaml/SidecarConfigurationImpl.java
```diff
@@ -292,6 +306,18 @@ public Builder healthCheckConfiguration(HealthCheckConfiguration healthCheckConf
...
+         * @return a reference to this Builder
+         */
+        public Builder cassandraInputValidationConfiguration(DriverConfiguration driverConfiguration)
```
frankgh:
```suggestion
        public Builder driverConfiguration(DriverConfiguration driverConfiguration)
```

src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -365,22 +381,21 @@ public SidecarStats sidecarStats()
      * @param vertx                      the vertx instance
      * @param cassandraInstance          the cassandra instance configuration
      * @param versionProvider            a Cassandra version provider
-     * @param healthCheckFrequencyMillis the health check frequency configuration in milliseconds
      * @param sidecarVersion             the version of the Sidecar from the current binary
      * @param jmxConfiguration           the configuration for the JMX Client
+     * @param session                         the CQL Session provider
```
frankgh: nit
```suggestion
     * @param session                    the CQL Session provider
```

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -75,6 +80,7 @@ public class CassandraAdapterDelegate implements ICassandraAdapter, Host.StateLi
     private volatile NodeSettings nodeSettings = null;
     private final AtomicBoolean registered = new AtomicBoolean(false);
     private final AtomicBoolean isHealthCheckActive = new AtomicBoolean(false);
+    private InetSocketAddress broadcastRpcAddress;
```
frankgh:
```suggestion
    private final InetSocketAddress broadcastRpcAddress;
```

src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -395,7 +410,9 @@ private static InstanceMetadata buildInstanceMetadata(Vertx vertx,
                                                                          versionProvider,
                                                                          session,
                                                                          jmxClient,
-                                                                         sidecarVersion);
+                                                                         sidecarVersion,
+                                                                         cassandraInstance.host(),
+                                                                         cassandraInstance.port());
```
frankgh:
```suggestion
                                                                         host,
                                                                         port);
```

common/src/main/java/org/apache/cassandra/sidecar/common/SidecarLoadBalancingPolicy.java
```diff
@@ -0,0 +1,204 @@
...
+/**
+ * The SidecarLoadBalancingPolicy is designed to ensure that the Cassandra Metadata objects associated with the
+ * CqlSessionProvider have enough non-local hosts in their allowed connections to be kept up-to-date
+ * even if the local Cassandra instances are down/have their native transport disabled.
+ * NOTE: This policy won't work with a child policy that is token-aware
+ */
+class SidecarLoadBalancingPolicy implements LoadBalancingPolicy
+{
+    public static final int MIN_ADDITIONAL_CONNECTIONS = 2;
+    private static final Logger logger = LoggerFactory.getLogger(SidecarLoadBalancingPolicy.class);
```
frankgh:
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(SidecarLoadBalancingPolicy.class);
```

common/src/main/java/org/apache/cassandra/sidecar/common/SidecarLoadBalancingPolicy.java
```diff
@@ -0,0 +1,204 @@
...
+        if (numAdditionalConnections < MIN_ADDITIONAL_CONNECTIONS)
+        {
+            logger.warn("Additional instances requested was {}, which is less than the minimum of {}. Using {}.",
+                        numAdditionalConnections, MIN_ADDITIONAL_CONNECTIONS, MIN_ADDITIONAL_CONNECTIONS);
+            numAdditionalConnections = MIN_ADDITIONAL_CONNECTIONS;
+        }
+        this.totalRequestedConnections = this.localHostAddresses.size() + numAdditionalConnections;
+    }
+
+    private LoadBalancingPolicy getChildPolicy(String localDc)
```
frankgh: nit
```suggestion
    private LoadBalancingPolicy childPolicy(String localDc)
```

common/src/main/java/org/apache/cassandra/sidecar/common/SidecarLoadBalancingPolicy.java
```diff
@@ -0,0 +1,204 @@
...
+    private synchronized void recalculateSelectedHosts()
+    {
+        // Copy the list to allow us to remove hosts as we build the list
+        List<Host> hosts = new ArrayList<>(this.allHosts);
+        addLocalHosts(hosts);
+        Collections.shuffle(hosts, this.secureRandom);
+        Iterator<Host> hostIterator = hosts.iterator();
+        while (selectedHosts.size() < totalRequestedConnections && hostIterator.hasNext())
+        {
+            Host h = hostIterator.next();
```
frankgh: nit
```suggestion
            Host host = hostIterator.next();
```

common/src/main/java/org/apache/cassandra/sidecar/common/SidecarLoadBalancingPolicy.java
```diff
@@ -0,0 +1,204 @@
...
+                selectedHosts.add(h);
+                hostIterator.remove();
+            }
+        }
+    }
+
+    private synchronized void recalculateSelectedHosts()
+    {
+        // Copy the list to allow us to remove hosts as we build the list
+        List<Host> hosts = new ArrayList<>(this.allHosts);
```
frankgh: should this be a set instead? given that we will add local hosts in the line below, we should avoid adding the local hosts twice if we ever encounter that situation

common/src/main/java/org/apache/cassandra/sidecar/common/SidecarLoadBalancingPolicy.java
```diff
@@ -0,0 +1,204 @@
...
+     * Adds any local hosts to the selected host list
+     *
+     * @param hosts the list of hosts provided by the control connection
+     */
+    private void addLocalHosts(List<Host> hosts)
+    {
+        Iterator<Host> hostIterator = hosts.iterator();
+        while (hostIterator.hasNext())
+        {
+            Host h = hostIterator.next();
```
frankgh: nit
```suggestion
            Host host = hostIterator.next();
```

common/src/main/java/org/apache/cassandra/sidecar/common/SidecarLoadBalancingPolicy.java
```diff
@@ -0,0 +1,204 @@
...
+    public synchronized void onUp(Host host)
+    {
+        if (selectedHosts.size() < totalRequestedConnections)
+        {
+            recalculateSelectedHosts();
+        }
+        childPolicy.onUp(host);
+    }
+
+    public synchronized void onDown(Host host)
```
frankgh: can we add override annotations in this class to identify methods from super classes/interfaces?

src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
```diff
@@ -103,7 +117,11 @@ void setup(AbstractCassandraTestContext cassandraTestContext) throws Interrupted
...
         server.start()
               .onSuccess(s -> {
+                  // TODO: DTR - Do we still need the new scheduleHealthCheck here (and the sleep below)?
+                  scheduleHealthCheck(5000); // Check health frequently so adapters are created in time for tests
                   sidecarTestContext.registerInstanceConfigListener(this::healthCheck);
+                  // Give everything a moment to get started and connected
+                  Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
                   if (!sidecarTestContext.isClusterBuilt())
                   {
                       context.completeNow();
```
frankgh:
```suggestion
                  if (!sidecarTestContext.isClusterBuilt())
                  {
                      vertx.setTimer(TimeUnit.SECONDS.toMillis(1), id -> context.completeNow());
```

src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
```diff
@@ -103,7 +117,11 @@ void setup(AbstractCassandraTestContext cassandraTestContext) throws Interrupted
 
         server.start()
               .onSuccess(s -> {
+                  // TODO: DTR - Do we still need the new scheduleHealthCheck here (and the sleep below)?
+                  scheduleHealthCheck(5000); // Check health frequently so adapters are created in time for tests
```
frankgh: this is an issue that should be fixed in CASSANDRASC-80

src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
```diff
@@ -68,28 +70,40 @@
...
-        sidecarTestContext = CassandraSidecarTestContext.from(vertx, cassandraTestContext, DnsResolver.DEFAULT);
+        sidecarTestContext = CassandraSidecarTestContext.from(vertx, cassandraTestContext, DnsResolver.DEFAULT,
+                                                              getNumInstancesToManage(clusterSize));
+
         integrationTestModule.setCassandraTestContext(sidecarTestContext);
 
+        instancesConfig = getInstancesConfig();
         server = injector.getInstance(Server.class);
-
+        vertx = injector.getInstance(Vertx.class);
```
frankgh: we already have the vertx instance
```suggestion
```

src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
```diff
@@ -114,6 +132,34 @@ void setup(AbstractCassandraTestContext cassandraTestContext) throws Interrupted
...
+    {
+        return clusterSize;
+    }
+
+    protected InstancesConfig getInstancesConfig()
+    {
+        return injector.getInstance(InstancesConfig.class);
+    }
+
+    private void scheduleHealthCheck(long intervalMillis)
```
frankgh: this is done already by the `HealthCheckPeriodicTask` but you are probably facing issues because of CASSANDRASC-80

src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
```diff
@@ -158,9 +216,30 @@ protected void createTestKeyspace()
...
-                        " WITH REPLICATION = { 'class' : 'NetworkTopologyStrategy', " + generateRfString(rf) + " };");
+        int attempts = 1;
+        ArrayList<Throwable> thrown = new ArrayList<>(5);
+        while (attempts <= 5)
+        {
+            try
+            {
+                Session session = maybeGetSession();
+
+                session.execute("CREATE KEYSPACE IF NOT EXISTS " + TEST_KEYSPACE +
```
frankgh, line 202: why is create keyspace failing? I haven't seen this before

src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
```diff
@@ -68,28 +70,40 @@
...
     private static final AtomicInteger TEST_TABLE_ID = new AtomicInteger(0);
+    protected Logger logger = LoggerFactory.getLogger(this.getClass());
+    protected Vertx vertx;
+    protected Server server;
     protected CassandraSidecarTestContext sidecarTestContext;
+    protected Injector injector;
+    private long healthCheckTimerId;
+    private InstancesConfig instancesConfig;
+
+    private static QualifiedTableName uniqueTestTableFullName()
```
frankgh: private static methods should go at the end

frankgh (commented):
I have a few more comments, but I've decided to submit the review so far.

src/test/integration/org/apache/cassandra/testing/AbstractCassandraTestContext.java
JeetKunDoug: :+1:

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
JeetKunDoug: There was a reason I had to do this before, but I think it was added before I actually started running scheduled health checks in the Integration tests -  without it, at the time, tests couldn't succeed if they tried to do things like creating a keyspace as they use the same session that the Sidecar does, and it wouldn't exist.

I've removed it and rerun the integration tests and it seems happy now.

src/main/java/org/apache/cassandra/sidecar/config/DriverConfiguration.java
JeetKunDoug: Yup

src/main/java/org/apache/cassandra/sidecar/config/DriverConfiguration.java
JeetKunDoug: 👍

src/main/java/org/apache/cassandra/sidecar/config/yaml/DriverConfigurationImpl.java
JeetKunDoug: It's done at usage - we take any number < the min additional connections (a constant, currently 2) and use 2 instead. Anything > the total number of instances in the cluster ends up being treated as "all the hosts" - both of these conditions log at `warn` level but allow the system to continue.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
JeetKunDoug: Yeah - there were some struggles with this.. at one point it _was_ the broadcastRpcAddress being used, which was why I used that particular name. I'll  change them all to `localNativeTransportAddress` which should make it clearer what we're asking for.

common/src/main/java/com/datastax/driver/core/DriverExtensions.java
JeetKunDoug: "native transport address" to be consistent with renaming the broadcastRpcAddress member/variable everywhere, but yeah.

common/src/main/java/org/apache/cassandra/sidecar/common/SidecarLoadBalancingPolicy.java
JeetKunDoug: looking at this, it isn't really a "getter" either - I renamed it `createChildPolicy` so it doesn't look like something that should be called repeatedly.

common/src/main/java/org/apache/cassandra/sidecar/common/SidecarLoadBalancingPolicy.java
JeetKunDoug: So I guess `addLocalHosts` is misnamed - we're not adding them to the `hosts` collection, we're adding them to `selectedHosts` and _removing_ them from the `hosts` list so we only consider non-local hosts when we do the rest of the selection. Let me see if I can rename some of these things to make sure that it's clearer.
`addLocalHosts` -> `addLocalHostsToSelected`
`hosts` -> `sourceHosts`

src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
JeetKunDoug: Happy to make this change - please tell me why so I can document it here.

src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
JeetKunDoug: Yeah - likely - I'll rebase once I make the other changes and see how it behaves without the health check here.

src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
JeetKunDoug: probably bad rebase - thanks for catching it.

src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestBase.java
JeetKunDoug: Because `maybeGetSession` returns `null` in certain caseses where it's not yet connected/the adapter isn't created. Again, may be helped with CASSANDRASC-80?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
```diff
@@ -90,20 +99,43 @@ public Metadata metadata()
...
+        {
+            LOGGER.debug("Could not find host in metadata for address {}", localNativeTransportAddress);
+            return null;
+        }
+        statement.setConsistencyLevel(ConsistencyLevel.ONE);
+        statement.setHost(host);
+        return activeSession.execute(statement);
+    }
+
+    public InetSocketAddress localNativeTransportPort()
```
frankgh, line 174: NIT
```suggestion
    @Override
    public InetSocketAddress localNativeTransportPort()
```

common/src/main/java/com/datastax/driver/core/DriverExtensions.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+public class DriverExtensions
+{
+    /**
+     * Check if a host has active connections
+     *
+     * @param host the host to check
+     * @return true if the host has active connections, false otherwise
+     */
+    public static boolean hasActiveConnections(Host host)
```
frankgh: should we annotate with visible for testing, since this is only used for tests

src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -107,6 +113,55 @@ public MainModule(Path confPath)
...
+     *
+     * @param vertx             the vertx instance
+     * @param cassandraInstance the cassandra instance configuration
+     * @param versionProvider   a Cassandra version provider
+     * @param sidecarVersion    the version of the Sidecar from the current binary
+     * @param jmxConfiguration  the configuration for the JMX Client
+     * @param session           the CQL Session provider
+     * @return the build instance metadata object
+     */
+    private static InstanceMetadata buildInstanceMetadata(Vertx vertx,
```
frankgh: can we keep private methods at the bottom of the class?

frankgh (commented):
Work in progress review

src/main/java/org/apache/cassandra/sidecar/cluster/CQLSessionProviderImpl.java
```diff
@@ -0,0 +1,198 @@
...
+    public CQLSessionProviderImpl(SidecarConfiguration configuration,
+                                  NettyOptions options)
+    {
+        DriverConfiguration driverConfiguration = configuration.driverConfiguration();
+        this.contactPoints = driverConfiguration.contactPoints();
+        this.localInstances = configuration.cassandraInstances()
+                                           .stream()
+                                           .map(i -> new InetSocketAddress(i.host(), i.port()))
+                                           .collect(Collectors.toList());
+        ;
```
frankgh: extra `;`?

frankgh (approved):

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
```diff
@@ -90,20 +99,43 @@ public Metadata metadata()
     @Nullable
     public NodeSettings nodeSettings()
     {
-        Session activeSession = session.localCql();
-        if (activeSession == null)
-        {
-            return null;
-        }
+        ResultSet rs = executeLocal("select release_version, partitioner from system.local");
+        if (rs == null) return null;
```
yifan-c: I do not think the inlined if statement is preferred. Can you break it into multiple lines?

common/src/main/java/com/datastax/driver/core/DriverExtensions.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+package com.datastax.driver.core;
+
+import java.net.InetSocketAddress;
+
+/**
+ * A collection of methods that require access to package-private members in the datastax driver.
+ */
+public class DriverExtensions
```
yifan-c: nit: `DriverUtils` might be a more conventional name.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraFactory.java
```diff
@@ -43,13 +45,16 @@ public CassandraFactory(DnsResolver dnsResolver, String sidecarVersion)
     /**
      * Returns a new adapter for Cassandra 4.0 clusters.
      *
-     * @param session   the session to the Cassandra database
-     * @param jmxClient the JMX client to connect to the Cassandra database
+     * @param session                     the session to the Cassandra database
+     * @param jmxClient                   the JMX client to connect to the Cassandra database
+     * @param localNativeTransportAddress the address and port on which this instance is configured to listen
```
yifan-c, line 50: There is no need to align the parameter descriptions, which leads to unrelated changes when adding a new param.

common/src/main/java/com/datastax/driver/core/DriverExtensions.java
yifan-c: +1 on adding the annotation

common/src/main/java/com/datastax/driver/core/DriverExtensions.java
```diff
@@ -0,0 +1,67 @@
...
+     * @return the {@link Host}           instance if found, else null
+     */
+    public static Host getHost(Metadata metadata, InetSocketAddress localNativeTransportAddress)
+    {
+        // Because the driver can sometimes mess up the broadcast address, we need to search by endpoint
+        // which is what it actually uses to connect to the cluster. Therefore, create a TranslatedAddressEndpoint
+        // to use for searching. It has to be one of these because that's what the driver is using internally,
+        // and the `.equals` method used when searching checks the type explicitly.
+        TranslatedAddressEndPoint endPoint = new TranslatedAddressEndPoint(localNativeTransportAddress);
+        return metadata.getHost(endPoint);
```
yifan-c: Consider caching the lookup result? The `metadata.getHost` lookup is an O(n) operation.

common/src/main/java/org/apache/cassandra/sidecar/common/CQLSessionProvider.java
```diff
@@ -18,159 +18,36 @@
...
-        // exception.
-        if (cause instanceof DriverException) throw ((DriverException) cause).copy();
-        else throw new DriverInternalError("Unexpected exception thrown", cause);
-    }
+    /**
+     * Gets the current Session object if it already exists.
+     * Otherwise, returns null.
+     * @return the connected {@link Session} object if available. Null otherwise.
+     */
+    Session getIfConnected();
```
yifan-c: Add `@Nullable` too?

common/src/main/java/org/apache/cassandra/sidecar/common/ICassandraAdapter.java
```diff
@@ -33,8 +38,35 @@ public interface ICassandraAdapter
      */
     Metadata metadata();
 
+    /**
+     * The {@link NodeSettings} instance for this instance.
```
yifan-c: nit: it is a bit hard to read with "instance" referring to different things. I would drop the first "instance" in the sentence.

src/main/java/org/apache/cassandra/sidecar/cluster/CQLSessionProviderImpl.java
```diff
@@ -0,0 +1,197 @@
...
+                    cluster.close();
+                }
+                catch (Exception ex)
+                {
+                    logger.error("Failed to close cluster in cleanup", ex);
+                }
+            }
+        }
+        return session;
+    }
```
yifan-c: How about minimizing the try-catch block by returning the non-null session early?

```suggestion
    public synchronized Session get()
    {
        Cluster cluster = null;
        if (session != null)
        {
            return session;
        }

        try
        {
            logger.info("Connecting to cluster using contact points {}", contactPoints);

            LoadBalancingPolicy lbp = new SidecarLoadBalancingPolicy(localInstances, localDc, numConnections);
            // Prevent spurious reconnects of ignored down nodes on `onUp` events
            QueryOptions queryOptions = new QueryOptions().setReprepareOnUp(false);
            cluster = Cluster.builder()
                             .addContactPointsWithPorts(contactPoints)
                             .withReconnectionPolicy(reconnectionPolicy)
                             .withoutMetrics()
                             .withLoadBalancingPolicy(lbp)
                             .withQueryOptions(queryOptions)
                             // tests can create a lot of these Cluster objects, to avoid creating HWTs and
                             // event thread pools for each we have the override
                             .withNettyOptions(nettyOptions)
                             .build();
            session = cluster.connect();
            logger.info("Successfully connected to Cassandra!");
        }
        catch (Exception e)
        {
            logger.error("Failed to reach Cassandra", e);
            if (cluster != null)
            {
                try
                {
                    cluster.close();
                }
                catch (Exception ex)
                {
                    logger.error("Failed to close cluster in cleanup", ex);
                }
            }
        }
        return session;
    }
```

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -196,7 +217,6 @@ private void healthCheckInternal()
             // Unregister the host listener and nullify the session in order to get a new object.
             markAsDownAndMaybeNotify();
             maybeUnregisterHostListener(activeSession);
-            cqlSessionProvider.close();
```
yifan-c, line 219: is it removed by mistake?

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -166,8 +177,18 @@ private void healthCheckInternal()
...
+            SimpleStatement healthCheckStatement =
+            new SimpleStatement("select release_version, partitioner from system.local");
+            Host host = DriverExtensions.getHost(activeSession.getCluster().getMetadata(), localNativeTransportAddress);
+            if (host == null)
+            {
+                LOGGER.warn("Could not find host in cluster metadata by address and port {}",
+                            localNativeTransportAddress);
+                return;
+            }
+            healthCheckStatement.setHost(host);
```
yifan-c, line 206: Consistency level needs to be `ONE`

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -166,8 +177,18 @@ private void healthCheckInternal()
 
         try
         {
-            Row oneResult = activeSession.execute("select release_version, partitioner from system.local")
-                                         .one();
+            // NOTE: We cannot use `executeLocal` here as there may be no adapter yet.
+            SimpleStatement healthCheckStatement =
+            new SimpleStatement("select release_version, partitioner from system.local");
+            Host host = DriverExtensions.getHost(activeSession.getCluster().getMetadata(), localNativeTransportAddress);
```
yifan-c: It is an indicator that the helper methods, `executeLocal`, do not belong to adapter. How about moving them into a utility class? So that the code is better reused. For example, the alternative is 

```java 
public class SessionUtils
{
    public ResultSet executeLocal(Session session,
                                  InetSocketAddress localNativeTransportAddress,
                                  String query)
    {
        return executeLocal(session, localNativeTransportAddress, new SimpleStatement(query));
    }

    public ResultSet executeLocal(Session session,
                                  InetSocketAddress localNativeTransportAddress,
                                  Statement statement)
    {
        Host host = DriverExtensions.getHost(session.getCluster().getMetadata(), localNativeTransportAddress);
        if (host == null)
        {
            LOGGER.warn("Could not find host in cluster metadata by address and port {}",
                        localNativeTransportAddress);
            throw new NoHostAvailableException(Collections.emptyMap());
        }
        statement.setHost(host);
        statement.setConsistencyLevel(ConsistencyLevel.ONE);
        return session.execute(statement);
    }
}
```

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -284,7 +314,7 @@ public boolean isUp()
     public void close()
     {
         markAsDownAndMaybeNotify();
-        Session activeSession = cqlSessionProvider.close();
+        Session activeSession = cqlSessionProvider.getIfConnected();
```
yifan-c, line 356: where is the session closed?

src/main/java/org/apache/cassandra/sidecar/cluster/SidecarLoadBalancingPolicy.java
```diff
@@ -0,0 +1,217 @@
...
+
+/**
+ * The SidecarLoadBalancingPolicy is designed to ensure that the Cassandra Metadata objects associated with the
+ * CqlSessionProvider have enough non-local hosts in their allowed connections to be kept up-to-date
+ * even if the local Cassandra instances are down/have their native transport disabled.
+ * NOTE: This policy won't work with a child policy that is token-aware
+ */
+class SidecarLoadBalancingPolicy implements LoadBalancingPolicy
+{
+    public static final int MIN_ADDITIONAL_CONNECTIONS = 2;
```
yifan-c: "Additional" is relative/implicit, how about `MIN_NON_LOCAL_CONNECTIONS`? Or add a comment to explain what is the additional connections.

src/main/java/org/apache/cassandra/sidecar/cluster/SidecarLoadBalancingPolicy.java
```diff
@@ -0,0 +1,217 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(SidecarLoadBalancingPolicy.class);
+    private final Set<Host> selectedHosts = new HashSet<>();
+    private final Set<InetSocketAddress> localHostAddresses;
+    private final LoadBalancingPolicy childPolicy;
+    private final int totalRequestedConnections;
+    private final Random secureRandom = new SecureRandom();
```
yifan-c: I do not think secure random is desired. The random instance is only used to shuffle. It does not require the output to be cryptographic. Using `Random` should suffice.

src/main/java/org/apache/cassandra/sidecar/cluster/SidecarLoadBalancingPolicy.java
```diff
@@ -0,0 +1,217 @@
...
+        Iterator<Host> hostIterator = sourceHosts.iterator();
+        while (selectedHosts.size() < totalRequestedConnections && hostIterator.hasNext())
+        {
+            Host host = hostIterator.next();
+            if (!selectedHosts.contains(host) && host.isUp())
+            {
+                selectedHosts.add(host);
+                hostIterator.remove();
+            }
+        }
```
yifan-c: The implementation would be easier to read by splitting `this.allHosts` into local and non-local host lists, and add the first N hosts from the shuffled non-local host list. 
The current one iterates through list and remove items (from different methods), which is a bit harder to digest. 

```java
        Map<Boolean, List<Host>> partitionedHosts = allHosts.stream()
                                                            .collect(Collectors.partitioningBy(host -> localHostAddresses.contains(host.getEndPoint().resolve())));
        List<Host> localHosts = partitionedHosts.get(true);
        if (localHosts == null)
        {
            // log warn
        }
        else
        {
            selectedHosts.addAll(localHosts);
        }
        List<Host> nonLocalHosts = partitionedHosts.get(false);
        if (nonLocalHosts == null)
        {
            // log warn and exit
        }

        Collections.shuffle(nonLocalHosts, this.secureRandom);
        // add first N hosts
```

src/main/java/org/apache/cassandra/sidecar/cluster/SidecarLoadBalancingPolicy.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    @Override
+    public synchronized void onDown(Host host)
+    {
+        // Don't remove local addresses from the selected host list
+        if (localHostAddresses.contains(host.getBroadcastRpcAddress()))
+        {
+            return;
+        }
```
yifan-c: Why prevent removing a down local host? It cannot serve queries. 
Besides that, how about adding a warn log?

src/main/java/org/apache/cassandra/sidecar/config/DriverConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.List;
+
+/**
+ * The driver configuration to use when connecting to Cassandra
+ */
+public interface DriverConfiguration
+{
+    /**
+     * A list of contact points to use for initial connection to Cassandra.
+     * At least 2 non-replica nodes are recommended.
```
yifan-c, line 31: Does "non-replica nodes" imply that the config is heterogeneous across sidecar instances? My understanding is that the contact points are the subset of the seed nodes and it is the same across all sidecar instances.

src/main/java/org/apache/cassandra/sidecar/config/DriverConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+     * @return a list of contact points
+     */
+    List<InetSocketAddress> contactPoints();
+
+    /**
+     * The number of connections other than locally-managed nodes to use.
+     * The minimum is 2 - if your value is &lt; 2, the Sidecar will use 2.
```
yifan-c: I would just use 'less than', instead of the html escaped characters.

src/main/java/org/apache/cassandra/sidecar/config/DriverConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+     * @return the number of connections to make to the cluster.
+     */
+    int numConnections();
+
+    /**
+     * The local datacenter to use for non-local queries to the cluster.
+     * @return the local datacenter, or null if no local datacenter is specified.
+     */
+    String localDc();
```
yifan-c, line 47: `String localDc()` means the sidecar instances in different DC have inconsistent configuration of the field. 
It could be a pain for operations. 

Sidecar should be able to figure out the name of the local DC via java driver. The config here could be `boolean shouldQueryLocalDcOnly()` (or some other name) to prevent the queries to the other DCs, when true.

src/test/integration/org/apache/cassandra/sidecar/cluster/SidecarLoadBalancingPolicyTest.java
```diff
@@ -0,0 +1,133 @@
...
+    {
+        return SIDECAR_MANAGED_INSTANCES; // we only want to manage the first 2 instances in the "cluster"
+    }
+
+    @CassandraIntegrationTest(nodesPerDc = 6)
+    public void shouldMaintainMinimumConnections() throws ExecutionException, InterruptedException
+    {
+        Set<Host> hosts = sidecarTestContext.session().getCluster().getMetadata().getAllHosts();
+        List<Host> connectedHosts = getConnectedHosts(hosts);
+        // We manage 2 hosts, and ask for an additional 4 (the default) for connections.
```
yifan-c: ask for an additional 2?

src/test/integration/org/apache/cassandra/sidecar/cluster/SidecarLoadBalancingPolicyTest.java
```diff
@@ -0,0 +1,133 @@
...
+        // Therefore, we expect 4 hosts to have connections at startup.
+        int expectedConnections = SIDECAR_MANAGED_INSTANCES + SidecarLoadBalancingPolicy.MIN_ADDITIONAL_CONNECTIONS;
+        assertThat(connectedHosts.size()).isEqualTo(expectedConnections);
+        // Now, shut down one of the hosts and make sure that we connect to a different node
+        UpgradeableCluster cluster = sidecarTestContext.cluster();
+        IUpgradeableInstance inst = shutDownNonLocalInstance(
+        cluster,
+        sidecarTestContext.instancesConfig().instances());
```
yifan-c: There is no indentation for the method parameters. 

```suggestion
        IUpgradeableInstance inst = shutDownNonLocalInstance(sidecarTestContext.cluster(),
                                                             sidecarTestContext.instancesConfig().instances());
```

src/test/integration/org/apache/cassandra/sidecar/common/CQLSessionProviderTest.java
```diff
@@ -0,0 +1,176 @@
...
+                                    .send()
+                                    .onSuccess(response -> assertHealthCheckOk(response, context))
+                           )
+                           .onSuccess(_ignored -> context.completeNow())
+                           .onFailure(context::failNow);
+                       }
+        );
+    }
+
+
```
yifan-c: - Less indentation 
- use `Future.future` instead of new thread in compose
```suggestion
        testWithClient(context, false, webClient -> {

            // To start, both instances are stopped, so we should get 503s for both
            buildInstanceHealthRequest(webClient, "1")
            .send()
            .onSuccess(response -> assertHealthCheckFailed(response, context))
            .compose(_ignored ->
                     buildInstanceHealthRequest(webClient, "2")
                     .send()
                     .onSuccess(response -> assertHealthCheckFailed(response, context)))
            .compose(_ignored ->
                     buildKeyspaceRequest(webClient)
                     .send()
                     // With no instances available in the cluster, keyspace requests should fail
                     .onSuccess(response -> assertKeyspaceFailed(response, context)))
            .compose(_ignored -> {
                // Start instance 1 and check both again
                return Future.future(promise -> {
                    cluster.get(1).startup();
                    // Instance 1 should now be up - wait for reconnect before testing
                    Uninterruptibles.sleepUninterruptibly(5000, TimeUnit.MILLISECONDS);
                    promise.complete();
                });
            })
            .compose(_ignored ->
                     buildInstanceHealthRequest(webClient, "1")
                     .send()
                     .onSuccess(response -> assertHealthCheckOk(response, context)))
            .compose(_ignored ->
                     buildInstanceHealthRequest(webClient, "2")
                     .send()
                     .onSuccess(response -> assertHealthCheckFailed(response, context))
            )
            .compose(_ignored ->
                     // Even with only 1 instance connected/up, we should still have keyspace metadata
                     buildKeyspaceRequest(webClient)
                     .send()
                     .onSuccess(response -> assertKeyspaceOk(response, context)))
            .compose(_ignored -> {
                // Start instance 2 and check both again
                return Future.future(promise -> {
                    cluster.get(2).startup();
                    ClusterUtils.assertInRing(cluster.get(1), cluster.get(2));
                    // Instance 2 should now be up - wait for gossip
                    // and reconnect before testing
                    Uninterruptibles.sleepUninterruptibly(5000, TimeUnit.MILLISECONDS);
                    promise.complete();
                });
            })
            .compose(_ignored ->
                     buildInstanceHealthRequest(webClient, "1")
                     .send()
                     .onSuccess(response -> assertHealthCheckOk(response, context)))
            .compose(_ignored ->
                     buildInstanceHealthRequest(webClient, "2")
                     .send()
                     .onSuccess(response -> assertHealthCheckOk(response, context))
            )
            .onSuccess(_ignored -> context.completeNow())
            .onFailure(context::failNow);
        });
    }
```

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
frankgh: this removal is correct. The session is now shared across all the instances managed by sidecar, so closing here is not desired. The lifecycle of the session is managed by the application, so on application shutdown it should close the session

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
frankgh: Session is closed here: https://github.com/apache/cassandra-sidecar/pull/74/files#diff-336fd041be872c420259a2577eacc637ac46d1ef58d6a5da85552e9d0aa6cc1eR268

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraFactory.java
JeetKunDoug: @frankgh  keeps making me do this - if we can pick one style or the other and stick to it I'm happy to not align or align - I tend to agree w/ @yifan-c here but don't have a strong opinion one way or another.  Git/GitHub are smart enough to be able to highlight that it's just a whitespace change when you do make those kinds of alignment changes, so I don't think it's a huge deal, but I also get the "fewer changes in a commit is good" vibe.

frankgh: this is the default behavior when applying code style to the file, so I'd prefer to use it since other developers might also apply code style this way.

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
yifan-c: Got it. 

This comment should be removed along with the code too. 
> and nullify the session in order to get a new object.

yifan-c:
#80 was just merged. It has conflicting changes. Please rebase.

common/src/main/java/com/datastax/driver/core/DriverExtensions.java
JeetKunDoug: This is a super-secret wish that Java would implement C#-style extension methods 😄 

I'll rename it.

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
JeetKunDoug: Good catch!

src/main/java/org/apache/cassandra/sidecar/cluster/SidecarLoadBalancingPolicy.java
JeetKunDoug: Local hosts should always be in the list even if they are down - the client will not use them if it thinks they are down, and this way we know they are always in the list and don't have to do anything more with down nodes.

I'm not sure if a node going down (local or otherwise) is worth a WARN - maybe debug, but it's not an uncommon occurrence so spitting out a WARN here seems overly noisy.

src/main/java/org/apache/cassandra/sidecar/config/DriverConfiguration.java
JeetKunDoug: It doesn't have to be seed nodes - that is _one way_ to configure it, but any set of non-replica nodes should allow the driver to stay connected through replica-safe operations.

src/main/java/org/apache/cassandra/sidecar/config/DriverConfiguration.java
JeetKunDoug: This is a "chicken and egg" problem - in order for the sidecar to get the local DC, it has to connect... to connect, it needs to build a load balancing policy that needs to know the local DC. We could connect, query a local instance for the local DC, disconnect, and connect again, but that seems like overkill to avoid configuration by DC.

src/test/integration/org/apache/cassandra/sidecar/cluster/SidecarLoadBalancingPolicyTest.java
JeetKunDoug: On a side note, I hate the fact that there's no indentation for method parameters in the default IntelliJ code style - it likes to do stuff like this. Not sure if there's a way to fix it, but then I suppose we'd have to reformat everything again.

src/test/integration/org/apache/cassandra/sidecar/common/CQLSessionProviderTest.java
JeetKunDoug: Unfortunately the indentation is what the code style does - I can de-dent it, but if you reformat the file or touch the curly-brace at the bottom it's just going to reformat back to this indentation.

`Future.future` just doesn't seem to do anything - it looks cool, but doesn't actually _run_ in the test, so it fails if you replace the thread with `Future.future(...)` instead.

common/src/main/java/com/datastax/driver/core/DriverExtensions.java
JeetKunDoug: Good call - I've pushed the caching to the Adapter/Delegate level as DCLP-backed instance members as each one checks its own host, rather than trying to cache in some static member here.

src/main/java/org/apache/cassandra/sidecar/cluster/SidecarLoadBalancingPolicy.java
```diff
@@ -193,25 +194,33 @@ private void addLocalHostsToSelected(List<Host> sourceHosts)
...
+        int requiredNewHosts = this.totalRequestedConnections - selectedHosts.size();
+        if (requiredNewHosts > 0)
         {
-            LOGGER.warn("Requested number of instances (local + additional) ({}) is greater " +
-                        "than the number of UP hosts in the cluster ({})",
-                        totalRequestedConnections, selectedHosts.size());
+            List<Host> nonLocalHosts = partitionedHosts.get(false);
+            if (nonLocalHosts == null)
+            {
+                LOGGER.warn("Did not find any local hosts in the complete host list!");
```
yifan-c:
```suggestion
                LOGGER.warn("Did not find any non-local hosts in the complete host list!");
```

src/main/java/org/apache/cassandra/sidecar/cluster/SidecarLoadBalancingPolicy.java
```diff
@@ -193,25 +194,33 @@ private void addLocalHostsToSelected(List<Host> sourceHosts)
...
-        // Copy the list to allow us to remove hosts as we build the list
-        List<Host> sourceHosts = new ArrayList<>(this.allHosts);
-        addLocalHostsToSelected(sourceHosts);
-        Collections.shuffle(sourceHosts, this.secureRandom);
-        Iterator<Host> hostIterator = sourceHosts.iterator();
-        while (selectedHosts.size() < totalRequestedConnections && hostIterator.hasNext())
+        Map<Boolean, List<Host>> partitionedHosts = allHosts.stream()
+                                                            .collect(Collectors.partitioningBy(
+                                                            host -> localHostAddresses.contains(
+                                                            host.getEndPoint().resolve())));
```
yifan-c: I found declare a local variable reads better than the inlined lambda that is formatted to a small block due to indentation. 
```suggestion
        Predicate<Host> hostPartitioner = host -> localHostAddresses.contains(host.getEndPoint().resolve());
        Map<Boolean, List<Host>> partitionedHosts = allHosts.stream()
                                                            .collect(Collectors.partitioningBy(hostPartitioner));
```

src/main/java/org/apache/cassandra/sidecar/cluster/SidecarLoadBalancingPolicy.java
```diff
@@ -193,25 +194,33 @@ private void addLocalHostsToSelected(List<Host> sourceHosts)
```
yifan-c: Method `addLocalHostsToSelected(...)` (above this line) is not longer used and can be deleted.

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/91f31399042a444c2ccfc55d28d64bfe36a72a74

PR #75, opened 2023-10-27
CASSANDRASC-80 HealthCheckPeriodicTask execute never completes the pr…

frankgh:
…omise when instances are empty

When the HealthCheckPeriodicTask executes, and the instances are null or empty, the promise never completes. This prevents subsequent scheduled health checks to take place because the PeriodicTaskExecutor will only schedule the new task only if no other tasks are active. This makes the HealthCheckPeriodicTask to never perform health checks when this condition is encountered.

In this commit, we fix the issue by completing the promise when this condition is encountered.

[CASSANDRASC-80](https://issues.apache.org/jira/browse/CASSANDRASC-80)

### DIFF
--- a/src/main/java/org/apache/cassandra/sidecar/cluster/InstancesConfig.java
+++ b/src/main/java/org/apache/cassandra/sidecar/cluster/InstancesConfig.java
@@ -22,6 +22,7 @@ import java.util.List;
 import java.util.NoSuchElementException;
 
 import org.apache.cassandra.sidecar.cluster.instance.InstanceMetadata;
+import org.jetbrains.annotations.NotNull;
 
 /**
  * Maintains metadata of instances maintained by Sidecar.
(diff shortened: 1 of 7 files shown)


src/main/java/org/apache/cassandra/sidecar/cluster/InstancesConfig.java
```diff
@@ -22,15 +22,20 @@
...
 import org.apache.cassandra.sidecar.cluster.instance.InstanceMetadata;
+import org.jetbrains.annotations.NotNull;
 
 /**
  * Maintains metadata of instances maintained by Sidecar.
  */
 public interface InstancesConfig
 {
     /**
+     * Returns metadata associated with the Cassandra instances managed by this Sidecar. The implementer
```
frankgh, line 33: Clarification for the implementer that the instances list must not be null. An empty collection is acceptable

src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTask.java
```diff
@@ -60,6 +60,11 @@ default TimeUnit initialDelayUnit()
      * Defines the task body.
      * The method can be considered as executing in a single thread.
      *
+     * <br><b>NOTE:</b> It is very important that the promise is completed (as either succeeded or failed), when the
```
frankgh: we need to add this note, to make sure implementing classes are aware of this detail

yifan-c:
closing since the patch has been merged.

PR #76, opened 2023-10-27
CASSANDRASC-81 Improve TokenRangeReplicasResponse payload

frankgh:
The `TokenRangeReplicasResponse` returns a list with `ReplicaMetadata`. This information is used by clients to lookup replica metadata. The lookup is done with the replica information, which consists of the `ip:port`.

Clients are looping over the `ReplicaMetadata` list and matching IP and port to retrieve the metadata object. Instead, we can improve the payload by changing the data structure from a list to a map, and have clients lookup by the replica (ip + port), without having to loop.

[CASSANDRASC-81](https://issues.apache.org/jira/browse/CASSANDRASC-81)


arjunashok:
Hey Francisco, Not sure I understand the need for this optimization. I see that it optimizes for the lookups we perform in the tests, but I do not think this should affect the API response structure.

I say this because AFAIK, all the use cases in the cassandra analytics (which is the only consumer of this endpoint) filter the replicas by the state or status of the instance and do not benefit from having a map keyed by the instance:port combination.

When you say `The lookup is done with the replica information, which consists of the ip:port`, which usage are you referring to?

frankgh:
> When you say `The lookup is done with the replica information, which consists of the ip:port`, which usage are you referring to?

@arjunashok yes, the usage is here: https://github.com/apache/cassandra-analytics/pull/17/files#diff-a9d37004bd1696575d9651feb1dc9a4782c3aa456ae9fc383cd158e45baeed9aR527 . I think we unnecessarily have to parse the ip and port using `HostAndPort` class, and then we loop over the metadata until we match ip + port.

arjunashok:
Agreed. Yes looks like we're depending on the lookup to categorize ranges by replicas. Makes sense.

arjunashok (approved):
LGTM, +1

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/d8e9e2359db1f5561eb872d28ab16098b5a62c1d

PR #77, opened 2023-11-07
CASSANDRASC-83 Require gossip to be enabled for ring and token ranges…

frankgh:
… mapping endpoints

The ring and token ranges mapping endpoints leverage JMX to produce the payload. Ring and token ranges mapping information is derived from gossip information available to the Cassandra instance. If gossip has been disabled, the information might provide an inconsistent view of the cluster which is undesirable. We should instead return a 503 (Service Unavailable) to the client whenever the client tries to access these endpoints and gossip has been disabled.

In this commit, we ensure the endpoints return a 503 (Service Unavailable) when gossip is disabled.

### DIFF
--- a/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/exception/OperationUnavailableException.java
+++ b/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/exception/OperationUnavailableException.java
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


dineshjoshi (approved):

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/exception/OperationUnavailableException.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * The requirement for the operation
+     */
+    public enum OperationType
+    {
+        /**
+         * Gossip is required for the operation, but gossip has been disabled
+         */
+        GOSSIP_DISABLED("Gossip is required for the operation but it is disabled"),
```
yifan-c: nit: `GOSSIP_DISABLED` is the reason, not the operation type. Maybe rename the enum to `Cause` or so, or rename the variant name to `GOSSIP`, which is a type. A follow up question is what if an operation is unavailable due to multiple causes, how is the exception going to represent it? It is a bit inflexible. Probably consider constructing the exception with just `string` error message (and `operationType()` is not used), but declare a few constants for common exceptions, and for simplicity. For example, 

```java
public class OperationUnavailableException extends RuntimeException
{
    public static final OperationUnavailableException GOSSIP_DISABLED = new OperationUnavailableException("Gossip is required for the operation but it is disabled");

    public OperationUnavailableException(String errorMessage)
    {
        super(errorMessage);
    }
}
```

frankgh: I think the suggestion looks cleaner. Applying the suggestion!

PR #78, opened 2023-11-14
CASSANDRASC-76: Sidecar does not handle keyspaces and table names wit…

frankgh:
…h mixed case

Cassandra Sidecar does not properly handle API requests for endpoints that have keypaces or table names that need to be quoted, for example when names have mixed case or when the name is a reserved keyword in Cassandra.

In this commit, we perform special handling when the keyspace or table names are quoted in the path params. We add tests to ensure that handling is correct.

Additionally, we fix the validation for keyspaces and table names without quotes and add special validation for quoted names.


build.gradle
```diff
@@ -323,6 +323,8 @@ tasks.register("integrationTest", Test) {
         jvmArgs(JDK11_OPTIONS)
         println("JVM arguments for $project.name are $allJvmArgs")
     }
+    // Disable direct memory allocator as it doesn't release properly
+    systemProperty "cassandra.netty_use_heap_allocator", "true"
```
frankgh, line 327: we can now run two integration tests concurrently in CI with this setting

common/src/main/java/org/apache/cassandra/sidecar/common/data/Keyspace.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.common.data;
+
+/**
+ * Represents a Cassandra keyspace
+ */
+public class Keyspace extends Name
```
yifan-c: This data class is unnecessary. The keyspace value can use the type `Name` directly.

common/src/main/java/org/apache/cassandra/sidecar/common/data/Table.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.common.data;
+
+/**
+ * Represents a Cassandra table
+ */
+public class Table extends Name
```
yifan-c: It is unnecessary too. I think it is valuable to have less code.

frankgh:
I had to rebase and force push to resolve conflicts

common/src/main/java/org/apache/cassandra/sidecar/common/data/Name.java
```diff
@@ -0,0 +1,110 @@
...
+    /**
+     * Constructs a {@link Name} object with the provided {@code unquotedName}, and
+     * {@code isQuotedFromSource} parameters. For example, if the user-provided input was
+     * {@code "\"QuotedName\""}, then the parameters for this constructor should be initialized with
+     * {@code new Name("QuotedTableName", true)}.
+     *
+     * @param unquotedName       the unquoted name for the table
+     * @param isQuotedFromSource whether the name was quoted from source
+     */
+    Name(String unquotedName, boolean isQuotedFromSource)
```
yifan-c: Should it be named as `isQuotedInSource`? Frankly, the "source" is a bit unclear when reading. The value is from the URL segment. Maybe just `isQuoted` is sufficient for the purpose of `Name`.

src/main/java/org/apache/cassandra/sidecar/utils/CassandraInputValidator.java
```diff
@@ -65,13 +67,16 @@ public CassandraInputValidator()
...
-        validatePattern(keyspace, "keyspace");
-        if (validationConfiguration.forbiddenKeyspaces().contains(keyspace))
+        String unquoted = removeQuotesIfNecessary(keyspace);
+        // whether the input is quoted from the source
+        boolean isQuoted = !unquoted.equals(keyspace);
+        validatePattern(unquoted, keyspace, "keyspace", isQuoted);
+        if (validationConfiguration.forbiddenKeyspaces().contains(unquoted))
             throw new HttpException(HttpResponseStatus.FORBIDDEN.code(), "Forbidden keyspace: " + keyspace);
-        return keyspace;
+        return new Keyspace(unquoted, isQuoted);
```
yifan-c: The implementation expects the value provided in the url to be explicitly quoted. Getting rid of the quotations and adding them back in `Name` seems unnecessary. 
Just passing the input value as-is (either a quoted string or not) should work the same. Right?

common/src/main/java/org/apache/cassandra/sidecar/common/data/Keyspace.java
frankgh: done

common/src/main/java/org/apache/cassandra/sidecar/common/data/Table.java
frankgh: done

common/src/main/java/org/apache/cassandra/sidecar/common/data/Name.java
frankgh: we no longer have this constructor, so it's no longer relevant

src/main/java/org/apache/cassandra/sidecar/utils/CassandraInputValidator.java
frankgh: done

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/31fa33fcb446e522947f899d948de4042be04c62

PR #79, opened 2023-11-16
CASSANDRASC-82 Expose additional SSL configuration options for the Sidecar Service

frankgh:


src/test/java/org/apache/cassandra/sidecar/config/SidecarConfigurationTest.java
```diff
@@ -338,6 +339,15 @@ void validateDefaultSslConfiguration(SslConfiguration config)
         assertThat(config.truststore().password()).isEqualTo("password");
         assertThat(config.truststore().reloadStore()).isFalse();
         assertThat(config.truststore().checkIntervalInSeconds()).isEqualTo(-1);
+        assertThat(config.secureTransportProtocols()).contains("TLSv1.3");
```
JeetKunDoug: If we're testing the default here, we should test that `secureTransportProtocols` contains all of the defaults (`TLSv1.2` _and_ `TLSv1.3`) and _nothing else_ - probably use `.containsExactly` here?

JeetKunDoug: We should also have a test that specifically tests setting secureTransportProtocols and cipherSuites to something other than the defaults, right?

JeetKunDoug: Ok - so `validateDefaultSslConfiguration` is a terribly-named method - this validates the values in `sidecar_ssl.yaml` not the _defaults_ - so, can we please add something that tests the _defaults_ (and maybe add some default cipher suites (or does providing an empty list just use whatever the JVM is configured to use?). Also, please use `containsExactly` to make sure you _aren't_ getting the defaults above.

JeetKunDoug (changes requested):
Just some comments on the config test - everything else looks reasonable.

src/test/java/org/apache/cassandra/sidecar/config/SidecarConfigurationTest.java
frankgh: > so validateDefaultSslConfiguration is a terribly-named method

Yeah, good point. I will rename it.

> can we please add something that tests the defaults

Yeah, I guess the default is that SSL is null, so maybe we need defaults when SSL is explicitly provided, so maybe it belongs in a unit test for the SSLConfiguration object instead?

> maybe add some default cipher suites

I didn't explicitly add any default cipher suites, I'm defaulting to the existing behavior in vertx which doesn't provide any defaults.

> please use containsExactly to make sure you aren't getting the defaults above

yeah, good point, will change it

src/main/java/org/apache/cassandra/sidecar/config/yaml/SslConfigurationImpl.java
```diff
@@ -227,8 +261,29 @@ public Builder handshakeTimeoutInSeconds(long handshakeTimeoutInSeconds)
...
+
+        /**
+         * Sets the {@code cipherSuites} and returns a reference to this Builder enabling method chaining.
+         *
+         * @param cipherSuites the {@code cipherSuites} to set
+         * @return a reference to this Builder
+         */
+        public Builder cipherSuites(List<String> cipherSuites)
+        {
+            return update(b -> b.cipherSuites = cipherSuites);
```
yifan-c: nit: copy the list instead?

PR #80, opened 2023-11-16
CASSANDRASC-84 Expose additional node settings

frankgh:
Sidecar exposes settings from the Cassandra node via the node settings API endpoint. The information exposed is limited, and we need to start exposing additional information from the `system.local` table, for example `datacenter` information, owned token ranges, and the local address and port for the native protocol. This information can be consumed by Sidecar itself, as well as the Cassandra Analytics library.

In this commit, we expose additional settings for the node.

### DIFF
--- a/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
+++ b/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
@@ -96,12 +96,23 @@ public class CassandraAdapter implements ICassandraAdapter
             return null;
         }
 
-        Row oneResult = activeSession.execute("select release_version, partitioner from system.local")
+        Row oneResult = activeSession.execute("select release_version, "
+                                              + "partitioner, "
+                                              + "data_center, "
+                                              + "rpc_address, "
+                                              + "rpc_port, "
+                                              + "tokens from system.local")
                                      .one();
 
-        return new NodeSettings(oneResult.getString("release_version"),
-                                oneResult.getString("partitioner"),
-                                sidecarVersion);
+        return NodeSettings.builder()
+                           .releaseVersion(oneResult.getString("release_version"))
+                           .partitioner(oneResult.getString("partitioner"))
+                           .sidecarVersion(sidecarVersion)
+                           .datacenter(oneResult.getString("data_center"))
+                           .tokens(oneResult.getSet("tokens", String.class))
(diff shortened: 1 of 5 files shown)


adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
```diff
@@ -96,12 +96,23 @@ public NodeSettings nodeSettings()
             return null;
         }
 
-        Row oneResult = activeSession.execute("select release_version, partitioner from system.local")
+        Row oneResult = activeSession.execute("select release_version, "
```
yifan-c: nit: create constants for the column names, since there are more fields queried.

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
```diff
@@ -103,15 +135,162 @@ public boolean equals(Object other)
...
         return Objects.equals(this.releaseVersion, that.releaseVersion)
-            && Objects.equals(this.partitioner,    that.partitioner)
-            && Objects.equals(this.sidecar,        that.sidecar);
+               && Objects.equals(this.partitioner, that.partitioner)
+               && Objects.equals(this.sidecar, that.sidecar)
+               && Objects.equals(this.datacenter, that.datacenter)
+               && Objects.equals(this.rpcAddress, that.rpcAddress)
+               && Objects.equals(this.rpcPort, that.rpcPort)
+               && Objects.equals(this.tokens, that.tokens)
+        ;
```
yifan-c, line 151: accidental line break?

yifan-c (commented):
some nits and overall looks good to me. 
I think it has some conflicts with #74, which is a much larger patch. Are you fine with merging #74 first and rebase?

common/src/main/java/org/apache/cassandra/sidecar/common/NodeSettings.java
frankgh: I actually did it on purpose, so subsequent fields we add to the list don't touch the `;` , it's just personal preference.

frankgh:
> I think it has some conflicts with #74, which is a much larger patch. Are you fine with merging #74 first and rebase?

Yeah, I am okay with waiting for #74 to be merged. I'll rebase once that's ready.

frankgh:
Closed via b16b89f7f4a15441af757fa5a84b7e1e0420dcba

PR #81, opened 2023-11-30
[CASSANDRASC-86] Startup Validation Failures when Checking Sidecar Connectivity

5:
Added a random delay between retries of health checks in Sidecar client.

### DIFF
--- a/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
+++ b/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
@@ -19,6 +19,7 @@
 
 package org.apache.cassandra.sidecar.client;
 
+import java.time.Duration;
 import java.util.List;
 import java.util.Objects;
 import java.util.UUID;
@@ -64,6 +65,9 @@ import org.jetbrains.annotations.Nullable;
 public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExtension
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(SidecarClient.class);
+    private static final Duration MINIMUM_RETRY_DELAY = Duration.ofSeconds(1L);
+    private static final Duration MAXIMUM_RETRY_DELAY = Duration.ofSeconds(5L);
+    private static final RetryPolicy ONCE_PER_INSTANCE_RETRY_POLICY = new OncePerInstanceRetryPolicy(MINIMUM_RETRY_DELAY, MAXIMUM_RETRY_DELAY);
 
     protected RequestExecutor executor;
     protected final RetryPolicy defaultRetryPolicy;
@@ -93,9 +97,9 @@ public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExt
     public CompletableFuture<HealthResponse> sidecarHealth()
     {
         return executor.executeRequestAsync(requestBuilder()
-                                            .sidecarHealthRequest()
-                                            .retryPolicy(new OncePerInstanceRetryPolicy())
-                                            .build());
+                .sidecarHealthRequest()
+                .retryPolicy(ONCE_PER_INSTANCE_RETRY_POLICY)
+                .build());
     }
 
     /**
@@ -108,9 +112,9 @@ public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExt
     public CompletableFuture<HealthResponse> cassandraHealth()
     {
         return executor.executeRequestAsync(requestBuilder()
-                                            .cassandraHealthRequest()
-                                            .retryPolicy(new OncePerInstanceRetryPolicy())
-                                            .build());
+                .cassandraHealthRequest()
+                .retryPolicy(ONCE_PER_INSTANCE_RETRY_POLICY)
+                .build());
     }
 
     /**
--- a/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
+++ b/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
@@ -19,7 +19,6 @@
 
 package org.apache.cassandra.sidecar.client;
 
-import java.time.Duration;
 import java.util.List;
 import java.util.Objects;
 import java.util.UUID;
@@ -65,13 +64,11 @@ import org.jetbrains.annotations.Nullable;
 public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExtension
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(SidecarClient.class);
-    private static final Duration MINIMUM_RETRY_DELAY = Duration.ofSeconds(1L);
-    private static final Duration MAXIMUM_RETRY_DELAY = Duration.ofSeconds(5L);
-    private static final RetryPolicy ONCE_PER_INSTANCE_RETRY_POLICY = new OncePerInstanceRetryPolicy(MINIMUM_RETRY_DELAY, MAXIMUM_RETRY_DELAY);
 
     protected RequestExecutor executor;
     protected final RetryPolicy defaultRetryPolicy;
     protected final RetryPolicy ignoreConflictRetryPolicy;
+    protected final RetryPolicy oncePerInstanceRetryPolicy;
     protected RequestContext.Builder baseBuilder;
 
     public SidecarClient(SidecarInstancesProvider instancesProvider,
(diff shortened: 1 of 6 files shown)


client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -80,9 +83,9 @@ public SidecarClient(SidecarInstancesProvider instancesProvider,
     public CompletableFuture<HealthResponse> sidecarHealth()
     {
         return executor.executeRequestAsync(requestBuilder()
-                                            .sidecarHealthRequest()
-                                            .retryPolicy(new OncePerInstanceRetryPolicy())
-                                            .build());
+                .sidecarHealthRequest()
+                .retryPolicy(oncePerInstanceRetryPolicy)
+                .build());
```
yifan-c: The indentation looks wrong.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -63,13 +64,15 @@ public SidecarClient(SidecarInstancesProvider instancesProvider,
...
-                      .retryPolicy(defaultRetryPolicy);
-        executor = requestExecutor;
+        this.ignoreConflictRetryPolicy = new IgnoreConflictRetryPolicy(sidecarClientConfig.maxRetries(),
+                                                                       sidecarClientConfig.retryDelayMillis(),
+                                                                       sidecarClientConfig.maxRetryDelayMillis());
+        this.oncePerInstanceRetryPolicy = new OncePerInstanceRetryPolicy(sidecarClientConfig.minimumHealthRetryDelay(),
+                                                                         sidecarClientConfig.maximumHealthRetryDelay());
+        this.baseBuilder = new RequestContext.Builder()
+                .instanceSelectionPolicy(new RandomInstanceSelectionPolicy(instancesProvider))
+                .retryPolicy(defaultRetryPolicy);
```
yifan-c: The indentation looks wrong. 
I also feel that adding the `this.` to instance variable unnecessarily increases the number of line of changes.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClientConfigImpl.java
```diff
@@ -68,19 +76,39 @@ public long maxRetryDelayMillis()
...
+     }
+
     public static Builder builder()
     {
         return new Builder();
     }
 
     /**
-     * {@code SidecarConfig} builder static inner class.
+     * {@code SidecarConfig} builder static inner class
```
yifan-c: removing the period is unnecessary.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClientConfigImpl.java
```diff
@@ -93,40 +121,62 @@ public Builder self()
...
-         * Sets the {@code maxRetries} and returns a reference to this Builder enabling method chaining.
+         * Sets the {@code maxRetries} and returns a reference to this Builder enabling method chaining
          *
          * @param maxRetries the {@code maxRetries} to set
          * @return a reference to this Builder
          */
         public Builder maxRetries(int maxRetries)
         {
-            return update(b -> b.maxRetries = maxRetries);
+            return update(builder -> builder.maxRetries = maxRetries);
```
yifan-c: unnecessary (and unrelated) change

common/src/main/java/org/apache/cassandra/sidecar/common/utils/TimeUtils.java
```diff
@@ -0,0 +1,36 @@
...
+     *
+     * @return random duration uniformly distributed between two provided durations, with millisecond precision
+     *
+     * @throws IllegalArgumentException if minimum duration is greater than maximum duration
+     */
+    public static Duration randomDuration(Duration minimum, Duration maximum)
+    {
+        Preconditions.checkArgument(minimum.compareTo(maximum) <= 0,
+                                    "Minimum duration must be less or equal to maximum duration");
+        return Duration.ofMillis(ThreadLocalRandom.current().nextLong(minimum.toMillis(), maximum.toMillis() + 1L));
```
yifan-c, line 54: I would like to suggest an alternative. 

Instead of having `min` and `max` as the parameters of the method (and the implementation has to check that `min` is indeed not greater than `max`), you can pass `base` and `range`.
In addition to the implementation change, I do not see the necessity of using thread local random. A shared `Random` object should be suffice. You do not need to create a random object per calling thread. 

```suggestion
    public static Duration randomDuration(long base, long range)
    {
        long randomDurationMillis = base + random.nextLong() % range;
        return Duration.ofMillis(randomDurationMillis);
```

yifan-c (changes requested):
There are bunch of unrelated changes. Please revert them.

common/src/test/java/org/apache/cassandra/sidecar/common/utils/TimeUtilsTest.java
```diff
@@ -0,0 +1,43 @@
+package org.apache.cassandra.sidecar.common.utils;
```
frankgh, line 20: license header is missing here

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
5: Reverted to the previous state here and in several other places.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClientConfigImpl.java
5: Added periods back here and in several other places.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClientConfigImpl.java
5: Reverted to single-letter names here and in several other places.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
5: Removed optional `this.`es, changed indentation.

common/src/test/java/org/apache/cassandra/sidecar/common/utils/TimeUtilsTest.java
5: Added license comment here and in one other place.

common/src/main/java/org/apache/cassandra/sidecar/common/utils/TimeUtils.java
```diff
@@ -0,0 +1,55 @@
...
+
+package org.apache.cassandra.sidecar.common.utils;
+
+import java.time.Duration;
+import java.util.concurrent.ThreadLocalRandom;
+
+/**
+ * Utility class for manipulating dates, times, and durations
+ */
+public final class TimeUtils {
```
frankgh: some formatting issues here maybe?
```suggestion
public final class TimeUtils
{
```

common/src/main/java/org/apache/cassandra/sidecar/common/utils/TimeUtils.java
5: I did make that change, but, to be honest, I didn't like what the code became.
Right now, "minimum duration" and "maximum duration" are easy to understand and explain.
With the change, "minimum duration" stays, but the actual meaning of the other value becomes confusing, at least to me.
"Maximum number of milliseconds to add on top of minimum duration" is a mouthful, and still needs to be checked for being non-negative.
Can we please keep the current implementation?

common/src/main/java/org/apache/cassandra/sidecar/common/utils/TimeUtils.java
5: Fixed line wrapping.

common/src/main/java/org/apache/cassandra/sidecar/common/utils/TimeUtils.java
yifan-c: ok.

frankgh (approved):
+1 looks good

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/2eb3474d7037a2887bcd9dee1f64c2a36a7e8d26

PR #83, opened 2023-12-11
CASSANDRASC-87 Add JMX health checks during the periodic health checks

frankgh:
In this commit, we add health checks based on the JMX connectivity to the managed Cassandra instances. Additionally, we construct the NodeSettings object based on JMX. This allows the Sidecar process to be able to determine an adapter for the node even if the node is in joining state, or its binary port has been disabled.


src/test/integration/org/apache/cassandra/testing/CassandraTestContext.java
```diff
@@ -43,9 +43,4 @@ public String toString()
                + ", cluster=" + cluster
                + '}';
     }
-
-    public UpgradeableCluster getCluster()
```
frankgh, line 47: this method was already implemented in the super class as `cluster()` . Removing here

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
```diff
@@ -107,30 +99,7 @@ public Metadata metadata()
...
-        return NodeSettings.builder()
-                           .releaseVersion(oneResult.getString(RELEASE_VERSION_COLUMN_NAME))
-                           .partitioner(oneResult.getString(PARTITIONER_COLUMN_NAME))
-                           .sidecarVersion(sidecarVersion)
-                           .datacenter(oneResult.getString(DATA_CENTER_COLUMN_NAME))
-                           .tokens(oneResult.getSet(TOKENS_COLUMN_NAME, String.class))
-                           .rpcAddress(oneResult.getInet(RPC_ADDRESS_COLUMN_NAME))
-                           .rpcPort(oneResult.getInt(RPC_PORT_COLUMN_NAME))
-                           .build();
+        throw new UnsupportedOperationException("Node settings are not provided by this adapter");
```
frankgh, line 102: `CassandraAdapterDelegate` manages its own node settings object, and never delegates the call to this method. We remove the implementation here for simplicity

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -205,46 +241,84 @@ private void healthCheckInternal()
...
+     */
+    protected List<String> maybeGetTokens(LimitedStorageOperations storageOperations)
+    {
+        try
+        {
+            return storageOperations.getTokens();
+        }
+        catch (AssertionError aex)
+        {
+            // On a joining node, the JMX call will with an AssertionError; we catch this scenario to prevent failure
```
yifan-c:
```suggestion
            // On a joining node, the JMX call will fail with an AssertionError; we catch this scenario to prevent failure
```

src/test/integration/org/apache/cassandra/sidecar/common/CQLSessionProviderTest.java
```diff
@@ -125,11 +125,115 @@ void testCqlSessionProviderWorksAsExpected(VertxTestContext context, CassandraTe
         );
     }
 
+//    @CassandraIntegrationTest(nodesPerDc = 2, newNodesPerDc = 1)
```
yifan-c: It this chunk to be removed?

yifan-c (commented):
I do not find the integration test for the new endpoints. Can you add them?

yifan-c:
> I do not find the integration test for the new endpoints. Can you add them?

The test is there.

src/test/integration/org/apache/cassandra/sidecar/common/CQLSessionProviderTest.java
frankgh: cleaned

frankgh:
> > I do not find the integration test for the new endpoints. Can you add them?
> 
> The test is there.

For reference the tests were added to the `DelegateIntegrationTest` class. They are exercising the new endpoints

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/529171b1f6dee277a9087eb9da7242ce17873643

PR #85, opened 2023-12-19
CASSANDRASC-85: Expose TTL option for the create snapshot endpoint

frankgh:
In this commit, we add the ability to request a TTL for a snapshot for Cassandra 4.1+ adapters. Additionally, we allow the client to specify the TTL option.

### DIFF
--- a/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
+++ b/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
@@ -41,17 +41,17 @@ import org.apache.cassandra.sidecar.common.utils.DriverUtils;
 import org.jetbrains.annotations.Nullable;
 
 /**
- * A {@link ICassandraAdapter} implementation for Cassandra 4.0 and later
+ * A {@link ICassandraAdapter} implementation for Cassandra 4.1 and later
  */
 public class CassandraAdapter implements ICassandraAdapter
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(CassandraAdapter.class);
     protected final DnsResolver dnsResolver;
     protected final JmxClient jmxClient;
-    private final CQLSessionProvider cqlSessionProvider;
-    private final InetSocketAddress localNativeTransportAddress;
-    private final DriverUtils driverUtils;
-    private volatile Host host;
+    protected final CQLSessionProvider cqlSessionProvider;
+    protected final InetSocketAddress localNativeTransportAddress;
+    protected final DriverUtils driverUtils;
+    protected volatile Host host;
 
     public CassandraAdapter(DnsResolver dnsResolver,
                             JmxClient jmxClient,
--- a/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
+++ b/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
@@ -51,7 +51,7 @@ public class CassandraAdapter implements ICassandraAdapter
     protected final CQLSessionProvider cqlSessionProvider;
     protected final InetSocketAddress localNativeTransportAddress;
     protected final DriverUtils driverUtils;
-    protected volatile Host host;
+    private volatile Host host;
 
     public CassandraAdapter(DnsResolver dnsResolver,
                             JmxClient jmxClient,
@@ -127,7 +127,7 @@ public class CassandraAdapter implements ICassandraAdapter
         return activeSession.execute(statement);
     }
 
-    private Host getHost(Metadata metadata)
+    protected Host getHost(Metadata metadata)
     {
         if (host == null)
         {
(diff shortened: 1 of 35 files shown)


src/test/integration/org/apache/cassandra/sidecar/routes/SnapshotsHandlerIntegrationTest.java
```diff
@@ -67,6 +82,69 @@ void createSnapshotEndpointFailsWhenTableDoesNotExist(VertxTestContext context)
...
 
+    @CassandraIntegrationTest
+    void createSnapshotWithTtlFailsWhenUnitsAreNotSpecified(VertxTestContext context,
+                                                            CassandraTestContext cassandraTestContext)
+    throws InterruptedException
+    {
+        if (cassandraTestContext.version.compareTo(SimpleCassandraVersion.create(4, 1, 0)) < 0)
+        {
+            // TTL is only supported in Cassandra 4.1
+            context.completeNow();
```
frankgh, line 93: skip testing for versions < 4.1

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
```diff
@@ -41,17 +41,17 @@
...
     protected final DnsResolver dnsResolver;
     protected final JmxClient jmxClient;
-    private final CQLSessionProvider cqlSessionProvider;
-    private final InetSocketAddress localNativeTransportAddress;
-    private final DriverUtils driverUtils;
-    private volatile Host host;
+    protected final CQLSessionProvider cqlSessionProvider;
+    protected final InetSocketAddress localNativeTransportAddress;
+    protected final DriverUtils driverUtils;
+    protected volatile Host host;
```
yifan-c: hmm... exposing a volatile field to be accessible from subclasses is prone to race error.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -40,7 +40,7 @@
 import static org.apache.cassandra.sidecar.adapters.base.StorageJmxOperations.STORAGE_SERVICE_OBJ_NAME;
 
 /**
- * An implementation of the {@link StorageOperations} that interfaces with Cassandra 4.0 and later
+ * An implementation of the {@link StorageOperations} that interfaces with Cassandra 4.1 and later
```
yifan-c: The comment feels misleading that the base is for cassandra 4.1+, but the subclass (`Cassandra40StorageOperations`) is for cassandra 4.0+. Can we revise the comment in the base class? I think it is not bound to specific version.

client/src/main/java/org/apache/cassandra/sidecar/client/request/SnapshotRequest.java
```diff
@@ -29,26 +30,38 @@ abstract class SnapshotRequest<T> extends DecodableRequest<T>
...
-        return requestUri + "?includeSecondaryIndexFiles=true";
+        if (!includeSecondaryIndexFiles)
+        {
+            return requestUri + "?ttl=" + snapshotTTL;
+        }
+
+        if (snapshotTTL == null)
+        {
+            return requestUri + "?includeSecondaryIndexFiles=true";
+        }
```
yifan-c: The logic here is a bit hard to read, since the condition does not relate well with the statement inside. 

I would have this.

```java
        if (!includeSecondaryIndexFiles && snapshotTTL == null)
        {
            return requestUri;
        }

        requestUri = requestUri + '?';

        if (includeSecondaryIndexFiles)
        {
            requestUri = requestUri + "includeSecondaryIndexFiles=true";
        }

        if (snapshotTTL != null)
        {
            requestUri = requestUri + "?ttl=" + snapshotTTL;
        }

        return requestUri;
```

adapters/cassandra40/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra40/Cassandra40Adapter.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.cassandra.sidecar.common.ICassandraAdapter;
+import org.apache.cassandra.sidecar.common.JmxClient;
+import org.apache.cassandra.sidecar.common.StorageOperations;
+import org.apache.cassandra.sidecar.common.dns.DnsResolver;
+import org.apache.cassandra.sidecar.common.utils.DriverUtils;
+
+/**
+ * A {@link ICassandraAdapter} implementation for Cassandra 4.0 and later
+ */
+public class Cassandra40Adapter extends CassandraAdapter
```
yifan-c: Why create `cassandra40`, instead of creating `cassandra41` and keep the base as 4.0+?

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/49723c720e0ae5762d77dfb7568c9b290a877560

PR #86, opened 2023-12-20
CASSANDRASC-89: Update Google Guice to 7.0.0

smiklosovic:
patch by Stefan Miklosovic; reviewed by TBD for CASSANDRASC-89


frankgh (approved):
@smiklosovic we should probably run this through CI again. I was able to run this locally successfully. it seems bumping the guava-testlib helped with the ClassNofFound errors we saw in CI

server/src/main/java/org/apache/cassandra/sidecar/cluster/locator/CachedLocalTokenRanges.java
```diff
@@ -205,7 +205,9 @@ private synchronized Map<Integer, Set<TokenRange>> getCacheOrReload(Metadata met
             if (isClusterTheSame && localTokenRangesCache != null && localTokenRangesCache.containsKey(ks.getName()))
             {
                 // we don't need to rebuild if already cached
-                perKeyspaceBuilder.put(ks.getName(), localTokenRangesCache.get(ks.getName()));
+                Map<Integer, Set<TokenRange>> value = localTokenRangesCache.get(ks.getName());
```
smiklosovic, line 208: this was failing spotbugs

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationListInstanceFilesHandler.java
```diff
@@ -20,11 +20,11 @@
 
 import java.io.IOException;
 import java.util.Set;
-import javax.inject.Inject;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.inject.Inject;
```
smiklosovic, line 27: replaced everything by Inject and Singleton from Guice. Guice ones are more prevalent already. After bumping Guice to 7.0.0, javax versions of these annotations were not found anymore.

PR #87, opened 2023-12-21
CASSANDRASC-90 Token-ranges endpoint fix to unwrap token-ranges by th…

arjunashok:
…e partitioner's boundary for a single node clusters

For single node clusters, the natural token range returned by the token-ranges endpoint looks like this (seems like an invalid range, without the partitioner’s minToken) -  [9223372036854775807, 9223372036854775807] . 
Verified that this is what the storage-service range-endpoints JMX returns

The contract for the token-ranges API is to make sure ranges are unwrapped around the partitioner’s min-max tokens, which is not honored for the above case. As a result, this is the fix in sidecar for the logic used to unwrap ranges.

### Testing
- Added range validation checks for existing API response validation and additional test case to validate the fix.

### DIFF
--- a/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
+++ b/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
@@ -216,10 +216,11 @@ public class TokenRangeReplicas implements Comparable<TokenRangeReplicas>
      */
     public static List<TokenRangeReplicas> normalize(List<TokenRangeReplicas> ranges)
     {
-
         if (ranges.stream().noneMatch(r -> r.partitioner.minToken.compareTo(r.start()) == 0))
         {
-            LOGGER.warn("Partitioner based minToken does not exist in the token ranges");
+            LOGGER.warn("{} based minToken does not exist in the token ranges", ranges.stream()
+                                                                                      .findFirst()
+                                                                                      .get().partitioner.name());
         }
 
         return deoverlap(ranges);
--- a/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
+++ b/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
@@ -69,7 +69,7 @@ public class TokenRangeReplicas implements Comparable<TokenRangeReplicas>
                                                                       Partitioner partitioner,
                                                                       Set<String> replicaSet)
     {
-        if (start.compareTo(end) > 0)
+        if (start.compareTo(end) >= 0)
         {
             return unwrapRange(start, end, partitioner, replicaSet);
         }
@@ -219,7 +219,7 @@ public class TokenRangeReplicas implements Comparable<TokenRangeReplicas>
 
         if (ranges.stream().noneMatch(r -> r.partitioner.minToken.compareTo(r.start()) == 0))
         {
-            LOGGER.warn("{} based minToken does not exist in the token ranges", Partitioner.class.getName());
+            LOGGER.warn("Partitioner based minToken does not exist in the token ranges");
         }
 
         return deoverlap(ranges);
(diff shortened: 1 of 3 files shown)


adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
```diff
@@ -219,7 +219,7 @@ public static List<TokenRangeReplicas> normalize(List<TokenRangeReplicas> ranges
 
         if (ranges.stream().noneMatch(r -> r.partitioner.minToken.compareTo(r.start()) == 0))
         {
-            LOGGER.warn("{} based minToken does not exist in the token ranges", Partitioner.class.getName());
+            LOGGER.warn("Partitioner based minToken does not exist in the token ranges");
```
frankgh: it seems that the intent of the logger was to log the partitioner name ? should we make the method non-static so we have access to the configured `partitioner`?

```
LOGGER.warn("{} based minToken does not exist in the token ranges", partitioner.name());
```

src/test/integration/org/apache/cassandra/sidecar/routes/tokenrange/BaseTokenRangeIntegrationTest.java
```diff
@@ -291,4 +294,17 @@ void assertMappingResponseOK(TokenRangeReplicasResponse mappingResponse,
             .isGreaterThanOrEqualTo(replicationFactor);
         }
     }
+
+    private void validateRanges(List<TokenRangeReplicasResponse.ReplicaInfo> replicaRanges)
+    {
+        // Ranges should not be empty
+        replicaRanges.stream().forEach(r -> assertThat(r.start()).isNotEqualTo(r.end()));
```
frankgh: nit:
```suggestion
        replicaRanges.forEach(r -> assertThat(r.start()).isNotEqualTo(r.end()));
```

src/test/integration/org/apache/cassandra/sidecar/routes/tokenrange/BaseTokenRangeIntegrationTest.java
```diff
@@ -291,4 +294,17 @@ void assertMappingResponseOK(TokenRangeReplicasResponse mappingResponse,
...
+        // Ranges should include partitioner start and end
+        replicaRanges.stream()
+                     .map(TokenRangeReplicasResponse.ReplicaInfo::start)
+                     .anyMatch(s -> s.equals(Murmur3Partitioner.MINIMUM.toString()));
+        replicaRanges.stream()
+                     .map(TokenRangeReplicasResponse.ReplicaInfo::end)
+                     .anyMatch(s -> s.equals(Long.toString(Murmur3Partitioner.MAXIMUM)));
```
frankgh: should we assert something here?
```suggestion
        assertThat(replicaRanges.stream()
                                .map(TokenRangeReplicasResponse.ReplicaInfo::start)
                                .anyMatch(s -> s.equals(Murmur3Partitioner.MINIMUM.toString()))).isTrue();
        assertThat(replicaRanges.stream()
                                .map(TokenRangeReplicasResponse.ReplicaInfo::end)
                                .anyMatch(s -> s.equals(Long.toString(Murmur3Partitioner.MAXIMUM)))).isTrue();
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
frankgh: ah, it looks like this method is called from other places as well..

frankgh (commented):
Looks good. I've left some minor comments

src/test/integration/org/apache/cassandra/sidecar/routes/tokenrange/BaseTokenRangeIntegrationTest.java
arjunashok: ah, yes, needs an assert, thanks.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/TokenRangeReplicas.java
arjunashok: We could extract a single range to get the partitioner to log here. I'll make that change.

frankgh (approved):
+1 Looks good to me

yifan-c (approved):

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/cd716c4c8a9f416e9af04f6e45ca7e89ace26829

PR #88, opened 2023-12-28
CASSANDRASC-91 Returning when AbstractHandler encounters exception while extracting params from the request.

nvharikrishna:
If AbstractHandler has encounted an exception while extracting params (using extractParamsOrThrow method), then it is returning appropriate error code to the client. But it is not stopping to process the request further by calling `handleInternal` method. Returning after processing the failure.

### DIFF
--- a/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
+++ b/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
@@ -89,14 +89,12 @@ public abstract class AbstractHandler<T> implements Handler<RoutingContext>
             requestParams = extractParamsOrThrow(context);
             logger.debug("{} received request={}, remoteAddress={}, instance={}",
                          this.getClass().getSimpleName(), requestParams, remoteAddress, host);
+            handleInternal(context, request, host, remoteAddress, requestParams);
         }
         catch (Exception exception)
         {
-            processFailure(exception, context, host, remoteAddress, null);
-            return;
+            processFailure(exception, context, host, remoteAddress, requestParams);
         }
-
-        handleInternal(context, request, host, remoteAddress, requestParams);
     }
 
     /**
@@ -166,7 +164,7 @@ public abstract class AbstractHandler<T> implements Handler<RoutingContext>
     }
 
     /**
-     * Processes the failure with the given parameters
+     * Processes the failure while handling the request.
      *
      * @param cause         the cause
      * @param context       the routing context
--- a/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
+++ b/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
@@ -93,6 +93,7 @@ public abstract class AbstractHandler<T> implements Handler<RoutingContext>
         catch (Exception exception)
         {
             processFailure(exception, context, host, remoteAddress, null);
+            return;
         }
 
         handleInternal(context, request, host, remoteAddress, requestParams);
--- a/src/test/java/org/apache/cassandra/sidecar/routes/StreamSSTableComponentHandlerTest.java
+++ b/src/test/java/org/apache/cassandra/sidecar/routes/StreamSSTableComponentHandlerTest.java
@@ -221,7 +221,8 @@ public class StreamSSTableComponentHandlerTest
     void failsWhenTableNameContainsInvalidCharacters(VertxTestContext context)
     {
         WebClient client = WebClient.create(vertx);
-        String testRoute = "/keyspaces/" + TEST_KEYSPACE + "/tables/i_❤_u/snapshots/snap/components/component-Data.db";
+        String testRoute = "/keyspaces/" + TEST_KEYSPACE +
+                           "/tables/i_❤_u/snapshots/snap/components/component-Data.db";
         client.get(server.actualPort(), "localhost", "/api/v1" + testRoute)
               .send(context.succeeding(response -> context.verify(() -> {
                   assertThat(response.statusCode()).isEqualTo(BAD_REQUEST.code());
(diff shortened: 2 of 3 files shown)


frankgh (commented):
Great find, thanks for the patch. May I suggest a unit test, something along these lines:

```java
/**
 * Unit tests for {@link AbstractHandler}
 */
@ExtendWith(VertxExtension.class)
class AbstractHandlerTest
{
    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractHandlerTest.class);
    Vertx vertx;
    Server server;
    Injector injector;

    @AfterEach
    void after() throws InterruptedException
    {
        CountDownLatch closeLatch = new CountDownLatch(1);
        server.close().onSuccess(res -> closeLatch.countDown());
        if (closeLatch.await(60, TimeUnit.SECONDS))
            LOGGER.info("Close event received before timeout.");
        else
            LOGGER.error("Close event timed out.");
    }

    @Test
    void failsWhileParsingParams(VertxTestContext context) throws InterruptedException
    {
        injector = Guice.createInjector(Modules.override(new MainModule())
                                               .with(new TestModule()));
        vertx = injector.getInstance(Vertx.class);

        Router router = injector.getInstance(Router.class);
        InstanceMetadataFetcher metadataFetcher = injector.getInstance(InstanceMetadataFetcher.class);
        ExecutorPools executorPools = injector.getInstance(ExecutorPools.class);
        router.get("/fail/parsing/params")
              .handler(new FailsOnParameterParsing(context, metadataFetcher, executorPools));

        VertxTestContext serverStartContext = new VertxTestContext();
        server = injector.getInstance(Server.class);
        server.start()
              .onSuccess(s -> serverStartContext.completeNow())
              .onFailure(serverStartContext::failNow);
        serverStartContext.awaitCompletion(5, TimeUnit.SECONDS);

        WebClient client = WebClient.create(vertx);
        String testRoute = "/fail/parsing/params";
        client.get(server.actualPort(), "127.0.0.1", testRoute)
                  assertThat(response.statusCode()).isEqualTo(INTERNAL_SERVER_ERROR.code());
                  vertx.setTimer(1000, tid -> context.completeNow());
              }));
    }

    /**
     * A handler that will simulate a failure in the {@link #extractParamsOrThrow(RoutingContext)} method
     */
    static class FailsOnParameterParsing extends AbstractHandler<Object>
    {
        private final VertxTestContext testContext;

        protected FailsOnParameterParsing(VertxTestContext testContext, InstanceMetadataFetcher metadataFetcher,
                                          ExecutorPools executorPools)
        {
            super(metadataFetcher, executorPools, null);
            this.testContext = testContext;
        }

        @Override
        protected Object extractParamsOrThrow(RoutingContext context)
        {
            throw new RuntimeException("Simulated failure");
        }

        @Override
        protected void handleInternal(RoutingContext context,
                                      HttpServerRequest httpRequest,
                                      String host,
                                      SocketAddress remoteAddress,
                                      Object request)
        {
            testContext.failNow("Should never reach here");
        }
    }
}
```

src/test/java/org/apache/cassandra/sidecar/routes/StreamSSTableComponentHandlerTest.java
```diff
@@ -221,7 +221,8 @@ void testAccessDeniedToCertainComponents(VertxTestContext context)
     void failsWhenTableNameContainsInvalidCharacters(VertxTestContext context)
     {
         WebClient client = WebClient.create(vertx);
-        String testRoute = "/keyspaces/" + TEST_KEYSPACE + "/tables/i_❤_u/snapshots/snap/components/component-Data.db";
```
nvharikrishna, line 224: The length of the line looks like 120 characters in IDE, but Checkstyle is complaining that it is having 121 characters due to the Unicode char. Added new line in between to avoid Checkstyle error.

nvharikrishna:
@frankgh Thanks for the test case. Added it (just changed the name of the test to in line with other tests).

frankgh (approved):

src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
```diff
@@ -93,6 +93,7 @@ public void handle(RoutingContext context)
         catch (Exception exception)
         {
             processFailure(exception, context, host, remoteAddress, null);
+            return;
```
yifan-c: Instead of `return` from the catch block, how about moving the `handleInternal` call inside the try block? So that it catches any exception throw during `handleInternal`.

nvharikrishna: If `handlerInternal` moved to try block, then processFailure will end up handling expections of `handleInternal` method too. As per my understand, processFailure is only for processing the failures of request parsing/exptract params. If `processFailure` starts handling `handlerInternal` exceptions too, too then isn't it a change of behavior for it?

yifan-c: I believe the intent here is to have a common catch block for any exception during `handle`. Moving `handlerInternal` inside the try block is the desired correction of the behavior. 
Maybe @frankgh can comment here, as you are the original author of the file.

frankgh: Yeah, `processFailure` is intended to process errors during the handleInternal method. So Yifan's suggestion makes sense

nvharikrishna: https://github.com/apache/cassandra-sidecar/blob/trunk/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java#L168 made me to think that it is only for parsing params. Making the changes and updating this description too.

nvharikrishna: @frankgh @yifan-c pushed the changes. [CI](https://app.circleci.com/pipelines/github/nvharikrishna/cassandra-sidecar/5/workflows/1f2da381-f29d-479c-8e56-a0eb157b3526) passed.

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/c874d713b987a5aa0e3420c7f09006e36a82a294

PR #89, opened 2024-01-02
CASSANDRASC-92 Add restore SSTables from S3 into Cassandra feature to Sidecar

sarankk:
Through this patch we want to add restore capability to Sidecar, for Sidecar to allow restoring data from S3. As part of this patch we want to add APIs for creating, updating and getting information about the restore jobs. We also want to add background tasks for managing these restore jobs.


frankgh (approved):
+1 Looks good to me

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/acea8725c0ff62cfcf538515f4b94a97e23414b4

PR #91, opened 2024-01-18
CASSANDRASC-94 Reduce filesystem call while streaming SSTables

frankgh:
Patch by Francisco Guerrero; Reviewed by TBD for CASSANDRASC-94


src/main/java/org/apache/cassandra/sidecar/cache/ToggleableCache.java
```diff
@@ -0,0 +1,227 @@
...
+        return delegate.getIfPresent(key);
+    }
+
+    /**
+     * When the cache is enabled, returns the value associated with the {@code key} in this cache,
+     * obtaining that value from the {@code mappingFunction} if necessary. When the cache is disabled,
+     * it will always obtain the value from the {@code mappingFunction}.
+     *
+     * @param key             the key with which the specified value is to be associated
+     * @param mappingFunction the function to compute a value
```
yifan-c: never a fan of the name `mappingFunction`, although it is named the same as in the guava lib. The name does not tell anything unless looking at the description. I would rename it to `valueLoader`. 
But it is just a nit and feel free to ignore.

src/main/java/org/apache/cassandra/sidecar/cache/ToggleableCache.java
```diff
@@ -0,0 +1,227 @@
...
+ * behavior remains the same as the provided behavior by the {@code delegate}.
+ *
+ * <p>When the cache is toggled to disabled, existing entries in the cache will <b>NOT</b> be
+ * removed, and if the cache is re-enabled and entries have not expired or have not been explicitly
+ * removed; the cache will serve the existing data from the {@code delegate} cache.
+ *
+ * @param <K> the type of keys maintained by this cache
+ * @param <V> the type of mapped values
+ */
+public class ToggleableCache<K, V> implements Cache<K, V>
```
yifan-c: The semantics of `Toggleable` implemented for this class is not clear to me. 
1. when the cache is disabled, the cache still serves read. Then, what exactly is "disabled"?
2. `getIfPresent` and `get` have different behavior. The former evaluate from cache regardlessly. The later, however, honor the flag value.
3. put, mutate and the other meta methods does not honor the value of the flag at all. Again the question, what does "disabled" mean? 

In the current implementation, I do not see how disabling a cache has any benefit, since it allows most of the operations anyway. IMO, if enabled, the cache can memorize values; if disabled, no operations are permitted at all. 

Let's have a clear definition of enabled and disabled, then refine the implementation.  Or maybe you did not mean "toggleable".

src/main/java/org/apache/cassandra/sidecar/cache/ToggleableCache.java
```diff
@@ -0,0 +1,227 @@
...
+
+/**
+ * A {@link Cache} where methods that take a mapping {@link Function} as a parameter can be toggled
+ * (enabled / disabled) such that when the cache is disabled, the {@link Function} will be called
+ * directly without retrieving cached values for any {@code key}, and when the cache is enabled the
+ * behavior remains the same as the provided behavior by the {@code delegate}.
+ *
+ * <p>When the cache is toggled to disabled, existing entries in the cache will <b>NOT</b> be
+ * removed, and if the cache is re-enabled and entries have not expired or have not been explicitly
+ * removed; the cache will serve the existing data from the {@code delegate} cache.
```
yifan-c: > if the cache is re-enabled and entries have not expired or have not been explicitly removed

I do not think it is a complete sentence. Do you meant to say "if the cache is re-enabled, the entries have not expired or have not been explicitly removed will become accessible again"?

src/main/java/org/apache/cassandra/sidecar/config/yaml/ServiceConfigurationImpl.java
```diff
@@ -458,6 +501,18 @@ public Builder ssTableImportConfiguration(SSTableImportConfiguration ssTableImpo
...
         }
 
+        /**
+         * Sets the {@code ssTableSnapshotConfiguration} and returns a reference to this Builder enabling
+         * method chaining.
+         *
+         * @param ssTableSnapshotConfiguration the {@code ssTableSnapshotConfiguration} to set
+         * @return a reference to this Builder
+         */
+        public Builder ssTableSnapshotConfiguration(SSTableSnapshotConfiguration ssTableSnapshotConfiguration)
```
yifan-c: nit: treat `sstable` as a single word. `ssTable` does not look right.

src/main/java/org/apache/cassandra/sidecar/models/HttpResponse.java
```diff
@@ -97,17 +97,23 @@ public void setInternalErrorStatus(String msg)
...
+        response.headersEndHandler(v -> {
+            // notify client we support range requests
+            response.putHeader(HttpHeaderNames.ACCEPT_RANGES, "bytes");
+
+            if (range.length() != fileLength)
+            {
+                setPartialContentStatus(range);
+            }
+            response.putHeader(HttpHeaderNames.CONTENT_TYPE, HttpHeaderValues.APPLICATION_OCTET_STREAM)
+                    .putHeader(HttpHeaderNames.CONTENT_LENGTH, Long.toString(range.length()));
```
yifan-c: Those 2 headers are put as part of `sendFile`. See `io.vertx.core.http.impl.Http2ServerResponse#sendFile:line#619-627`. And then at line#628 `checkSendHeaders(false);` the `headersEndHandler` is called. 

Other than that, the `CONTENT_LENGTH` might not always be the `range.length()`. It is the minimum of `Math.min(length, file_.length() - offset);` as implemented in `io.vertx.core.http.impl.HttpUtils#resolveFile`

src/main/java/org/apache/cassandra/sidecar/routes/FileStreamHandler.java
```diff
@@ -74,40 +103,114 @@ protected String extractParamsOrThrow(RoutingContext context)
...
-            logger.error("The requested file '{}' does not exist", localFile);
-            return Future.failedFuture(wrapHttpException(NOT_FOUND, "The requested file does not exist"));
+            FileProps fileProps = filePropsCache.getIfPresent(localFile);
+
+            if (fileProps != null)
+            {
+                return Future.succeededFuture(fileProps);
+            }
+
+            return ensureValidFileNonCached(fs).apply(localFile)
```
yifan-c: I am not convinced on the usefulness of the cache.

1. When the cache is update to date, the cache helps to save 1 look up of the file attributes. However, during `sendFile`, there are more I/O operations. 
2. When the cache is stale, the file does not exist. It is going to acquire permit and try to `sendFile` falsely, which could undesirably increase CPU usage.

src/main/java/org/apache/cassandra/sidecar/snapshots/CachedSnapshotPathBuilder.java
```diff
@@ -0,0 +1,207 @@
...
+
+    /**
+     * The snapshot path cache maintains a cache of recently streamed snapshot SSTable components. This cache
+     * avoids having to resolve the snapshot SSTable component file during bulk reads. Since bulk reads stream
+     * sub-ranges of an SSTable component, the resolution can happen multiple times during bulk reads for a single
+     * SSTable component. This cache aims to reduce filesystem access to resolve the SSTable component path
+     * during bulk reads.
+     */
+    @VisibleForTesting
+    final Cache<String, Future<String>> snapshotPathCache;
```
yifan-c: Similar question as for `fileProps` cache, what if the value is stable?

src/main/java/org/apache/cassandra/sidecar/models/HttpResponse.java
frankgh: ok, maybe we should let vertx set those headers. Thanks for catching it

src/main/java/org/apache/cassandra/sidecar/config/yaml/ServiceConfigurationImpl.java
frankgh: will keep it as is for consistency

frankgh: updated

src/main/java/org/apache/cassandra/sidecar/routes/FileStreamHandler.java
frankgh: Removed this cache

src/main/java/org/apache/cassandra/sidecar/routes/StreamSSTableComponentHandler.java
```diff
@@ -42,6 +52,7 @@
 @Singleton
 public class StreamSSTableComponentHandler extends AbstractHandler<StreamSSTableComponentRequest>
 {
+    private static final Logger LOGGER = LoggerFactory.getLogger(StreamSSTableComponentHandler.class);
```
yifan-c: `AbstractHandler` defines `logger` already. I would prefer to this change though. But for this patch, I think we should use `logger`.

src/main/java/org/apache/cassandra/sidecar/routes/StreamSSTableComponentHandler.java
```diff
@@ -61,32 +72,93 @@ public void handleInternal(RoutingContext context,
...
+                promise.fail(wrapHttpException(HttpResponseStatus.BAD_REQUEST,
+                                               "Invalid data directory index=" + dataDirIndex));
+                return;
+            }
+
+            String path = snapshotPathBuilder.resolveComponentPath(dataDirList.get(dataDirIndex), request);
+
+            promise.complete(path);
+        });
+    }
```
yifan-c: Let's not introduce new use cases of the deprecated API. Execute with a callable simplify the code a little bit too. We also need to check the return value of `dataDirectoryIndex` is not null, as compiler cannot infer from the callsites. 

```suggestion
    private Future<String> resolveComponentPathFromRequest(String host, StreamSSTableComponentRequest request)
    {
        return executorPools.internal().executeBlocking(() -> {
            CassandraAdapterDelegate delegate = metadataFetcher.delegate(host);
            StorageOperations storageOperations = delegate == null ? null : delegate.storageOperations();
            if (storageOperations == null)
            {
                throw cassandraServiceUnavailable();
            }

            List<String> dataDirList = storageOperations.dataFileLocations();
            Integer dataDirIndex = request.dataDirectoryIndex();
            if (dataDirIndex == null || dataDirIndex < 0 || dataDirIndex >= dataDirList.size())
            {
                throw wrapHttpException(HttpResponseStatus.BAD_REQUEST,
                                        "dataDirectoryIndex out of boundary. index=" + dataDirIndex);
            }

            return snapshotPathBuilder.resolveComponentPath(dataDirList.get(dataDirIndex), request);
        });
    }
```

src/main/java/org/apache/cassandra/sidecar/data/StreamSSTableComponentRequest.java
```diff
@@ -112,6 +127,36 @@ public String toString()
...
                '}';
     }
+
+    public static StreamSSTableComponentRequest from(QualifiedTableName qualifiedTableName, RoutingContext context)
+    {
+        String snapshotName = context.pathParam("snapshot");
+        String secondaryIndexName = context.pathParam("index");
+        String componentName = context.pathParam("component");
+        String tableUuid = maybeGetTableUuid(context.pathParam("table"));
+        Integer dataDirectoryIndex = RequestUtils.parseIntegerQueryParam(context.request(), "dataDirectory", null);
```
yifan-c: Should the param name be "dataDirectoryIndex"?

src/main/java/org/apache/cassandra/sidecar/data/StreamSSTableComponentRequest.java
```diff
@@ -112,6 +127,36 @@ public String toString()
...
+                                                 componentName,
+                                                 tableUuid,
+                                                 dataDirectoryIndex);
+    }
+
+    static String maybeGetTableUuid(String table)
+    {
+        if (table != null)
+        {
+            int index = table.indexOf("-");
```
yifan-c, line 139: add a comment here. 

```
// Cassandra disallow having '-' as part of the table name, even if the table name is quoted. 
// If the string contains '-', it follows by the tableId.
// See https://github.com/apache/cassandra/blob/c33c8ebab444209a9675f273448110afd0787faa/src/java/org/apache/cassandra/schema/TableId.java#L88
```

common/src/main/java/org/apache/cassandra/sidecar/common/data/ListSnapshotFilesResponse.java
```diff
@@ -73,6 +74,7 @@ public FileInfo(@JsonProperty("size") long size,
                         @JsonProperty("snapshotName") String snapshotName,
                         @JsonProperty("keySpaceName") String keySpaceName,
                         @JsonProperty("tableName") String tableName,
+                        @JsonProperty("tableUid") String tableUid,
```
yifan-c: How about naming it `tableId`? just the same as Cassandra

src/main/java/org/apache/cassandra/sidecar/routes/FileStreamHandler.java
```diff
@@ -77,40 +81,57 @@ protected String extractParamsOrThrow(RoutingContext context)
...
+
+        super.processFailure(cause, context, host, remoteAddress, localFile);
+    }
+
+    protected Future<Long> fileSize(RoutingContext context, String path)
+    {
+        Long fileSize = RequestUtils.parseLongQueryParam(context.request(), "size", null);
+        if (fileSize != null)
+        {
+            return Future.succeededFuture(fileSize);
```
yifan-c: Discussed offline and agreed on removing client supplied file size.

src/main/java/org/apache/cassandra/sidecar/data/StreamSSTableComponentRequest.java
```diff
@@ -112,6 +127,39 @@ public String toString()
...
                '}';
     }
+
+    public static StreamSSTableComponentRequest from(QualifiedTableName qualifiedTableName, RoutingContext context)
+    {
+        String snapshotName = context.pathParam("snapshot");
+        String secondaryIndexName = context.pathParam("index");
+        String componentName = context.pathParam("component");
+        String tableId = maybeGetTableId(context.pathParam("table"));
+        Integer dataDirectoryIndex = RequestUtils.parseIntegerQueryParam(context.request(), "dataDirectoryIndex", null);
```
yifan-c: By assigning the default value 0 here, the implement can be simply by avoiding the null case. 

```suggestion
        int dataDirectoryIndex = RequestUtils.parseIntegerQueryParam(context.request(), "dataDirectoryIndex", 0);
```

common/src/main/java/org/apache/cassandra/sidecar/common/data/ListSnapshotFilesResponse.java
```diff
@@ -82,14 +84,25 @@ public FileInfo(@JsonProperty("size") long size,
...
+            String tableName = this.tableId != null
+                               ? this.tableName + "-" + this.tableId
+                               : this.tableName;
+
+            return ApiEndpointsV1.COMPONENTS_ROUTE
+                   .replaceAll(ApiEndpointsV1.KEYSPACE_PATH_PARAM, keySpaceName)
+                   .replaceAll(ApiEndpointsV1.TABLE_PATH_PARAM, tableName)
+                   .replaceAll(ApiEndpointsV1.SNAPSHOT_PATH_PARAM, snapshotName)
+                   .replaceAll(ApiEndpointsV1.COMPONENT_PATH_PARAM, fileName)
+                   + "?dataDirectoryIndex=" + dataDirIndex;
```
yifan-c: nit: extract the URI as a instance variable.

src/main/java/org/apache/cassandra/sidecar/routes/snapshots/ListSnapshotHandler.java
```diff
@@ -0,0 +1,245 @@
...
+    protected Cache<String, Future<ListSnapshotFilesResponse>> initializeCache(CacheConfiguration cacheConfiguration)
+    {
+        if (cacheConfiguration == null)
+        {
+            return null;
+        }
+        return Caffeine.newBuilder()
+                       .maximumSize(cacheConfiguration.maximumSize())
+                       .expireAfterAccess(cacheConfiguration.expireAfterAccessMillis(), TimeUnit.MILLISECONDS)
+                       .recordStats()
```
yifan-c: There is no code to collect from `cache.state()`. It will be nice to know the cache hit rate.

src/test/resources/config/sidecar_multiple_instances.yaml
```diff
@@ -91,6 +91,10 @@ sidecar:
     cache:
       expire_after_access_millis: 7200000 # 2 hours
       maximum_size: 10000
+  sstable_snapshot:
+    snapshot_list_cache:
+      expire_after_access_millis: 350
```
yifan-c, line 96: The expiry is really short.

frankgh: ah this is only a resource for tests, the actual value in `sidecar.yaml` is `7200000` or 2 hours

PR #94, opened 2024-01-25
CASSANDRASC-97: Add support for additional checksum validation during…

frankgh:
… SSTable upload

In this commit we add the ability to support additional checksum algorithms for verification during SSTable uploads. We introduce the `ChecksumVerifierFactory` that allows configuring different implementations of the `ChecksumVerifier` interface.

This commit also adds support for XXHash32 checksums. Clients can now send the XXHash32 checksum instead of MD5. This would allow both the clients and server the flexibility to utilize a more performant algorithm.

Patch by Francisco Guerrero; Reviewed by TBD for CASSANDRASC-97


src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -430,9 +434,14 @@ public DnsResolver dnsResolver()
 
     @Provides
     @Singleton
-    public ChecksumVerifier checksumVerifier(Vertx vertx)
+    public ChecksumVerifierFactory checksumVerifierFactory(Vertx vertx)
     {
-        return new MD5ChecksumVerifier(vertx.fileSystem());
+        FileSystem fs = vertx.fileSystem();
+        List<ChecksumVerifier> verifierList = new ArrayList<>();
+        verifierList.add(new XXHash32ChecksumVerifier(fs));
```
frankgh: now Sidecar can be customized to use different checksum verifiers

yifan-c:
longing for this change

src/main/java/org/apache/cassandra/sidecar/utils/SSTableUploader.java
```diff
@@ -76,14 +78,14 @@ public SSTableUploader(Vertx vertx,
...
-     * @param expectedChecksum  for verifying upload integrity, passed in through request
+     * @param headers           request headers
      * @param filePermissions   specifies the posix file permissions used to create the SSTable file
      * @return path of SSTable component to which data was uploaded
      */
     public Future<String> uploadComponent(ReadStream<Buffer> readStream,
                                           String uploadDirectory,
                                           String componentFileName,
-                                          String expectedChecksum,
+                                          MultiMap headers,
```
yifan-c: Passing http request headers to this layer feels odd to me. 

Request handling should be done in the corresponding handler, e.g. extracting values from header and convert them into a digest object. 

The uploader should care about its own responsibility only, which is upload/store files on local disk. 
(Ideally, checksum verification should be a dedicated step, but it has been there already, so no need to change)

common/src/main/java/org/apache/cassandra/sidecar/common/data/Digest.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.apache.cassandra.sidecar.common.data;
+
+import java.util.Map;
+import java.util.function.Consumer;
+
+/**
+ * Interface that represents a checksum digest
+ */
+public interface Digest extends Consumer<Map<String, String>>
```
yifan-c: - The `Digest` interface is only used from the client side to put headers. Can you relocate the interface and its implementations to client package? 
- Despite `Consumer` does the job seemingly, can we have a more dedicated interface for digest?  

```java
Digest 
{
   // supply the header to be used in http request
   Map<String, String> headers();
}

// usage, e.g. in UploadSSTableRequest#headers() method, merge the headers with others
Map<String, String> headers = new HashMap<>(super.headers());
headers.putAll(digest.headers()); 
```

src/main/java/org/apache/cassandra/sidecar/utils/ChecksumVerifier.java
```diff
@@ -30,9 +31,19 @@
...
     /**
-     * @param checksum  expected checksum value
-     * @param filePath  path to SSTable component
-     * @return String   component path, if verification is a success, else a failed future is returned
+     * @param options  a map with options required for validation
+     * @param filePath path to SSTable component
+     * @return a future String with the component path if verification is a success, otherwise a failed future
      */
-    Future<String> verify(String checksum, String filePath);
+    Future<String> verify(MultiMap options, String filePath);
```
yifan-c: The checksum verifier should not be responsible for parsing headers for the same reason mentioned in the other comment. 

I think you are missing a server-side `Digest` data structure, which supplies the input for verification. 

One approach is this.

```java
public interface ChecksumVerifier<D extends Digest>
{
    Future<String> verify(D digest, String filePath);

    default boolean canVerify(D digest)
    {
        return false;
    }
}

// And then,
public class MD5ChecksumVerifier extends AsyncFileChecksumVerifier<MD5Digest>
```

Beside that, as a nit, I would rename `ChecksumVerifier` to `DigestVerifier` on server code for consistency and accuracy.

yifan-c (commented):
Looks good in general. Some suggestions for refactoring.

common/src/main/java/org/apache/cassandra/sidecar/common/data/Digest.java
frankgh: we are now using it in both the client / server code. so leaving it in common for now

yifan-c (approved):

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/e329f3232fae867879aba2cd0a766404aaf9a427

PR #96, opened 2024-01-30
CASSANDRASC-99 Break restore job into stage and import phases and persist restore sl…

yifan-c:
…ice status on phase completion

### DIFF
--- a/common/src/main/java/org/apache/cassandra/sidecar/common/data/CreateRestoreJobRequestPayload.java
+++ b/common/src/main/java/org/apache/cassandra/sidecar/common/data/CreateRestoreJobRequestPayload.java
@@ -28,6 +28,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.cassandra.sidecar.common.utils.Preconditions;
 
 import static org.apache.cassandra.sidecar.common.data.RestoreJobConstants.JOB_AGENT;
+import static org.apache.cassandra.sidecar.common.data.RestoreJobConstants.JOB_CONSISTENCY_LEVEL;
 import static org.apache.cassandra.sidecar.common.data.RestoreJobConstants.JOB_EXPIRE_AT;
 import static org.apache.cassandra.sidecar.common.data.RestoreJobConstants.JOB_ID;
 import static org.apache.cassandra.sidecar.common.data.RestoreJobConstants.JOB_IMPORT_OPTIONS;
@@ -43,6 +44,7 @@ public class CreateRestoreJobRequestPayload
     private final RestoreJobSecrets secrets;
     private final SSTableImportOptions importOptions;
     private final long expireAtInMillis;
+    private final String consistencyLevel; // Nullable / optional field
 
     /**
      * Builder to build a CreateRestoreJobRequest
@@ -65,13 +67,15 @@ public class CreateRestoreJobRequestPayload
      * @param secrets          secrets to be used by restore job to download data
      * @param importOptions    the configured options for SSTable import
      * @param expireAtInMillis a timestamp in the future when the job is considered expired
+     * @param consistencyLevel consistency level a job should satisfy
      */
     @JsonCreator
     public CreateRestoreJobRequestPayload(@JsonProperty(JOB_ID) UUID jobId,
                                           @JsonProperty(JOB_AGENT) String jobAgent,
                                           @JsonProperty(JOB_SECRETS) RestoreJobSecrets secrets,
                                           @JsonProperty(JOB_IMPORT_OPTIONS) SSTableImportOptions importOptions,
-                                          @JsonProperty(JOB_EXPIRE_AT) long expireAtInMillis)
+                                          @JsonProperty(JOB_EXPIRE_AT) long expireAtInMillis,
+                                          @JsonProperty(JOB_CONSISTENCY_LEVEL) String consistencyLevel)
     {
         Preconditions.checkArgument(jobId == null || jobId.version() == 1,
                                     "Only time based UUIDs allowed for jobId");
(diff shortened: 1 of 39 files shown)


src/main/java/org/apache/cassandra/sidecar/db/RestoreJob.java
```diff
@@ -156,13 +123,28 @@ private RestoreJob(Builder builder)
...
 
+    /**
+     * When a job object is materialized from database, and it has non-null consistencyLevel, the range of the slices
+     * of the restore is managed by Sidecar server, meaning that server should assign slices to sidecar instances and
+     * check whether the job has met the consistency level to complete the job; otherwise, sidecar instances are just
+     * simple workers and rely on client for decision-making.
+     * Depending on the return value, call-sites select the correct handling for the restore jobs and slices
+     * @return true when sidecar server manages the slices based on the token range; otherwise, return false.
+     */
+    public boolean isRangeManagedByServer()
```
yifan-c: naming is hard. Open to suggestions

src/main/java/org/apache/cassandra/sidecar/config/RestoreJobConfiguration.java
```diff
@@ -48,4 +48,11 @@ public interface RestoreJobConfiguration
      * @return time to live for restore job tables: restore_job and restore_slice
      */
     long restoreJobTablesTtlSeconds();
+
+    /**
+     * Controls whether the feature to allow server-side the slice assignment and verification based on the token range
+     * is enabled or not.
+     * @return true if enabled; false otherwise
+     */
+    boolean isServerManagedRangeEnabled();
```
yifan-c: Naming is hard. Open to suggestions.

common/src/main/java/org/apache/cassandra/sidecar/common/data/CreateRestoreJobRequestPayload.java
```diff
@@ -43,6 +44,7 @@ public class CreateRestoreJobRequestPayload
     private final RestoreJobSecrets secrets;
     private final SSTableImportOptions importOptions;
     private final long expireAtInMillis;
+    private final String consistencyLevel; // Nullable / optional field
```
frankgh: we can use annotations if we want to mark them as nullable? no?

common/src/main/java/org/apache/cassandra/sidecar/common/data/RestoreSliceStatus.java
```diff
@@ -18,15 +18,44 @@
...
     FAILED,
-    ABORTED
+    ABORTED,
+    COMMITTING(SUCCEEDED, FAILED, ABORTED),
+    STAGED(COMMITTING, FAILED, ABORTED),
+    PROCESSING(STAGED, FAILED, ABORTED),
+    EMPTY(PROCESSING, FAILED, ABORTED);
+
+    // Do not use EnumSet, since validTargetStatuses is assigned on constructing and enums are not available yet.
+    private final Set<RestoreSliceStatus> validTargetStatuses;
```
frankgh: NIT
```suggestion
    private final Set<RestoreSliceStatus> validTargetStatusSet;
```

common/src/test/java/org/apache/cassandra/sidecar/common/data/RestoreSliceStatusTest.java
```diff
@@ -0,0 +1,78 @@
...
+import static org.apache.cassandra.sidecar.common.data.RestoreSliceStatus.STAGED;
+import static org.apache.cassandra.sidecar.common.data.RestoreSliceStatus.SUCCEEDED;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class RestoreSliceStatusTest
+{
+    @Test
+    void testStatusAdvancing()
+    {
+        EMPTY.advanceTo(PROCESSING);
```
frankgh: maybe assert?
```suggestion
        assertThat(EMPTY.advanceTo(PROCESSING)).isEqualTo(PROCESSING);
```

src/main/java/org/apache/cassandra/sidecar/config/RestoreJobConfiguration.java
```diff
@@ -48,4 +48,11 @@ public interface RestoreJobConfiguration
      * @return time to live for restore job tables: restore_job and restore_slice
      */
     long restoreJobTablesTtlSeconds();
+
+    /**
+     * Controls whether the feature to allow server-side the slice assignment and verification based on the token range
```
frankgh: I'm not following the description of this javadoc. Can you elaborate what we are trying to achieve here?

frankgh (commented):
Looks good! I've left some minor comments

src/main/java/org/apache/cassandra/sidecar/config/RestoreJobConfiguration.java
yifan-c: The goal is to having a boolean flag to toggle the "server-side the slice assignment and verification based on the token range". Server side slice assignment is to select and download the slices based on its owning token range. Verification is to check whether the slices imported can satisfy the specified consistency level.

src/main/java/org/apache/cassandra/sidecar/db/RestoreJob.java
```diff
@@ -76,10 +75,13 @@ public static RestoreJob from(@NotNull Row row) throws DataObjectMappingExceptio
                .jobSecrets(decodeJobSecrets(row.getBytes("blob_secrets")))
                .expireAt(row.getTimestamp("expire_at"))
                .sstableImportOptions(decodeSSTableImportOptions(row.getBytes("import_options")));
+        String consistencyLevel = row.getString("consistency_level");
+        builder.manager(consistencyLevel == null ? Manager.SPARK : Manager.SIDECAR)
```
JeetKunDoug: Took me a moment to figure out why consistencyLevel would be null if the manager is SPARK - can we document why this is the case here please?

frankgh (approved):
+1 the patch looks good

PR #98, opened 2024-02-08
CASSANDRASC-100 Fix flaky SSTableImportHandlerTest

sarankk:


src/test/java/org/apache/cassandra/sidecar/routes/sstableuploads/BaseUploadsHandlerTest.java
```diff
@@ -222,4 +229,53 @@ public SidecarConfiguration configuration()
             return sidecarConfiguration;
         }
     }
+
+    static class TestCassandraAdapterDelegate extends CassandraAdapterDelegate
```
frankgh, line 233: can we call this `FakeCassandraAdapterDelegate` instead?

sarankk: But we already follow a pattern of calling such dummy classes Test<>. Example https://github.com/apache/cassandra-sidecar/blob/c72f2179143e7e031f247d3e8385a29c5e64c1c3/src/test/java/org/apache/cassandra/sidecar/routes/restore/BaseRestoreJobTests.java#L168

frankgh (approved):
+1 looks good to me

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/2914d57f0428643b3a92b6af8f4da1b209d80c2a

PR #99, opened 2024-02-11
CASSANDRASC-103 Introduce yaml type binding annotation for deserialization

yifan-c:


frankgh (commented):
PR looks good. Should we backfill tests in `SidecarConfigurationTest` as well with the configuration classes that are missing?

yifan-c:
> PR looks good. Should we backfill tests in `SidecarConfigurationTest` as well with the configuration classes that are missing?

I do not think there are additional test cases to be added. The error is caused by meeting both conditions, 1) having `schema` section in yaml and 2) missing the type mapping. If the yaml file does not have the `schema` section, the produced configuration picks up the default value from the builder. 

This patch prevents the condition 2). But for 1), it will be our / reviewers' job to make sure the new configuration section is added to the yaml file in the related PR during review.

src/test/java/org/apache/cassandra/sidecar/config/SidecarConfigurationTest.java
```diff
@@ -190,6 +190,22 @@ void testDriverParameters() throws IOException
         assertThat(driverConfiguration.numConnections()).isEqualTo(6);
     }
 
+    @Test
+    void testReadCustomSchemaKeyspaceConfiguration() throws IOException
```
frankgh, line 194: 👍

src/main/java/org/apache/cassandra/sidecar/config/yaml/ServiceConfigurationImpl.java
```diff
@@ -174,7 +174,7 @@ public int port()
      * {@inheritDoc}
      */
     @Override
-    @JsonProperty(value = REQUEST_IDLE_TIMEOUT_MILLIS_PROPERTY, defaultValue = DEFAULT_REQUEST_IDLE_TIMEOUT_MILLIS + "")
```
frankgh, line 177: we don't have builders for all classes, I wonder what's the side effect to remove defaultValue for classes that don't have builders?

frankgh (commented):
Looks good, I have a question regarding default values missing for classes that don't use builders

src/main/java/org/apache/cassandra/sidecar/config/yaml/ServiceConfigurationImpl.java
yifan-c: Right. For those implementations, there is a no-args constructor defined with the default values for the fields. For example, `CassandraInputValidationConfigurationImpl`. 
Jackson always create an instance with such constructor and set the fields, during deserialization. So the default value is already present when the field is absent in the json text

frankgh (approved):
+1 Looks good to me

PR #102, opened 2024-02-29
CASSANDRASC-111 Improve observability in Sidecar with Dropwizard metrics

sarankk:


src/main/dist/conf/sidecar.yaml
```diff
@@ -142,6 +142,15 @@ healthcheck:
   initial_delay_millis: 0
   poll_freq_millis: 30000
 
+metrics:
+  registry_name: cassandra_sidecar
+  vertx:
+    enabled: true
+    registry_name: cassandra_sidecar
```
yifan-c: is it a duplication of the `registry_name` in the outer structure?

src/main/dist/conf/sidecar.yaml
```diff
@@ -142,6 +142,15 @@ healthcheck:
...
+    jmx_enabled: false
+    jmx_domain_name: sidecar.vertx.jmx_domain
+    monitored_server_routes_regex: /api/v1/*      # all routes are monitored
```
yifan-c: I do not think having a single matcher is sufficient. There are likely use cases that exclude some route and add some selectively. 
It probably need a list of matchers

src/main/java/org/apache/cassandra/sidecar/metrics/MetricName.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * Representation of a metric name.
+ */
+public class MetricName
+{
+    private final String feature;
+    private final  String name;
```
yifan-c: extra space

src/main/java/org/apache/cassandra/sidecar/metrics/MetricName.java
```diff
@@ -0,0 +1,96 @@
...
+    {
+        return feature;
+    }
+
+    /**
+     * If applicable, additional name tags added to metric name for more clarity. Tags are usually like,
+     * component=data, route=/stream/component, etc.
+     * @return a set containing additional metric name tags
+     */
+    public Set<String> tags()
```
yifan-c: If tag is always in the form of `key=value`, can you have proper data class instead of `String`?

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceMetricRegistry.java
```diff
@@ -0,0 +1,76 @@
...
+    public int instanceId()
+    {
+        return instanceId;
+    }
+
+    /**
+     * A factory for creating instance specific {@link MetricRegistry} provided an instance id.
+     */
+    @Singleton
+    public static class Factory
```
yifan-c: The name `Factory` is too general.

yifan-c (commented):
First, thank you for adding a metrics framework! 

I only skimmed through the first dozen files. But it becomes hard to follow, as this patch contains both new framework and changes to the existing metric publish. 

I would suggest to trim this patch to contain only the new framework, which is the most valuable. And we can focus on getting the framework right first. 
Then, you can create the follow-up patches that migrate the existing metrics. 

Let me know if it makes sense to you.

src/main/dist/conf/sidecar.yaml
sarankk: We can set registry name separately for the metrics that Vert.x provides. Currently sidecar's metrics are recorded in a single registry, hence both are same

src/main/dist/conf/sidecar.yaml
sarankk: Makes sense, will modify it to include a list

sarankk:
Thanks for the review Yifan, removed changes to existing metrics

src/main/dist/conf/sidecar.yaml
sarankk: Changed to using global registry for vert.x metrics too.

src/main/java/org/apache/cassandra/sidecar/config/VertxMetricsConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * @return JMX domain to be used when metrics are exposed via JMX
+     */
+    String jmxDomainName();
+
+    /**
+     * @return List of regexes used for identifying what server routes will be monitored
+     */
+    List<String> monitoredServerRouteRegexes();
```
yifan-c, line 46: Can you explain why choosing server route match over server URI match?

sarankk: Yifan and I did some testing, URI match, matches with the resolved URI (after resolving path params). This would publish many metrics, since path params change. Hence route matching with regex, works better, since it  would publish for generic routes like `/api/v1/keyspaces/:keyspace`

src/test/java/org/apache/cassandra/sidecar/tasks/HealthCheckPeriodicTaskTest.java
```diff
@@ -136,7 +160,11 @@ void testDelegateExceptionDoesntPreventChecksOnOtherInstances(VertxTestContext c
         when(mockInstancesConfig.instances()).thenReturn(mockInstanceMetadata);
         Promise<Void> promise = Promise.promise();
         healthCheck.execute(promise);
-        promise.future().onComplete(context.succeedingThenComplete());
+        promise.future().onComplete(context.failing(v -> {
```
yifan-c, line 171: The assertion change makes sense. How was the test passing before? hmm...

src/main/java/org/apache/cassandra/sidecar/metrics/NamedMetric.java
```diff
@@ -0,0 +1,171 @@
...
+        this.metric = builder.register(canonicalName);
+    }
+
+    public static <T extends Metric> Builder<T> builder(Function<String, T> metricCreator)
+    {
+        return new Builder<>(metricCreator);
+    }
+
+    /**
+     * {@link NamedMetric} builder static inner class.
```
yifan-c: nit: the sentence is not helpful to mention that the builder is a static inner class. It of course is! Let's just say `Builder for {@link NamedMetric}`. Wdyt?

src/main/java/org/apache/cassandra/sidecar/metrics/NamedMetric.java
```diff
@@ -0,0 +1,171 @@
...
+        {
+            this.metricCreator = metricCreator;
+        }
+
+        /**
+         * Sets {@code domain} of metric and returns a reference to this builder to enable method chaining.
+         *
+         * @param domain domain metric is part of
+         * @return a reference to this Builder
+         */
```
yifan-c: " and returns a reference to this builder to enable method chaining." is not helpful for the same reason. I am aware that other existing builder methods have the same style javadoc. So feel free to ignore this nit.

src/main/java/org/apache/cassandra/sidecar/metrics/NamedMetric.java
```diff
@@ -0,0 +1,171 @@
...
+ * {@link NamedMetric} is for creating {@link Metric} with a structured name. Metric name should contain domain it
+ * is captured for and optional tags for additional context.
+ *
+ * @param <T>   {@link NamedMetric} can be created for any {@link Metric} type. T represents the Metric type
+ *              {@link NamedMetric} represents
+ */
+public class NamedMetric<T extends Metric>
+{
+    public final String canonicalName;
+    public final T metric;
```
yifan-c, line 39: can you add a test case for `NamedMetric` to assert the retrieved metric with `canonicalName` is the same metric?

src/main/java/org/apache/cassandra/sidecar/metrics/NamedMetric.java
```diff
@@ -0,0 +1,171 @@
...
+
+import com.codahale.metrics.Metric;
+import org.apache.cassandra.sidecar.common.DataObjectBuilder;
+
+/**
+ * {@link NamedMetric} is for creating {@link Metric} with a structured name. Metric name should contain domain it
+ * is captured for and optional tags for additional context.
+ *
+ * @param <T>   {@link NamedMetric} can be created for any {@link Metric} type. T represents the Metric type
+ *              {@link NamedMetric} represents
```
yifan-c: This can be simplified to the below.

```suggestion
 * @param <T> Metric type
```

src/main/java/org/apache/cassandra/sidecar/metrics/ServerMetrics.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Tracks metrics related to server initialization, periodic tasks related to server.
+ */
+@Singleton
+public class ServerMetrics
+{
+    public static final String DOMAIN = "sidecar.server";
+    protected final MetricRegistry metricRegistry;
+    public final DefaultSettableGauge<Integer> cassandraInstancesUp;
+    public final DefaultSettableGauge<Integer> cassandraInstancesDown;
```
yifan-c: It is sufficient to only keep the reference of metrics for collection purpose. However, the metric names are lost and seemingly should be kept. Admitted that the metric names are not used anywhere. So I am on the fence.

src/main/java/org/apache/cassandra/sidecar/metrics/NamedMetric.java
```diff
@@ -0,0 +1,171 @@
...
+        return new Builder<>(metricCreator);
+    }
+
+    /**
+     * {@link NamedMetric} builder static inner class.
+     *
+     * @param <T>    {@link NamedMetric} can be created for any {@link Metric} type. T represents the Metric type
+     *               {@link NamedMetric} represents
+     */
+    public static class Builder<T extends Metric> implements DataObjectBuilder<Builder<T>, NamedMetric<T>>
```
yifan-c, line 57: Can you change the builder to this (meaning it builds a metric directly), if metric name is not going to be discarded?
```suggestion
    public static class Builder<T extends Metric> implements DataObjectBuilder<Builder<T>, T>
```

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceMetricProvider.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.metrics.instance;
+
+/**
+ * Provider for getting instance specific metrics given instance id or host.
+ */
+public interface InstanceMetricProvider
```
yifan-c: how about removing this interface and its implementation, and provide `InstanceMetrics` from `InstanceMetadata`? i.e. adding a new interface method in `InstanceMetadata` that return `InstanceMetrics`

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceMetricProviderImpl.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.apache.cassandra.sidecar.cluster.InstancesConfig;
+import org.apache.cassandra.sidecar.cluster.instance.InstanceMetadata;
+
+/**
+ * Provider for getting instance specific metrics given Cassandra instance id or host name.
+ */
+public class InstanceMetricProviderImpl implements InstanceMetricProvider
```
yifan-c: it can be deleted.

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceMetricRegistry.java
```diff
@@ -0,0 +1,76 @@
...
+import com.codahale.metrics.SharedMetricRegistries;
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import org.apache.cassandra.sidecar.cluster.InstancesConfig;
+import org.apache.cassandra.sidecar.cluster.instance.InstanceMetadata;
+
+/**
+ * A dropwizard {@link MetricRegistry} associated with a specific Cassandra instance.
+ */
+public class InstanceMetricRegistry extends MetricRegistry
```
yifan-c: once `InstanceMetrics` is part of `InstanceMetadata`, this class can be delete too.

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceMetrics.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package org.apache.cassandra.sidecar.metrics.instance;
+
+/**
+ * {@link InstanceMetrics} tracks metrics related to a Cassandra instance that Sidecar maintains.
+ */
+public interface InstanceMetrics
+{
+    String INSTANCE_PREFIX = "sidecar.instance";
```
yifan-c: tbh, the meaning of the word `instance` is vague. W/o looking at the code comment, one can be easily confused by the string "sidecar.instance", especially they just look at the metrics name and have no familiarity with the code base. It reads as sidecar instance. 

How about changing it to `sidecar.cass_instance`

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceMetrics.java
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * @return resource metrics tracked for instance
+     */
+    InstanceResourceMetrics resource();
+
+    /**
+     * @param component     SSTable component such as data
+     * @return metrics that are tracked during streaming of provided SSTable component
+     */
+    StreamSSTableComponentMetrics forStreamComponent(String component);
```
yifan-c: `StreamSSTableComponentMetrics` should contain the metrics of each component. The input `component` should be removed from the interface method. Otherwise, the implementation have to maintain metrics per component, which is not the right place.

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceMetrics.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * @param component     SSTable component such as data
+     * @return metrics that are tracked during upload of provided SSTable component
+     */
+    UploadSSTableComponentMetrics forUploadComponent(String component);
```
yifan-c: same here. `UploadSSTableComponentMetrics` should contain the metrics of each component.

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceMetricsImpl.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Tracks metrics related to a Cassandra instance that Sidecar maintains.
+ */
+public class InstanceMetricsImpl implements InstanceMetrics
+{
+    protected final InstanceMetricRegistry instanceMetricRegistry;
+    protected final Map<String, StreamSSTableComponentMetrics> streamComponentMetrics = new ConcurrentHashMap<>();
+    protected final Map<String, UploadSSTableComponentMetrics> uploadComponentMetrics = new ConcurrentHashMap<>();
```
yifan-c: They do not belong in the `InstanceMetricsImpl`

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceResourceMetrics.java
```diff
@@ -0,0 +1,48 @@
...
+    public final Meter insufficientStorageErrors;
+
+    public InstanceResourceMetrics(InstanceMetricRegistry metricRegistry)
+    {
+        this.metricRegistry = Objects.requireNonNull(metricRegistry, "Metric registry can not be null");
+
+        insufficientStorageErrors
+        = NamedMetric.builder(metricRegistry::meter)
+                     .withDomain(DOMAIN)
+                     .withName("insufficient_storage_errors")
```
yifan-c: the name has to be more specific to avoid misleading. The metric is reporting on insufficient space for staging. 

my suggestion of new name is `insufficient_staging_space`

src/main/java/org/apache/cassandra/sidecar/metrics/instance/MetricRegistryNameFactory.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import org.apache.cassandra.sidecar.config.SidecarConfiguration;
+
+/**
+ * Provides registry name for instance specific registry {@link InstanceMetricRegistry}.
+ */
+@Singleton
+public class MetricRegistryNameFactory
```
yifan-c: This class look unnecessary. It can be a private method in `InstanceMetadata`, when exposing metrics from there.

src/main/java/org/apache/cassandra/sidecar/metrics/instance/StreamSSTableComponentMetrics.java
```diff
@@ -0,0 +1,73 @@
...
+                     .withName("rate_limited_calls_429")
+                     .addTag(componentTag)
+                     .build().metric;
+        timeTakenForSendFile
+        = NamedMetric.builder(instanceMetricRegistry::timer)
+                     .withDomain(DOMAIN)
+                     .withName("time_taken_for_sendfile")
+                     .addTag(componentTag)
+                     .build().metric;
+        waitTimeSent
```
yifan-c: is it really needed to report the retryAfter duration values that are sent to client?

src/main/java/org/apache/cassandra/sidecar/metrics/instance/StreamSSTableComponentMetrics.java
```diff
@@ -0,0 +1,73 @@
...
+        rateLimitedCalls
+        = NamedMetric.builder(instanceMetricRegistry::meter)
+                     .withDomain(DOMAIN)
+                     .withName("rate_limited_calls_429")
+                     .addTag(componentTag)
+                     .build().metric;
+        timeTakenForSendFile
+        = NamedMetric.builder(instanceMetricRegistry::timer)
+                     .withDomain(DOMAIN)
+                     .withName("time_taken_for_sendfile")
```
yifan-c: "sendfile_latency"?

src/main/java/org/apache/cassandra/sidecar/metrics/instance/StreamSSTableComponentMetrics.java
```diff
@@ -0,0 +1,73 @@
...
+        {
+            throw new IllegalArgumentException("SSTableComponent required for component specific metrics capture");
+        }
+
+        NamedMetric.Tag componentTag = NamedMetric.Tag.of("component", sstableComponent);
+
+        rateLimitedCalls
+        = NamedMetric.builder(instanceMetricRegistry::meter)
+                     .withDomain(DOMAIN)
+                     .withName("rate_limited_calls_429")
```
yifan-c: "throttled_429"?

src/main/java/org/apache/cassandra/sidecar/metrics/instance/StreamSSTableComponentMetrics.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.cassandra.sidecar.metrics.NamedMetric;
+
+import static org.apache.cassandra.sidecar.metrics.instance.InstanceMetrics.INSTANCE_PREFIX;
+
+/**
+ * {@link StreamSSTableComponentMetrics} tracks metrics captured during streaming of SSTable component from Sidecar.
+ */
+public class StreamSSTableComponentMetrics
+{
+    public static final String DOMAIN = INSTANCE_PREFIX + ".stream";
```
yifan-c: "stream" is too general. and in the future, we could stream other things. Let's put "stream_sstable"

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableComponentMetrics.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.cassandra.sidecar.metrics.NamedMetric;
+
+import static org.apache.cassandra.sidecar.metrics.instance.InstanceMetrics.INSTANCE_PREFIX;
+
+/**
+ * {@link UploadSSTableComponentMetrics} tracks metrics captured during upload of SSTable component into Sidecar.
+ */
+public class UploadSSTableComponentMetrics
+{
+    public static final String DOMAIN = INSTANCE_PREFIX + ".upload";
```
yifan-c: "upload_sstable"

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableComponentMetrics.java
```diff
@@ -0,0 +1,66 @@
...
+        {
+            throw new IllegalArgumentException("SSTableComponent required for component specific metrics capture");
+        }
+
+        NamedMetric.Tag componentTag = NamedMetric.Tag.of("component", sstableComponent);
+
+        rateLimitedCalls
+        = NamedMetric.builder(instanceMetricRegistry::meter)
+                     .withDomain(DOMAIN)
+                     .withName("rate_limited_calls_429")
```
yifan-c: "throttled_429"?

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableComponentMetrics.java
```diff
@@ -0,0 +1,66 @@
...
+                     .addTag(componentTag)
+                     .build().metric;
+        diskUsageHighErrors
+        = NamedMetric.builder(instanceMetricRegistry::meter)
+                     .withDomain(DOMAIN)
+                     .withName("disk_usage_high_errors")
```
yifan-c: it is the same as insufficient_staging_space. can you remove this metric? and collect with that one instead.

yifan-c (changes requested):

src/test/java/org/apache/cassandra/sidecar/tasks/HealthCheckPeriodicTaskTest.java
sarankk: Yes I saw a flaky test on circleci. Locally it was passing when I ran all tests together, if I ran just that test it was always failing

src/main/java/org/apache/cassandra/sidecar/metrics/NamedMetric.java
sarankk: Yes, makes sense. I added it to be consistent in doc style

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceMetricProvider.java
sarankk: Makes sense. @frankgh suggested this too. Will update this

frankgh: `InstanceMetadata` seems like the natural place to have the instance-specific metrics. +1

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableComponentMetrics.java
sarankk: It is not the same currently. Code in upload handler to check if sufficient space is there, is different from `DiskprotectionHandler`.  If we want to use same metric, we can modify `SSTableUploadHandler`

src/main/java/org/apache/cassandra/sidecar/metrics/instance/StreamSSTableComponentMetrics.java
sarankk: We can remove it, I thought it would indicate too many request overload, but we have `throttled_429` count

src/main/java/org/apache/cassandra/sidecar/metrics/ServerMetrics.java
sarankk: Changed to maintaining `NamedMetric`, incase in code we want to use metric name

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceMetrics.java
sarankk: If a new component is added we have to add metrics for it. This way, retrieval of metrics for a component is dynamic?

sarankk: Discussed and addressed this, moved map to `StreamSSTableComponentMetrics`

yifan-c (approved):

src/main/dist/conf/sidecar.yaml
```diff
@@ -142,6 +142,15 @@ healthcheck:
   initial_delay_millis: 0
   poll_freq_millis: 30000
 
+metrics:
+  registry_name: cassandra_sidecar
+  vertx:
+    enabled: true
+    expose_via_jmx: false
+    jmx_domain_name: sidecar.vertx.jmx_domain
+    monitored_server_route_regexes:               # list of regexes for matching server routes
```
frankgh: NIT:
```suggestion
    monitored_server_route_regexes:               # regex list to match server routes
```

src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
```diff
@@ -58,4 +58,9 @@ public interface InstanceMetadata
      * @return a {@link CassandraAdapterDelegate} specific for the instance
      */
     @Nullable CassandraAdapterDelegate delegate();
+
+    /**
+     * @return {@link InstanceMetrics} metrics specific for the cassandra instance
```
frankgh: NIT
```suggestion
     * @return {@link InstanceMetrics} metrics specific for the Cassandra instance
```

frankgh: Should we annotate with `@NotNull`?

src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
```diff
@@ -188,6 +203,29 @@ public Builder delegate(CassandraAdapterDelegate delegate)
...
+            String instanceRegistryName = instanceRegistryName(registryName);
+            MetricRegistry instanceMetricRegistry = SharedMetricRegistries.getOrCreate(instanceRegistryName);
+            InstanceMetricsImpl instanceMetrics = new InstanceMetricsImpl(instanceMetricRegistry);
+
+            return update(b -> b.metrics = instanceMetrics);
+        }
+
+        private String instanceRegistryName(String globalRegistryName)
+        {
+            return globalRegistryName + "_" + id;
```
frankgh: this will not work if you set the globalMetricRegistryName first and then the id. Should we compute this during build time instead?

src/main/java/org/apache/cassandra/sidecar/config/yaml/MetricsConfigurationImpl.java
```diff
@@ -0,0 +1,67 @@
...
+package org.apache.cassandra.sidecar.config.yaml;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.cassandra.sidecar.config.MetricsConfiguration;
+import org.apache.cassandra.sidecar.config.VertxMetricsConfiguration;
+
+/**
+ * Configuration needed for capturing metrics.
+ */
+public class MetricsConfigurationImpl implements MetricsConfiguration
```
frankgh, line 28: can we add tests for yaml parsing? we have a set of tests under `org.apache.cassandra.sidecar.config.SidecarConfigurationTest`

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableMetrics.java
```diff
@@ -0,0 +1,88 @@
...
+            {
+                throw new IllegalArgumentException("SSTableComponent required for component specific metrics capture");
+            }
+
+            NamedMetric.Tag componentTag = NamedMetric.Tag.of("component", sstableComponent);
+
+            rateLimitedCalls
+            = NamedMetric.builder(metricRegistry::meter)
+                         .withDomain(DOMAIN)
+                         .withName("throttled_429")
```
frankgh, line 76: should we make these names constants of the class with public visibility?

src/main/java/org/apache/cassandra/sidecar/routes/DiskSpaceProtectionHandler.java
```diff
@@ -99,6 +100,8 @@ protected void handleInternal(RoutingContext context,
         .onFailure(throwable -> {
             if (throwable instanceof InsufficientStorageException)
             {
+                InstanceMetrics metrics = instance.metrics();
+                metrics.resource().insufficientStagingSpaceErrors.metric.mark();
```
frankgh: NIT:
```suggestion;
                instance.metrics().resource().insufficientStagingSpaceErrors.metric.mark();
```

src/main/java/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableUploadHandler.java
```diff
@@ -116,10 +119,15 @@ public void handleInternal(RoutingContext context,
         // accept the upload.
         httpRequest.pause();
 
+        InstanceMetrics instanceMetrics = metadataFetcher.instance(host).metrics();
```
frankgh, line 122: InstanceMetadata can be null, we usually return 503 Unavailable when this occurs. I think we run into  the issue of an NPE here.

src/main/java/org/apache/cassandra/sidecar/tasks/HealthCheckPeriodicTask.java
```diff
@@ -101,6 +108,11 @@ public void execute(Promise<Void> promise)
 
         // join always waits until all its futures are completed and will not fail as soon as one of the future fails
         Future.join(futures)
+              .onComplete(v -> {
+                  int instancesUp = instancesConfig.instances().size() - instanceDown.get();
```
frankgh, line 112: you don't really need an AtomicInteger to keep track of the down instances. You can check the results instead.

`v.result().causes().stream().filter(Objects::nonNull).count()`

src/test/java/org/apache/cassandra/sidecar/restore/RestoreJobManagerTest.java
```diff
@@ -66,7 +68,7 @@ class RestoreJobManagerTest
     @BeforeEach
     void setup()
     {
-        Injector injector = Guice.createInjector(new MainModule());
+        Injector injector = Guice.createInjector(Modules.override(new MainModule()).with(new TestModule()));
```
frankgh, line 71: why do we need the test module here? the metrics should be provided in the main module

src/test/java/org/apache/cassandra/sidecar/server/ServerSSLTest.java
```diff
@@ -124,6 +123,7 @@ void tearDown() throws InterruptedException
     void failsWhenKeyStoreIsNotConfigured()
     {
         builder.sslConfiguration(SslConfigurationImpl.builder().enabled(true).build());
+        vertx = vertx();
```
frankgh, line 126: why is this change needed?

frankgh (commented):
This is a great contribution. Thanks for the patch. I've added some comments

src/main/java/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableUploadHandler.java
sarankk: In that case we can have `instance(host)` throw unavailable exception. Else we will have to do null check everywhere before retrieving metrics

sarankk: It does it currently `throw new NoSuchElementException("Instance id " + id + " not found")`

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableMetrics.java
sarankk: we expose the full name, if we want to use the metric name during testing. I don't see a use case for just the name extension elsewhere, wdyt?

src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
sarankk: since `id` is an`int`. Same could happen if no id was set before build. Maybe we should convert it into an integer and check if set, before setting registry

src/main/java/org/apache/cassandra/sidecar/tasks/HealthCheckPeriodicTask.java
sarankk: But what if the futures failed due to some other reason, other than what is caught in `catch` block?

src/test/java/org/apache/cassandra/sidecar/server/ServerSSLTest.java
sarankk: Since now creating a Vert.x instance needs `SidecarConfiguration` to get metrics options. Changed this. Because in these tests, the configuration is updated inside test as per each test. After the configuration update, building the vert.x instance

sarankk: It is not needed for all tests cases, example some of the failure test cases. But changed it consistently for easier test reading

src/test/java/org/apache/cassandra/sidecar/restore/RestoreJobManagerTest.java
sarankk: Here it is needed because, the test doesn't provide all the classes we need. It only provides `mock(RestoreJobConfiguration.class)`. But for building Vert.x instance we need sidecar configuration object. The test throws error saying, yaml not found if we use `MainModule`. `TestModule` provides  `SidecarConfiguration` hence we avoid yaml error.

src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
sarankk: Setting registry before building makes sense. Modified this

sarankk:
Thanks for the review @frankgh addressed your comments.

src/main/java/org/apache/cassandra/sidecar/tasks/HealthCheckPeriodicTask.java
frankgh: how is that possible? we explicitly set the promise to either failed or succeeded

src/main/java/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableUploadHandler.java
frankgh: We handle nulls coming from metadata fetcher in the codebase already , so I think we should continue handling them

src/test/java/org/apache/cassandra/sidecar/restore/RestoreJobManagerTest.java
frankgh: maybe we need to add these fields to the yaml?

src/main/java/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableUploadHandler.java
sarankk: ok

src/main/java/org/apache/cassandra/sidecar/tasks/HealthCheckPeriodicTask.java
yifan-c: `AtomicInteger instanceDown` provides better readability w/o perf sacrifice. I am leaning to the current implementation

sarankk: I tried these changes, we have to process the value of result to get cause. Maintaining the count outside seems simpler, wdyt?

frankgh (approved):
+1 Thanks for addressing the comments.

PR #103, opened 2024-03-06
CASSANDRASC-112 Fix ClosedChannelException when downloading from S3

yifan-c:
patch by Yifan Cai; reviewed by Francisco Guerrero for CASSANDRASC-112


frankgh (approved):
+1 Thanks for the fix and the additional tests

PR #104, opened 2024-03-07
CASSANDRASC-106: Add restore task watcher to report long running tasks

JeetKunDoug:


src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
```diff
@@ -36,6 +36,10 @@ public class RestoreJobConfigurationImpl implements RestoreJobConfiguration
     public static final int DEFAULT_JOB_DISCOVERY_RECENCY_DAYS = 5;
     public static final int DEFAULT_PROCESS_MAX_CONCURRENCY = 20; // process at most 20 slices concurrently
     public static final long DEFAULT_RESTORE_JOB_TABLES_TTL_SECONDS = TimeUnit.DAYS.toSeconds(90);
+    public static final String RESTORE_JOB_LONG_RUNNING_HANDLER_THRESHOLD_SECONDS =
+    "restore_job_long_running_threshold_seconds";
+    // A restore job handler is considered long-running if it has been in the "active" list for 10 minutes.
+    private static final long DEFAULT_RESTORE_JOB_LONG_RUNNING_HANDLER_THRESHOLD_SECONDS = 600;
```
frankgh, line 42: Let's use TimeUnit to define the time for consistency and accuracy
```suggestion
    private static final long DEFAULT_RESTORE_JOB_LONG_RUNNING_HANDLER_THRESHOLD_SECONDS = TimeUnit.MINUTES.toSeconds(10);
```

src/main/java/org/apache/cassandra/sidecar/restore/RestoreSliceTask.java
```diff
@@ -513,4 +524,48 @@ private void logWarnIfHasHttpExceptionCauseOnCommit(Throwable throwable, Restore
...
+        private final RestoreJobException cause;
+        private final RestoreSlice slice;
+
+        public Failed(RestoreJobException cause, RestoreSlice slice)
+        {
+            this.cause = cause;
+            this.slice = slice;
+        }
+
+        public void handle(Promise<RestoreSlice> promise)
```
frankgh, line 566: can we add missing `@Override` annotations here?

src/main/java/org/apache/cassandra/sidecar/restore/RestoreSliceTask.java
```diff
@@ -236,6 +242,11 @@ else if (s3Exception.statusCode() == 400 &&
         });
     }
 
+    private long currentTimeInNanos()
```
frankgh, line 253: missing `@Override` annotations

src/test/java/org/apache/cassandra/sidecar/stats/TestRestoreJobStats.java
```diff
@@ -118,4 +121,9 @@ public void captureTokenRefreshed()
     {
         tokenRefreshCount += 1;
     }
+
+    public void captureLongRunningRestoreHandler(int instanceId, long handlerDuration)
```
frankgh, line 126: NIT missing`@Override` tag

frankgh (commented):
Looks good, I have some minor suggestions

src/main/java/org/apache/cassandra/sidecar/restore/RestoreSliceTask.java
JeetKunDoug: I'm doing a global "fix all the missing `@Override` annotations" on the project as part of this PR - there were 30, and rather than piecemeal fixing them as we go I figure I'd just fix them all :)

src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
JeetKunDoug: So I tried this while implementing it, but it ends up that using a non-constant value means you can't use it as the default value as part of the @JsonProperty annotations... so I'd end up having to put `"600"` in the `@JsonProperty` if we used `TimeUnit.MINUTES.toSeconds(10)` here. 😢

frankgh (approved):
+1 Looks good to me

PR #107, opened 2024-03-14
CASSANDRASC-105 RestoreSliceTask could be stuck due to missing exception handling

sarankk:


frankgh (approved):
+1 Thanks for the fix

PR #109, opened 2024-03-28
CASSANDRASC-115 Add FilteringMetricRegistry to allow filtering of metrics

sarankk:


src/main/dist/conf/sidecar.yaml
```diff
@@ -150,6 +150,12 @@ metrics:
     jmx_domain_name: sidecar.vertx.jmx_domain
     monitored_server_route_regexes:                # regex list to match server routes
      - /api/v1/.*
+  include:
+    - type: "regex"                                # other possible value is "equals"
```
yifan-c: other possible ~value~ _type_ is "equals"

yifan-c (commented):
I have several comments that requires to largely refactor the code, hence sending the commit directly. Besides that, can you add test to check the vertx server metrics can be muted? When the server metrics are excluded (via exclusion configuration)
https://github.com/apache/cassandra-sidecar/commit/60d78f2651d0e735c36ec48c343707acd3384f36

sarankk:
Thanks Yifan!, addressed feedback

src/main/java/org/apache/cassandra/sidecar/config/yaml/MetricsFilteringConfigurationImpl.java
```diff
@@ -0,0 +1,63 @@
...
+        this.value = value;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String type()
+    {
+        return type;
```
frankgh, line 65: type is not validated. We should validate the allowed strings.

src/main/java/org/apache/cassandra/sidecar/metrics/MetricRegistryFactory.java
```diff
@@ -0,0 +1,116 @@
...
+    public MetricRegistry getOrCreate(String name)
+    {
+        // the metric registry already exists
+        if (SharedMetricRegistries.names().contains(name))
+        {
+            return SharedMetricRegistries.getOrCreate(name);
+        }
+
+        FilteringMetricRegistry metricRegistry;
+        readWriteLock.readLock().lock();
```
frankgh: this lock doesn't make sense. I recommend synchronized or a simple lock instead

src/main/java/org/apache/cassandra/sidecar/metrics/MetricRegistryFactory.java
```diff
@@ -0,0 +1,116 @@
...
+    @Inject
+    public MetricRegistryFactory(SidecarConfiguration sidecarConfiguration)
+    {
+        this(sidecarConfiguration.metricsConfiguration().registryName(),
+             MetricFilter.parse(sidecarConfiguration.metricsConfiguration().includeConfigurations()),
+             MetricFilter.parse(sidecarConfiguration.metricsConfiguration().excludeConfigurations()));
+    }
+
+    @VisibleForTesting
+    public MetricRegistryFactory(String globalRegistryName, List<MetricFilter> inclusions, List<MetricFilter> exclusions)
```
frankgh, line 51: use package private for testing?
```suggestion
   MetricRegistryFactory(String globalRegistryName, List<MetricFilter> inclusions, List<MetricFilter> exclusions)
```

src/main/java/org/apache/cassandra/sidecar/metrics/MetricRegistryFactory.java
```diff
@@ -0,0 +1,116 @@
...
+     * <p>
+     * Note that an empty inclusions means including all
+     *
+     * @param name  metric name
+     * @return true if allowed; false otherwise
+     */
+    private boolean isAllowed(String name)
+    {
+        boolean included = inclusions.isEmpty() || inclusions.stream().anyMatch(filter -> filter.matches(name));
+        boolean excluded = exclusions.stream().anyMatch(filter -> filter.matches(name));
```
frankgh, line 100: `if (included && excluded)` should we throw a configuration error?

src/main/java/org/apache/cassandra/sidecar/metrics/FilteringMetricRegistry.java
```diff
@@ -0,0 +1,226 @@
...
+import com.codahale.metrics.Timer;
+import io.vertx.core.impl.ConcurrentHashSet;
+
+/**
+ * Allows filtering of metrics based on configured allow list. Metrics are filtered out before registering them.
+ */
+public class FilteringMetricRegistry extends MetricRegistry
+{
+    private static final NoopMetricRegistry NO_OP_METRIC_REGISTRY = new NoopMetricRegistry(); // supplies no-op metrics
+    private final Predicate<String> isAllowed;
```
frankgh, line 42: Maybe call it predicate?
```suggestion
    private final Predicate<String> isAllowedPredicate;
```

frankgh (commented):
Not a full review, but some progress

src/main/java/org/apache/cassandra/sidecar/config/yaml/MetricsFilteringConfigurationImpl.java
sarankk: You want to fail? We validate while parsing and then only create. If they dont match expected, we function as if that filter was not valid

src/main/java/org/apache/cassandra/sidecar/metrics/MetricRegistryFactory.java
sarankk: Yes discussed, without reconfiguring filters we don't need a lock at all. So removing this

src/main/java/org/apache/cassandra/sidecar/metrics/MetricRegistryFactory.java
sarankk: This is useful in general, if people want to pass filters and create the class, instead of always wanting `SidecarConfiguration` for creating a factor.

src/main/java/org/apache/cassandra/sidecar/metrics/MetricRegistryFactory.java
sarankk: In an example where configuration is 

```
 include:                                   
    - type: "regex"
      value: "vertx.*"
  exclude:                                  
    - type: "regex"                         
      value: "vertx.eventbus.*"
```

We want to include all vert.x metrics except the event bus metrics. In that case, both included and excluded will be true. So it not necessarily a configuration error.

src/main/java/org/apache/cassandra/sidecar/config/yaml/MetricsFilteringConfigurationImpl.java
frankgh: I think it is a reasonable expectation to fail the service startup if misconfigured. As opposed to ignoring the error and have an operator spend a lot of time trying to figure out why something they think should be working is not really working. Take a look at this for example: [testInvalidClientAuth](https://github.com/apache/cassandra-sidecar/blob/trunk/src/test/java/org/apache/cassandra/sidecar/config/SidecarConfigurationTest.java#L176-L184)

src/main/java/org/apache/cassandra/sidecar/metrics/FilteringMetricRegistry.java
sarankk: I feel adding `Predicate` to name is redundant. For consistency, lets keep it this way? since we don't add that for other members

src/main/java/org/apache/cassandra/sidecar/metrics/MetricFilter.java
```diff
@@ -0,0 +1,89 @@
...
+    {
+        private final Pattern pattern;
+
+        public Regex(String regex)
+        {
+            Objects.requireNonNull(regex);
+            this.pattern = Pattern.compile(regex);
+        }
+
+        public boolean matches(String name)
```
frankgh, line 72: can you add override annotations ?

src/main/java/org/apache/cassandra/sidecar/metrics/MetricFilter.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.regex.Pattern;
+
+import org.apache.cassandra.sidecar.config.MetricsFilteringConfiguration;
+
+/**
+ * Filter for deciding whether a metric should be captured
+ */
+public abstract class MetricFilter
+{
+    public abstract boolean matches(String name);
```
frankgh, line 39: javadocs?

src/main/java/org/apache/cassandra/sidecar/metrics/FilteringMetricRegistry.java
frankgh: I don't feel the variable name fully conveys what the variable is supposed to do. Maybe another name is possible too?

src/main/java/org/apache/cassandra/sidecar/metrics/MetricRegistryFactory.java
frankgh: fair, makes sense

sarankk:
Thanks @frankgh for the comments. Addressed them and your comment about remembering the predicate results with a `CachedPredicate`

src/test/java/org/apache/cassandra/sidecar/metrics/FilteringMetricRegistryTest.java
```diff
@@ -0,0 +1,216 @@
...
+              .onSuccess(v -> {
+                  MetricRegistryFactory registryFactory = injector.getInstance(MetricRegistryFactory.class);
+                  Pattern excludedPattern = Pattern.compile("vertx.eventbus.*");
+                  FilteringMetricRegistry globalRegistry = (FilteringMetricRegistry) registryFactory.getOrCreate();
+                  assertThat(globalRegistry.getIncludedMetrics().size()).isGreaterThanOrEqualTo(1);
+                  assertThat(globalRegistry.getIncludedMetrics()
+                                           .keySet()
+                                           .stream()
+                                           .noneMatch(key -> excludedPattern.matcher(key).matches()))
+                  .isTrue();
```
frankgh: You can instead directly do `noneMatch` within the assertion.
```suggestion
                  assertThat(globalRegistry.getIncludedMetrics()
                                           .keySet()
                                           .stream())
                                           .noneMatch(key -> excludedPattern.matcher(key).matches()));
```

src/main/java/org/apache/cassandra/sidecar/config/yaml/MetricsFilteringConfigurationImpl.java
sarankk: Got it will add the check to configImpl

src/test/java/org/apache/cassandra/sidecar/metrics/FilteringMetricRegistryTest.java
sarankk: I already use `noneMatch` are you suggesting removing `isTrue`?

sarankk: Addressed this

src/main/java/org/apache/cassandra/sidecar/metrics/FilteringMetricRegistry.java
sarankk: For readability added `Predicate` extension

src/main/java/org/apache/cassandra/sidecar/metrics/MetricRegistryFactory.java
frankgh: thanks for the explanation, yeah this makes sense

PR #111, opened 2024-04-09
CASSANDRASC-117 Record existing and additional metrics with dropwizard

sarankk:


src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceHealthMetrics.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Tracks health related metrics of a Cassandra instance maintained by Sidecar.
+ */
+public class InstanceHealthMetrics
+{
+    public static final String DOMAIN = INSTANCE_PREFIX + ".health";
+    protected final MetricRegistry metricRegistry;
+    public final NamedMetric<Meter> jmxDown;
+    public final NamedMetric<Meter> nativeDown;
```
yifan-c: why are they meters? should operators care about the rate of it gets down? Should it be a gauge and publish the number of down instances?

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceRestoreMetrics.java
```diff
@@ -0,0 +1,135 @@
...
+                     .build();
+        sliceUncompressedSizeInBytes
+        = NamedMetric.builder(name -> metricRegistry.gauge(name, () -> new DefaultSettableGauge<>(0L)))
+                     .withDomain(DOMAIN)
+                     .withName("slice_uncompressed_size_bytes")
+                     .build();
+        longRunningRestoreHandler
+        = NamedMetric.builder(metricRegistry::timer)
+                     .withDomain(DOMAIN)
+                     .withName("long_running_restore_handler")
```
yifan-c: rename to `SlowRestoreTask`

src/main/java/org/apache/cassandra/sidecar/metrics/RestoreMetrics.java
```diff
@@ -0,0 +1,93 @@
...
+        successfulJobs
+        = NamedMetric.builder(metricRegistry::meter)
+                     .withDomain(DOMAIN)
+                     .withName("successful_jobs")
+                     .build();
+        failedJobs
+        = NamedMetric.builder(metricRegistry::meter)
+                     .withDomain(DOMAIN)
+                     .withName("failed_jobs")
+                     .build();
```
yifan-c: I think they should be gauge, publish 1 to mark it as succeeded/failed. We do not need the rate.

src/main/java/org/apache/cassandra/sidecar/metrics/RestoreMetrics.java
```diff
@@ -0,0 +1,93 @@
...
+        activeJobs
+        = NamedMetric.builder(name -> metricRegistry.gauge(name, () -> new DefaultSettableGauge<>(0)))
+                     .withDomain(DOMAIN)
+                     .withName("active_jobs")
+                     .build();
+        tokenRefreshes
+        = NamedMetric.builder(metricRegistry::meter)
+                     .withDomain(DOMAIN)
+                     .withName("token_refreshes")
+                     .build();
```
yifan-c: `TokenRefreshed`

src/main/java/org/apache/cassandra/sidecar/metrics/RestoreMetrics.java
```diff
@@ -0,0 +1,93 @@
...
+        tokenUnauthorized
+        = NamedMetric.builder(metricRegistry::meter)
+                     .withDomain(DOMAIN)
+                     .withName("token_unauthorized")
+                     .build();
+        tokenExpired
+        = NamedMetric.builder(metricRegistry::meter)
+                     .withDomain(DOMAIN)
+                     .withName("token_expired")
+                     .build();
```
yifan-c: They should be gauge.

src/main/java/org/apache/cassandra/sidecar/metrics/SchemaMetrics.java
```diff
@@ -0,0 +1,54 @@
...
+        failedInitializations
+        = NamedMetric.builder(metricRegistry::meter)
+                     .withDomain(DOMAIN)
+                     .withName("failed_initializations")
+                     .build();
+        failedModifications
+        = NamedMetric.builder(metricRegistry::meter)
+                     .withDomain(DOMAIN)
+                     .withName("failed_modifications")
+                     .build();
```
yifan-c: use gauge

src/test/java/org/apache/cassandra/sidecar/concurrent/ExecutorPoolsTest.java
```diff
@@ -27,31 +27,40 @@
...
+import org.apache.cassandra.sidecar.metrics.ResourceMetrics;
 
+import static org.apache.cassandra.sidecar.utils.TestMetricUtils.registry;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 /**
  * Test {@link ExecutorPools}
  */
+@ExtendWith(VertxExtension.class)
```
yifan-c: `VertxExtension` is not required. 
You can move `assertThat(metrics.longTasks.metric.getCount()).isEqualTo(200);` after countdown latch has passed, so you can get rid of VertxTestContext from the test.

src/main/java/org/apache/cassandra/sidecar/concurrent/ExecutorPools.java
```diff
@@ -186,11 +208,31 @@ public long setTimer(long delay, Handler<Long> handler)
...
+            }
+
+            if (workerPoolName.equals(SERVICE_POOL))
+            {
+                metrics.shortTaskTimeTaken.metric.update(duration, TimeUnit.NANOSECONDS);
+            }
+            else if (workerPoolName.equals(INTERNAL_POOL))
+            {
+                metrics.longTaskTimeTaken.metric.update(duration, TimeUnit.NANOSECONDS);
+            }
```
yifan-c: Using pool name within the pool implementation to determine the instance type is smelly. Let's not do that.

You can change `TaskExecutorPool` to be abstract with abstract method `recordTimeTaken` declared. The abstract class does not need to hold a reference to the metrics. The following code snippet shows the idea. 

```java
    public abstract static class TaskExecutorPool implements WorkerExecutor
    {
        private final Vertx vertx;
        private final WorkerExecutor workerExecutor;

        private TaskExecutorPool(Vertx vertx, WorkerPoolConfiguration config)
        {
            ...
        }

        // record time taken (timer) already records the count. there is no need to declare recordTask().
        protected abstract void recordTimeTaken(long duration); 
    }

    private static class ServiceTaskExecutorPool extends TaskExecutorPool
    {
        private final ResourceMetrics metrics;
        
        private ServiceTaskExecutorPool(Vertx vertx, WorkerPoolConfiguration config, ResourceMetrics metrics)
        {
            super(vertx, config);
            this.metrics = metrics;
        }

        @Override
        protected void recordTimeTaken(long duration)
        {
            // actual publishing code
        }
    }

    // similarly, add the pool for internal
    private static class InternalTaskExecutorPool extends TaskExecutorPool
```

src/main/java/org/apache/cassandra/sidecar/metrics/ResourceMetrics.java
```diff
@@ -0,0 +1,68 @@
...
+ */
+@Singleton
+public class ResourceMetrics
+{
+    public static final String DOMAIN = "sidecar.resource";
+    protected final MetricRegistry metricRegistry;
+    public final NamedMetric<Meter> shortLivedTasks;
+    public final NamedMetric<Meter> longTasks;
+    public final NamedMetric<Timer> shortTaskTimeTaken;
+    public final NamedMetric<Timer> longTaskTimeTaken;
```
yifan-c: I think only timer is needed here. Timer has a meter inside. 

Can you also rename the metrics to `serviceTaskTimer` and `internalTaskTimer`? Service pool is to serve tasks from service API. Internal pool is to serve the background/operational/internal/ tasks.

yifan-c (commented):
Not a complete review, but submit what I have so far.

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceHealthMetrics.java
sarankk: Meters also give count of instances. But it is more expensive since it captures moving averages. Will change these to Gauges

src/main/java/org/apache/cassandra/sidecar/concurrent/TaskExecutorPool.java
```diff
@@ -0,0 +1,277 @@
...
+     * @param handler the handler that will be called with the timer ID when the timer fires
+     * @return the unique identifier for the timer
+     */
+    public long setTimer(long delay, Handler<Long> handler)
+    {
+        return vertx.setTimer(delay, id ->
+                                     workerExecutor.executeBlocking(promise -> {
+                                         long startTime = System.nanoTime();
+                                         handler.handle(id);
+                                         recordTimeTaken(System.nanoTime() - startTime);
```
frankgh: unrelated to the PR, but do we need to complete the `promise` here?

src/main/java/org/apache/cassandra/sidecar/metrics/Timer.java
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-package org.apache.cassandra.sidecar.stats;
+package org.apache.cassandra.sidecar.metrics;
```
frankgh: we have some unused methods in this class. Should we clean up?

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceRestoreMetrics.java
```diff
@@ -0,0 +1,113 @@
...
+        = NamedMetric.builder(metricRegistry::timer).withDomain(DOMAIN).withName("SliceValidationTime").build();
+        sliceDownloadTimeouts
+        = NamedMetric.builder(name -> metricRegistry.gauge(name, () -> new DefaultSettableGauge<>(0)))
+                     .withDomain(DOMAIN)
+                     .withName("SliceDownloadTimeouts")
+                     .build();
+        sliceDownloadRetries
+        = NamedMetric.builder(name -> metricRegistry.gauge(name, () -> new DefaultSettableGauge<>(0)))
+                     .withDomain(DOMAIN)
+                     .withName("SliceCompletionRetries")
```
frankgh: The field is called `sliceDownloadRetries`, should the name match the field ?

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceRestoreMetrics.java
```diff
@@ -0,0 +1,113 @@
...
+        pendingSliceCount
+        = NamedMetric.builder(name -> metricRegistry.gauge(name, () -> new DefaultSettableGauge<>(0)))
+                     .withDomain(DOMAIN)
+                     .withName("PendingSliceCount")
+                     .build();
+        dataSSTableComponentSize
+        = NamedMetric.builder(name -> metricRegistry.gauge(name, () -> new DefaultSettableGauge<>(0L)))
+                     .withDomain(DOMAIN)
+                     .withName("RestoreDataSizeBytes")
+                     .addTag("Component", "db")
```
frankgh: Should this be Data.db instead? db is not unique for the data file

src/main/java/org/apache/cassandra/sidecar/metrics/instance/SSTableImportMetrics.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Tracks metrics during SSTable import operation done in Cassandra instance
+ */
+public class SSTableImportMetrics
+{
+    public static final String DOMAIN = INSTANCE_PREFIX + ".SSTableImport";
+    protected final MetricRegistry metricRegistry;
+    public final NamedMetric<DefaultSettableGauge<Integer>> cassandraUnavailable;
+    public final NamedMetric<DefaultSettableGauge<Integer>> pendingImports;
```
frankgh, line 37: I think it makes sense to track successful imports as well, what do you think?

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableMetrics.java
```diff
@@ -55,8 +62,9 @@ public static class UploadSSTableComponentMetrics
     {
         protected final MetricRegistry metricRegistry;
         public final String sstableComponent;
-        public final NamedMetric<Meter> rateLimitedCalls;
-        public final NamedMetric<Meter> diskUsageHigh;
+        public final NamedMetric<DefaultSettableGauge<Integer>> rateLimitedCalls;
+        public final NamedMetric<DefaultSettableGauge<Integer>> diskUsageHigh;
```
frankgh: nit
```suggestion
        public final NamedMetric<DefaultSettableGauge<Integer>> highDiskUsage;
```

src/main/java/org/apache/cassandra/sidecar/metrics/RestoreMetrics.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+
+/**
+ * Tracks metrics related to restore functionality provided by Sidecar.
+ */
+@Singleton
+public class RestoreMetrics
+{
+    public static final String DOMAIN = "sidecar.restore";
```
frankgh: NIT
```suggestion
    public static final String DOMAIN = "Sidecar.Restore";
```

src/main/java/org/apache/cassandra/sidecar/metrics/instance/InstanceRestoreMetrics.java
```diff
@@ -0,0 +1,113 @@
...
+import org.apache.cassandra.sidecar.metrics.NamedMetric;
+
+import static org.apache.cassandra.sidecar.metrics.instance.InstanceMetrics.INSTANCE_PREFIX;
+
+/**
+ * {@link InstanceRestoreMetrics} contains metrics to track restore task for a Cassandra instance maintained by Sidecar.
+ */
+public class InstanceRestoreMetrics
+{
+    public static final String DOMAIN = INSTANCE_PREFIX + ".restore";
```
frankgh: NIT: for consistency with the uppercasing in other metrics
```suggestion
    public static final String DOMAIN = INSTANCE_PREFIX + ".Restore";
```

src/main/java/org/apache/cassandra/sidecar/restore/RestoreProcessor.java
```diff
@@ -141,21 +146,26 @@ public void execute(Promise<Void> promise)
...
                                                          requiredUsableSpacePercentage,
-                                                         sliceDatabaseAccessor, stats,
-                                                         restoreJobUtil);
+                                                         sliceDatabaseAccessor,
+                                                         restoreJobUtil,
+                                                         restoreMetrics);
             activeTasks.add(task);
             pool.executeBlocking(task, false) // unordered; run in parallel
             .onSuccess(restoreSlice -> {
                 int instanceId = slice.owner().id();
```
frankgh: `instanceId` is now unused

src/main/java/org/apache/cassandra/sidecar/restore/RestoreProcessor.java
```diff
@@ -312,13 +325,17 @@ void decrementActiveSliceCount(RestoreSlice slice)
         void captureImportQueueLength()
         {
             activeSliceCounterPerInstance.forEach((instanceId, counter) ->
-                                                  stats.captureSliceImportQueueLength(instanceId, counter.get()));
+                                                  metadataFetcher.instance(instanceId)
```
frankgh: should capture be specific to the instance we are processing at the moment? i.e. just pass the `slice` as parameter and report it for the individual slice?

src/main/java/org/apache/cassandra/sidecar/config/yaml/VertxMetricsConfigurationImpl.java
```diff
@@ -32,34 +29,28 @@ public class VertxMetricsConfigurationImpl implements VertxMetricsConfiguration
     public static final boolean DEFAULT_ENABLED = true;
     public static final boolean DEFAULT_EXPOSE_VIA_JMX = false;
     public static final String DEFAULT_JMX_DOMAIN_NAME = "sidecar.vertx.jmx_domain";
-    public static final List<String> DEFAULT_MONITORED_SERVER_ROUTE_REGEXES = Collections.singletonList("/api/v1/.*");
```
frankgh, line 35: where will this go?

src/main/java/org/apache/cassandra/sidecar/concurrent/TaskExecutorPool.java
```diff
@@ -0,0 +1,277 @@
...
+
+    @Override
+    public <T> Future<T> executeBlocking(Handler<Promise<T>> blockingCodeHandler)
+    {
+        return Timer.measureTimeTaken(workerExecutor.executeBlocking(blockingCodeHandler), this::recordTimeTaken);
+    }
+
+    /**
+     * Records time taken for tasks executed by {@link TaskExecutorPool}
+     * @param duration time taken by a task
```
frankgh: we are missing the units here

src/main/java/org/apache/cassandra/sidecar/concurrent/TaskExecutorPool.java
```diff
@@ -0,0 +1,277 @@
...
+     */
+    protected abstract void recordTimeTaken(long duration);
```
frankgh:
```suggestion
    protected abstract void recordTimeTaken(long durationNanos);
```

src/main/java/org/apache/cassandra/sidecar/concurrent/TaskExecutorPool.java
```diff
@@ -0,0 +1,277 @@
...
+        ServiceTaskExecutorPool(Vertx vertx, WorkerPoolConfiguration config, ResourceMetrics metrics)
+        {
+            super(vertx, config);
+            this.metrics = metrics;
+        }
+
+        @Override
+        protected void recordTimeTaken(long duration)
+        {
+            if (metrics == null)
```
frankgh, line 224: will `metrics` ever be null? it doesn't seem like we do null checks for other metrics anywhere else in the code

src/main/java/org/apache/cassandra/sidecar/utils/SSTableImporter.java
```diff
@@ -159,16 +159,45 @@ protected ImportQueue initializeQueue(String key)
...
+
+    /**
+     * Host is extracted from key for publishing Cassandra instance specific metrics related to {@link ImportQueue}
+     *
+     * @param key a key for one of the import queues
+     * @return host for which the import queue is maintained
+     */
+    protected String hostFromKey(String key)
+    {
+        String[] keyParts = key.split("\\$");
```
frankgh: do we really want to parse the key here? is there a better way to do this?

src/main/java/org/apache/cassandra/sidecar/metrics/instance/SSTableImportMetrics.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.apache.cassandra.sidecar.metrics.instance.InstanceMetrics.INSTANCE_PREFIX;
+
+/**
+ * Tracks metrics during SSTable import operation done in Cassandra instance
+ */
+public class SSTableImportMetrics
+{
+    public static final String DOMAIN = INSTANCE_PREFIX + ".SSTableImport";
+    protected final MetricRegistry metricRegistry;
+    public final NamedMetric<DefaultSettableGauge<Integer>> cassandraUnavailable;
```
frankgh: These metrics are a bit confusing. For `cassandraUnavailable` , it tracks on the handler but it doesn't track on the `SSTableImporter`. For `pendingImports`, it tracks on `SSTableImporter` . Should `cassandraUnavailable` track for the unavailable in `SSTableImporter` as well?

src/main/java/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableUploadHandler.java
```diff
@@ -127,7 +123,7 @@ public void handleInternal(RoutingContext context,
         if (!limiter.tryAcquire())
         {
             String message = String.format("Concurrent upload limit (%d) exceeded", limiter.limit());
-            componentMetrics.rateLimitedCalls.metric.mark();
+            componentMetrics.rateLimitedCalls.metric.setValue(1);
```
frankgh: The component metrics seem wrong or misleading. When I think of components I think of SSTable Components. https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/io/sstable/format/big/BigFormat.java#L206  But you are really tracking by extension, which is different than component. I think we should either rename the metric, or if we really want to track SSTable components, we should properly determine the SSTable component.

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableMetrics.java
frankgh: Why is the disk usage associated to the SSTable component? Wouldn't this be a metric that tracks the staging directory instead?

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableMetrics.java
```diff
@@ -71,15 +79,21 @@ public UploadSSTableComponentMetrics(MetricRegistry metricRegistry, String sstab
             NamedMetric.Tag componentTag = NamedMetric.Tag.of("component", sstableComponent);
 
             rateLimitedCalls
-            = NamedMetric.builder(metricRegistry::meter)
+            = NamedMetric.builder(name -> metricRegistry.gauge(name, () -> new DefaultSettableGauge<>(0)))
```
frankgh: I'm not sure if rate limited calls are tied to a specific component. Is there a reason it matters to make it component specific?

frankgh (commented):
I've left a few comments, but I think we need to address the SSTable Component metrics, because it seems they are not really SSTable Component per se, but rather based on the extension. So either the metric will need to be extension based metric, or we need to correctly determine the SSTable Component

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableMetrics.java
sarankk: This is what we track specifically for upload handler

src/main/java/org/apache/cassandra/sidecar/metrics/instance/SSTableImportMetrics.java
sarankk: I missed the unavailable exceptions thrown there. Will add there

src/main/java/org/apache/cassandra/sidecar/utils/SSTableImporter.java
sarankk: As discussed will capture the metric while draining the queue

src/main/java/org/apache/cassandra/sidecar/concurrent/TaskExecutorPool.java
sarankk: This one can be, because we might want to allow creation of `ExecutorPools` without metrics. Since `ExecutorPools` is also created from integration tests. Thats why we do the null check only here.

src/main/java/org/apache/cassandra/sidecar/config/yaml/VertxMetricsConfigurationImpl.java
sarankk: We decided to not use this. Instead use the include and exclude pattern we provide for filtering.

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableMetrics.java
frankgh: It doesn't make much sense to track high disk usage on a per component basis. Maybe a drive makes sense, but it doesn't make sense for a specific component

src/main/java/org/apache/cassandra/sidecar/config/yaml/VertxMetricsConfigurationImpl.java
frankgh: should we add it in  `sidecar.yaml` then?

src/main/java/org/apache/cassandra/sidecar/metrics/Timer.java
sarankk: These methods were unused before too. I moved them as it is. @yifan-c

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableMetrics.java
sarankk: We want to know for which SSTable components highDiskUsage metric is captured. We also track usable space in staging directory in general and that is a separate metric in `InstanceResourceMetrics`

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableMetrics.java
sarankk: Moved them out. But I feel for highDiskUsage it might be valuable to keep component detail

src/main/java/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableUploadHandler.java
sarankk: Addressed this. Switched to using SSTable component name.

sarankk:
Thanks @frankgh for the review. Addressed them. Modified to using SSTable component name instead of extension while tagging metrics

src/main/java/org/apache/cassandra/sidecar/metrics/Timer.java
yifan-c: yep. let's keep them as is. Not related with this PR anyway

src/main/java/org/apache/cassandra/sidecar/config/yaml/VertxMetricsConfigurationImpl.java
sarankk: Discussed with Fracisco, we will use the exclude list instead of configuring monitored server routes

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableMetrics.java
sarankk: Discussed with Francisco, removing the metric per component.

src/main/java/org/apache/cassandra/sidecar/concurrent/TaskExecutorPool.java
yifan-c: It is wrong w/o fulfilling the promise. I have a feedback commit (to be pushed) to address it.

src/main/java/org/apache/cassandra/sidecar/metrics/instance/SSTableImportMetrics.java
yifan-c: When `cassandraUnavailable`, the import fails, which can be classified under `failedImports`. Can we remove `cassandraUnavailable` from `SSTableImportMetrics`? 
Besides that, `InstanceHealthMetrics` already captures `jmxDown`, which leads to `cassandraUnavailable` here.

src/main/java/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableImportHandler.java
```diff
@@ -162,18 +164,20 @@ protected SSTableImportRequest extractParamsOrThrow(RoutingContext context)
...
         {
+            metrics.cassandraUnavailable.metric.setValue(1);
             return Future.failedFuture(cassandraServiceUnavailable());
         }
 
         TableOperations tableOperations = cassandra.tableOperations();
 
         if (tableOperations == null)
         {
+            metrics.cassandraUnavailable.metric.setValue(1);
```
yifan-c: I still think we should use `failedImports` only. But if keeping `cassandraUnavailable`, I want to call out that the metric value is never set to 0.

src/main/java/org/apache/cassandra/sidecar/metrics/instance/UploadSSTableMetrics.java
```diff
@@ -33,13 +34,26 @@
  */
 public class UploadSSTableMetrics
 {
-    public static final String DOMAIN = INSTANCE_PREFIX + ".upload_sstable";
+    public static final String DOMAIN = INSTANCE_PREFIX + ".UploadSSTable";
     protected final MetricRegistry metricRegistry;
     protected final Map<String, UploadSSTableComponentMetrics> uploadComponentMetrics = new ConcurrentHashMap<>();
+    public final NamedMetric<Meter> totalBytesUploadedRate;
+    public final NamedMetric<DefaultSettableGauge<Integer>> rateLimitedCalls;
```
yifan-c: 1. how about just rename it to `Throttled`? The same as the metric name.
2. The metric type needs to be `DeltaGauge`. We want to capture the number of throttled calls between the publishing intervals.

src/main/java/org/apache/cassandra/sidecar/metrics/instance/StreamSSTableMetrics.java
```diff
@@ -34,13 +34,26 @@
  */
 public class StreamSSTableMetrics
 {
-    public static final String DOMAIN = INSTANCE_PREFIX + ".stream_sstable";
+    public static final String DOMAIN = INSTANCE_PREFIX + ".StreamSSTable";
     protected final MetricRegistry metricRegistry;
     protected final Map<String, StreamSSTableComponentMetrics> streamComponentMetrics = new ConcurrentHashMap<>();
+    public final NamedMetric<Meter> totalBytesStreamedRate;
+    public final NamedMetric<DefaultSettableGauge<Integer>> rateLimitedCalls;
```
yifan-c: same here. Rename and change type to `DeltaGauge`

yifan-c (commented):
The comments and the feedback commit at https://github.com/yifan-c/cassandra-sidecar/commit/b12517bd09cd18538a577a0a9cfc389d222ab5c5

src/main/java/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableImportHandler.java
sarankk: Removed `cassandraUnavailable`

sarankk:
Thanks @yifan-c for the feedback, addressed it.

yifan-c (approved):

src/main/java/org/apache/cassandra/sidecar/metrics/instance/SSTableImportMetrics.java
sarankk: Removed this metric

frankgh (approved):
+1 thanks for addressing all the comments!

PR #117, opened 2024-04-25
CASSANDRASC-125: Import Queue pendingImports metrics is reporting an …

frankgh:
…incorrect value

The pending imports metric does not aggregate across all keyspaces/tables, in this commit we aggregate the queue sizes and report on a per host basis.

Patch by Francisco Guerrero; Reviewed by Yifan Cai for CASSANDRASC-125

### DIFF
--- a/src/main/java/org/apache/cassandra/sidecar/snapshots/SnapshotPathBuilder.java
+++ b/src/main/java/org/apache/cassandra/sidecar/snapshots/SnapshotPathBuilder.java
@@ -259,6 +259,7 @@ public class SnapshotPathBuilder extends BaseFileSystem
         public final long size;
         public final int dataDirectoryIndex;
         public final String tableId;
+        private final int hashCode;
 
         @VisibleForTesting
         SnapshotFile(Path path, long size, int dataDirectoryIndex, String tableId)
@@ -273,6 +274,7 @@ public class SnapshotPathBuilder extends BaseFileSystem
             this.size = size;
             this.dataDirectoryIndex = dataDirectoryIndex;
             this.tableId = tableId;
+            this.hashCode = Objects.hash(name, size, dataDirectoryIndex, tableId);
         }
 
         @Override
@@ -301,7 +303,7 @@ public class SnapshotPathBuilder extends BaseFileSystem
         @Override
         public int hashCode()
         {
-            return Objects.hash(name, size, dataDirectoryIndex, tableId);
+            return hashCode;
         }
     }
 
--- a/src/main/java/org/apache/cassandra/sidecar/snapshots/SnapshotPathBuilder.java
+++ b/src/main/java/org/apache/cassandra/sidecar/snapshots/SnapshotPathBuilder.java
@@ -128,7 +128,7 @@ public class SnapshotPathBuilder extends BaseFileSystem
                                 }
 
                                 String snapshotFileName = snapshotFile.subpath(snapshotDirNameCount, snapshotFile.getNameCount()).toString();
-                                return new SnapshotFile(snapshotFileName, snapshotFile, attrs.size(), dataDirectoryIndex, tableId);
+                                return new SnapshotFile(snapshotFileName, attrs.size(), dataDirectoryIndex, tableId);
                             }
                             catch (IOException e)
                             {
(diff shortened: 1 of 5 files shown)


src/main/java/org/apache/cassandra/sidecar/utils/SSTableImporter.java
```diff
@@ -623,4 +645,38 @@ public ImportOptions build()
             }
         }
     }
+
+    /**
+     * Key used for the map of queues
+     */
+    protected static class ImportKey
```
yifan-c: 1. Maybe call it `ImportId`?
2. Can change the access modifier to `private`? I think it is unexpected to create its subclasses. `protected` adds maintenance burden

src/main/java/org/apache/cassandra/sidecar/utils/SSTableImporter.java
```diff
@@ -623,4 +645,38 @@ public ImportOptions build()
...
+            ImportKey importKey = (ImportKey) o;
+            return Objects.equals(host, importKey.host)
+                   && Objects.equals(keyspace, importKey.keyspace)
+                   && Objects.equals(table, importKey.table);
+        }
+
+        @Override
+        public int hashCode()
+        {
+            return Objects.hash(host, keyspace, table);
```
yifan-c: NIT: The object is only going to be used as map keys. We can store the hashcode as an instance variable. (so compute only once)

src/main/java/org/apache/cassandra/sidecar/snapshots/SnapshotPathBuilder.java
```diff
@@ -298,14 +295,13 @@ public boolean equals(Object object)
...
                    && Objects.equals(name, that.name)
-                   && Objects.equals(path, that.path)
                    && Objects.equals(tableId, that.tableId);
         }
 
         @Override
         public int hashCode()
         {
-            return Objects.hash(name, path, size, dataDirectoryIndex, tableId);
+            return Objects.hash(name, size, dataDirectoryIndex, tableId);
```
yifan-c: similar nit: the `hashcode` can be pre-computed.

src/main/java/org/apache/cassandra/sidecar/utils/SSTableImporter.java
```diff
@@ -310,6 +307,31 @@ private void cleanup(ImportOptions options)
                                                  cause));
     }
 
+    /**
+     * Aggregates pending imports per host from multiple keyspaces and tables
+     */
+    private void reportPendingImportAggregates()
```
yifan-c: nit: the method name is a bit confusing with "Aggregates". How about `reportPendingImportPerHost`?

src/main/java/org/apache/cassandra/sidecar/utils/SSTableImporter.java
```diff
@@ -310,6 +307,31 @@ private void cleanup(ImportOptions options)
...
+            {
+                // Report aggregate metrics for the queues
+                InstanceMetadata instance = metadataFetcher.instance(host);
+                instance.metrics().sstableImport().pendingImports.metric.setValue(count);
+            }
+            catch (Exception e)
+            {
+                LOGGER.warn("Unable to report metrics for host={} pendingImports={}", host, count, e);
+            }
+        });
```
yifan-c: also a nit
```suggestion
        importQueuePerHost
        .entrySet()
        .stream()
        .collect(Collectors.groupingBy(e -> e.getKey().host, Collectors.counting()))
        .forEach((host, count) -> {
            try
            {
                // Report aggregate metrics for the queues
                InstanceMetadata instance = metadataFetcher.instance(host);
                instance.metrics().sstableImport().pendingImports.metric.setValue(count.intValue());
            }
            catch (Exception e)
            {
                LOGGER.warn("Unable to report metrics for host={} pendingImports={}", host, count, e);
            }
        });
```

yifan-c (commented):
lgtm, some nits

src/main/java/org/apache/cassandra/sidecar/utils/SSTableImporter.java
frankgh: I find this not as readable , so that's why I went the other way.

PR #123, opened 2024-05-24
CASSANDRASC-132: Add restore job progress endpoint and consistency check on restore ranges

yifan-c:
The patch adds a new web API to retrieve the restore job progress that is computed using the current Cassandra cluster topology and the consistency level required for the restore job.

Patch by Yifan Cai; Reviewed by Francisco Guerrero for CASSANDRASC-132


client-common/src/main/java/org/apache/cassandra/sidecar/common/data/RestoreJobConstants.java
```diff
@@ -48,4 +49,12 @@ public class RestoreJobConstants
...
     public static final String CREDENTIALS_SESSION_TOKEN = "sessionToken";
     public static final String CREDENTIALS_REGION = "region";
+
+    // -- restore job status fields
+    public static final String JOB_STATUS_MESSAGE = "message";
+    public static final String JOB_STATUS_SUMMARY = "summary";
+    public static final String JOB_STATUS_FAILED_RANGE = "failedRanges";
+    public static final String JOB_STATUS_ABORTED_RANGE = "abortedRanges";
+    public static final String JOB_STATUS_PENDING_RANGE = "pendingRanges";
+    public static final String JOB_STATUS_SUCCEEDED_RANGE = "succeededRanges";
```
bbotella: Nit: Should we end the const names as plurals? `*_RANGES`

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CreateRestoreJobRequestPayload.java
```diff
@@ -205,31 +226,48 @@ public static class Builder
...
+        public Builder self()
         {
-            this.consistencyLevel = consistencyLevel;
             return this;
         }
 
         public CreateRestoreJobRequestPayload build()
         {
+            Preconditions.checkArgument(!consistencyLevel.isLocalDcOnly || (localDc != null && !localDc.isEmpty()),
+                                        "Must specify a non-empty localDc for consistency level: " + consistencyLevel);
```
bbotella: Nit: `"Must specify a non-empty " + JOB_LOCAL_DATA_CENTER + " for consistency level: "`?

client-common/src/main/java/org/apache/cassandra/sidecar/common/utils/Preconditions.java
```diff
@@ -32,10 +34,29 @@ public class Preconditions
...
+    public static void expect(boolean validCondition, Supplier<RuntimeException> exceptionSupplier)
     {
         if (!validCondition)
         {
-            throw new IllegalArgumentException(errorMessage);
+            throw exceptionSupplier.get();
+        }
+    }
+
+    public static void throwsOn(boolean invalidCondition, Supplier<RuntimeException> exceptionSupplier)
```
bbotella: Javadoc?

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/CQLSessionProvider.java
```diff
@@ -31,23 +31,24 @@
 public interface CQLSessionProvider
 {
     /**
-     * Provides a Session connected to the cluster. If null it means the connection was
-     * could not be established. The session still might throw a NoHostAvailableException if the
-     * cluster is otherwise unreachable.
+     * Provides a Session connected to the cluster; otherwise, it tries to connect to the cluster.
+     * Returning null means the connection can not be established.
```
bbotella, line 35: Why null? Would it we worth to throw a different exception? Something like `ConnectionCouldNotBeStablishedException`?

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/cluster/locator/TokenRange.java
```diff
@@ -75,18 +76,62 @@ public TokenRange(long start, long end)
...
-    public BigInteger end()
+    /**
+     * @return the first token enclosed in the range
+     */
+    public Token firstToken()
+    {
+        return range.lowerEndpoint().increment();
+    }
+
+    public boolean encloses(TokenRange other)
```
bbotella: Add some javadocs here?

server-common/src/test/java/org/apache/cassandra/sidecar/common/server/utils/ThrowableUtilsTest.java
```diff
@@ -69,4 +69,10 @@ void testGetCauseWithPredicate()
                                                         && cause.getMessage().equals("non-existing exception")))
         .isNull();
     }
+
+    @Test
+    void testPropogate()
```
bbotella: Test logic missing

src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
```diff
@@ -56,11 +55,12 @@ public class RestoreJobConfigurationImpl implements RestoreJobConfiguration
     @JsonProperty(value = "restore_job_tables_ttl_seconds")
     protected final long restoreJobTablesTtlSeconds;
 
-
-    @JsonProperty(value = RESTORE_JOB_LONG_RUNNING_HANDLER_THRESHOLD_SECONDS,
-    defaultValue = DEFAULT_RESTORE_JOB_LONG_RUNNING_HANDLER_THRESHOLD_SECONDS + "")
+    @JsonProperty(value = "restore_job_long_running_threshold_seconds")
```
bbotella: Why are we removing the constant to hardcode the string?

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/CQLSessionProvider.java
yifan-c: The API has been like this that can return nullable result. I am only updating the JavaDoc here to clarify the behavior. Any API change here is unrelated and should be avoided. 

I like the idea of throwing exception instead of returning null in this case. It seems to be able to improve the code clarity. Let's give it more thought. And probably do it in a follow-up.

src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
yifan-c: It is mainly to remove the `defaultValue` declared in the annotation. The constants are removed to be consistent with other fields.

bbotella:
Thanks for addressing the comments!
+1

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -96,17 +96,19 @@ public final class ApiEndpointsV1
     public static final String KEYSPACE_TOKEN_MAPPING_ROUTE = API_V1 + PER_KEYSPACE + "/token-range-replicas";
 
     // Blob Transport Extension
-    public static final String JOB_ID_PATH_PARAM = ":jobId";
+    // todo: refactor other path param use cases; and replaceAll in requestURI building
```
frankgh: I think this can just be a JIRA, instead of a comment in the code.

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -96,17 +96,19 @@ public final class ApiEndpointsV1
...
+    public static final PathParam JOB_ID_PATH_PARAM = PathParam.of("jobId");
```
frankgh: I think this is overly complex, maybe it deserves to be its own patch if you wanted to change it. I'd keep it simple here following the existing pattern

client-common/src/main/java/org/apache/cassandra/sidecar/common/data/ConsistencyLevel.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    /**
+     * Try to get the enum value based on the input name.
+     * @param name string literal of the enum name
+     * @return enum, or null if the input is null
+     */
+    public static @Nullable ConsistencyLevel fromString(@Nullable String name)
+    {
+        return name == null ? null : valueOf(name.toUpperCase(Locale.ROOT));
```
frankgh, line 56: can `name` be null in practicality?

src/main/java/org/apache/cassandra/sidecar/routes/restore/RestoreJobProgressHandler.java
```diff
@@ -0,0 +1,88 @@
...
+        super(metadataFetcher, executorPools, validator);
+        this.consistencyLevelChecker = consistencyLevelChecker;
+    }
+
+    @Override
+    protected RestoreJobProgressFetchPolicy extractParamsOrThrow(RoutingContext context)
+    {
+        List<String> fetchPolicyValues = context.queryParam(ApiEndpointsV1.FETCH_POLICY);
+        Preconditions.checkArgument(fetchPolicyValues.size() == 1, "Must define exactly one RestoreJobProgressFetchPolicy");
+        return RestoreJobProgressFetchPolicy.fromString(fetchPolicyValues.get(0));
```
frankgh, line 80: I think it makes sense to have a default policy defined when the user does not provide one. After all, query params are optional

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -413,24 +413,28 @@ public TableOperations tableOperations()
     @Override
     public void onAdd(Host host)
     {
+        LOGGER.info("Host added. host={}", host);
```
frankgh: are these needed at info level? maybe debug?

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/RestoreJobProgressRequestParams.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.cassandra.sidecar.common.data.RestoreJobProgressFetchPolicy;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Parameters to fetch the restore job progress
+ */
+public class RestoreJobProgressRequestParams
+{
+    public final String keyspace;
+    public final String table;
```
frankgh, line 30: QualifiedName?

frankgh (commented):
56/129 files so far. This is WIP review

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
yifan-c: fair

client-common/src/main/java/org/apache/cassandra/sidecar/common/data/ConsistencyLevel.java
yifan-c: it can. When restore ranges are not managed by sidecar, the consistency level field is absent, and the name is null.

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
yifan-c: I think it is useful to log the lifecycle events at info. They should now appear too often. 
They are not related to the patch. I am open to remove them from the patch. Otherwise, I will keep them as is.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/RestoreJobProgressRequestParams.java
yifan-c: It does not seem to be necessary, although it would be valid to wrap them in `QualifiedName`.

The values are used in `org.apache.cassandra.sidecar.common.request.RestoreJobProgressRequest` mainly. Other request object take string value for keyspace and table too.

src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
frankgh: My concern is from something I've seen in the past which is cql connection flapping. That's when log will become excessive . For that reason I'd rather have this at debug level and only enable it when needed

client-common/src/test/java/org/apache/cassandra/sidecar/common/request/CreateRestoreJobRequestPayloadTest.java
```diff
@@ -186,4 +197,19 @@ void testBuilder()
...
+    void testCreateLocalQuorumJobWithoutLocalDCFails()
+    {
+        RestoreJobSecrets secrets = RestoreJobSecretsGen.genRestoreJobSecrets();
+        assertThatThrownBy(() -> CreateRestoreJobRequestPayload.builder(secrets, System.currentTimeMillis())
+                                                               .consistencyLevel(ConsistencyLevel.LOCAL_QUORUM)
+                                                               .build())
+        .hasMessage("Must specify a non-empty localDatacenter for consistency level: LOCAL_QUORUM");
+
+        assertThatThrownBy(() -> CreateRestoreJobRequestPayload.builder(secrets, System.currentTimeMillis())
+                                                               .consistencyLevel(ConsistencyLevel.LOCAL_QUORUM, "")
```
frankgh: maybe try with LOCAL_ONE here instead?

src/main/java/org/apache/cassandra/sidecar/db/schema/RestoreRangesSchema.java
```diff
@@ -0,0 +1,143 @@
...
+                             "  slice_start_token," +
+                             "  slice_end_token," +
+                             "  status_by_replica" +
+                             ") VALUES (?, ?, ?, ?, ?, ?, ?, ?)", config);
+        }
+
+        // ALLOW FILTERING within the same partition should have minimum impact on read performance.
+        static String findAll(SchemaKeyspaceConfiguration config)
+        {
+            return withTable("SELECT job_id, bucket_id, slice_id, " +
```
frankgh: I think we are missing `slice_start_token` and `slice_end_token` from this query. I'm looking at `RestoreRange#from(Row)` and it seems that it will need those fields. Maybe a test is missing here as well to catch these issues?

src/main/java/org/apache/cassandra/sidecar/db/RestoreRangeDatabaseAccessor.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+        return range;
+    }
+
+    // todo: change to stream api and paginate
+    public List<RestoreRange> findAll(UUID jobId, short bucketId)
+    {
+        sidecarSchema.ensureInitialized();
+
+        BoundStatement statement = tableSchema.findAll()
```
frankgh, line 85: I think you touch on this in your TODO, but concern here is a timeout from the C* query if there are a lot of ranges for a given `jobId` and `bucketId`

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/data/Name.java
```diff
@@ -24,33 +24,26 @@
...
-     * parameters.
-     *
-     * @param unquotedName    the unquoted name
-     * @param maybeQuotedName the maybe quoted name
-     */
-    public Name(String unquotedName, String maybeQuotedName)
-    {
-        this.unquotedName = Objects.requireNonNull(unquotedName, "the unquoted name is required");
-        this.maybeQuotedName = Objects.requireNonNull(maybeQuotedName, "the _maybe_ quoted name is required");
+        Objects.requireNonNull(name, "name must not be null");
```
frankgh: can we rename the param to `maybeQuotedName`?

src/main/java/org/apache/cassandra/sidecar/cluster/ConsistencyVerifiers.java
```diff
@@ -0,0 +1,356 @@
...
+     */
+    public static class ForLocalOne extends BaseVerifier
+    {
+        public static final ForLocalOne INSTANCE = new ForLocalOne();
+
+        @Override
+        public Result verify(Set<String> succeeded, Set<String> failed, InstanceSetByDc all)
+        {
+            Preconditions.checkArgument(all.size() == 1, ONLY_LOCAL_DC_CHECK_ERR_MSG);
+            String dcName = all.keySet().iterator().next();
```
frankgh: ok,  I might be a bit confused here, but it seems we are only checking for the DC given by `InstanceSetByDC` , you make sure that the param only has a single DC?

src/main/java/org/apache/cassandra/sidecar/cluster/ConsistencyVerifiers.java
```diff
@@ -0,0 +1,356 @@
...
+            Set<String> localReplicas = all.get(dcName);
+            UnaryOperator<String> dcClassifier = s -> localReplicas.contains(s) ? dcName : UNKNOWN_DC;
```
frankgh: when would we have an unknown DC?

src/test/java/org/apache/cassandra/sidecar/cluster/ConsistencyVerifiersTest.java
```diff
@@ -0,0 +1,211 @@
...
+        Set<String> set = Collections.emptySet();
+        assertThatThrownBy(() -> verifier.verify(set, set, allReplicasByDc))
+        .isExactlyInstanceOf(IllegalArgumentException.class)
+        .hasMessageContaining("Parameter 'all' should contain the write replicas in local datacenter only");
+    }
+
+    @Test
+    void testClEachQuorum()
+    {
+        ConsistencyVerifier verifier = ConsistencyVerifiers.forConsistencyLevel(ConsistencyLevel.EACH_QUORUM);
```
frankgh, line 157: hmm, are we really testing multiple DCs here. I'm not understanding this test maybe?

src/main/java/org/apache/cassandra/sidecar/cluster/ConsistencyVerifiers.java
```diff
@@ -0,0 +1,356 @@
...
+            {
+                return Result.SATISFIED;
+            }
+
+            if (!failed.isEmpty())
+            {
+                return Result.FAILED;
+            }
+
+            return Result.PENDING;
```
frankgh, line 300: Ideally the failed and succeeded sets are mutually exclusive, but I"m thinking of the case where something is wrong from the caller, we can return an accurate Result by doing this:

```suggestion
            Result result = Result.PENDING;
            if (succeeded.size() == sum(all))
            {
                result = Result.SATISFIED;
            }

            if (!failed.isEmpty())
            {
                result = Result.FAILED;
            }

            return result;
```

src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobConsistencyLevelChecker.java
```diff
@@ -0,0 +1,215 @@
...
+    {
+        this.ringTopologyRefresher = ringTopologyRefresher;
+        this.rangeDatabaseAccessor = rangeDatabaseAccessor;
+        this.taskExecutorPool = executorPools.internal();
+    }
+
+    public Future<RestoreJobProgress> check(RestoreJob restoreJob, RestoreJobProgressFetchPolicy fetchPolicy)
+    {
+        RestoreJobProgressCollector collector = RestoreJobProgressCollectors.create(restoreJob, fetchPolicy);
+        RestoreRangeStatus successCriteria = restoreJob.expectedNextRangeStatus();
```
frankgh, line 85: it looks like we are restricting making calls to the job status endpoint when the status is ready see `RestoreJob#expectedNextRangeStatus`. The first thing we do in the method is check that status is ready. Now, assume we have a succeeded or failed job, we won't be able to call this endpoint

src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobConsistencyLevelChecker.java
```diff
@@ -0,0 +1,215 @@
...
+
+    private Future<RestoreJobProgress> findRangesAndConclude(RestoreJob restoreJob,
+                                                             RestoreRangeStatus successCriteria,
+                                                             TokenRangeReplicasResponse topology,
+                                                             ConsistencyVerifier verifier,
+                                                             RestoreJobProgressCollector collector)
+    {
+        return taskExecutorPool
+               .executeBlocking(() -> {
+                   short bucketId = 0; // todo: replace with looping through all bucketIds
```
frankgh, line 99: will there be a follow up PR or were you planning on addressing it in this PR?

frankgh (commented):
Still reviewing (80/126). I've added a few more comments

src/main/java/org/apache/cassandra/sidecar/db/RestoreRangeDatabaseAccessor.java
yifan-c: The data amount of the expected/reasonable job should be fine. Let's say the job restores 1 TB data, and each slice has 5 GB, and all in a single bucket. There are in total 205 slices. When there are token movement, the reasonable max number of ranges is 410. And likely the number of ranges is close to the 205 end. Each range object should take roughly (or less than) 1KB when reading from C*.

src/test/java/org/apache/cassandra/sidecar/cluster/ConsistencyVerifiersTest.java
yifan-c: i-1, i-2, i-3 are in dc-1.
i-4, i-5, i-6 are in dc-2.

src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobConsistencyLevelChecker.java
yifan-c: I think 0 is good for now. Will address in a follow up PR

src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobManagerGroup.java
```diff
@@ -54,39 +52,32 @@ public RestoreJobManagerGroup(SidecarConfiguration configuration,
...
     {
         this.restoreJobConfig = configuration.restoreJobConfiguration();
         this.restoreProcessor = restoreProcessor;
-        this.jobDiscoverer = jobDiscoverer;
         this.executorPools = executorPools;
         initializeManagers(instancesConfig);
-        jobDiscoverer.signalRefresh();
         periodicTaskExecutor.schedule(jobDiscoverer);
         periodicTaskExecutor.schedule(restoreProcessor);
+        periodicTaskExecutor.schedule(ringTopologyRefresher);
```
frankgh, line 64: It feels like the ring topology refresher is not necessarily tied to the restore feature. I'd preferably schedule it somewhere else where we don't couple it with this feature.

src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobManager.java
```diff
@@ -240,4 +245,10 @@ boolean isObsoleteRestoreJobDir(Path path)
         long gapInMillis = TimeUnit.DAYS.toMillis(restoreJobConfig.jobDiscoveryRecencyDays());
         return delta > gapInMillis;
     }
+
+    @VisibleForTesting
+    RestoreJobProgressTracker progressTrackerUnsafe(RestoreJob restoreJob)
```
frankgh, line 250: why unsafe here?

src/main/java/org/apache/cassandra/sidecar/utils/InstanceMetadataFetcher.java
```diff
@@ -99,11 +103,32 @@ public CassandraAdapterDelegate delegate(int instanceId)
...
+     */
+    public InstanceMetadata anyInstance()
+    {
+        ensureInstancesConfigured();
+        List<InstanceMetadata> instances = instancesConfig.instances();
+        if (instances.size() == 1)
+        {
+            return instances.get(0);
+        }
+        int randomPick = RANDOM.nextInt(instances.size());
```
frankgh: can we use `ThreadLocalRandom.current().nextInt()` here instead?

src/main/java/org/apache/cassandra/sidecar/utils/InstanceMetadataFetcher.java
```diff
@@ -99,11 +103,32 @@ public CassandraAdapterDelegate delegate(int instanceId)
...
+    {
+        ensureInstancesConfigured();
+        return instancesConfig.instances().get(0);
+    }
+
+    /**
+     * @return any instance from the list of configured instances
+     * @throws IllegalStateException when there are no configured instances
+     */
+    public InstanceMetadata anyInstance()
```
frankgh, line 113: looking at the caller of this method , it seems what we really need is the "first available instance". Having a random instance being returned here could prevent the topology refresher to make any progress for a prolonged period of time, if say all instances except one are available and you keep getting an unavailable instance on each iteration. Can we instead have a new API method that returns the firstAvailableInstance instead?

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -96,17 +97,17 @@ public final class ApiEndpointsV1
     public static final String KEYSPACE_TOKEN_MAPPING_ROUTE = API_V1 + PER_KEYSPACE + "/token-range-replicas";
 
     // Blob Transport Extension
-    public static final String JOB_ID_PATH_PARAM = ":jobId";
     public static final String RESTORE_JOBS = "/restore-jobs";
     public static final String SLICES = "/slices";
     public static final String ABORT = "/abort";
+    public static final String PROGRESS = "/progress";
+    public static final String FETCH_POLICY = "fetch-policy";
```
frankgh: should we name this `FETCH_POLICY_QUERY_PARAM` ?

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CreateRestoreJobRequestPayload.java
```diff
@@ -94,17 +102,19 @@ public CreateRestoreJobRequestPayload(@JsonProperty(JOB_ID) UUID jobId,
                              ? SSTableImportOptions.defaults()
                              : importOptions;
         this.expireAtInMillis = expireAtInMillis;
-        this.consistencyLevel = consistencyLevel;
+        this.consistencyLevel = ConsistencyLevel.fromString(consistencyLevel);
+        this.localDatacenter = localDatacenter;
```
frankgh: should we validate here that if (consistencyLevel.isLocal) then localDC cannot be null?

client-common/src/test/java/org/apache/cassandra/sidecar/common/request/CreateRestoreJobRequestPayloadTest.java
```diff
@@ -49,10 +56,14 @@ void testSerDeser() throws JsonProcessingException
         Date date = Date.from(Instant.ofEpochMilli(time));
         CreateRestoreJobRequestPayload req = CreateRestoreJobRequestPayload.builder(secrets, time)
                                                                            .jobId(UUID.fromString(id))
-                                                                           .consistencyLevel("QUORUM")
+                                                                           .consistencyLevel(ConsistencyLevel.QUORUM)
                                                                            .jobAgent("agent")
                                                                            .build();
         String json = MAPPER.writeValueAsString(req);
+        assertThat(json).describedAs("Null value fields should be excluded").doesNotContain(JOB_LOCAL_DATA_CENTER)
```
frankgh, line 61: so here we just assert that the string is contained in the serialized payload? is that desired?

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/utils/StringUtils.java
```diff
@@ -0,0 +1,45 @@
...
+    public static boolean isNullOrEmpty(@Nullable String string)
+    {
+        return string == null || string.isEmpty();
+    }
+
+    /**
+     * @param string string value to test
+     * @return true only when the string is not null and its length is not 0; false otherwise
+     */
+    public static boolean notEmpty(@Nullable String string)
```
frankgh: why not use commons lang3 StringUtils ? This method is only used from the server which has a dependency to lang3

src/main/java/org/apache/cassandra/sidecar/db/schema/RestoreRangesSchema.java
```diff
@@ -0,0 +1,143 @@
...
+    {
+        insert = prepare(insert, session, CqlLiterals.insert(keyspaceConfig));
+        findAll = prepare(findAll, session, CqlLiterals.findAll(keyspaceConfig));
+        update = prepare(update, session, CqlLiterals.update(keyspaceConfig));
+    }
+
+    @Override
+    protected boolean exists(@NotNull Metadata metadata)
+    {
+        return false;
```
frankgh: is it okay to return false here?

frankgh (commented):
Some additional comments

src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobManagerGroup.java
yifan-c: It is not 0 cost to refresh the topology. My take is that until we find more use cases, let's keep it scoped in the restore feature. At the end, the refreshing requires "keyspace", which is provided in the restore jobs.

src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobManager.java
yifan-c: The method can changes the `jobs` states directly without going through the intended path, i.e. `trySubmit`. It is used for testing. Having the "unsafe" in the method name, in addition to the annotation, to make it more outstanding and not invoke it in production code.

src/main/java/org/apache/cassandra/sidecar/utils/InstanceMetadataFetcher.java
yifan-c: The idea of picking any instance is to avoid making a node hotter than others. Ideally, all callers of `firstInstance` should be replaced by `anyInstance`. 

It is good of calling out the availability. Depending on the callsites, the requirement can differ. We can have an overload method that takes an optional filter parameter to remove the unwanted instances. For example, `public InstanceMetadata anyInstance(Predicate<InstanceMetadata> filter)`

frankgh (approved):
+1 no further comments for now

PR #124, opened 2024-05-24
CASSANDRASC-133: Allow optional reason to abort restore jobs

yifan-c:


client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -517,10 +518,11 @@ public CompletableFuture<Void> updateRestoreJob(String keyspace,
      * {@inheritDoc}
      */
     @Override
-    public CompletableFuture<Void> abortRestoreJob(String keyspace, String table, UUID jobId)
+    public CompletableFuture<Void> abortRestoreJob(String keyspace, String table, UUID jobId,
```
frankgh, line 521: I feel we should preserve the existing signature and add another method that takes payload. For the existing signature, the payload would be null.

src/main/java/org/apache/cassandra/sidecar/routes/restore/AbortRestoreJobHandler.java
```diff
@@ -69,33 +73,47 @@ protected void handleInternal(RoutingContext context,
...
+
+        try
+        {
+            return Json.decodeValue(bodyString, AbortRestoreJobRequestPayload.class);
+        }
+        catch (Exception anyException)
+        {
+            if (bodyString != null)
+            {
+                logger.warn("Failed  to deserialize json string into AbortRestoreJobRequestPayload", anyException);
```
frankgh: Remove extra space here
```suggestion
                logger.warn("Failed to deserialize json string into AbortRestoreJobRequestPayload", anyException);
```

src/main/java/org/apache/cassandra/sidecar/routes/restore/AbortRestoreJobHandler.java
```diff
@@ -69,33 +73,47 @@ protected void handleInternal(RoutingContext context,
...
+    protected AbortRestoreJobRequestPayload extractParamsOrThrow(RoutingContext context)
     {
-        return context.pathParam("jobId");
+        String bodyString = context.body().asString(); // nullable
+
+        try
+        {
+            return Json.decodeValue(bodyString, AbortRestoreJobRequestPayload.class);
+        }
+        catch (Exception anyException)
```
frankgh: maybe just call it exception? I feel this is introducing a different style that differs from the existing codebase
```suggestion
        catch (Exception exception)
```

frankgh (commented):
Looks good, I've added some minor comments

frankgh (approved):
+1 Looks good

PR #125, opened 2024-06-03
CASSANDRASC-134: Detect out of range data and cleanup using nodetool

yifan-c:
The patch adds a step to check the data ownership before importing sstable. When fully out of range sstables are found, the sstables are removed before importing. When partially out of range sstables are found, running nodetool cleanup is requested on job completion, including both success and failure cases.

Patch by Yifan Cai; Reviewed by TBD for CASSANDRASC-134


server-common/src/main/java/org/apache/cassandra/sidecar/common/server/cluster/locator/Token.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public BigInteger toBigInteger()
+    {
+        return value;
+    }
+
+    public Token increment()
+    {
+        return new Token(value.add(BigInteger.ONE));
```
bbotella, line 86: Do we need to create a new object instead of just incrementing `value` and returning this?

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/cluster/locator/TokenRange.java
```diff
@@ -75,18 +75,83 @@ public TokenRange(long start, long end)
...
+     * @return the first token enclosed in the range
+     */
+    @Nullable
+    public Token firstToken()
+    {
+        if (range.isEmpty())
+        {
+            return null;
+        }
+        return range.lowerEndpoint().increment();
```
bbotella: Curious. Why returning the increment here? Shouldn't it return `range.lowerEndpoint()`?

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/cluster/locator/TokenRange.java
```diff
@@ -75,18 +75,83 @@ public TokenRange(long start, long end)
 
     public TokenRange(BigInteger start, BigInteger end)
     {
-        this.start = start;
-        this.end = end;
+        this(Token.from(start), Token.from(end));
```
bbotella, line 93: Should we validate that end is bigger than start?

server-common/src/test/java/org/apache/cassandra/sidecar/common/server/cluster/locator/TokenTest.java
```diff
@@ -0,0 +1,54 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+class TokenTest
+{
+    @Test
+    void testCreateToken()
+    {
+        Token t1 = Token.from(1);
+        Token t2 = Token.from(BigInteger.ONE);
+        Token t3 = Token.from("1");
```
bbotella, line 32:
```
Token t4 = Token.from(1L);
```
?

src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -264,7 +264,7 @@ private void findSlicesOfRangeAndSubmit(InstanceMetadata instance, RestoreJob re
     {
         short bucketId = 0; // TODO: update the implementation to pick proper bucketId
         restoreSliceDatabaseAccessor
-        .selectByJobByBucketByTokenRange(restoreJob.jobId, bucketId, range.start, range.end)
+        .selectByJobByBucketByTokenRange(restoreJob.jobId, bucketId, range.start().toBigInteger(), range.end().toBigInteger())
```
bbotella: Should we make `selectByJobByBucketByTokenRange` just accept a `TokenRange range` instead of a `range.start()` and a `range.end()`?

src/main/java/org/apache/cassandra/sidecar/restore/RestoreSliceTask.java
```diff
@@ -430,22 +450,79 @@ private Future<File> validateFiles(File directory)
...
+            if (!hasOverlap)
+            {
+                // remove the entry from manifest
+                it.remove();
+                // delete the files
+                for (String fileName : entry.componentsChecksum().keySet())
+                {
+                    Path path = directory.toPath().resolve(fileName);
+                    Files.deleteIfExists(path);
+                }
```
bbotella, line 507: Nit: Maybe extracting this to a private method? It may help with readability (the method is already big)

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -201,4 +202,12 @@ public List<String> dataFileLocations()
         }
         return dataFileLocations;
     }
+
+    @Override
+    public void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String table, int concurrency)
+    throws IOException, ExecutionException, InterruptedException
+    {
```
frankgh, line 210: For consistency with the behavior in other methods in this class and generally good practice 
```suggestion
    {
        requireNonNull(keyspace, "keyspace must be non-null");
        requireNonNull(table, "table must be non-null");
```

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/StorageOperations.java
```diff
@@ -76,4 +78,27 @@ TokenRangeReplicasResponse tokenRangeReplicas(@NotNull Name keyspace,
...
+    void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String table, int concurrency)
+    throws IOException, ExecutionException, InterruptedException;
+
+    /**
+     * Similar to {@link #outOfRangeDataCleanup(String, String, int)}, but use 1 for concurrency
+     */
+    default void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String table)
+    throws IOException, ExecutionException, InterruptedException
+    {
+        outOfRangeDataCleanup(keyspace, table, 1);
```
frankgh, line 102: the default for the `nodetool cleanup` feature is 2 , should we stick to that default?

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/cluster/locator/Token.java
```diff
@@ -0,0 +1,108 @@
...
+        this.value = value;
+        this.hashCode = value.hashCode();
+    }
+
+    public BigInteger toBigInteger()
+    {
+        return value;
+    }
+
+    public Token increment()
```
frankgh, line 84: maybe document the contract here, to make sure consumers understand this offer immutability and they will need to track the reference if they desire to use it

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/cluster/locator/TokenRange.java
frankgh: This should already be taking care by `Range.openClosed`

server-common/src/test/java/org/apache/cassandra/sidecar/common/server/cluster/locator/TokenTest.java
frankgh: t1 is using the long constructor

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/cluster/locator/Token.java
```diff
@@ -0,0 +1,108 @@
...
+
+    public static Token from(long value)
+    {
+        return new Token(BigInteger.valueOf(value));
+    }
+
+    private Token(BigInteger value)
+    {
+        this.value = value;
+        this.hashCode = value.hashCode();
```
frankgh: wondering if this early optimization is needed. If we are always comparing tokens this makes sense.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/cluster/locator/Token.java
yifan-c: `Token` is used with `Range`. If the token is made mutable, it can change the range unnoticed, causing unexpected behavior. Imagine a range (1, 10] changes to (2, 10].

src/main/java/org/apache/cassandra/sidecar/restore/RestoreProcessor.java
```diff
@@ -78,6 +81,7 @@ public RestoreProcessor(ExecutorPools executorPools,
                             SSTableImporter importer,
                             RestoreSliceDatabaseAccessor sliceDatabaseAccessor,
                             RestoreJobUtil restoreJobUtil,
+                            CachedLocalTokenRanges localTokenRangesProvider,
```
frankgh, line 84: does this need to be the concrete implementation? Ideally we should just use the interface here: `org.apache.cassandra.sidecar.locator.LocalTokenRangesProvider`

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/cluster/locator/TokenRange.java
yifan-c: The token range is always open-closed. The lower end is not included. So the first token is the lower end + 1.

frankgh (commented):
WIP review

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/cluster/locator/TokenRange.java
frankgh: we create an object every time, should we keep a reference?

server-common/src/test/java/org/apache/cassandra/sidecar/common/server/cluster/locator/TokenRangeTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+        TokenRange emptyRange = new TokenRange(1, 1);
+        assertThat(emptyRange.firstToken()).isNull();
+    }
+
+    @Test
+    void testCreateFromJavaDriverTokenRange()
+    {
+        com.datastax.driver.core.TokenRange ordinaryRange = mockRange(1L, 100L);
+        when(ordinaryRange.isWrappedAround()).thenReturn(false);
```
frankgh, line 69: I think isWrappedAround has an issue when the end range is the MIN_TOKEN: i.e. Murmur3, this would fail:

```
        com.datastax.driver.core.TokenRange range = mockRange(10L, Long.MIN_VALUE);
        when(range.isWrappedAround()).thenReturn(false);
         ranges = TokenRange.from(range);
        assertThat(ranges).hasSize(1)
                          .isEqualTo(Collections.singletonList(Range.openClosed(BigInteger.valueOf(10), Long.MAX_VALUE)));
```

src/main/java/org/apache/cassandra/sidecar/restore/RestoreProcessor.java
yifan-c: It is not required, but the current form of code is succinct, as `CachedLocalTokenRanges` is annotated with `@Singleton`. I would keep it for now, until it gets complex (unlikely).

frankgh: this wouldn't allow us to swap out to a new implementation and  it would strictly tie it to the cached implementation.  That's an anti pattern for dependency injection frameworks like guice

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/cluster/locator/Token.java
yifan-c: reverted. I do not think it is used.

src/main/java/org/apache/cassandra/sidecar/restore/RestoreSliceTracker.java
```diff
@@ -102,22 +108,56 @@ public void fail(RestoreJobFatalException exception)
...
+     */
+    private void runOnCompletion()
+    {
+        if (cleanupOutOfRangeRequested)
+        {
+            CassandraAdapterDelegate delegate = instanceMetadata.delegate();
+            StorageOperations operations = delegate == null ? null : delegate.storageOperations();
+            if (operations == null)
+            {
+                LOGGER.warn("Out of range data cleanup for the restore job is requested. It failed to start the operation. jobId={}", restoreJob.jobId);
```
frankgh, line 148: just wondering if we should have a retry here ?

src/test/integration/org/apache/cassandra/sidecar/common/JmxClientIntegrationTest.java
```diff
@@ -37,46 +40,77 @@ public class JmxClientIntegrationTest
...
+    {
+        String releaseVersion = jmxClient.proxy(SSProxy.class, SS_OBJ_NAME)
+                                         .getReleaseVersion();
+        assertThat(releaseVersion).startsWith(majorVersion);
+    }
+
+    // a test to ensure the jmx client can invoke the MBean method
+    private void testTableCleanup(JmxClient jmxClient, UpgradeableCluster cluster)
+    {
+        cluster.schemaChange("CREATE KEYSPACE jmx_client_test WITH REPLICATION = {'class' : 'SimpleStrategy', 'replication_factor' : 1}");
```
frankgh, line 95: can you really do a cleanup with a single node? although I'm not sure we want to test that the out of range token is executed, but rather than the cleanup interaction with JMX occurs

frankgh (approved):
Discussed offline about the remaining items and we agree that we don't need to address them yet. +1 looks good to me

bbotella (approved):

PR #126, opened 2024-06-11
CASSANDRASC-135: Added additional JVM options to increase max open FD limit for MacOS

rwelgosh:
Patch by Raymond Welgosh; Reviewed by Francisco Guerrero, Yifan Cai for CASSANDRASC-135


.java-version
```diff
@@ -0,0 +1 @@
+graalvm64-11.0.20
```
yifan-c: we need to exclude this file and add a new entry to gitignore. 
The file is produced by `jenv`

gradlew
```diff
@@ -105,8 +105,10 @@ if [ "$cygwin" = "false" -a "$darwin" = "false" -a "$nonstop" = "false" ] ; then
 fi
 
 # For Darwin, add options to specify how the application appears in the dock
+# In addition, we want to disable the MaxFDLimit for MacOS, which is set to a lower limit than the actual system maximum.
```
frankgh:
```suggestion
# In addition, we want to increase the file descriptor limit to the MaxFDLimit in MacOS which, by default, is set to a lower limit than the actual system's maximum.
```

vertx-client-shaded/build.gradle
```diff
@@ -83,6 +83,8 @@ shadowJar {
     }
 }
 
+test.dependsOn shadowJar
```
bbotella: I'm curious. Why does it need the shadow jar now but didn't before?

src/test/java/org/apache/cassandra/sidecar/UlimitTest.java
```diff
@@ -0,0 +1,44 @@
...
+    void ensureUlimitMaxFilesIsNotTooLow() throws Exception
+    {
+        Process p = new ProcessBuilder("ulimit", "-n").start();
+        p.waitFor();
+        InputStreamReader inputStreamReader = new InputStreamReader(p.getInputStream());
+        BufferedReader reader = new BufferedReader(inputStreamReader);
+        String line = reader.readLine();
+        long maxFD = Long.parseLong(line);
+        System.out.println(maxFD);
+        assertThat(maxFD).isGreaterThan(10240);
```
frankgh, line 51: will this always be true? also, can we scope this test to macOS ?

vertx-client-shaded/build.gradle
frankgh: thanks for fixing this dependency! I have a small NIT:
```suggestion
tasks.named('test') {
    dependsOn 'shadowJar'
}
```

frankgh (commented):
Looks good. I have some minor small comments

gradlew
yifan-c: To clarify, the JVM option `-XX:-MaxFDLimit` is to _disable_ the low `MaxFDLimit` set in JDK. By disabling, it can pick up the system FD limit, which is larger and proper. 

The suggested sentence looks incorrect.

vertx-client-shaded/build.gradle
yifan-c:
```
LibraryTest > Ensures that the shading is correct for the vertx-client-shaded project FAILED
```

It is to fix the flaky test, `LibraryTest`. The test requires shadow jar being built. However, when the test runs, the jar is empty. 

That being said, test already depends on shadowJar. It needs a different fix.

gradlew
frankgh: I noticed there are places where we enable the flag, and other places where we disable the flag. Is that intentional?

Reading this documentation: https://docs.oracle.com/cd/E19683-01/806-7930/filestructure-11/index.html it looks like setting the flag i.e. doing `-XX:+MaxFDLimit` should `Bump the number of file descriptors to max`. Isn't this what we are trying to accomplish here?

yifan-c: No. The max FD constant for macOS is low. It is `OPEN_MAX == 10240` https://github.com/openjdk/jdk/blob/dfd3da3f52480f68f653beb1e720691f8232ace7/src/hotspot/os/bsd/os_bsd.cpp#L1997-L2002

You want to _disable_ the max FD, so that it can pick up the system's limit.

client-common/build.gradle
```diff
@@ -41,6 +44,9 @@ repositories {
 
 test {
     useJUnitPlatform()
+    if (Os.isFamily(Os.FAMILY_MAC)) {
+        jvmArgs "-XX:+MaxFDLimit"
```
yifan-c: disable. use `'-XX:-MaxFDLimit` instead.

src/test/java/org/apache/cassandra/sidecar/UlimitTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test to ensure that the max file descriptors is large enough to build Cassandra Sidecar
+ */
+@Disabled("ulimit does not exist at the CI runtime")
+public class UlimitTest
```
frankgh, line 35: can we instead try annotating the test with `@EnabledOnOs({OS.MAC})`?

yifan-c: Some annotations in junit5 are amazing!

frankgh (approved):
+1 Looks good to me

PR #127, opened 2024-06-27
CASSANDRASC-136: Add delay between reporting of the same slow task

yifan-c:
Patch by Yifan Cai; Reviewed by Francisco Guerrero for CASSANDRASC-136

### DIFF
--- a/src/main/dist/conf/sidecar.yaml
+++ b/src/main/dist/conf/sidecar.yaml
@@ -172,11 +172,13 @@ cassandra_input_validation:
   allowed_chars_for_restricted_component_name: "[a-zA-Z0-9_-]+(.db|TOC.txt)"
 
 blob_restore:
-  job_discovery_active_loop_delay_millis: 5
-  job_discovery_idle_loop_delay_millis: 10
+  job_discovery_active_loop_delay_millis: 300000
+  job_discovery_idle_loop_delay_millis: 600000
   job_discovery_recency_days: 5
   slice_process_max_concurrency: 20
   restore_job_tables_ttl_seconds: 7776000
+  slow_task_threshold_seconds: 600
+  slow_task_report_delay_seconds: 60
 
 s3_client:
   concurrency: 4
--- a/src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
+++ b/src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
@@ -37,9 +37,9 @@ public class RestoreJobConfigurationImpl implements RestoreJobConfiguration
     private static final int DEFAULT_PROCESS_MAX_CONCURRENCY = 20; // process at most 20 slices concurrently
     private static final long DEFAULT_RESTORE_JOB_TABLES_TTL_SECONDS = TimeUnit.DAYS.toSeconds(90);
     // A restore task is considered slow if it has been in the "active" list for 10 minutes.
-    private static final long DEFAULT_RESTORE_JOB_SLOW_TASK_THRESHOLD_SECONDS = 600;
+    private static final long DEFAULT_RESTORE_JOB_SLOW_TASK_THRESHOLD_SECONDS = TimeUnit.MINUTES.toSeconds(10);
     // report once a minute
-    private static final long DEFAULT_RESTORE_JOB_SLOW_TASK_REPORT_DELAY_SECONDS = 60;
+    private static final long DEFAULT_RESTORE_JOB_SLOW_TASK_REPORT_DELAY_SECONDS = TimeUnit.MINUTES.toSeconds(1);
 
     @JsonProperty(value = "job_discovery_active_loop_delay_millis")
     protected final long jobDiscoveryActiveLoopDelayMillis;
--- a/src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
+++ b/src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
@@ -31,15 +31,15 @@ public class RestoreJobConfigurationImpl implements RestoreJobConfiguration
 {
     private static final long MIN_RESTORE_JOB_TABLES_TTL_SECONDS = TimeUnit.DAYS.toSeconds(14);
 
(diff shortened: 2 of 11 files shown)


src/main/dist/conf/sidecar.yaml
```diff
@@ -172,11 +172,13 @@ cassandra_input_validation:
   allowed_chars_for_restricted_component_name: "[a-zA-Z0-9_-]+(.db|TOC.txt)"
 
 blob_restore:
-  job_discovery_active_loop_delay_millis: 5
-  job_discovery_idle_loop_delay_millis: 10
+  job_discovery_active_loop_delay_millis: 300000
```
yifan-c, line 175: The setting was wrong. It meant to have 5 minutes.

src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
```diff
@@ -31,15 +31,15 @@ public class RestoreJobConfigurationImpl implements RestoreJobConfiguration
...
-    "restore_job_long_running_threshold_seconds";
-    // A restore job handler is considered long-running if it has been in the "active" list for 10 minutes.
-    private static final long DEFAULT_RESTORE_JOB_LONG_RUNNING_HANDLER_THRESHOLD_SECONDS = 600;
+    private static final long DEFAULT_JOB_DISCOVERY_ACTIVE_LOOP_DELAY_MILLIS = TimeUnit.MINUTES.toMillis(5);
+    private static final long DEFAULT_JOB_DISCOVERY_IDLE_LOOP_DELAY_MILLIS = TimeUnit.MINUTES.toMillis(10);
+    private static final int DEFAULT_JOB_DISCOVERY_RECENCY_DAYS = 5;
+    private static final int DEFAULT_PROCESS_MAX_CONCURRENCY = 20; // process at most 20 slices concurrently
+    private static final long DEFAULT_RESTORE_JOB_TABLES_TTL_SECONDS = TimeUnit.DAYS.toSeconds(90);
+    // A restore task is considered slow if it has been in the "active" list for 10 minutes.
+    private static final long DEFAULT_RESTORE_JOB_SLOW_TASK_THRESHOLD_SECONDS = 600;
```
frankgh: NIT, use `TimeUnit` similar to what we do for other properties

src/main/java/org/apache/cassandra/sidecar/restore/RestoreProcessor.java
```diff
@@ -248,6 +223,44 @@ public void close()
         sliceQueue.close();
     }
 
+    private void checkForLongRunningTasks()
```
yifan-c, line 227: moved the private method after all public ones.

src/main/java/org/apache/cassandra/sidecar/exceptions/ThrowableUtils.java
```diff
@@ -101,12 +102,16 @@ public static Throwable getCause(Throwable throwable, Predicate<Throwable> predi
...
-                // This way we are sure that all exceptions/causes are visited at least once.
-                stop = cause;
+                // once stop is set; updating fast tracer is no longer required
+                if (cause == fastTracer)
+                {
+                    // Mark the position to stop, and continue tracing the cause up until hitting stop the next time.
+                    // This way we are sure that all exceptions/causes are visited at least once.
+                    stop = cause;
+                }
+                fastTracer = getCause(fastTracer, 2);
```
frankgh: it seems we need an else here
```suggestion
                }
                else
                {
                    fastTracer = getCause(fastTracer, 2);
                 }
```

src/main/java/org/apache/cassandra/sidecar/exceptions/ThrowableUtils.java
```diff
@@ -101,12 +102,16 @@ public static Throwable getCause(Throwable throwable, Predicate<Throwable> predi
                 return cause;
             }
 
-            fastTracer = getCause(fastTracer, 2);
-            if (cause == fastTracer && stop == null)
+            if (stop == null)
```
frankgh, line 105: another optimization 
```suggestion
            if (fastTracer != null && predicate.test(fastTracer))
            {
                return fastTracer;
            }

            if (stop == null)
```

src/main/java/org/apache/cassandra/sidecar/restore/RestoreProcessor.java
```diff
@@ -151,7 +152,8 @@ public void execute(Promise<Void> promise)
                                                          restoreJobUtil,
                                                          localTokenRangesProvider,
                                                          metrics);
-            activeTasks.add(task);
+
+            activeTasks.put(task, slowTaskThresholdInSeconds);
             pool.executeBlocking(task, false) // unordered; run in parallel
             .onSuccess(restoreSlice -> {
```
frankgh, line 159: do we need to remove the task from `activeTasks` when a failure occurs?

src/main/java/org/apache/cassandra/sidecar/exceptions/ThrowableUtils.java
yifan-c: It is probably not. Because it adds another test of the predicate.

src/main/java/org/apache/cassandra/sidecar/restore/RestoreProcessor.java
yifan-c: task is removed in the `onComplete` block, regardless of success or failure.

src/main/java/org/apache/cassandra/sidecar/restore/RestoreProcessor.java
```diff
@@ -66,8 +65,10 @@ public class RestoreProcessor implements PeriodicTask
     private final double requiredUsableSpacePercentage; // value range: [0.0, 1.0)
     private final RestoreSliceDatabaseAccessor sliceDatabaseAccessor;
     private final RestoreJobUtil restoreJobUtil;
-    private final Set<RestoreSliceHandler> activeTasks = ConcurrentHashMap.newKeySet();
-    private final long longRunningHandlerThresholdInSeconds;
+    // mapping of task to the time when it should be reported as 'slow' if it is still active
+    private final Map<RestoreSliceHandler, Long> activeTasks = new HashMap<>();
```
frankgh: concurrent map here?

frankgh (approved):
+1 looks good

PR #130, opened 2024-08-06
Fixes updating traffic shaping options throws IllegalStateException

frankgh:
Patch by Francisco Guerrero; Reviewed by Yifan Cai for CASSANDRASC-140

### DIFF
--- a/src/test/java/org/apache/cassandra/sidecar/server/ServerTest.java
+++ b/src/test/java/org/apache/cassandra/sidecar/server/ServerTest.java
@@ -25,7 +25,6 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.Timeout;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.api.io.TempDir;
 import org.slf4j.Logger;
--- a/src/test/java/org/apache/cassandra/sidecar/server/ServerTest.java
+++ b/src/test/java/org/apache/cassandra/sidecar/server/ServerTest.java
@@ -22,6 +22,7 @@ import java.nio.file.Path;
 import java.util.concurrent.TimeUnit;
 
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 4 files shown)


src/test/resources/config/sidecar_single_instance_default_port.yaml
```diff
@@ -0,0 +1,106 @@
...
+  staging_dir: /ccm/test/node1/sstable-staging
+  jmx_host: 127.0.0.1
+  jmx_port: 7199
+  jmx_role: controlRole
+  jmx_role_password: controlPassword
+  jmx_ssl_enabled: true
+
+sidecar:
+  host: 0.0.0.0
+  port: 9043
```
frankgh, line 21: configuring the port to be non-zero, surfaces the issue because the logic in vert.x treats the zero port configuration differently

sarankk (approved):
+1 Looks good to me

arjunashok (approved):
LGTM, +1

src/test/java/org/apache/cassandra/sidecar/server/ServerTest.java
```diff
@@ -181,6 +194,25 @@ void updatingTrafficShapingOptions(VertxTestContext context)
...
+        {
+            server.start().toCompletionStage().toCompletableFuture().get(30, TimeUnit.SECONDS);
+            TrafficShapingOptions update = new TrafficShapingOptions()
+                                           .setOutboundGlobalBandwidth(100 * 1024 * 1024);
+            server.updateTrafficShapingOptions(update);
+        }
+        catch (Exception exception)
+        {
+            Assertions.fail(exception);
+        }
```
yifan-c: I guess the catch is not necessary. Is it added because IDE complains no assertion is performed? 
If so, how about this?
```java
        assertThatNoException().isThrownBy(() -> {
            server.start().toCompletionStage().toCompletableFuture().get(30, TimeUnit.SECONDS);
            TrafficShapingOptions update = new TrafficShapingOptions()
                                           .setOutboundGlobalBandwidth(100 * 1024 * 1024);
            server.updateTrafficShapingOptions(update);
        });
```

yifan-c (approved):
One nit. lgtm

src/test/java/org/apache/cassandra/sidecar/server/ServerTest.java
frankgh: Without the fix, the exception was thrown and I would fail the test. I will update the test. I was running `@RepeatedTest` so it was necessary to fail the test so the server was properly closed during teardown

PR #132, opened 2024-08-15
CASSANDRASC-142: Improve S3 download throttling with range-GetObject

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSANDRASC-142


src/main/java/org/apache/cassandra/sidecar/restore/HttpRangesIterator.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        if (offset < totalBytes)
+        {
+            long end = Math.min(offset + rangeSize, totalBytes) - 1;
+            current = HttpRange.of(offset, end);
+            offset = end + 1;
+        }
+
+        return current != null;
```
sarankk: Nit: Since next() is called in only 1 place at a time, how about this?

```
 @Override
 public boolean hasNext()
 {
return offset < totalBytes;
}

@Override
public HttpRange next()
 {
       if (!hasNext())
       {
           throw new NoSuchElementException("No more HttpRanges");
       }
      long end = Math.min(offset + rangeSize, totalBytes) - 1;
      HttpRange range = HttpRange.of(offset, end);
      offset = end + 1;
      return range;
}
```

src/test/containerTest/org/apache/cassandra/sidecar/restore/StorageClientTest.java
```diff
@@ -23,31 +23,41 @@
...
+import java.time.Duration;
+import java.util.Arrays;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 
 import com.google.common.util.concurrent.SidecarRateLimiter;
+import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.AfterEach;
```
sarankk: Nit: Unused

src/test/containerTest/org/apache/cassandra/sidecar/restore/StorageClientTest.java
```diff
@@ -23,31 +23,41 @@
...
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
```
sarankk: Nit: Unused

src/test/containerTest/org/apache/cassandra/sidecar/restore/StorageClientTest.java
```diff
@@ -23,31 +23,41 @@
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URI;
+import java.net.URISyntaxException;
```
sarankk: Nit: Unused

src/test/containerTest/org/apache/cassandra/sidecar/restore/StorageClientTest.java
```diff
@@ -192,42 +210,68 @@ void testGetObject() throws Exception
...
         assertThat(downloaded.getAbsolutePath()).isEqualTo(existingPath.resolve("key").toString());
     }
 
     @Test
     void testGetObjectThroughputRateLimited() throws Exception
     {
-        // only allow 1/4 the speed of transfer
-        client = new StorageClient(s3AsyncClient, SidecarRateLimiter.create(LARGE_FILE_IN_BYTES >> 2));
+        // only allow 1/4 the speed of transfer; each request downloads 1024 bytes (it is an extremely low value only for testing)
+        client = new StorageClient(s3AsyncClient, 1024, SidecarRateLimiter.create(LARGE_FILE_IN_BYTES >> 2));
```
sarankk: Won't this be a problem if we run tests parallelly? spot bugs is complaining about this

sarankk (approved):
+1 Looks good to me, some checkstyle related comments

src/main/java/org/apache/cassandra/sidecar/restore/HttpRangesIterator.java
yifan-c: sounds good. The impl is more succinct

sarankk (approved):
+1 LGTM

src/main/java/org/apache/cassandra/sidecar/config/yaml/S3ClientConfigurationImpl.java
```diff
@@ -51,26 +68,34 @@ public S3ClientConfigurationImpl()
...
     }
 
     public S3ClientConfigurationImpl(String threadNamePrefix,
                                      int concurrency,
                                      long threadKeepAliveSeconds,
+                                     int rangeGetObjectBytesSize,
+                                     long apiCallTimeoutMillis,
                                      S3ProxyConfiguration proxyConfig)
     {
+        Preconditions.checkArgument(apiCallTimeoutMillis > TimeUnit.SECONDS.toMillis(10),
```
JeetKunDoug, line 83: Why is 10 seconds the minimum? Just an AWS client limitation or ?

src/main/java/org/apache/cassandra/sidecar/restore/StorageClientPool.java
```diff
@@ -103,16 +105,22 @@ private StorageClient storageClient(String region)
...
+                                                      .apiCallTimeout(apiCallTimeout))
                          .asyncConfiguration(b -> b.advancedOptions(advancedOptions));
             S3ProxyConfiguration s3ProxyConfiguration = clientConfig.proxyConfig();
             URI endpointOverride = s3ProxyConfiguration.endpointOverride();
             if (endpointOverride != null) // set for local testing only
                 clientBuilder.endpointOverride(endpointOverride)
                              .forcePathStyle(true);
 
+            NettyNioAsyncHttpClient.Builder nettyClientBuilder = NettyNioAsyncHttpClient.builder()
+                                                                                        .sslProvider(SslProvider.OPENSSL);
```
JeetKunDoug: Do we need to check if OpenSSL is available before setting the provider, or is Netty OK with this state?

yifan-c: Looks like netty does the auto-selection. There is no need to set the `sslProvider` explicitly. I will remove the option and add a warning message when openSSL is not available.

src/main/java/org/apache/cassandra/sidecar/config/yaml/S3ClientConfigurationImpl.java
yifan-c: AWS client do not enforce any timeout for api calls by default. The 10 seconds is an arbitrary value that should prevent unexpected early exit in most case. The default api call timeout set as part of the patch is 60 seconds.

JeetKunDoug (approved):

PR #133, opened 2024-08-22
CASSANDRA-19854 add .asf.yaml

smiklosovic:
patch by Stefan Miklosovic; reviewed by TBD for CASSANDRA-19854

This is same as for cassandra-gocql-driver.

Notice autolink is CASSANDRASC which will make link to JIRA out of `CASSANDRA-123` in the commit message. This is already done in main Cassandra repository as well as in gocql driver. (It was done manually for Cassandra repository though before this was configurable by asf.yaml means).

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,37 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+notifications:
+  commits:      commits@cassandra.apache.org
+  issues:       commits@cassandra.apache.org
+  pullrequests: pr@cassandra.apache.org
+  jira_options: link worklog
+
+github:
+  description: "Sidecar for Apache Cassandra®"
+  homepage: https://cassandra.apache.org/
+  enabled_merge_buttons:
+    squash:  false
+    merge:   false
+    rebase:  true
+  features:
+    wiki: false
+    issues: false
+    projects: false
+  autolink_jira:
+    - CASSANDRASC
+    - CASSANDRA


.asf.yaml
```diff
@@ -0,0 +1,36 @@
...
+  enabled_merge_buttons:
+    squash:  false
+    merge:   false
+    rebase:  true
+  features:
+    wiki: false
+    issues: false
+    projects: false
+  autolink_jira:
+    - CASSANDRASC
```
michaelsembwever, line 36: maybe `CASSANDRA` too …? (i doubt CASSANDRA-nnnn references would mean anything else here…)

michaelsembwever (approved):

.asf.yaml
yifan-c: When there are 2 projects listed, does it link to both projects? Given all tickets for sidecar are supposed to be created in `CASSANDRASC`.

.asf.yaml
```diff
@@ -0,0 +1,37 @@
...
+  commits:      commits@cassandra.apache.org
+  issues:       commits@cassandra.apache.org
+  pullrequests: pr@cassandra.apache.org
+  jira_options: link worklog
+
+github:
+  description: "Sidecar for Apache Cassandra®"
+  homepage: https://cassandra.apache.org/
+  enabled_merge_buttons:
+    squash:  false
```
yifan-c, line 28: hmm. "squash and merge" option is used regularly. Is there any reason to disable it due to policy? @smiklosovic
cc: @michaelsembwever

michaelsembwever: avoid meaningless dev-to-mainline merge commits. (they are not tracking, so have no purpose.)
fast-forward merges are cleaner to history.  project-wide there is precedence to avoid merge commits, unless they are tracking forward merges across mainlines.

smiklosovic: https://the-asf.slack.com/archives/C0459N9R5C6/p1724359199270789

.asf.yaml
michaelsembwever: even if work is done under a `CASSANDRASC` ticket, other tickets may still be mentioned.

the following implies multiple project linking works…
https://github.com/apache/infrastructure-asfyaml?tab=readme-ov-file#autolinks-for-jira

smiklosovic: no, it just links to what you create a commit message for. I get that 99% it is CASSANDRASC but there _can_ be a ticket CASSANDRA-XYZ which _might_ touch sidecar as well (like CASSANDRA-19854, right?) so it will link that one. If we kept just CASSANDRASC then this one would not be linked.

smiklosovic: what Mick said :)

.asf.yaml
yifan-c: Thank you for the timely reply Mick!

The "Squash and Merge" option does not produce the meaningless merge commit. Most of the commits made this year are using the "squash and merge" option. It avoids the work to squash locally and force push. The option is effectively the same as squash and rebase. 
I would like to enable "squash".

On a related topic, I want to disable "rebase" option. We never want to have multiple commits from a single patch. With "rebase" option enabled, there is a risk that multiple commits being merged into the mainline w/o paying attention. 

Let me know if the above is sounding.

michaelsembwever: ah, thanks @smiklosovic 

yes i forgot, "squash and merge" is ok if "required_linear_history: true" is also used.

smiklosovic: Look, I think that as a subproject you are very welcome to just do it like you please. I think if that works for your workflow we should not be rigid about that. Do you want me to make a follow up ticket or you are OK to do that yourself? Up to you. I don't mind either.

michaelsembwever: > The "Squash and Merge" option does not produce the meaningless merge commit. 

and yes, my bad, your quotes are important.  was confused because the options here and squash and merge.  the erroneous merge commits are just from the merge option.

michaelsembwever: >  I want to disable "rebase" option. We never want to have multiple commits from a single patch. With "rebase" option enabled, there is a risk that multiple commits being merged into the mainline w/o paying attention.

Personally, I would argue (and hope for) that this is part of the peer review process.  That is the squashed and finalised commit message is also reviewed, or at least visible to others before being committed.

But if your repo has `required_linear_history` and an established practice, go for it :)

yifan-c: Sounds good! Thanks all for the input. Interesting that there could be race for the squash option. I can patch the file to enable both squash and required_linear_history. @smiklosovic would you mind to review once I have the patch?

.asf.yaml
yifan-c: Thanks for the example.

.asf.yaml
yifan-c: I took another look at the option "required_linear_history: true". It is to prevent have the "meaningless" merge commits. It is not a prerequisite to enable rebase/squash. Instead, it is the other way around. Either rebase or squash must be enabled, in order to enable "required_linear_history". 

Given that, I do not think "required_linear_history" is necessary, as long as the "merge" option is disabled. 

In the future, if this project ever gets multiple branches, the merge commit is wanted for merges across branches. In that case, "required_linear_history" is undesired.

michaelsembwever: > I do not think "required_linear_history" is necessary, as long as the "merge" option is disabled.

With "squash" enabled (which is what you want), without "required_linear_history" you run the risk still (IIUC) of introducing erroneous merge commits when there's a race.

> if this project ever gets multiple branches, the merge commit is wanted for merges across branches. 

you cannot do our "only merge commits for tracking between the mainlines" practice.
what you would be introducing is erroneous merge commits for every commit from every PR.  this creates rather unpleasant (IMHO) trees in the git history, rather than the much simpler and clean fast-forwarded  tree (where the only branches correlate to each mainline).

to have linear history and only-tracking-merge commits, and permit both command line and github approaches, is difficult.  see https://github.com/michaelsembwever/atomic-linear-history-with-forward-merges-enforced-repo-demo 

and of course, atomically committing to multiple branches (which is kinda a safety requirement to the above example), can only be done form the command line.  long story short, if you have multiple mainlines and want fast-forward tracking stick with the command line.

(mainlines are your "multiple branches")

PR #135, opened 2024-08-29
CASSANDRASC-144 Introduces sidecar endpoint to vend statistics associ…

arjunashok:
…ated with the client connection

## Changes
- Introduces new handler for client-stats
- Supports query params for `client-options`, `by-protocol`, `list-connections` and `verbose`.
- Added integration tests

## Update: New commits
- Switched to fetch client stats data from virtual table `system_views.clients`
- Supports only the boolean param `all` based on usage use-cases
- Default mode (no args) returns the total no. conenctions and connections by user.
- Code structure - Client stats specific accessor, pojo are in the base `adapter` module, as any the schema  change across version should be captured here
-  Code structure - Dependent accessor interfaces migrated to `common` module.
- Other changes related to renaming `SidecarSchema` to `SidecarSchemaInitializer`, for clarity


src/test/integration/org/apache/cassandra/sidecar/routes/ClientStatsHandlerIntegrationTest.java
```diff
@@ -0,0 +1,170 @@
...
+                  }));
+        });
+    }
+
+    @CassandraIntegrationTest
+    void retrieveClientStatsVerbose(VertxTestContext context, CassandraTestContext cassandraTestContext)
+    throws Exception
+    {
+        Map<String, Boolean> params = Collections.singletonMap("verbose", true);
+        String testRoute = "/api/v1/storage/client-stats?verbose=true";
```
bbotella: Maybe add tests with more than one param?

src/test/integration/org/apache/cassandra/sidecar/routes/ClientStatsHandlerIntegrationTest.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    @CassandraIntegrationTest
+    void retrieveClientStatsByProtocol(VertxTestContext context, CassandraTestContext cassandraTestContext)
+    throws Exception
+    {
+        Map<String, Boolean> params = Collections.singletonMap("by-protocol", true);
+        String testRoute = "/api/v1/storage/client-stats?by-protocol=true";
+        testWithClient(context, client -> {
+            client.get(server.actualPort(), "127.0.0.1", testRoute)
```
bbotella: Can we get the actual `127.0.0.1` from `server` instead of hardcoded?

src/main/java/org/apache/cassandra/sidecar/routes/ClientStatsHandler.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+        MetricsOperations operations = delegate.metricsOperations();
+        Metadata metadata = delegate.metadata();
+        if (operations == null || metadata == null)
+        {
+            context.fail(cassandraServiceUnavailable());
+            return;
+        }
+
+        Map<String, String> params = httpRequest.params().entries()
```
bbotella: Do we need some kind of validation here? I see that we are using `Boolean.valueOf` on the actual delegate. That means that if, for example, `verbose=foo` param is passed, it would be evaluated to false. Is that the intended behavior?

Similar comment for the fact that we can pass in arbitrary params (that will be ignored). Is that something we want?

src/test/integration/org/apache/cassandra/sidecar/routes/ClientStatsHandlerIntegrationTest.java
arjunashok: Will do

src/test/integration/org/apache/cassandra/sidecar/routes/ClientStatsHandlerIntegrationTest.java
arjunashok: Good point. However, since all the existing integration tests are defined this way, and the `Server` implementation does not expose the host API, I'd like to keep this as-is to keep the scope of this PR confined to the changes related to this API.

src/main/java/org/apache/cassandra/sidecar/routes/ClientStatsHandler.java
arjunashok: Yes, ignoring arbitrary params is intentional, so as to not fail-fast, which is typically common-practice. 

As for the invalid/non-boolean values, the existing sidecar APIs default them to `false`. I'll update this to use the existing `parseBooleanQueryParam` utility invoked from the handler.

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -109,6 +109,8 @@ public final class ApiEndpointsV1
     public static final String ABORT_RESTORE_JOB_ROUTE = RESTORE_JOB_ROUTE + ABORT;
     public static final String RESTORE_JOB_PROGRESS_ROUTE = RESTORE_JOB_ROUTE + PROGRESS;
 
+    public static final String CLIENT_STATS_ROUTE = API_V1 + CASSANDRA + "/client-stats";
```
arjunashok: I would like to have additional "context" in the URI between `cassandra` and the operation - `client-stats`. This was `storage` previously. Reviewers - Lmk your thoughts.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/JmxGaugeOperations.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.adapters.base;
+
+/**
+ * An interface that pulls methods from the Cassandra Gauge Metrics Proxy
+ */
+public interface JmxGaugeOperations
```
arjunashok: Not needed anymore. Will remove.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/db/ConnectedClientStats.java
```diff
@@ -0,0 +1,210 @@
...
+    public final Map<String, String> authMetadata;
+
+    public static ConnectedClientStats.Builder builder()
+    {
+        return new ConnectedClientStats.Builder();
+    }
+
+    public static ConnectedClientStats from(@NotNull Row row) throws DataObjectMappingException
+    {
+        ConnectedClientStats.Builder builder = new ConnectedClientStats.Builder();
```
frankgh: builder seems unnecessary here. Can we have a simple constructor and simplify?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/db/ConnectedClientStatsSummary.java
```diff
@@ -0,0 +1,102 @@
...
+        int totalConnections = resultMap.values().stream().mapToInt(Math::toIntExact).sum();
+
+        ConnectedClientStatsSummary.Builder builder = new ConnectedClientStatsSummary.Builder();
+        builder.connectionsByUser(resultMap);
+        builder.totalConnectedClients(totalConnections);
+
+        return builder.build();
+    }
+
+    private ConnectedClientStatsSummary(ConnectedClientStatsSummary.Builder builder)
```
frankgh: simplify by removing the builder?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraMetricsOperations.java
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Default implementation that pulls methods from the Cassandra Metrics Proxy
+ */
+public class CassandraMetricsOperations implements MetricsOperations
+{
+    private ConnectedClientStatsDatabaseAccessor dbAccessor;
+
+    /**
+     * Creates a new instance with the provided {@link JmxClient}
```
JeetKunDoug: Correct JavaDoc
```suggestion
     * Creates a new instance with the provided {@link CQLSessionProvider}
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraMetricsOperations.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.stream.Collectors;
+
+import org.apache.cassandra.sidecar.adapters.base.db.ConnectedClientStats;
+import org.apache.cassandra.sidecar.adapters.base.db.ConnectedClientStatsDatabaseAccessor;
+import org.apache.cassandra.sidecar.adapters.base.db.ConnectedClientStatsSummary;
+import org.apache.cassandra.sidecar.adapters.base.db.schema.ClientStatsSchema;
+import org.apache.cassandra.sidecar.common.response.ConnectedClientStatsResponse;
+import org.apache.cassandra.sidecar.common.response.data.ClientConnectionEntry;
+import org.apache.cassandra.sidecar.common.server.CQLSessionProvider;
+import org.apache.cassandra.sidecar.common.server.JmxClient;
```
JeetKunDoug: Once the JavaDoc below is corrected, this is an unused import.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraMetricsOperations.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.cassandra.sidecar.common.server.MetricsOperations;
+
+/**
+ * Default implementation that pulls methods from the Cassandra Metrics Proxy
+ */
+public class CassandraMetricsOperations implements MetricsOperations
+{
+    private ConnectedClientStatsDatabaseAccessor dbAccessor;
```
JeetKunDoug: NIT: Should be `final`

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/db/ConnectedClientStats.java
```diff
@@ -0,0 +1,210 @@
...
+               .port(row.getInt("port"))
+               .hostname(row.getString("hostname"))
+               .username(row.getString("username"))
+               .connectionStage(row.getString("connection_stage"))
+               .protocolVersion(Integer.toString(row.getInt("protocol_version")))
+               .driverName(row.getString("driver_name"))
+               .driverVersion(row.getString("driver_version"))
+               .sslEnabled(row.getBool("ssl_enabled"))
+               .sslProtocol(row.getString("ssl_protocol"))
+               .sslCipherSuite(row.getString("ssl_cipher_suite"));
```
JeetKunDoug: There are several members of the ConnetedClientStats class missing here, and it appears the corresponding Builder methods are never called anywhere else. Why do we have all of the extra fields?

```
        this.keyspaceName 
        this.authenticationMode
        this.requestCount
        this.authMetadata
        this.clientOptions
```

Further, it seems odd to create a Builder in a `from` method that ends up just calling the constructor anyway - you could just pass a `row` to a `ConnectedClientStats` constructor and pull out the data directly in the constructor, which allows you to create one of these without the builder at all (and one less object per row of junk to create). This may become significant when you have thousands of connected clients, which we often do. Maybe this was the previously-established pattern in the data access layer, but if we're talking about up to 20k connected clients per request, creating a bunch of extra junk (even if it's short-lived and gets GCed out of eden space) doesn't make a lot of sense if we can avoid it quite easily.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/db/ConnectedClientStatsDatabaseAccessor.java
```diff
@@ -0,0 +1,66 @@
...
+    public Set<ConnectedClientStats> stats()
+    {
+        tableSchema.prepareStatements(session());
+        BoundStatement statement = tableSchema.listAll().bind();
+        ResultSet resultSet = execute(statement);
+        return resultSet.all()
+                        .stream()
+                        .map(ConnectedClientStats::from)
+                        .collect(Collectors.toSet());
+    }
```
JeetKunDoug: Rather than call `.all()` (which obviously retrieves the whole resultset into memory), how about streaming over the resultSet:

```suggestion
    /**
     * Query for all the client connection metadata with an entry per connection
     * @return {@link ConnectedClientStats} for each connection
     */
    public Set<ConnectedClientStats> stats()
    {
        tableSchema.prepareStatements(session());
        BoundStatement statement = tableSchema.listAll().bind();
        ResultSet resultSet = execute(statement);
        return StreamSupport.stream(resultSet.spliterator(),false)
                            .map(ConnectedClientStats::from)
                            .collect(Collectors.toSet());
    }
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/db/ConnectedClientStatsSummary.java
```diff
@@ -0,0 +1,102 @@
...
+
+    public static ConnectedClientStatsSummary.Builder builder()
+    {
+        return new ConnectedClientStatsSummary.Builder();
+    }
+
+    public static ConnectedClientStatsSummary from(@NotNull ResultSet resultSet) throws DataObjectMappingException
+    {
+
+        Map<String, Long> resultMap = resultSet.all().stream()
```
JeetKunDoug: Same as other call-out of `all`- we should stream the results

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreJobDatabaseAccessor.java
```diff
@@ -55,13 +55,16 @@
 public class RestoreJobDatabaseAccessor extends DatabaseAccessor<RestoreJobsSchema>
 {
     private static final ObjectMapper MAPPER = new ObjectMapper();
+    public final SidecarSchema sidecarSchema;
```
JeetKunDoug, line 58: Pull this into `SidecarDatabaseAccessor` parent for the all of the classes that use `SidecarSchema`

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -83,8 +80,8 @@ private void configureSidecarServerEventListeners()
     {
         EventBus eventBus = vertx.eventBus();
 
-        eventBus.localConsumer(ON_CASSANDRA_CQL_READY.address(), message -> startSidecarSchemaInitializer());
-        eventBus.localConsumer(ON_SERVER_STOP.address(), message -> cancelTimer(initializationTimerId.get()));
+        eventBus.localConsumer(SidecarServerEvents.ON_CASSANDRA_CQL_READY.address(), message -> startSidecarSchemaInitializer());
```
JeetKunDoug: TODO: Should we make these back to static imports?

JeetKunDoug (commented):
As discussed, I'll send a patch with our pair programming changes for you.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/data/SidecarSchemaModificationException.java
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-package org.apache.cassandra.sidecar.exceptions;
+package org.apache.cassandra.sidecar.common.response.data;
```
frankgh: can we preserve the package ?

frankgh (approved):
+1 Thanks for the PR.

server/src/test/integration/org/apache/cassandra/sidecar/routes/ConnectedConnectedClientStatsHandlerIntegrationTest.java
```diff
@@ -0,0 +1,193 @@
...
+import org.apache.cassandra.sidecar.utils.SimpleCassandraVersion;
+import org.apache.cassandra.testing.CassandraIntegrationTest;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Test the client-stats endpoint with cassandra container.
+ */
+@ExtendWith(VertxExtension.class)
+public class ConnectedConnectedClientStatsHandlerIntegrationTest extends IntegrationTestBase
```
JeetKunDoug: NIT: `ConnectedConnected`?
```suggestion
public class ConnectedClientStatsHandlerIntegrationTest extends IntegrationTestBase
```

JeetKunDoug (commented):
One small nit that should be taken care of but otherwise looks great. Thanks again!

JeetKunDoug (approved):

PR #137, opened 2024-10-03
CASSANDRASC-146 Add vertx auth subproject for mTLS authentication

sarankk:


vertx-auth-mtls/build.gradle
```diff
@@ -0,0 +1,82 @@
...
+configurations {
+    all*.exclude(group: 'ch.qos.logback')
+}
+
+dependencies {
+    implementation(group: 'io.vertx', name: 'vertx-auth-common', version: '4.5.8')
+    testImplementation(group: 'io.vertx', name: 'vertx-junit5', version: '4.5.8')
+    testImplementation group: 'org.assertj', name: 'assertj-core', version: '3.26.3'
+    testImplementation group: 'org.mockito', name: 'mockito-core', version: '2.1.0'
+    testImplementation group: 'org.bouncycastle', name: 'bcpkix-jdk15on', version: '1.70'
```
yifan-c: nit: use the same style

```suggestion
    testImplementation(group: 'org.assertj', name: 'assertj-core', version: '3.26.3')
    testImplementation(group: 'org.mockito', name: 'mockito-core', version: '2.1.0')
    testImplementation(group: 'org.bouncycastle', name: 'bcpkix-jdk15on', version: '1.70')
```

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/authentication/CertificateCredentials.java
```diff
@@ -0,0 +1,92 @@
...
+import java.security.cert.Certificate;
+import java.util.Collections;
+import java.util.List;
+import javax.net.ssl.SSLPeerUnverifiedException;
+
+/**
+ * Certificates based {@link Credentials} implementation, carries user's certificates, which can be used for
+ * authenticating or authorizing users.
+ */
+public class CertificateCredentials implements Credentials
```
yifan-c, line 33: Do you need to implement `applyHttpChallenge` and `toHttpAuthorization` methods? It is hard to see the need since this patch only adds the code that is not used.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/SpiffeIdentityExtractor.java
```diff
@@ -0,0 +1,82 @@
...
+    private static final int SUBJECT_ALT_NAME_URI_TYPE = 6;
+    private static final String SPIFFE_PREFIX = "spiffe://";
+
+    @Override
+    public String identity(Certificate[] certificateChain) throws CredentialValidationException, CertificateParsingException
+    {
+        X509Certificate[] castedCerts = castCertsToX509(certificateChain);
+        if (castedCerts.length == 0)
+        {
+            throw new CredentialValidationException("No based X509Certificate found for validating");
```
yifan-c: Is the word `based` added to the error message by accident?

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/SpiffeIdentityExtractor.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * {@link CertificateIdentityExtractor} implementation for SPIFFE certificates for extracting SPIFFE identity.
+ * SPIFFE is a URI, present as part of SAN of client certificates, it uniquely identifies client.
+ */
+public class SpiffeIdentityExtractor implements CertificateIdentityExtractor
+{
+    private static final int SUBJECT_ALT_NAME_URI_TYPE = 6;
```
yifan-c, line 37: Consider add a comment to further explain what is 6. 

According to https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6, type 6 is for URI.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/SpiffeIdentityExtractor.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Override
+    public String identity(Certificate[] certificateChain) throws CredentialValidationException, CertificateParsingException
+    {
+        X509Certificate[] castedCerts = castCertsToX509(certificateChain);
+        if (castedCerts.length == 0)
+        {
+            throw new CredentialValidationException("No based X509Certificate found for validating");
+        }
+        X509Certificate privateCert = castedCerts[0];
```
yifan-c: why checking only the first cert from the array?

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/SpiffeIdentityExtractor.java
```diff
@@ -0,0 +1,82 @@
...
+     * @return an array of {@link X509Certificate} certificates
+     */
+    private X509Certificate[] castCertsToX509(Certificate[] certificateChain) throws CredentialValidationException
+    {
+        if (certificateChain == null || certificateChain.length == 0)
+        {
+            throw new CredentialValidationException("Certificate chain shared is empty");
+        }
+        return Arrays.stream(certificateChain).filter(certificate -> certificate instanceof X509Certificate).toArray(X509Certificate[]::new);
+    }
```
yifan-c: I would suggest doing this instead. You can avoid the throw statement.

```suggestion
    private List<X509Certificate> castCertsToX509(Certificate[] certificateChain) throws CredentialValidationException
    {
        List<X509Certificate> x509Certs = new ArrayList<>();
        for (Certificate cert : certificateChain)
        {
            if (cert instanceof X509Certificate)
            {
                x509Certs.add((X509Certificate) cert);
            }
        }
        return x509Certs;
    }
```

In fact, it

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/CertificateIdentityExtractor.java
```diff
@@ -0,0 +1,47 @@
...
+     *  <li>an identifier in SAN of the certificate like SPIFFE
+     *  <li>CN of the certificate
+     *  <li>any other fields in the certificate can be combined and be used as identifier of the certificate
+     * </ul>
+     *
+     * @param certificateChain certificate chain of user
+     * @return identity {@code String} extracted from certificate
+     * @throws CredentialValidationException when identity cannot be extracted
+     */
+    String identity(Certificate[] certificateChain) throws CredentialValidationException, CertificateParsingException;
```
yifan-c: 1. Can we use `List<Certificate>` instead of array? 
The only callsite, i.e. `MutualTlsAuthenticationProvider#authenticate` converts from List to array to call the method. 

2. How about we refine the contract of this API that it only returns non-null string? When it cannot extract any identity, it throws the `CredentialValidationException`. So there is no scenario, that the API would return null. In other words, this method either extracts some identity or throws.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/CertificateIdentityExtractor.java
```diff
@@ -0,0 +1,47 @@
...
+     * <p>An example of identity could be the following:
+     * <ul>
+     *  <li>an identifier in SAN of the certificate like SPIFFE
+     *  <li>CN of the certificate
+     *  <li>any other fields in the certificate can be combined and be used as identifier of the certificate
+     * </ul>
+     *
+     * @param certificateChain certificate chain of user
+     * @return identity {@code String} extracted from certificate
+     * @throws CredentialValidationException when identity cannot be extracted
```
yifan-c: missing javadoc for `CertificateParsingException`

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/MutualTlsAuthenticationProvider.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+        catch (Exception e)
+        {
+            return Future.failedFuture("Invalid certificate passed");
+        }
+
+        if (identity == null || identity.isEmpty())
+        {
+            return Future.failedFuture("Could not extract identity from certificate");
+        }
```
yifan-c: This block can be deleted, if you update the API to return non-null value only.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/MutualTlsAuthenticationProvider.java
```diff
@@ -0,0 +1,93 @@
...
+        CertificateCredentials certificateCredentials = (CertificateCredentials) credentials;
+        String identity;
+        try
+        {
+            certificateValidator.isValidCertificate(certificateCredentials);
+            identity = identityExtractor.identity(certificateCredentials.certificateChain().toArray(new Certificate[0]));
+        }
+        catch (Exception e)
+        {
+            return Future.failedFuture("Invalid certificate passed");
```
yifan-c: do not swallow the exception.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/CertificateIdentityValidator.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public interface CertificateIdentityValidator
+{
+    /**
+     * Verifies that identity extracted from certificate is a valid identity.
+     *
+     * @param identity {@code String} string extracted from certificate, uniquely represents client
+     * @return {@code true} if the identity is valid, {@code false otherwise}
+     */
+    boolean isValidIdentity(String identity);
```
yifan-c: Prefer throwing exception instead of returning boolean for the validation. The exception conveys better context when it throws. 

```suggestion
    void validateIdentity(String identity) throws InvalidIdentityException;
```

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/MutualTlsAuthenticationProvider.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        if (identity == null || identity.isEmpty())
+        {
+            return Future.failedFuture("Could not extract identity from certificate");
+        }
+        if (!identityValidator.isValidIdentity(identity))
+        {
+            return Future.failedFuture("Certificate identity is not a valid identity");
+        }
```
yifan-c: You can merge the validation code with the try-catch block above, once you made the interface to throw exception, instead of returning boolean.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/MutualTlsAuthenticationProvider.java
```diff
@@ -0,0 +1,93 @@
...
+        return Future.succeededFuture(User.fromName(identity));
+    }
+
+    /**
+     * @deprecated use {@code authenticate(Credentials credentials, Handler<AsyncResult<User>> resultHandler)} instead
```
yifan-c: 1. IIRC, codestyle plugin does not like `@deprecated`
2. the message `use {@code authenticate(Credentials credentials, Handler<AsyncResult<User>> resultHandler)} instead` is incorrect. It is referring to itself. You probably want to have `use {@code authenticate(Credentials credentials} instead`

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/CertificateValidator.java
```diff
@@ -0,0 +1,47 @@
...
+import io.vertx.ext.auth.authentication.CertificateCredentials;
+
+import java.security.cert.CertificateParsingException;
+
+/**
+ * Interface for validating certificates for mutual TLS authentication.
+ * <p>
+ * This interface can be implemented to provide logic for validating various fields from Certificates.
+ */
+public interface CertificateValidator
```
yifan-c, line 29: I feel that all three interfaces `CertificateValidator`, `CertificateIdentityExtractor` and `CertificateIdentityValidator` can merge together to reduce cognitive complexity. They are tightly coupled logically, even if they are 3 individual interfaces as of now. For example, you always pair `SpiffeIdentityExtractor` with `SpiffeIdentityValidator`. For another example, it is probably never a valid scenario to have this combination, `AllowAllCertValidator + SpiffeIdentityExtractor + AllowAllIndentityValidator`, but the `MutualTlsAuthenticationProvider` constructor would allow it.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/CertificateValidator.java
```diff
@@ -0,0 +1,47 @@
...
+     *  <li>Verifying CA information
+     *  <li>Checking CN information
+     *  <li>Validating Issuer information
+     *  <li>Checking organization information etc
+     * </ul>
+     *
+     * @param credentials user certificate credentials shared
+     * @return {@code true} if the credentials are valid, {@code false} otherwise
+     */
+    boolean isValidCertificate(CertificateCredentials credentials) throws CertificateParsingException;
```
yifan-c: if it throws exception, you do not need to return boolean. If it is valid, no throw; it throws otherwise.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/CertificateValidatorImpl.java
```diff
@@ -0,0 +1,189 @@
...
+    private final String trustedIssuerCountry;
+
+    public CertificateValidatorImpl()
+    {
+        this.trustedCNs = Collections.emptySet();
+        this.trustedIssuerOrganization = null;
+        this.trustedIssuerOrganizationUnit = null;
+        this.trustedIssuerCountry = null;
+    }
+
```
yifan-c: not used

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/CertificateValidatorImpl.java
```diff
@@ -0,0 +1,189 @@
...
+            catch (CertificateExpiredException e)
+            {
+                throw new CredentialValidationException("Expired certificates shared for authentication");
+            }
+            catch (Exception e)
+            {
+                return false;
+            }
+        }
+        return false;
```
yifan-c: It requires the first certificate to be a `X509Certificate`, otherwise returns false? Also, why only the first cert?

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/SpiffeIdentityValidator.java
```diff
@@ -0,0 +1,62 @@
...
+
+import io.vertx.ext.auth.mtls.CertificateIdentityValidator;
+
+/**
+ * {@link CertificateIdentityValidator} implementation for SPIFFE certificates for validating SPIFFE identities.
+ * SPIFFE is a URI, present as part of SAN of client certificates, it uniquely identifies client.
+ */
+public class SpiffeIdentityValidator implements CertificateIdentityValidator
+{
+    private static final String SPIFFE_ID_PREFIX = "spiffe://";
```
yifan-c: The same constant is define in multiple classes. Another sign that those classes should be a single one.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/SpiffeIdentityExtractor.java
sarankk: We assume the first cert is always the private key entry in cert chain arrary. I initially had a loop to go through certs and check their type to get private key specific details. But  there is no method to check if an entry is private key or CA cert entry.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/SpiffeIdentityExtractor.java
sarankk: Yes removing it

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/authentication/CertificateCredentials.java
sarankk: No we don't need those, hence left it to default. in `MutualTlsAuthenticationProvider` we get the certificate chain from `CertificateCredentials` and authenticate with it. We probably dont need `CertificateCredentials` except for `checkValid`. It maintains consistency with other auth subprojects which provide `TokenCredentials` `UsernamePasswordCredentials`

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/SpiffeIdentityExtractor.java
yifan-c: Let's think more about it. If we go with this route, definitely need to call out the implicit in javadoc.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/CertificateValidator.java
sarankk: `CertificateValidator` and `CertificateIdentityValidator` are independent validators. Former validates details in certificates (such as issuer, owner, validity) and latter validates the identity extracted from certificate is trusted identity. We could have `AllowAllCertificateValidator` with custom `SpiffeIdentityValidator`, in scenario where users are not particular about certificate details, but only care about identity within a certificate. Likewise we could have custom `CertificateValidator` and `AllowAllIdentityValidator` if user cares about only details within the certificate. 

We can combine `CertificateIdentityExtractor` and `CertificateIdentityValidator` interfaces. Reason I had the identity extractor separately is, we can use the identity extracted for authorization. But we could combine both the interfaces under `CertificateIdentityExtractor` and have the valid method return the identity. for e.g.

`String validIdentity(Certificate certificate) throws IdentityValidationException` 

wdyt?

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/CertificateValidatorImpl.java
sarankk: In interface methods, we take in `Certificate` but for default implementations we support checking only `X509Certificate` . This is because `X509Certificate` implementation has methods to get SAN detail. Else we have to parse the certificate content and extract SAN detail. Regarding first cert, we check with the assumption that first cert is private key entry. When we extract the chain from `HttpRequest` that is the case too.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/CertificateValidator.java
yifan-c: It sounds good. Thanks for the explanation.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/impl/MutualTlsAuthenticationProvider.java
sarankk: thanks, added `@Deprecated` annotation to method instead. The doc is not referring to itself. It is instead pointing to use `authenticate(Credentials credentials, Handler<AsyncResult<User>> resultHandler)`, to use `Credentials` instead of `JsonObject`.

vertx-auth-mtls/build.gradle
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+configurations {
+    all*.exclude(group: 'ch.qos.logback')
+}
+
+dependencies {
+    implementation(group: 'io.vertx', name: 'vertx-auth-common', version: '4.5.8')
+    testImplementation(group: 'io.vertx', name: 'vertx-junit5', version: '4.5.8')
```
frankgh: Any reason we are using a different version from the main project? Maybe you want to move the entire project to this vertx version?
```suggestion
    implementation(group: 'io.vertx', name: 'vertx-auth-common', version: "${project.vertxVersion}")
    testImplementation(group: 'io.vertx', name: 'vertx-junit5', version: "${project.vertxVersion}")
```

vertx-auth-mtls/build.gradle
```diff
@@ -0,0 +1,82 @@
...
+}
+
+group 'org.apache.cassandra.sidecar'
+version project.version
+
+sourceCompatibility = 1.8
+
+repositories {
+    mavenCentral()
+}
```
frankgh: this is not necessary

vertx-auth-mtls/build.gradle
```diff
@@ -0,0 +1,82 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.nio.file.Paths
+
+plugins {
+    id('java-library')
+    id('idea')
+    id('maven-publish')
```
frankgh, line 24: we can remove if we decide not to publish, which I think makes sense since this is an internal dependency

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/authentication/CertificateCredentials.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private X509Certificate getPeerCertificate()
+    {
+        // First certificate in the chain is peer's own cert
+        if (!certificateChain.isEmpty() && certificateChain.get(0) instanceof X509Certificate)
+        {
+            return (X509Certificate) certificateChain.get(0);
+        }
+
+        return null;
```
frankgh, line 113: Should we allow for the first X509Certificate in the chain?
```suggestion
                // First certificate in the chain is peer's own cert
        return certificateChain.stream()
                               .filter(cert -> cert instanceof X509Certificate)
                               .map(cert -> (X509Certificate) cert)
                               .findFirst()
                               .orElse(null);
```

vertx-auth-mtls/build.gradle
sarankk: Makes sense. Will remove it

vertx-auth-mtls/build.gradle
sarankk: Updating this to latest Vert.x version for easier contribution back.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/authentication/CertificateCredentials.java
sarankk: Since the first entry is usually the private key entry, we are checking for that

yifan-c: > Should we allow for the first X509Certificate in the chain?

No. The behavior is get the first cert and proceed if it is a X509Certificate

PR #138, opened 2024-10-08
CASSANDRASC-147: Expose vert.x filesystem options configuration

frankgh:
By default, vert.x will attempt to resolve files from the application classpath when it is unable to find them in the local filesystem. Additionally, by default vert.x will cache any files that it reads from the classpath into the local filesystem.

For Sidecar, this optimization is unnecessary as Sidecar doesn't package anything in the classpath that might be used while running the application.

In this commit, we disable this optimization by default, but expose configuration options to tune these options on need-basis.

Patch by Francisco Guerrero; Reviewed by Yifan Cai for CASSANDRASC-147

### DIFF
--- a/server/src/main/dist/conf/sidecar.yaml
+++ b/server/src/main/dist/conf/sidecar.yaml
@@ -68,9 +68,6 @@ sidecar:
   tcp_keep_alive: false
   accept_backlog: 1024
   server_verticle_instances: 1
-  vertx_filesystem_options:
-    class_path_resolving_enabled: false
-    file_caching_enabled: false
   throttle:
     stream_requests_per_sec: 5000
     timeout_sec: 10
@@ -112,6 +109,11 @@ sidecar:
     keyspace: sidecar_internal
     replication_strategy: SimpleStrategy
     replication_factor: 1
+
+vertx:
+  filesystem_options:
+    classpath_resolving_enabled: false
+    file_caching_enabled: false
 
 #
 # Enable SSL configuration (Disabled by default)
--- a/server/src/main/dist/conf/sidecar.yaml
+++ b/server/src/main/dist/conf/sidecar.yaml
@@ -68,6 +68,9 @@ sidecar:
   tcp_keep_alive: false
   accept_backlog: 1024
   server_verticle_instances: 1
+  vertx_filesystem_options:
+    class_path_resolving_enabled: false
+    file_caching_enabled: false
   throttle:
     stream_requests_per_sec: 5000
     timeout_sec: 10
--- a/server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
@@ -587,13 +587,13 @@ public class MainModule extends AbstractModule
      * Builds the {@link InstanceMetadata} from the {@link InstanceConfiguration},
      * a provided {@code  versionProvider}, and {@code healthCheckFrequencyMillis}.
      *
-     * @param vertx             the vertx instance
-     * @param cassandraInstance the cassandra instance configuration
-     * @param versionProvider   a Cassandra version provider
-     * @param sidecarVersion    the version of the Sidecar from the current binary
-     * @param jmxConfiguration  the configuration for the JMX Client
-     * @param session           the CQL Session provider
-     * @param registryFactory   factory for creating cassandra instance specific registry
+     * @param vertx                 the vertx instance
+     * @param cassandraInstance     the cassandra instance configuration
+     * @param versionProvider       a Cassandra version provider
+     * @param sidecarVersion        the version of the Sidecar from the current binary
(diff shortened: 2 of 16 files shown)


server/src/main/dist/conf/sidecar.yaml
```diff
@@ -68,6 +68,9 @@ sidecar:
   tcp_keep_alive: false
   accept_backlog: 1024
   server_verticle_instances: 1
+  vertx_filesystem_options:
+    class_path_resolving_enabled: false
+    file_caching_enabled: false
```
yifan-c: nit: how about having the structure `vertx -> filesystem_options` to be future-proof? In case, there are other vertx configurations to be exposed via yaml.

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -162,7 +164,25 @@ public Vertx vertx(SidecarConfiguration sidecarConfiguration, MetricRegistryFact
                                         // Monitor all V1 endpoints.
                                         // Additional filtering is done by configuring yaml fields 'metrics.include|exclude'
                                         .addMonitoredHttpServerRoute(serverRouteMatch);
-        return Vertx.vertx(new VertxOptions().setMetricsOptions(dropwizardMetricsOptions));
+
+        FileSystemOptions fsOptions;
+        VertxFilesystemOptionsConfiguration configuredFSOptions = sidecarConfiguration.serviceConfiguration()
+                                                                                      .vertxFilesystemOptionsConfiguration();
```
yifan-c: Can `vertxFilesystemOptionsConfiguration` return null? It does not according to the implementation.
If it can, it would be good to add `@Nullable` to the signature.

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -162,7 +164,25 @@ public Vertx vertx(SidecarConfiguration sidecarConfiguration, MetricRegistryFact
...
+                        .setFileCachingEnabled(configuredFSOptions.fileCachingEnabled());
+        }
+        else
+        {
+            fsOptions = new FileSystemOptions();
+        }
+
+        VertxOptions vertxOptions = new VertxOptions().setMetricsOptions(dropwizardMetricsOptions)
+                                                      .setFileSystemOptions(fsOptions);
+        return Vertx.vertx(vertxOptions);
```
yifan-c: nit: this is more succinct. 

```java
        VertxOptions vertxOptions = new VertxOptions().setMetricsOptions(dropwizardMetricsOptions);
        VertxFilesystemOptionsConfiguration configuredFSOptions = sidecarConfiguration.serviceConfiguration()
                                                                                      .vertxFilesystemOptionsConfiguration();
        if (configuredFSOptions != null)
        {
            FileSystemOptions fsOptions = new FileSystemOptions()
                                          .setClassPathResolvingEnabled(configuredFSOptions.classPathResolvingEnabled())
                                          .setFileCacheDir(configuredFSOptions.fileCacheDir())
                                          .setFileCachingEnabled(configuredFSOptions.fileCachingEnabled());
            vertxOptions.setFileSystemOptions(fsOptions);
        }

        return Vertx.vertx(vertxOptions);
```

server/src/test/java/org/apache/cassandra/sidecar/config/SidecarConfigurationTest.java
```diff
@@ -439,6 +457,14 @@ void validateCassandraInputValidationConfigurationFromYaml(CassandraInputValidat
         assertThat(config.allowedPatternForRestrictedComponentName()).isEqualTo("[a-zA-Z0-9_-]+(.db|TOC.txt)");
     }
 
+    void validateVertxFilesystemOptionsConfiguration(VertxFilesystemOptionsConfiguration config)
```
yifan-c: maybe `validateVertxFilesystemOptionsClasspathResolvingDisabled`?

server/src/test/resources/config/sidecar_multiple_instances.yaml
```diff
@@ -71,6 +71,9 @@ sidecar:
   tcp_keep_alive: false
   accept_backlog: 1024
   server_verticle_instances: 1
+  vertx_filesystem_options:
+    class_path_resolving_enabled: false
```
yifan-c: I thought `classpath` is commonly recognized as a single word :p https://en.wikipedia.org/wiki/Classpath

yifan-c (commented):
looks good overall.

server/src/main/dist/conf/sidecar.yaml
frankgh: Yeah, I thought about this, I think it is a good idea, I was mostly trying to avoid nesting.

server/src/test/resources/config/sidecar_multiple_instances.yaml
frankgh: good call, I missed this

PR #139, opened 2024-10-23
CASSSIDECAR-150: Adds job management framework and APIs

arjunashok:
These changes are part of the effort to introduce bespoke Sidecar APIs to support key operational functionality currently managed through nodetool commands. The APIs for some of the commands can trigger long-running tasks on Cassandra. 
The job management framework allows  management of the async jobs submitted via the command APIs while coordinating downstream with the Cassandra node.

## Changes
- Job management framework
- Introduces APIs for `job status` and `list jobs`, with corresponding Sidecar client classes and tests


client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -109,6 +109,9 @@ public final class ApiEndpointsV1
     public static final String ABORT_RESTORE_JOB_ROUTE = RESTORE_JOB_ROUTE + ABORT;
     public static final String RESTORE_JOB_PROGRESS_ROUTE = RESTORE_JOB_ROUTE + PROGRESS;
 
+    public static final String LIST_JOBS_ROUTE = API_V1 + CASSANDRA + "/jobs";
+    public static final String JOB_STATUS_ROUTE = API_V1 + CASSANDRA + "/jobs/" + JOB_ID_PATH_PARAM + "/status";
```
yifan-c: `/status` feels redundant. 

`.../jobs/:jobId` can be used to retrieve the job details, including status. The job details API is not there. Is it reserved for later?

arjunashok: > The job details API is not there. Is it reserved for later?

This is the only job API since it would return the job status, operation and failure reason when applicable. I do not see a use-case for a separate details API, although I do agree we could do away with the `status` suffix.

client/src/main/java/org/apache/cassandra/sidecar/client/RequestContext.java
```diff
@@ -486,6 +489,29 @@ public Builder connectedClientStatsRequest()
             return request(new ConnectedClientStatsRequest());
         }
 
+        /**
+         * Sets the {@code request} to be a {@link JobStatusRequest} and returns a reference to this Builder
+         * enabling method chaining.
+         *
+         * @return a reference to this Builder
+         */
+        public Builder jobStatusRequest(String jobId)
```
skoppu22: Minor, to be in.line with listJobsRequest() naming style below, can we have this named statusJobRequest?

server/src/main/java/org/apache/cassandra/sidecar/job/JobTracker.java
```diff
@@ -0,0 +1,254 @@
...
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected boolean removeEldestEntry(final Map.Entry<UUID, Job> eldest)
+    {
+        if (size() <= capacity)
```
skoppu22: Let's say capacity is 5 (just assuming for the discussion), all 5 jobs are ongoing, if a new job comes up do we force delete the eldest ongoing job from tracker? If so, how can a user track that deleted eldest job?
Instead of having capacity, if we are just worried about too many ongoing jobs and jobtracker taking up too much memory, can we have a cap on ongoing jobs and reject incoming requests if jobtracker reached the maximum limit and ask user to wait to finish an ongoing job to submit a new one?

server/src/main/java/org/apache/cassandra/sidecar/routes/JobStatusHandler.java
```diff
@@ -0,0 +1,115 @@
...
+                             logger.info(response);
+                             context.fail(wrapHttpException(HttpResponseStatus.NOT_FOUND, response));
+                         }
+                         return job;
+                     })
+                     .onFailure(cause -> processFailure(cause, context, host, remoteAddress, request))
+                     .onSuccess(job -> sendStatusBasedResponse(context, jobUUID, job));
+    }
+
+    private UUID verifyJobIdPresent(RoutingContext context)
```
skoppu22: From the name verifyJobIdPresent, I was expecting this function to lookup jobId in the ongoing jobIds list and error if not found. Whereas this function only ensuring given jobid is in UUID format. Not sure am I missing something.
If we just want to ensure jobId string is a valid UUID, can we rename this function as validateJobId (or similar)?

skoppu22:
Good work 👍

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
frankgh: I think we should qualify these endpoints as admin. Also, I think we should not have the `status` path part . I propose we use the following instead:

```
/api/v1/admin/cassandra/jobs
/api/v1/admin/cassandra/jobs/<job_id>
```

client-common/src/main/java/org/apache/cassandra/sidecar/common/http/SidecarHttpHeaderNames.java
```diff
@@ -31,4 +31,9 @@ public final class SidecarHttpHeaderNames
      * {@code "cassandra-content-xxhash32-seed"}
      */
     public static final String CONTENT_XXHASH32_SEED = "cassandra-content-xxhash32-seed";
+
+    /**
+     * {@code "cassandra-sidecar-job-uuid"}
+     */
+    public static final String ASYNC_JOB_UUID = "cassandra-sidecar-job-uuid";
```
frankgh: I don't think this should be a header name, instead we should return a the UUID as part of the JSON payload

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -588,6 +599,13 @@ public LocalTokenRangesProvider localTokenRangesProvider(InstancesConfig instanc
         return new CachedLocalTokenRanges(instancesConfig, dnsResolver);
     }
 
+    @Provides
+    @Singleton
+    public JobTracker jobTracker()
+    {
+        return new JobTracker(64);
```
frankgh: should we make this configurable?

server/src/main/java/org/apache/cassandra/sidecar/routes/ListJobsHandler.java
```diff
@@ -0,0 +1,70 @@
...
+
+    protected Void extractParamsOrThrow(RoutingContext context)
+    {
+        return null;
+    }
+
+    @Override
+    protected void handleInternal(RoutingContext context, HttpServerRequest httpRequest, String host, SocketAddress remoteAddress, Void request)
+    {
+        executorPools.service()
```
frankgh: There's no need to run this on the service executor pool. These jobs are in memory.

server/src/main/java/org/apache/cassandra/sidecar/routes/ListJobsHandler.java
```diff
@@ -0,0 +1,70 @@
...
+{
+    private final JobManager jobManager;
+    @Inject
+    public ListJobsHandler(InstanceMetadataFetcher metadataFetcher, ExecutorPools executorPools, CassandraInputValidator validator, JobManager jobManager)
+    {
+        super(metadataFetcher, executorPools, validator);
+        this.jobManager = jobManager;
+    }
+
+    protected Void extractParamsOrThrow(RoutingContext context)
```
frankgh: let's add the `@Override` annotation.

server/src/main/java/org/apache/cassandra/sidecar/routes/JobStatusHandler.java
```diff
@@ -0,0 +1,115 @@
...
+    {
+        return null;
+    }
+
+    @Override
+    public void handleInternal(RoutingContext context, HttpServerRequest httpRequest, String host, SocketAddress remoteAddress, Void request)
+    {
+        UUID jobUUID = verifyJobIdPresent(context);
+
+        executorPools.service().executeBlocking(() -> {
```
frankgh: this is getting data from memory, no need to execute it in the service executor pool

server/src/main/java/org/apache/cassandra/sidecar/routes/JobStatusHandler.java
```diff
@@ -0,0 +1,115 @@
...
+{
+    private final JobManager jobManager;
+    @Inject
+    public JobStatusHandler(InstanceMetadataFetcher metadataFetcher, ExecutorPools executorPools, CassandraInputValidator validator, JobManager jobManager)
+    {
+        super(metadataFetcher, executorPools, validator);
+        this.jobManager = jobManager;
+    }
+
+    protected Void extractParamsOrThrow(RoutingContext context)
```
frankgh: this should return the jobUUID

server/src/main/java/org/apache/cassandra/sidecar/job/JobTracker.java
```diff
@@ -0,0 +1,254 @@
...
+ * Tracks and stores the results of long-running jobs running on the sidecar
+ */
+public class JobTracker extends LinkedHashMap<UUID, Job>
+{
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(JobTracker.class);
+    int capacity;
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
+
+    public JobTracker(final int initialCapacity)
```
frankgh: let's remove unnecessary `final` usage in this class

server/src/main/java/org/apache/cassandra/sidecar/job/JobManager.java
```diff
@@ -0,0 +1,102 @@
...
+
+    /**
+     * Creates a manager instance with a default sized job-tracker.
+     * @param executorPools
+     */
+    @Inject
+    public JobManager(ExecutorPools executorPools)
+    {
+        this.executorPools = executorPools;
+        jobTracker = new JobTracker(64);
```
frankgh: shouldn't job tracker be injected and we should use the singleton instance?

server/src/main/java/org/apache/cassandra/sidecar/job/JobManager.java
```diff
@@ -0,0 +1,102 @@
...
+     * @param jobSupplier supplier used to create an instance of the job
+     * @return the instance of the job that is either being tracked or was just submitted
+     */
+    public Job trySubmitJob(UUID jobId, Supplier<Job> jobSupplier)
+    {
+
+        return jobTracker.computeIfAbsent(jobId, id -> {
+            Job job = jobSupplier.get();
+            LOGGER.info("Created job with ID: {}, operation: {}", job.jobId(), job.operation());
+            jobTracker.put(job.jobId(), job);
```
frankgh: this seems wrong, you shouldn't call put here, that's the job of computeIfAbsent, no?

server/src/main/java/org/apache/cassandra/sidecar/job/Job.java
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Execute the job behavior as specified in the operation supplier {@link #jobOperationSupplier()},
+     * while tracking the status of the job's lifecycle.
+     */
+    public void execute()
+    {
+        try
+        {
+            LOGGER.info("Executing job with ID: {}", jobId);
+            JobResult result = jobOperationSupplier().get();
```
frankgh: it feels we should use Future primitives here instead, execute should return  a future as well, instead of doing latch countdown

server/src/main/java/org/apache/cassandra/sidecar/job/JobTracker.java
```diff
@@ -0,0 +1,254 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Tracks and stores the results of long-running jobs running on the sidecar
+ */
+public class JobTracker extends LinkedHashMap<UUID, Job>
```
frankgh: do we really need to expose all the methods from Map here? I only see usages of `getJobsView` , `get`, `computeIfAbsent` (which is not even guarded by the lock), and `put` which seems wrong to have . I think I'd prefer to have a smaller API, that exposes only the things we need. Maybe we can simplify by doing this:

```
        Map<UUID, Job> internalTracker = Collections.synchronizedMap(new LinkedHashMap<UUID, Job>()
        {
            @Override
            protected boolean removeEldestEntry(Map.Entry<UUID, Job> eldest)
            {
                if (size() <= capacity)
                {
                    return false;
                }

                LOGGER.warn("Job tracker reached max size, so expiring job uuid={}", eldest.getKey());
                return true;
            }
        });
```

frankgh (commented):
Added some initial comments

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
yifan-c: Yep. I mean exactly this, replacing `jobs/:jobId/status` with `jobs/:jobId`

server/src/main/java/org/apache/cassandra/sidecar/job/JobTracker.java
arjunashok: So, the intention behind the tracker is only to maintain the state for ongoing jobs without growing unbounded, but not to throttle/limit the no. of jobs that can be posted.

Since jobs are typically short-lived, the entry's age based eviction seems reasonable as a capping mechanism. 

> how can a user track that deleted eldest job

Ideally, we should size this optimally to minimize evictions of ongoing jobs.
If a job that is inflight on the C* node gets evicted from the cache, the subsequent polling will return a 404. 

The expected behavior for the client at this point is to retry the command (eg. decommission), here the downstream check will determine that there is an ongoing operation and return a 409 Conflict, and start tracking it with a new UUID (which is part of the response).

server/src/main/java/org/apache/cassandra/sidecar/routes/JobStatusHandler.java
arjunashok: Fair point, will update.

server/src/main/java/org/apache/cassandra/sidecar/job/JobManager.java
arjunashok: yep redundant, removed.

server/src/main/java/org/apache/cassandra/sidecar/job/JobTracker.java
arjunashok: In addition to overriding the LHM, we also intend to make it thread-safe with read/write locks around the operations, so we will need all the methods.

yifan-c:
There are empty lines in many files. Those empty lines should be removed.

server/src/main/java/org/apache/cassandra/sidecar/job/JobTracker.java
yifan-c: How about _composing_ a concurrent hash map, instead of _inheriting_ a map. Composition provides a better flexibility to expose only the necessary APIs. 
I am also not convinced with removing the old jobs. If bounded maps is required, it is clearer to me to reject the creation of new jobs. Otherwise, it (i.e. removing jobs underneath) loses the purpose of tracking. 

```java
public class JobTracker
{
    private ConcurrentHashMap<UUID, Job> jobsById;
```

(somehow my last comment got swallowed. Writing up the same comment again...)

frankgh: Wrapping the map in `Collections.synchronizedMap` makes it thread safe. and we reduce the exposure of the internals

frankgh: I think @yifan-c and I are suggesting the same.

yifan-c: The tracker does not need to be a map. It just uses map internally, and the map should be encapsulated within the tracker, not exposing the map APIs.

client-common/src/main/java/org/apache/cassandra/sidecar/common/http/SidecarHttpHeaderNames.java
arjunashok: As discussed offline, the responses for operation APIs will have a payload. 

The UUID header, however is meant to provide a consistent interface for all long running jobs, independent of whether it has a response or not (eg. long-running GET operations).

So, for a client triggering a operation, the expectation is to use the header value to poll for completion (using the job status API) without having to parse the response when the status is 202.

server/src/main/java/org/apache/cassandra/sidecar/job/JobTracker.java
skoppu22: I am still not convinced, as this doesn't feel like smooth user experience. Returning 404 after Sidecar restart sounds fine. But without any process restart, kicking out an ongoing job from the map and user seeing 404 doesn't feel like great user experience. It will confuse the user who just submitted the job, polling for status and seeing 404.

client-common/src/main/java/org/apache/cassandra/sidecar/common/http/SidecarHttpHeaderNames.java
```diff
@@ -31,4 +31,9 @@ public final class SidecarHttpHeaderNames
      * {@code "cassandra-content-xxhash32-seed"}
      */
     public static final String CONTENT_XXHASH32_SEED = "cassandra-content-xxhash32-seed";
+
+    /**
+     * {@code "cassandra-operations-job-uuid"}
+     */
+    public static final String OPERATIONS_JOB_HEADER_NAME = "cassandra-operations-jobid";
```
yifan-c: the header seems to be duplication since there is `OperationsJobsResponse` that contains jobId too.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/ListOperationsJobsRequest.java
```diff
@@ -0,0 +1,45 @@
...
+package org.apache.cassandra.sidecar.common.request;
+
+import io.netty.handler.codec.http.HttpMethod;
+import org.apache.cassandra.sidecar.common.ApiEndpointsV1;
+import org.apache.cassandra.sidecar.common.response.ListOperationsJobsResponse;
+
+/**
+ * Represents a request to retrieve the list of async operations jobs
+ */
+public class ListOperationsJobsRequest extends JsonRequest<ListOperationsJobsResponse>
```
yifan-c: "OperationsJobs" reads unnatural. Should it be "OperationalJobs"?

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/ListOperationsJobsResponse.java
```diff
@@ -0,0 +1,77 @@
...
+    {
+        return jobs;
+    }
+
+    /**
+     * Structure of the operations job instance within the list operations jobs API response
+     */
+    public static class OperationsJobsResponse
+    {
+        public final UUID jobId;
```
yifan-c: I think the jobId does not necessarily be a `UUID`, since it is produced by the local instance and queried against the instance only. It can be just an integer. 
But not necessarily to be addressed neither.

client-common/src/main/java/org/apache/cassandra/sidecar/common/utils/OperationsJobResult.java
```diff
@@ -0,0 +1,68 @@
...
+ * Holder class for the results of a job execution. Captures the job status and reason
+ * as a result of performing the downstream operation.
+ */
+public class OperationsJobResult
+{
+    /**
+     * Encapsulates the states of the job lifecycle. All new jobs are in Pending state.
+     */
+    public enum OperationsJobStatus
+    { Pending, Running, Completed, Failed }
```
yifan-c: nit: one line per variant, and `Completed --> Succeeded` 

Do we really need to distinguish between `pending` and `running`?

client-common/src/main/java/org/apache/cassandra/sidecar/common/utils/OperationsJobResult.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private OperationsJobStatus status;
+    private String reason;
```
yifan-c: `public final` is usually preferred over `private`

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/OperationsJobsResponse.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Response structure of the operations jobs API
+ */
+public class OperationsJobsResponse
+{
+    private final UUID jobId;
+    private final OperationsJobResult.OperationsJobStatus status;
+    private final String operation;
+    private final String reason;
```
yifan-c: `OperationsJobResult` consists of status and reason. Why flatten here?

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/OperationsJobsResponse.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.cassandra.sidecar.common.utils.OperationsJobResult;
+
+/**
+ * Response structure of the operations jobs API
+ */
+public class OperationsJobsResponse
```
yifan-c: Add `@JsonIgnoreProperties(ignoreUnknown = true)` for compatibility.

Add `@JsonInclude(JsonInclude.Include.NON_NULL)` if any field is nullable.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/OperationsJobsRequest.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Represents a request to retrieve the status of a async operations job
+ */
+public class OperationsJobsRequest extends JsonRequest<OperationsJobsResponse>
+{
+
+    /**
+     * Constructs a request to retrieve status for a specified operations job
+     */
+    public OperationsJobsRequest(String jobId)
```
yifan-c: if a jobId is a uuid, then the input type here should be uuid.

yifan-c (commented):
not a full review. submitting my comments so far.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    public abstract String operation();
+
+    /**
+     * Specifies the downstream operation to be performed to check if the job is running on the Cassandra node/cluster.
+     * This functionality is provided by the Job when it is asynchronously triggering the job via the {@code JobManager}
+     * For synchronous jobs this should always return false.
+     * @return true if the job is running downstream
+     */
+    public abstract boolean checkInflightJob();
```
yifan-c: nit:

```suggestion
    public abstract boolean isInflight();
```

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
```diff
@@ -0,0 +1,152 @@
...
+        return jobId;
+    }
+
+    /**
+     * Supplier specifying the functionality of the job to be triggered when the job is executed. Subclasses to
+     * provide operation-specific implementations
+     * @return a function with the operation implementation that returns a {@code JobResult}
+     * @throws Exception
+     */
+    public abstract Supplier<OperationsJobResult> jobOperationSupplier() throws Exception;
```
yifan-c: Returning a function seems to complex. 

The equivalent is probably easier to understand. 

```java
    protected abstract OperationsJobResult executeInternal() throws OperationalJobException;
```

Note that the access modifier can be `protected` and prefer a domain specific exception type.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
```diff
@@ -0,0 +1,152 @@
...
+     * For synchronous jobs this should always return false.
+     * @return true if the job is running downstream
+     */
+    public abstract boolean checkInflightJob();
+
+    /**
+     * Execute the job behavior as specified in the operation supplier {@link #jobOperationSupplier()},
+     * while tracking the status of the job's lifecycle.
+     */
+    public void execute()
```
yifan-c: nit: consider adding `final` to forbidden overriding this method, since the concrete logic is implemented in `jobOperationSupplier`

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * Returns if the job execution completes within the provided wait time (in seconds)
+     * @param waitSeconds no. of seconds to wait for the job execution to complete
+     * @return true if the job completed within the specified time
+     */
+    public boolean isResultAvailable(long waitSeconds)
+    {
+        try
+        {
+            return (!latch.await(waitSeconds, TimeUnit.SECONDS)) ? false : true;
```
yifan-c:
```suggestion
            return latch.await(waitSeconds, TimeUnit.SECONDS);
```

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+        catch (final InterruptedException e)
+        {
+            return false;
```
yifan-c: preserve the interrupted flag. 
```java
Thread.currentThread().interrupt();
```

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+    }
```
yifan-c: Not at a fan of this blocking API.

Consider this option. Schedule a future task / timer, and check the result after timer times out. 

```
    public boolean isComplete()
    {
        // of course, not necessary to use CountDownLatch in this case
        return latch.getCount() == 0;
    }

    // at call-site
    Vertx vertx;
    OperationsJob operationsJob;
    vertx.timer(waitSeconds, TimeUnit.SECONDS).map(v -> operationsJob.isComplete())...
```

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJobTracker.java
```diff
@@ -0,0 +1,254 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Tracks and stores the results of long-running jobs running on the sidecar
+ */
+public class OperationsJobTracker extends LinkedHashMap<UUID, OperationsJob>
```
yifan-c: The class can use some simplifications.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJobManager.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Asynchronously submit (and lazily create, via the supplier) the job, if it is not currently being
+     * tracked and is not running downstream. The job is triggered on a separate internal thread-pool.
+     * The job execution failure behavior is tracked within the {@link OperationsJob}.
+     * @param jobId job identifier
+     * @param jobSupplier supplier used to create an instance of the job
+     * @return the instance of the job that is either being tracked or was just submitted
+     */
+    public OperationsJob trySubmitJob(UUID jobId, Supplier<OperationsJob> jobSupplier)
```
yifan-c: Passing a `Supplier` would have correctness issue. (Explained at the end) 
My suggestion is to have this signature instead.

```suggestion
public boolean trySubmitJob(OperationsJob job);
```

1. Use boolean to indicate the submission result.
2. `OperationsJob` already contains jobId. The parameter `UUID jobId` can be omitted.

Now, why there is potential correctness issue. 
Say there are 3 threads submitting the same jobId with different suppliers at the same time. Let's say thread A wins, thread B and C lose. Since the signature returns a `OperationsJob`, there is no way to tell which thread has succeeded. All A, B and C will consider they have created the job successfully. The ambiguity can break the subsequent execution.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/ListOperationsJobsResponse.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Response structure of the list operations jobs API
+ */
+public class ListOperationsJobsResponse
+{
+    private final List<OperationsJobsResponse> jobs;
```
yifan-c: It read confusing to me that the array in the json response body is also named "Response". Can you place the current `OperationsJobsResponse` class in package `org.apache.cassandra.sidecar.common.response.data` and rename the class to `OperationalJobEntry`? In other words, the `ListOperationalJobsResponse` contains an array of `OperationalJobEntry`

server/src/main/java/org/apache/cassandra/sidecar/routes/OperationsJobsHandler.java
```diff
@@ -0,0 +1,120 @@
...
+    public OperationsJobsHandler(InstanceMetadataFetcher metadataFetcher,
+                                 ExecutorPools executorPools,
+                                 CassandraInputValidator validator,
+                                 OperationsJobManager jobManager)
+    {
+        super(metadataFetcher, executorPools, validator);
+        this.jobManager = jobManager;
+    }
+
+    protected Void extractParamsOrThrow(RoutingContext context)
```
yifan-c: `@Override`

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJobManager.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    /**
+     * Fetch the job using its UUID
+     * @param jobId identifier of the job
+     * @return instance of the job or null
+     */
+    public OperationsJob getJobIfExists(UUID jobId)
+    {
+        return jobTracker.get(jobId);
```
yifan-c: It is odd that the `allInflightJobs` is backed by jobView, but `getJobIfExists` reads from jobTracker directly. It makes me wondering that `getJobsView` is unnecessary to have.

server/src/main/java/org/apache/cassandra/sidecar/routes/OperationsJobsHandler.java
```diff
@@ -0,0 +1,120 @@
...
+    {
+        return null;
+    }
+
+    @Override
+    public void handleInternal(RoutingContext context, HttpServerRequest httpRequest, String host, SocketAddress remoteAddress, Void request)
+    {
+        UUID jobUUID = validateJobIdParam(context);
```
yifan-c: nit: rename to `jobId`

server/src/main/java/org/apache/cassandra/sidecar/routes/OperationsJobsHandler.java
```diff
@@ -0,0 +1,120 @@
...
+
+        executorPools.service().executeBlocking(() -> {
+                         OperationsJob job = jobManager.getJobIfExists(jobUUID);
+                         if (job == null)
+                         {
+                             String response = String.format("Unknown job with ID:%s. Please retry the operation.", jobUUID);
+                             logger.info(response);
+                             context.fail(wrapHttpException(HttpResponseStatus.NOT_FOUND, response));
```
yifan-c:
```suggestion
                             throw wrapHttpException(HttpResponseStatus.NOT_FOUND, response);
```

Otherwise, it is going to return job and hit the `onSucesss` handler (unnecessarily).

client-common/src/main/java/org/apache/cassandra/sidecar/common/utils/OperationsJobResult.java
yifan-c: Enum variants name should be in upper case.

yifan-c:
Submitted the comments of the full review. Thanks for the patch.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/ListOperationsJobsRequest.java
arjunashok: I don't have a strong opinion here. The reason I did not choose "Operational" is because it seems to suggest something is "working" or "in-use" and not specifically jobs that are used in Operations context (which seemed more appropriate). Lmk what you think.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/OperationsJobsRequest.java
arjunashok: Done

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/ListOperationsJobsResponse.java
arjunashok: Will do

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/ListOperationsJobsResponse.java
arjunashok: I see your point, but it does help for it to be guaranteed to be globally unique, when we are debugging these jobs across multiple nodes (even though they are currently node-local) to avoid confusion.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/OperationsJobsResponse.java
arjunashok: Ack

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/OperationsJobsResponse.java
arjunashok: Not sure I understand. Are you asking why we're not using `OperationsJobResult` as a member here?  

If so, I did not see a reason to have the hierarchy since all these are attributes of the jobs API response.

client-common/src/main/java/org/apache/cassandra/sidecar/common/utils/OperationsJobResult.java
arjunashok: Ack

client-common/src/main/java/org/apache/cassandra/sidecar/common/utils/OperationsJobResult.java
arjunashok: Why have public attributes, when we have the accessors?

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
arjunashok: I had this renamed to `isRunningDownstream` in my earlier draft. Will update to that.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
arjunashok: Ack

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
arjunashok: Much better :)

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
arjunashok: Ack

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJobManager.java
arjunashok: `allInflightJobs` iterates over all entries, so doing this without holding locks is beneficial to not affect other actions. 

`getJobIfExists` on the other hand is a point lookup.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJobManager.java
arjunashok: I changed this to use a creator function (`Function<UUID, OperationsJob> creator`) that takes UUID argument, to improve clarity on which ID is used for the job.  This should address the correctness issue as the returned job always corresponds to the headerId passed in (or a new UUID when absent).

As for the return-type, we need to job status and id to send the appropriate response.

server/src/main/java/org/apache/cassandra/sidecar/routes/OperationsJobsHandler.java
arjunashok: Ack

server/src/main/java/org/apache/cassandra/sidecar/routes/OperationsJobsHandler.java
arjunashok: Ack

server/src/main/java/org/apache/cassandra/sidecar/routes/OperationsJobsHandler.java
arjunashok: Ack

client-common/src/main/java/org/apache/cassandra/sidecar/common/utils/OperationsJobResult.java
frankgh: coding style conventions for Cassandra

frankgh: https://cassandra.apache.org/_/development/code_style.html

```
Boilerplate

Prefer public final fields to private fields with getters (but prefer encapsulating behavior in "real" methods to either).


```

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
arjunashok: I like this idea. Will give it a shot.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
arjunashok: I see the concern, but in this case, we will have - 
1) handler of long-running job invoking `trySubmitJob` within `executorPools.service().executeBlocking(()`
2) Job execution/submission done within `executorPools.internal().runBlocking()`

So we have a multi-level indirection. For vertx to wait for the execution to complete, we would need some form of Future chaining to detect completion of the inner most execution thread. 

We will have to weigh the complexity of that to the existing implementation. Thoughts?

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJobTracker.java
frankgh: +1 we don't need a full blown map here. Also, I think internally we should keep track of Futures

arjunashok: @yifan-c  Can you elaborate on simplifications?

Regarding the need for map, we already had a discussion in this thread about this - the LRU eviction policy and thread-safety is driving this need. https://github.com/apache/cassandra-sidecar/pull/139#discussion_r1830081175. 

>  I think internally we should keep track of Futures

Not sure I understand, there are no Futures referenced in this class. Can you elaborate?

nvharikrishna:
I see operations and jobs coming together often in this PR. Small question on terminology being used. Is a job seen as a set of operations? Or are they the same?

arjunashok:
> I see operations and jobs coming together often in this PR. Small question on terminology being used. Is a job seen as a set of operations? Or are they the same?

So, the term "OperationsJob" or "OperationalJob" in the context of this PR refers to a long-running sidecar functionality to perform operational tasks for the node, such as `decommission`, `repair`, `rebuild` etc. Such jobs, when initiated, would return an identifier that can be used by the client to poll for completion.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/ListOperationsJobsRequest.java
yifan-c: "Operational" also describes something that is related to operations.

> of, relating to, or based on operations

See https://www.merriam-webster.com/dictionary/operational

It is confusing to see two nouns putting together. Looks like Hari has the same confusion. 

Please update the name for clarity.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/OperationsJobsResponse.java
yifan-c: `OperationsJobsResponse.status.operation == OperationsJobsResponse.operation` and `OperationsJobsResponse.status.reason == OperationsJobsResponse.reason`. 

They are duplicated.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationsJob.java
yifan-c: `isRunning` also sounds good to me. But without the `Downstream` part.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/ListOperationsJobsRequest.java
arjunashok: sounds reasonable. "Operational" it is.

server/src/main/java/org/apache/cassandra/sidecar/job/Job.java
arjunashok: Addressed in latest revision

server/src/main/java/org/apache/cassandra/sidecar/job/JobTracker.java
arjunashok: Discussed offline. Simplified tracker API in latest revision

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/OperationsJobsResponse.java
arjunashok: Latest revision removes the need for status and reason as they are inferred from the Future state which is cached.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/data/OperationalJobsEntry.java
```diff
@@ -0,0 +1,48 @@
...
+package org.apache.cassandra.sidecar.common.response.data;
+
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Structure of the operational job instance within the list operational jobs API response
+ */
+public class OperationalJobsEntry
```
frankgh: is this class needed?

server/src/main/java/org/apache/cassandra/sidecar/routes/ListOperationalJobsHandler.java
```diff
@@ -0,0 +1,69 @@
...
+        return null;
+    }
+
+    @Override
+    protected void handleInternal(RoutingContext context, HttpServerRequest httpRequest, String host, SocketAddress remoteAddress, Void request)
+    {
+        ListOperationalJobsResponse listResponse = new ListOperationalJobsResponse();
+        jobManager.allInflightJobs()
+                  .stream()
+                  .map(job -> new OperationalJobResponse(job.jobId, RUNNING, job.name(), ""))
```
frankgh: can the reason be null here? or do we have some constraint that it Neds to be an empty string?

server/src/main/java/org/apache/cassandra/sidecar/routes/OperationalJobHandler.java
```diff
@@ -0,0 +1,120 @@
...
+        String reason = null;
+        switch(status)
+        {
+            case SUCCEEDED:
+                context.response().setStatusCode(HttpResponseStatus.OK.code()); break;
+            case CREATED:
+            case RUNNING:
+                context.response().setStatusCode(HttpResponseStatus.ACCEPTED.code()); break;
+            case FAILED:
+                reason = job.asyncResult().cause().getMessage();
```
frankgh: do we need to send a status code here?

frankgh (commented):
Looking good . I've left a few remaining comments

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/data/OperationalJobsEntry.java
arjunashok: Not needed. WIll remove.

server/src/main/java/org/apache/cassandra/sidecar/routes/ListOperationalJobsHandler.java
arjunashok: Yes. Will update.

server/src/main/java/org/apache/cassandra/sidecar/routes/OperationalJobHandler.java
arjunashok: Added

frankgh: does 200 make sense here?

arjunashok: Yes, since 200 is the status of the (current) job API request, which succeeded. The response body has the status of the job being queried which is `FAILED`.

yifan-c: yep. 200 to indicate the request is successful. The status of the job is expressed in the response body.

frankgh (approved):
+1 Thanks for working through all the feedback. This is a key piece in the coming features for Sidecar!

PR #143, opened 2024-10-30
CASSANDRASC-155: Fix dangling restore jobs

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSANDRASC-155

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/db/RestoreRange.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/db/RestoreRange.java
@@ -281,7 +281,7 @@ public class RestoreRange
 
         if (tracker.restoreJob().hasExpired(System.currentTimeMillis()))
         {
-            return RestoreRangeTask.failed(RestoreJobExceptions.ofFatal("Restore job has expired at " + tracker.restoreJob().expireAt,
+            return RestoreRangeTask.failed(RestoreJobExceptions.ofFatal("Restore job has expired at " + tracker.restoreJob().expireAt.toInstant(),
                                                                         this, null), this);
         }
 
--- a/server/src/main/java/org/apache/cassandra/sidecar/db/RestoreRange.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/db/RestoreRange.java
@@ -281,7 +281,7 @@ public class RestoreRange
 
         if (tracker.restoreJob().hasExpired(System.currentTimeMillis()))
         {
-            return RestoreRangeTask.failed(RestoreJobExceptions.ofFatal("Restore job has expired at " + tracker.restoreJob().expireAt.toInstant(),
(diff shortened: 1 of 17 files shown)


server/src/test/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscovererTest.java
```diff
@@ -290,6 +291,33 @@ void testCleanupJobIdsByDay()
...
+    void testAdjustRecencyDays()
+    {
+        when(sidecarSchema.isInitialized()).thenReturn(true);
+        assertThat(loop.jobDiscoveryRecencyDays()).isEqualTo(10);
+        loop.registerPeriodicTaskExecutor(executor);
+
+        executeBlocking();
+
+        assertThat(loop.jobDiscoveryRecencyDays())
+        .describedAs("Recency days is adjusted to 1 since there is no jobs running")
```
frankgh:
```suggestion
        .describedAs("Recency days is adjusted to 1 since there are no jobs running")
```

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreRange.java
```diff
@@ -463,11 +469,11 @@ public long estimatedSpaceRequiredInBytes()
 
     public String shortDescription()
     {
-        return "StartToken: " + startToken + ", EndToken: " + endToken +
-               ", SliceId: " + sliceId +
-               ", Key: " + sliceKey() +
-               ", Bucket: " + sliceBucket() +
-               ", Checksum: " + sliceChecksum();
+        return "RestoreRange{" +
```
frankgh, line 472: should we preserve the checksum?

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreProcessor.java
```diff
@@ -170,9 +169,6 @@ public void execute(Promise<Void> promise)
                 break;
             }
 
-            Preconditions.checkState(range.canProduceTask(),
```
frankgh, line 173: do we not need this check anymore?

frankgh (commented):
Looks good, I have some minor comments

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreRange.java
yifan-c: I removed checksum as it is not helpful and verbose. The startToken and endToken are removed for the same reason. The tokens are part of the sliceKey already.

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreProcessor.java
yifan-c: It is checked in `org.apache.cassandra.sidecar.db.RestoreRange#toAsyncTask` already. The check here is redundant and disruptive.

frankgh (approved):
+1

PR #128, opened 2024-06-27
CASSANDRASC-137: Bump AWS SDK version to 2.26.10

yifan-c:
Patch by Yifan Cai; Reviewed by Francisco Guerrero for CASSANDRASC-137

PR #144, opened 2024-11-02
CASSSIDECAR-151: Adds sidecar endpoint for node decommissioning operation

arjunashok:
These changes are part of the effort to introduce bespoke Sidecar APIs to support key operational functionality currently managed through nodetool commands. 

These changes specifically incorporate the node decommission functionality in a sidecar endpoint.

### Update 12/19
- Rebased changes from job management framework after CASSSIDECAR-150 was merged.

### Changes
- Incorporates Decommission API `api/v1/cassandra/node/decommission`
- Leverages the job-management framework (CASSANDRASC-150) via a concrete `DecommissionJob`, which invokes the JMX API to trigger `decommission` operation.
- Unit tests and integration tests
- SidecarClient changes and tests for the new API


client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -111,6 +111,10 @@ public final class ApiEndpointsV1
 
     public static final String CONNECTED_CLIENT_STATS_ROUTE = API_V1 + CASSANDRA + "/stats/connected-clients";
 
+    public static final String LIST_JOBS_ROUTE = API_V1 + CASSANDRA + "/jobs";
+    public static final String JOB_STATUS_ROUTE = API_V1 + CASSANDRA + "/jobs/" + JOB_ID_PATH_PARAM + "/status";
+    public static final String NODE_DECOMMISSION_ROUTE = API_V1 + CASSANDRA + "/node/decommission";
```
skoppu22: We are using `management` in the prefix, for example /api/v1/cassandra/management/node/decommission for JMX and nodetool operations. IIRC, we discussed this during Sidecar's JMX APIs design meeting

arjunashok: Right, and I feel these are different from those (and we arrived at this design during the nodetool APIs review), as they are specific to the administrative node operations like decommission, repair, rebuild etc. We can converge on the URI offline if you'd like.

yifan-c:
Please update `CHANGES.txt`

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -213,4 +214,38 @@ public void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tabl
...
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean isDecommissioning()
+    {
+        StorageJmxOperations ssProxy = jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME);
+        String nodeOperationMode = ssProxy.getOperationMode();
+        // The following check is equivalent to the operation-mode checks within Cassandra to determine if a
+        // decommission operation is in progress
+        return nodeOperationMode.equals("LEAVING") || nodeOperationMode.equals("DECOMMISSION_FAILED");
```
yifan-c: Cassandra code use this condition to check if the node is decommissioning. Why do we differ from the implementation? 

```java
    public boolean isDecommissioning()
    {
        return operationMode() == LEAVING;
    }
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -213,4 +214,38 @@ public void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tabl
...
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean isDecommissioning()
+    {
+        StorageJmxOperations ssProxy = jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME);
+        String nodeOperationMode = ssProxy.getOperationMode();
```
yifan-c: How about extract out into a method to retrieve the operationMode? It is used in other locations.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -213,4 +214,38 @@ public void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tabl
...
+    @Override
+    public OperationalJobStatus decommission(boolean force)
+    {
+        StorageJmxOperations ssProxy = jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME);
+        String mode = ssProxy.getOperationMode();
+        switch(mode)
+        {
+            case "LEAVING":
+            case "DECOMMISSION_FAILED":
+                return OperationalJobStatus.RUNNING;
```
yifan-c: Can you comment why `DECOMMISSION_FAILED` is considered `RUNNING`?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -213,4 +214,38 @@ public void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tabl
...
+     * {@inheritDoc}
+     */
+    @Override
+    public OperationalJobStatus decommission(boolean force)
+    {
+        StorageJmxOperations ssProxy = jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME);
+        String mode = ssProxy.getOperationMode();
+        switch(mode)
+        {
+            case "LEAVING":
```
yifan-c: nit: define enum for the modes.

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -126,6 +126,7 @@ public final class ApiEndpointsV1
     public static final String LIST_OPERATIONAL_JOBS_ROUTE = API_V1 + CASSANDRA + OPERATIONAL_JOBS;
     public static final String OPERATIONAL_JOB_ROUTE = API_V1 + CASSANDRA + PER_OPERATIONAL_JOB;
 
+    public static final String NODE_DECOMMISSION_ROUTE = API_V1 + CASSANDRA + "/node/decommission";
```
yifan-c: I am trying to understand the role the url segment `/node` plays here. It seems as clear as `/api/v1/cassandra/decommission`, but this is shorter w/o the `/node` segment. 
Or, if we want to follow the `resource/identifier` structure strictly, `/operations` is a better placement for `/node`, i.e. `/api/v1/cassandra/operations/decommission`, where `decommission` is viewed as an identifier of the operation.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/NodeDecommissionResponse.java
```diff
@@ -0,0 +1,87 @@
...
+ * Response structure of the jobs status API
+ */
+public class NodeDecommissionResponse
+{
+    private final UUID jobId;
+    private final OperationalJobStatus status;
+
+    private final String instance;
+
+    private final String reason;
```
yifan-c: nit: remove empty lines

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -664,6 +665,16 @@ public CompletableFuture<ListOperationalJobsResponse> listOperationalJobs(Sideca
...
 
+    /**
+     * Executes the node decommission request using the default retry policy and configured selection policy
+     *
+     * @return a completable future of the jobs list
+     */
+    public CompletableFuture<NodeDecommissionResponse> nodeDecommission()
+    {
+        return executor.executeRequestAsync(requestBuilder().nodeDecommissionRequest().build());
+    }
```
yifan-c: Given decommission is per node, the client should want to talk to a specific instance, i.e. taking `SidecarInstance` as parameter. 

Please take `org.apache.cassandra.sidecar.client.SidecarClient#operationalJobs` for example.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/exceptions/OperationalJobException.java
```diff
@@ -36,7 +36,8 @@ public static OperationalJobException wraps(Throwable throwable)
         }
         else
         {
-            return new OperationalJobException(throwable.getMessage(), throwable);
+            String msg = throwable.getCause() != null ? throwable.getCause().getMessage() : throwable.getMessage();
+            return new OperationalJobException(msg, throwable);
```
yifan-c: Would it look odd to use the message of the throwable's cause, but set the throwable as the cause of `OperationalJobException`? The message of the cause of throwable is going to repeat in the stack trace.

server/src/main/java/org/apache/cassandra/sidecar/job/DecommissionJob.java
```diff
@@ -0,0 +1,78 @@
...
+    {
+        LOGGER.info("Executing decommission operation. jobId={}", jobId);
+        return storageOperations.decommission(isForce);
+    }
+
+    @Override
+    public void close()
+    {
+        super.close();
+    }
```
yifan-c: seems redundant, if there is no concrete close action.

server/src/main/java/org/apache/cassandra/sidecar/job/DecommissionJob.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+
+    protected OperationalJobStatus executeInternal() throws OperationalJobException
+    {
+        LOGGER.info("Executing decommission operation. jobId={}", jobId);
+        return storageOperations.decommission(isForce);
```
yifan-c: The method signature defines that the thrown exception is `OperationalJobException`. There is no catch block to convert the exceptions to `OperationalJobException`.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
```diff
@@ -95,12 +94,18 @@ public boolean isStale(long referenceTimestampInMillis, long ttlInMillis)
...
     }
 
+    /**
+     * The concrete-job-specific implementation to determine if the job is running on the Cassandra node.
+     * @return true if the job is running on the Cassandra node. For example, node decommission is tracked by the
+     * operationMode exposed from Cassandra.
+     */
+    public abstract boolean isRunningOnCassandra();
+
+
```
yifan-c: nit: usually keep just one empty line after the end of method

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
```diff
@@ -44,7 +43,7 @@ public abstract class OperationalJob implements Task<Void>
     // use v1 time-based uuid
     public final UUID jobId;
 
-    private final Promise<Void> executionPromise;
+    private final Promise<OperationalJobStatus> executionPromise;
```
yifan-c: I am not convinced of having `Promise<OperationalJobStatus>`. The future state is already good enough to map to `OperationalJobStatus`. 

From the semantics, status should not be read from future. The callsite should be able to query the status directly, w/o extracting from future.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
```diff
@@ -193,11 +199,41 @@ public void execute(Promise<Void> promise)
...
+
+
+
+
         try
         {
             // Blocking call to perform concrete job-specific execution, returning the status
-            executeInternal();
+            OperationalJobStatus result = executeInternal();
+            executionPromise.tryComplete(result);
```
yifan-c: The change does not look correct.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
```diff
@@ -76,22 +81,33 @@ public OperationalJob getJobIfExists(UUID jobId)
...
 
         // New job is submitted for all cases when we do not have a corresponding downstream job
-        return jobTracker.computeIfAbsent(job.jobId, jobId -> job);
+        jobTracker.computeIfAbsent(job.jobId, jobId -> {
+            internalExecutorPool.executeBlocking(() -> {
+                job.execute(Promise.promise());
+                return null;
+            });
+            return job;
+        });
```
yifan-c: Although `executeBlocking` with promise is deprecated, let's still use it for the succinct code it produces here. We can migrate away from the deprecated method all together in future refactoring. 

```suggestion
        jobTracker.computeIfAbsent(job.jobId, jobId -> {
            internalExecutorPool.executeBlocking(job::execute);
            return job;
        });
```

server/src/main/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandler.java
```diff
@@ -0,0 +1,132 @@
...
+    {
+        return null;
+    }
+
+    protected void handleInternal(RoutingContext context, HttpServerRequest httpRequest, String host, SocketAddress remoteAddress, Void request)
+    {
+        StorageOperations operations = metadataFetcher.delegate(host).storageOperations();
+        isForce = parseBooleanQueryParam(context.request(), "force", false);
+
+        OperationalJob job = new DecommissionJob(UUIDs.timeBased(), operations, isForce);
```
yifan-c: `isForce` can be local variable, instead of being an instance variable.

server/src/main/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandler.java
```diff
@@ -0,0 +1,132 @@
...
+        job.asyncResult(executorPools.service(),
+                        Duration.of(config.operationalJobExecutionMaxWaitTimeInMillis(), ChronoUnit.MILLIS))
+           .onSuccess(f -> sendResponse(context, job, request, remoteAddress, host))
+           .onFailure(cause -> processFailure(cause, context, host, remoteAddress, request));
+    }
+
+    private void sendResponse(RoutingContext context, OperationalJob job, Void request, SocketAddress remoteAddress, String host)
+    {
+        OperationalJobStatus jobStatus = job.status();
+        logger.info("Job completion status={}, request={}, remoteAddress={}, instance={}",
```
yifan-c: remove commas.
```suggestion
        logger.info("Job completion status={} request={} remoteAddress={} instance={}",
```

server/src/main/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandler.java
```diff
@@ -0,0 +1,132 @@
...
+        String reason = null;
+        switch(jobStatus)
+        {
+            case SUCCEEDED:
+                context.response().setStatusCode(HttpResponseStatus.OK.code());
+                break;
+            case FAILED:
+                reason = job.asyncResult().cause().getMessage();
+                context.response().setStatusCode(HttpResponseStatus.INTERNAL_SERVER_ERROR.code());
+                break;
```
yifan-c: Should the status of failed job be `OK` too? 

It is the job that fails. The result should be indicated in the response payload. The http request is successful, i.e. 200. 

There is almost the same logic in `org.apache.cassandra.sidecar.routes.OperationalJobHandler`

server/src/test/integration/org/apache/cassandra/sidecar/common/CQLSessionProviderTest.java
```diff
@@ -60,6 +60,7 @@ void testCqlSessionProviderWorksAsExpected(VertxTestContext context, CassandraTe
     throws Exception
     {
         UpgradeableCluster cluster = cassandraTestContext.cluster();
+
```
yifan-c: nit: unrelated change

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/NodeDecommissionResponse.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.UUID;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.cassandra.sidecar.common.data.OperationalJobStatus;
+
+/**
+ * Response structure of the jobs status API
+ */
+public class NodeDecommissionResponse
```
yifan-c: Is it a dup of `org.apache.cassandra.sidecar.common.response.OperationalJobResponse`? Can we only use `OperationalJobResponse`? 

The response payload should be the same between the endpoint that triggers an operation and the endpoint that queries the operation status.

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+/**
+ * Test the node decommission endpoint with cassandra container.
+ */
+@ExtendWith(VertxExtension.class)
+public class NodeDecommissionIntegrationTest extends IntegrationTestBase
+{
+    private static final String DECOMMISSION_FAILED_MESSAGE = "Failed to decommission node";
+
+    @CassandraIntegrationTest(nodesPerDc = 5)
```
yifan-c: can you use only 2 nodes cluster for this and other test cases in the class? I think it is good enough for testing decommission. We want to create clusters with less nodes to use less testing resources.

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
```diff
@@ -0,0 +1,176 @@
...
+        pollStatusForState(context, jobId[0], FAILED, DECOMMISSION_FAILED_MESSAGE);
+        context.awaitCompletion(2, TimeUnit.MINUTES);
+    }
+
+    private void pollStatusForState(VertxTestContext context,
+                                    String uuid,
+                                    OperationalJobStatus expectedStatus,
+                                    String expectedReason)
+    {
+
```
yifan-c: nit: remove the empty line.

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
```diff
@@ -0,0 +1,176 @@
...
+        int attempts = 10;
+        String status = "/api/v1/cassandra/operational-jobs/" + uuid;
+        AtomicBoolean stateReached = new AtomicBoolean(false);
+        logger.info("Job Stats Attempt:" + attempts);
+        final int[] counter = {0};
```
yifan-c: how about just use `AtomicInteger`? We do not need the atomicity property here, but its API is way more expressive than `counter[0]++`.

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
```diff
@@ -0,0 +1,176 @@
...
+    @CassandraIntegrationTest(nodesPerDc = 5, network = true, buildCluster = false)
+    void decommissionNodeWithFailure(VertxTestContext context,
+                                     ConfigurableCassandraTestContext cassandraTestContext) throws InterruptedException
+    {
+        BBHelperDecommissionNode.reset();
+        cassandraTestContext.configureAndStartCluster(builder -> {
+            builder.withInstanceInitializer(BBHelperDecommissionNode::install);
+        });
+
+        final String[] jobId = new String[1];
```
yifan-c: AtomicReference<String>?

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
```diff
@@ -0,0 +1,176 @@
...
+    private void pollStatusForState(VertxTestContext context,
+                                    String uuid,
+                                    OperationalJobStatus expectedStatus,
+                                    String expectedReason)
+    {
+
+        int attempts = 10;
+        String status = "/api/v1/cassandra/operational-jobs/" + uuid;
+        AtomicBoolean stateReached = new AtomicBoolean(false);
+        logger.info("Job Stats Attempt:" + attempts);
```
yifan-c: use placeholder `{}` for logging messages

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
```diff
@@ -0,0 +1,176 @@
...
+                                                      }
+                                                      logger.info("Request completed");
+                                                  })));
+            if (stateReached.get() || counter[0] == finalAttempts)
+            {
+                vertx.cancelTimer(id);
+                assertThat(stateReached.get()).isTrue();
+                context.completeNow();
+            }
+        });
```
yifan-c: This looks odd. 

`WebClient` sends requests asyncly. `testWithClient` returns immediately. It checks the condition `stateReached.get() || counter[0] == finalAttempts` prematurely. 


I would simplify the method to this with blocking client to make it easier to script the repeated check logic. Note that `VertxTestContext` is no longer needed when using the client blockingly.  

```java
    private void pollStatusForState(String uuid,
                                    OperationalJobStatus expectedStatus,
                                    String expectedReason)
    {
        int attempts = 10;
        String status = "/api/v1/cassandra/operational-jobs/" + uuid;
        AtomicBoolean stateReached = new AtomicBoolean(false);
        logger.info("Job Stats Attempt: {}", attempts);
        AtomicInteger counter = new AtomicInteger(0);
        loopAssert(30, () -> {
            counter.incrementAndGet();
            // todo: create a helper method in the base class to get response in the blocking manner
            HttpResponse<Buffer> resp = client.get(server.actualPort(), "127.0.0.1", status)
                                              .send().toCompletionStage().toCompletableFuture().get();
            logger.info("Success Status Response code: {}", resp.statusCode());
            logger.info("Status Response: {}", resp.bodyAsString());
            if (resp.statusCode() == HttpResponseStatus.OK.code())
            {
                stateReached.set(true);
                OperationalJobResponse jobStatusResp = resp.bodyAsJson(OperationalJobResponse.class);
                assertThat(jobStatusResp.jobId()).isEqualTo(UUID.fromString(uuid));
                assertThat(jobStatusResp.status()).isEqualTo(expectedStatus);
                assertThat(jobStatusResp.reason()).isEqualTo(expectedReason);
                assertThat(jobStatusResp.operation()).isEqualTo("decommission");
            }
            else
            {
                assertThat(resp.statusCode()).isEqualTo(HttpResponseStatus.ACCEPTED.code());
                OperationalJobResponse jobStatusResp = resp.bodyAsJson(OperationalJobResponse.class);
                assertThat(jobStatusResp.jobId()).isEqualTo(UUID.fromString(uuid));
                // assertThat(jobStatusResp).isNull();
            }
            logger.info("Request completed");
            assertThat(stateReached.get()).isTrue();
        });
    }
```

server/src/test/java/org/apache/cassandra/sidecar/job/OperationalJobManagerTest.java
```diff
@@ -55,26 +54,29 @@
...
     {
         vertx = Vertx.vertx();
+        executorPool = new ExecutorPools(vertx, new ServiceConfigurationImpl());
         MockitoAnnotations.openMocks(this);
-        ServiceConfiguration mockServiceConfig = mock(ServiceConfiguration.class);
-        when(mockConfig.serviceConfiguration()).thenReturn(mockServiceConfig);
-        when(mockServiceConfig.operationalJobExecutionMaxWaitTimeInMillis()).thenReturn(5000L);
+//        ServiceConfiguration mockServiceConfig = mock(ServiceConfiguration.class);
+//        when(mockConfig.serviceConfiguration()).thenReturn(mockServiceConfig);
+//        when(mockServiceConfig.operationalJobExecutionMaxWaitTimeInMillis()).thenReturn(5000L);
```
yifan-c: Update?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
arjunashok: Looks like this was changed in Cassandra recently, will update.
https://github.com/apache/cassandra/commit/c679b4730332ef67102ec7e47db891be2f8feabf

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
arjunashok: Addressed to make it blocking.

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
arjunashok: Addressed after offline sync in latest commits.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
arjunashok: Same reason as above. THis used to be this way in C*. Addressed.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
arjunashok: Addressed after offline sync in latest commits

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/exceptions/OperationalJobException.java
arjunashok: Updated to use the corresponding cause with the message.

server/src/main/java/org/apache/cassandra/sidecar/job/DecommissionJob.java
arjunashok: Yep, not needed here. The `execute` method is the one that wraps all execution exceptions into `OperationalJobException`

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
arjunashok: Addressed based on offline sync

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
arjunashok: Thi change goes away with the `Void` execution promise

server/src/main/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandler.java
arjunashok: Updated. As discussed offline, we will need to document the reasoning to treat job errors as OK response is because the API is considering the "job trigger/execution" as the resource, which completed.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -213,4 +213,34 @@ public void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tabl
...
+//     */
+//    @Override
+//    public boolean isDecommissioning()
+//    {
+//        StorageJmxOperations ssProxy = jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME);
+//        String nodeOperationMode = ssProxy.getOperationMode();
+//        // The following check is equivalent to the operation-mode checks within Cassandra to determine if a
+//        // decommission operation is in progress
+//        return nodeOperationMode.equals("LEAVING");
+//    }
```
yifan-c: Is it to be removed?

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -126,6 +126,7 @@ public final class ApiEndpointsV1
     public static final String LIST_OPERATIONAL_JOBS_ROUTE = API_V1 + CASSANDRA + OPERATIONAL_JOBS;
     public static final String OPERATIONAL_JOB_ROUTE = API_V1 + CASSANDRA + PER_OPERATIONAL_JOB;
 
+    public static final String NODE_DECOMMISSION_ROUTE = API_V1 + CASSANDRA + "/operations/decommission";
```
yifan-c, line 128: nit: move to line #128. Basically, remove the empty line before this statement and add the empty line after it.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/exceptions/OperationalJobException.java
```diff
@@ -36,7 +36,8 @@ public static OperationalJobException wraps(Throwable throwable)
         }
         else
         {
-            return new OperationalJobException(throwable.getMessage(), throwable);
+            Throwable failingCause = throwable.getCause() != null ? throwable.getCause() : throwable;
+            return new OperationalJobException(failingCause.getMessage(), failingCause);
```
yifan-c: It is still unclear about trimming here. Conservatively thinking, retaining more information is usually preferred. Can you explain why skipping `throwable` and using its cause?

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDecommissionJob.java
```diff
@@ -0,0 +1,101 @@
...
+        this.storageOperations = storageOps;
+        this.isForce = isForce;
+    }
+
+    @Override
+    public boolean isRunningOnCassandra()
+    {
+        String operationMode = storageOperations.getOperationMode();
+        return operationMode.equals("LEAVING") || operationMode.equals("DECOMMISSIONED");
+
```
yifan-c: nit: remove empty line

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDecommissionJob.java
```diff
@@ -0,0 +1,101 @@
...
+        else
+        {
+            return super.status();
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+
```
yifan-c: remove empty line

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDecommissionJob.java
```diff
@@ -0,0 +1,101 @@
...
+    protected void executeInternal()
+    {
+        if (isRunningOnCassandra())
+        {
+            LOGGER.info("Not executing job as an ongoing or completed decommission operation was found jobId={}", jobId);
```
yifan-c: how about retaining the `operationMode` value read during calling `isRunningOnCassandra()` (as an instance variable) and log it here?

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
```diff
@@ -179,10 +183,8 @@ public Future<Void> asyncResult(TaskExecutorPool executorPool, Duration waitTime
 
     /**
      * OperationalJob body. The implementation is executed in a blocking manner.
-     *
-     * @throws OperationalJobException OperationalJobException that wraps job failure
      */
-    protected abstract void executeInternal() throws OperationalJobException;
+    protected abstract void executeInternal();
```
yifan-c, line 192: 👍  the `throws` is unnecessary. It is handled by `execute()` already.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/StorageOperations.java
```diff
@@ -102,4 +102,16 @@ default void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tab
     {
         outOfRangeDataCleanup(keyspace, table, 1);
     }
+
+    /**
+     * @return the operation-mode of the Cassandra instance
+     */
+    String getOperationMode();
```
yifan-c: we could consider cache the operationMode in the future if it is queried a lot. It is not necessary to be addressed for now.

server/src/main/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandler.java
```diff
@@ -0,0 +1,134 @@
...
+        catch (OperationalJobConflictException oje)
+        {
+            logger.error("Conflicting job encountered. reason={}", oje.getMessage());
+            context.fail(wrapHttpException(HttpResponseStatus.CONFLICT, oje.getMessage(), oje.getCause()));
+        }
+
+        // Get the result, waiting for the specified wait time for result
+        job.asyncResult(executorPools.service(),
+                        Duration.of(config.operationalJobExecutionMaxWaitTimeInMillis(), ChronoUnit.MILLIS))
+        .onComplete(v -> sendResponse(context, job, request, remoteAddress, host));
```
yifan-c: align with the the dot at `job.asyncResult...`

server/src/main/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandler.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    private void sendResponse(RoutingContext context, OperationalJob job, Void request, SocketAddress remoteAddress, String host)
```
yifan-c: What prevents to reusing the code in `org.apache.cassandra.sidecar.routes.OperationalJobHandler`?

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    private void pollStatusForState(String uuid,
+                                    OperationalJobStatus expectedStatus,
+                                    String expectedReason)
+    {
+        int attempts = 10;
+        String status = "/api/v1/cassandra/operational-jobs/" + uuid;
+        AtomicBoolean stateReached = new AtomicBoolean(false);
+        logger.info("Job Stats Attempt: {}", attempts);
```
yifan-c: maybe you want to place it in the `loopAssert`? Right now, it only logs once with the `attempts` value 10

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
```diff
@@ -0,0 +1,134 @@
...
+        AtomicInteger counter = new AtomicInteger(0);
+        loopAssert(30, () -> {
+            counter.incrementAndGet();
+            // todo: create a helper method in the base class to get response in the blocking manner
```
yifan-c: Are you going to address this todo?

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
```diff
@@ -0,0 +1,134 @@
...
+                                       .send(context.succeeding(response -> {
+                                           logger.info("Response Status:" + response.statusCode());
+                                           OperationalJobResponse decommissionResponse = response.bodyAsJson(OperationalJobResponse.class);
+                                           assertThat(decommissionResponse.status()).isEqualTo(RUNNING);
+                                           jobId[0] = String.valueOf(decommissionResponse.jobId());
+                                       })));
+        Uninterruptibles.sleepUninterruptibly(10, TimeUnit.SECONDS);
+        pollStatusForState(jobId[0], SUCCEEDED, null);
+        context.completeNow();
+        context.awaitCompletion(2, TimeUnit.MINUTES);
```
yifan-c: `context.awaitCompletion` is no longer necessary since it is blocking style.

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
```diff
@@ -0,0 +1,134 @@
...
+        final String[] jobId = new String[1];
+        String testRoute = "/api/v1/cassandra/operations/decommission?force=true";
+        testWithClient(client -> client.put(server.actualPort(), "127.0.0.1", testRoute)
+                                       .send(context.succeeding(response -> {
+                                           logger.info("Response Status:" + response.statusCode());
+                                           OperationalJobResponse decommissionResponse = response.bodyAsJson(OperationalJobResponse.class);
+                                           assertThat(decommissionResponse.status()).isEqualTo(RUNNING);
+                                           jobId[0] = String.valueOf(decommissionResponse.jobId());
+                                       })));
+        Uninterruptibles.sleepUninterruptibly(10, TimeUnit.SECONDS);
```
yifan-c, line 64: nit: sleep is unnecessary since `pollStatusForState` retries.

server/src/test/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandlerTest.java
```diff
@@ -0,0 +1,216 @@
...
+    @Test
+    void testDecommissionConflict(VertxTestContext context)
+    {
+        when(mockStorageOperations.getOperationMode()).thenReturn("LEAVING");
+        WebClient client = WebClient.create(vertx);
+        String testRoute = "/api/v1/cassandra/operations/decommission";
+        client.put(server.actualPort(), "127.0.0.1", testRoute)
+              .expect(ResponsePredicate.SC_CONFLICT)
+              .send(context.succeeding(response -> {
+                  assertThat(response.statusCode()).isEqualTo(CONFLICT.code());
```
yifan-c, line 166: nit: add the assertion for the returned jobId, so that client can use this jobId to track the existing inflight job.

server/src/main/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandler.java
```diff
@@ -0,0 +1,134 @@
...
+        OperationalJob job = new NodeDecommissionJob(UUIDs.timeBased(), operations, isForce);
+        try
+        {
+            jobManager.trySubmitJob(job);
+        }
+        catch (OperationalJobConflictException oje)
+        {
+            logger.error("Conflicting job encountered. reason={}", oje.getMessage());
+            context.fail(wrapHttpException(HttpResponseStatus.CONFLICT, oje.getMessage(), oje.getCause()));
+        }
```
yifan-c: Can you include the jobId of the conflicting job in the response json payload?

server/src/test/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandlerTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+    /**
+     * Test guice module for Node Decommission handler tests
+     */
+    static class NodeDecommissionTestModule extends AbstractModule
+    {
+
+        public NodeDecommissionTestModule()
+        {
+            MockitoAnnotations.openMocks(this);
```
yifan-c: This openMocks has no effect. Can you remove it?

server/src/test/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandlerTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+    @Mock
+    static
+    StorageOperations mockStorageOperations;
+
+    @BeforeEach
+    void before() throws InterruptedException
+    {
+        Injector injector;
+        MockitoAnnotations.openMocks(this);
```
yifan-c: The test code is responsible for closing the opened mocks when completing. Since there is only a single field (i.e. mockStorageOperations) gets mock, consider mocking it the simple way?

server/src/test/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandlerTest.java
```diff
@@ -0,0 +1,216 @@
...
+ */
+@ExtendWith(VertxExtension.class)
+public class NodeDecommissionHandlerTest
+{
+    static final Logger LOGGER = LoggerFactory.getLogger(NodeDecommissionHandlerTest.class);
+    Vertx vertx;
+    Server server;
+
+    @Mock
+    static
```
yifan-c: remove `static`. It is updated by each test case.

arjunashok: This is made static as the Guice `InstanceMetadata` provider returns the mock.

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
arjunashok: This was added to ensure the decommission API response (and jobId) is available for the subsequent status polls. Its set to `10s` so it passes deterministically.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/exceptions/OperationalJobException.java
arjunashok: yep, simply wrapping the throwable as OperationalJobException should suffice

server/src/main/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandler.java
arjunashok: Addressed

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
arjunashok: Leaving this as-is for now, since this is the only usage of the blocking call, and since it's a simple `.get()` on the future. We can abstract as needed when it's reused.

Will remove the comment, if needed

adapters/cassandra41/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra41/Cassandra41StorageOperations.java
```diff
@@ -34,6 +37,8 @@
  */
 public class Cassandra41StorageOperations extends CassandraStorageOperations
 {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(Cassandra41StorageOperations.class);
```
frankgh: seems unused, can we remove it?

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/NodeDecommissionRequest.java
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public NodeDecommissionRequest()
+    {
+        super(ApiEndpointsV1.NODE_DECOMMISSION_ROUTE);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public HttpMethod method()
```
frankgh, line 42: let's make sure you add override annotations

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/StorageOperations.java
frankgh: I don't think we can cache the value because it can change during the lifecycle of the C* process. also this method should be `operationMode`

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDecommissionJob.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public OperationalJobStatus status()
+    {
+        String operationMode = storageOperations.getOperationMode();
+
+        if (operationMode.equals("LEAVING"))
```
frankgh: use switch instead?

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDecommissionJob.java
```diff
@@ -0,0 +1,99 @@
...
+        else
+        {
+            return super.status();
+        }
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    protected void executeInternal()
```
frankgh, line 79: needs override annotations

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDecommissionJob.java
```diff
@@ -0,0 +1,99 @@
...
+    {
+        if (isRunningOnCassandra())
+        {
+            LOGGER.info("Not executing job as an ongoing or completed decommission operation was found jobId={}", jobId);
```
frankgh: info is the right level here?

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
```diff
@@ -42,7 +41,7 @@ public abstract class OperationalJob implements Task<Void>
     private static final Logger LOGGER = LoggerFactory.getLogger(OperationalJob.class);
 
     // use v1 time-based uuid
-    public final UUID jobId;
+    protected final UUID jobId;
```
frankgh: why is this protected if we already expose it via a public method?

server/src/test/java/org/apache/cassandra/sidecar/routes/ListOperationalJobsHandlerTest.java
```diff
@@ -149,6 +149,11 @@ protected SampleOperationalJob(UUID jobId)
             super(jobId);
         }
 
+        public boolean isRunningOnCassandra()
```
frankgh, line 153: override

server/src/test/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandlerTest.java
frankgh: you can pass the mock to the ctor of the module and use it there.

frankgh (commented):
partial review in place

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDecommissionJob.java
arjunashok: This currently only explicitly handles the 2 modes, so its excessive to add all modes here. Also, this used to be a switch previously, I believe @yifan-c had a comment to do something like this?

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDecommissionJob.java
arjunashok: Seems appropriate, as this is not an error, but a conflicting operation.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
arjunashok: `jobId` was being accessed from the concrete job. But ya, makes sense to keep this private scope.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/StorageOperations.java
yifan-c: Adding a bit more details to my comment. 

_Only if_ operationMode is queries a lot, it could cache the value and periodically refresh internally, to avoid loading Cassandra every single time.

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
yifan-c: let's keep the todo. I am adding the helper in another patch. Find it useful. I will update here when rebasing.

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDecommissionJob.java
yifan-c: I think my last nit comment was about defining enum for the modes. So you parse the string once into an enum, and compare enum, instead of comparing string multiple times.

arjunashok: My preference is to keep this as-is in this change, for the reason mentioned above and that operationModes vary across C* versions (https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/service/StorageService.java#L450). When we do incorporate a enum, we would have to ensure the parsing is specific to each version, which seems out of scope for this change.

adapters/cassandra41/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra41/Cassandra41StorageOperations.java
```diff
@@ -34,6 +34,7 @@
  */
 public class Cassandra41StorageOperations extends CassandraStorageOperations
 {
+
```
yifan-c: nit: empty line..

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDecommissionJob.java
yifan-c: yep. no worries. as said, it is a nit.

server/src/main/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandler.java
```diff
@@ -0,0 +1,94 @@
...
+            String reason = oje.getMessage();
+            logger.error("Conflicting job encountered. reason={}", reason);
+            context.response().setStatusCode(HttpResponseStatus.CONFLICT.code());
+            context.json(new OperationalJobResponse(job.jobId(), OperationalJobStatus.FAILED, job.name(), reason));
+        }
+
+        // Get the result, waiting for the specified wait time for result
+        job.asyncResult(executorPools.service(),
+                        Duration.of(config.operationalJobExecutionMaxWaitTimeInMillis(), ChronoUnit.MILLIS))
+            .onComplete(v -> sendStatusBasedResponse(context, job));
```
yifan-c: I think it is still off by 1 space? Is it?

server/src/main/java/org/apache/cassandra/sidecar/routes/NodeDecommissionHandler.java
```diff
@@ -0,0 +1,94 @@
...
+        try
+        {
+            jobManager.trySubmitJob(job);
+        }
+        catch (OperationalJobConflictException oje)
+        {
+            String reason = oje.getMessage();
+            logger.error("Conflicting job encountered. reason={}", reason);
+            context.response().setStatusCode(HttpResponseStatus.CONFLICT.code());
+            context.json(new OperationalJobResponse(job.jobId(), OperationalJobStatus.FAILED, job.name(), reason));
```
yifan-c, line 92: It can/should return early from the catch block.

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
yifan-c: I see. I thought it could speed up the test a little bit w/o waiting for the complete 10 seconds by using a count down latch. But it does not show a big difference. Feel free to either update or keep the current as-is.

```java
        AtomicReference<String> jobId = new AtomicReference<>();
        CountDownLatch jobIdRetrieved = new CountDownLatch(1);
        String testRoute = "/api/v1/cassandra/operations/decommission?force=true";
        testWithClient(client -> client.put(server.actualPort(), "127.0.0.1", testRoute)
                                           logger.info("Response Status:" + response.statusCode());
                                           logger.info("Response Body:" + response.bodyAsString());
                                           OperationalJobResponse decommissionResponse = response.bodyAsJson(OperationalJobResponse.class);
                                           assertThat(decommissionResponse.status()).isEqualTo(RUNNING);
                                           jobId.set(String.valueOf(decommissionResponse.jobId()));
                                           jobIdRetrieved.countDown();
                                       })));
        awaitLatchOrThrow(jobIdRetrieved, 10, TimeUnit.SECONDS, "Wait for the conflict jobId");
        assertThat(jobId.get()).isNotNull();
        pollStatusForState(jobId.get(), SUCCEEDED, null);
        context.completeNow();
```

server/src/test/java/org/apache/cassandra/sidecar/routes/OperationalJobHandlerTest.java
```diff
@@ -124,6 +124,8 @@ void testGetJobStatusRunningJob(VertxTestContext context)
               .expect(ResponsePredicate.SC_ACCEPTED)
               .send(context.succeeding(response -> {
                   assertThat(response.statusCode()).isEqualTo(ACCEPTED.code());
+                  LOGGER.info("Response Status:" + response.statusCode());
+                  LOGGER.info("Response Body:" + response.bodyAsString());
```
yifan-c: nit: use `{}` to eval log message... I know that it is just test code. But for the sake of consistency.

yifan-c (approved):
Looks good to me

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDecommissionJob.java
frankgh: you can use a switch to handle the two modes + default behavior. No need to be exhaustive. In any case, to avoid any potential NPE let's reverse the comparison i.e. `"LEAVING".equals(operationMode)`

frankgh: same comment for other places in this patch for example line 50 of this file

server/src/test/java/org/apache/cassandra/sidecar/routes/OperationalJobHandlerTest.java
```diff
@@ -124,6 +124,8 @@ void testGetJobStatusRunningJob(VertxTestContext context)
               .expect(ResponsePredicate.SC_ACCEPTED)
               .send(context.succeeding(response -> {
                   assertThat(response.statusCode()).isEqualTo(ACCEPTED.code());
+                  LOGGER.info("Response Status: {}", response.statusCode());
```
frankgh: is this useful in tests? we already know the status code is accepted by this point?

server/src/test/java/org/apache/cassandra/sidecar/routes/OperationalJobHandlerTest.java
```diff
@@ -124,6 +124,8 @@ void testGetJobStatusRunningJob(VertxTestContext context)
...
+                  LOGGER.info("Response Body: {}", response.bodyAsString());
```
frankgh: I find that logging this here is not useful, maybe just validate the payload

server/src/test/java/org/apache/cassandra/sidecar/routes/OperationalJobHandlerTest.java
```diff
@@ -138,6 +140,8 @@ void testGetJobStatusCompletedJob(VertxTestContext context)
               .send(context.succeeding(response -> {
                   assertThat(response.statusCode()).isEqualTo(OK.code());
                   OperationalJobResponse jobStatus = response.bodyAsJson(OperationalJobResponse.class);
+                  LOGGER.info("Response Status: {}", response.statusCode());
```
frankgh: same comment about logging

server/src/test/integration/org/apache/cassandra/sidecar/routes/NodeDecommissionIntegrationTest.java
```diff
@@ -0,0 +1,134 @@
...
+public class NodeDecommissionIntegrationTest extends IntegrationTestBase
+{
+    @CassandraIntegrationTest(nodesPerDc = 2)
+    void decommissionNodeDefault(VertxTestContext context)
+    {
+        final AtomicReference<String> jobId = new AtomicReference<>();
+        String testRoute = "/api/v1/cassandra/operations/decommission?force=true";
+        testWithClient(client -> client.put(server.actualPort(), "127.0.0.1", testRoute)
+                                       .send(context.succeeding(response -> {
+                                           logger.info("Response Status:" + response.statusCode());
```
frankgh: not sure what the real value of log here is , maybe just assert ?

frankgh (commented):
Some final comments

frankgh (approved):
+1 thanks for addressing all the comments

PR #145, opened 2024-11-05
CASSANDRASC-156: Add mTLS Authentication in Sidecar

sarankk:


server/src/main/dist/conf/sidecar.yaml
```diff
@@ -136,6 +136,22 @@ vertx:
...
+      certificate_identity_extractor: org.apache.cassandra.sidecar.accesscontrol.authentication.CassandraIdentityExtractor
+  admin_identities:
+#    - spiffe://authorized/admin/identities
+  permission_cache:
+    enabled: true
+    expire_after_access_millis: 300000
+    maximum_size: 10000
+    warming_retries: 5
+    warming_retry_interval_millis: 2000
+
```
yifan-c: It would be nice to add comments to describe the configurations. For example, https://github.com/apache/cassandra/blob/7feb03e75b7249a8f86023713b28471a0fb6d309/conf/cassandra.yaml#L185-L219
The yaml file is operator-oriented. It would be nice for them to learn the feature without navigating through code.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * {@link AuthCache} caches information needed for authenticating sidecar users.
+ *
+ * @param <K> key stored in cache for retrieving value
+ * @param <V> value cached
+ *
+ * for {@link IdentityRoleCache} key will be String representing identity extracted from certificate and value will
+ * be String representing Cassandra role associated with the identity
+ */
```
yifan-c:
```suggestion
/**
 * Caches information needed for authenticating sidecar users.
 *
 * @param <K> Key type
 * @param <V> Value type
 */
```

I removed the paragraph about "for {@link IdentityRoleCache}" because it is specific to one implementation, and similar comment has added to the class in question.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+    protected volatile LoadingCache<K, V> cache;
+
+    protected AuthCache(String name,
+                        boolean enabled,
+                        Function<K, V> loadFunction,
+                        Supplier<Map<K, V>> bulkLoadFunction,
+                        int warmingRetries,
+                        long warmingRetryIntervalInMillis,
+                        long expireAfterMillis,
+                        long maxEntries)
```
yifan-c: How about just passing cacheConfig as input? Making a more succinct version

```java
    protected AuthCache(String name,
                        Function<K, V> loadFunction,
                        Supplier<Map<K, V>> bulkLoadFunction,
                        CacheConfiguration cacheConfiguration)
```

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+                this.expireAfterMillis = expireAfterMillis;
+                cache = initCache(cache);
+            }
+        }
+    }
+
+    public long getExpireAfterMillis()
+    {
+        return this.expireAfterMillis;
+    }
```
yifan-c: The method is not used. Its value should be retrieved from `CacheConfiguration`.
Maybe remove the method.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+        this.loadFunction = loadFunction;
+        this.bulkLoadFunction = bulkLoadFunction;
+        this.warmingRetries = warmingRetries;
+        this.warmingRetryIntervalInMillis = warmingRetryIntervalInMillis;
+        this.expireAfterMillis = expireAfterMillis;
+        this.maxEntries = maxEntries;
+        this.cache = initCache(null);
+    }
+
+    public void setExpireAfterMillis(long expireAfterMillis)
```
yifan-c: `@VisibleForTesting`?

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+    protected final boolean enabled;
+    protected final Function<K, V> loadFunction;
+    protected final Supplier<Map<K, V>> bulkLoadFunction;
+    protected final int warmingRetries;
+    protected final long warmingRetryIntervalInMillis;
+
+    protected long expireAfterMillis;
+    protected long maxEntries;
+
+    protected volatile LoadingCache<K, V> cache;
```
yifan-c: add comment to explain when cache can be null.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+                this.maxEntries = maxEntries;
+                cache = initCache(cache);
+            }
+        }
+    }
+
+    public long getMaxEntries()
+    {
+        return this.maxEntries;
+    }
```
yifan-c: Unused methods in prod code.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+     * Retrieve a value from the cache. Will call {@link LoadingCache#get(Object)} which will
+     * "load" the value if it's not present, thus populating the key.
+     * @param k key
+     * @return The current value of {@code K} if cached or loaded.
+     *
+     * See {@link LoadingCache#get(Object)} for possible exceptions.
+     */
+    public V get(K k)
+    {
+        if (cache == null)
```
yifan-c: It is probably easier to understand with the condition, `if (!enabled)`

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    public long getMaxEntries()
+    {
+        return this.maxEntries;
+    }
+
+    /**
+     * Retrieve a value from the cache. Will call {@link LoadingCache#get(Object)} which will
+     * "load" the value if it's not present, thus populating the key.
```
yifan-c: Write javadoc to explain the behavior of the method when cache is disabled.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+     * @return a map of cached key-value pairs
+     */
+    public Map<K, V> getAll()
+    {
+        if (cache == null)
+        {
+            return Collections.emptyMap();
+        }
+        return Collections.unmodifiableMap(cache.asMap());
+    }
```
yifan-c: The implementation conflicting with `get(k)`

When the cache is disabled, `get(k)` returns something, but `getAll()` returns empty map. Can you make it consistent? Either call `bulkLoadFunction` in this method or return null in `get(k)`

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+
+    // We might need to add support for invalidating cache entries when we support removing auth entries from
+    // Cassandra auth tables through sidecar
+    public void invalidate(K k)
+    {
+        throw new UnsupportedOperationException("Invalidate functionality is not yet supported for AuthCache");
+    }
```
yifan-c: If it is not supported, just remove the method. 
It can always be introduced later when adding the functionality described in the comment.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+                                   .build(loadFunction::apply);
+        }
+        else
+        {
+            updatedCache = cache;
+            // Always set as mandatory
+            cache.policy().expireAfterWrite().ifPresent(policy -> policy.setExpiresAfter(expireAfterMillis, TimeUnit.MILLISECONDS));
+            cache.policy().eviction().ifPresent(policy -> policy.setMaximum(getMaxEntries()));
+        }
+        return updatedCache;
```
yifan-c: Simplify it once the cache config cannot be updated.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    public void warm()
+    {
+        if (cache == null)
+        {
+            LOGGER.warn("Cache {} not enabled, skipping pre-warming", name);
```
yifan-c: Maybe just `info`

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,214 @@
...
+        try
+        {
+            Map<K, V> entries = bulkLoadFunction.get();
+            cache.putAll(entries);
+        }
+        catch (Exception e)
+        {
+            LOGGER.warn("Unexpected error encountered during cache {} pre-warming, ", name, e);
+            Uninterruptibles.sleepUninterruptibly(warmingRetryIntervalInMillis, TimeUnit.MILLISECONDS);
+            warm(retry - 1);
```
yifan-c: avoid tail recursion. it can be easily rewrite as a while loop, which is more performant.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/IdentityRoleCache.java
```diff
@@ -0,0 +1,57 @@
...
+package org.apache.cassandra.sidecar.accesscontrol;
+
+import org.apache.cassandra.sidecar.config.CacheConfiguration;
+import org.apache.cassandra.sidecar.db.SystemAuthDatabaseAccessor;
+
+/**
+ * Caches entries from identity_to_role table. identity_to_role table maps valid certificate identity to Cassandra
+ * role the identity holds. identity_to_role table is created in Cassandra versions 5.x and above
+ */
+public class IdentityRoleCache extends AuthCache<String, String>
```
yifan-c: Let's rename the class to `IdentityToRoleCache`. It is import to keep "TO" to avoid ambiguity.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/IdentityRoleCache.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package org.apache.cassandra.sidecar.accesscontrol;
+
+import org.apache.cassandra.sidecar.config.CacheConfiguration;
+import org.apache.cassandra.sidecar.db.SystemAuthDatabaseAccessor;
+
+/**
+ * Caches entries from identity_to_role table. identity_to_role table maps valid certificate identity to Cassandra
+ * role the identity holds. identity_to_role table is created in Cassandra versions 5.x and above
```
yifan-c:
```suggestion
 * Caches entries from system_auth.identity_to_role table. The table maps valid certificate identities to Cassandra
 * roles. identity_to_role table is available since Cassandra versions 5.0
```

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/IdentityRoleCache.java
```diff
@@ -0,0 +1,57 @@
...
+              config.warmingRetries(),
+              2000,
+              config.expireAfterAccessMillis(),
+              config.maximumSize());
+        this.systemAuthDatabaseAccessor = systemAuthDatabaseAccessor;
+    }
+
+    public boolean contains(String identity)
+    {
+
```
yifan-c: remove the empty line

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCacheService.java
```diff
@@ -0,0 +1,77 @@
...
+import com.google.common.base.Preconditions;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import org.apache.cassandra.sidecar.config.SidecarConfiguration;
+
+/**
+ * {@link AuthCacheService} allows for easier maintenance of all {@link AuthCache}.
```
yifan-c:
```suggestion
 * Manages all {@link AuthCache}.
```

Avoid "easier" as it does not compare to anything else. Hopefully, the simple sentence in the suggestion captures what this class does.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCacheService.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+@Singleton
+public class AuthCacheService
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(AuthCacheService.class);
+    private final boolean enabled;
+    private final Set<AuthCache<?, ?>> caches = new HashSet<>();
```
yifan-c: Using `Set` here? How to retrieve a specific `AuthCache`? 

hmmm... looks like `AuthCacheService` does not support retrieval, which makes the service quite useless. It does not manage anything. 

How about removing the class?

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCacheService.java
```diff
@@ -0,0 +1,77 @@
...
+        for (AuthCache<?, ?> cache : caches)
+        {
+            cache.warm();
+        }
+    }
+
+    public synchronized void invalidateCaches()
+    {
+        throw new UnsupportedOperationException("Currently invalidate of caches not supported");
+    }
```
yifan-c: remove the unsupported method

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/IdentityRoleCache.java
```diff
@@ -0,0 +1,57 @@
...
+        if (cache == null)
+        {
+            return false;
+        }
+        String role = get(identity);
+        return  role != null && !role.isEmpty();
```
yifan-c: Should it be an invalid state if the retrieved `role` string is empty?

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/IdentityRoleCache.java
```diff
@@ -0,0 +1,57 @@
...
+              systemAuthDatabaseAccessor::findRoleFromIdentity,
+              systemAuthDatabaseAccessor::findAllIdentityRoles,
+              config.warmingRetries(),
+              2000,
+              config.expireAfterAccessMillis(),
+              config.maximumSize());
+        this.systemAuthDatabaseAccessor = systemAuthDatabaseAccessor;
+    }
+
+    public boolean contains(String identity)
```
yifan-c: Plz rename to `containsKey`. The method is to check the existence of a key. 
In the context of map-like data structure, the `contains` can also mean the value existence check. Rename to `containsKey` to be clear.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/authentication/CassandraIdentityExtractor.java
```diff
@@ -0,0 +1,65 @@
...
+    protected final Set<String> adminIdentities = new HashSet<>();
+
+    public CassandraIdentityExtractor(IdentityRoleCache identityRoleCache,
+                                      Set<String> adminIdentities)
+    {
+        this.identityRoleCache = identityRoleCache;
+        this.adminIdentities.addAll(adminIdentities);
+    }
+
+    public List<String> validIdentities(CertificateCredentials certificateCredentials) throws CredentialValidationException
```
yifan-c: Add `@Override`

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/authentication/MutualTlsAuthenticationHandler.java
```diff
@@ -0,0 +1,85 @@
...
+import io.vertx.ext.web.handler.impl.AuthenticationHandlerImpl;
+import org.apache.cassandra.sidecar.concurrent.ExecutorPools;
+
+/**
+ * Handler for verifying user certificates for Mutual TLS authentication. {@link MutualTlsAuthenticationHandler} can be
+ * chained with other {@link io.vertx.ext.web.handler.AuthenticationHandler} implementations.
+ */
+public class MutualTlsAuthenticationHandler extends AuthenticationHandlerImpl<MutualTlsAuthentication>
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(MutualTlsAuthenticationHandler.class);
```
yifan-c: `LOGGER` not used

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/authentication/MutualTlsAuthenticationHandler.java
```diff
@@ -0,0 +1,85 @@
...
+                            }
+                            else
+                            {
+                                // In case of other failures, reject the promise with an UNAUTHORIZED error
+                                promise.fail(new HttpException(HttpResponseStatus.UNAUTHORIZED.code()));
+                            }
+                        });
+        })
+        .onSuccess(result -> handler.handle(Future.succeededFuture((User) result)))
+        .onFailure(cause -> handler.handle(Future.failedFuture(cause)));
```
yifan-c:
```suggestion
        authProvider.authenticate(certificateCredentials)
                    .onSuccess(result -> handler.handle(Future.succeededFuture(result)))
                    .onFailure(cause -> ctx.fail(new HttpException(HttpResponseStatus.UNAUTHORIZED.code(), cause)));
```

1. `authProvider.authenticate` is already async. No need to dispatch to another executor pool again. 
2. All exceptions should be mapped to `UNAUTHORIZED`, IMO. Invalid cert is not a bad request. 
3. Do not call the next handler when auth failed. Simply fail the request instead.

server/src/main/java/org/apache/cassandra/sidecar/config/AccessControlConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.util.Set;
+
+/**
+ * Configuration stored for controlling user access to Sidecar.
+ */
+public interface AccessControlConfiguration
+{
+    /**
+     * @return whether access control is enabled, if yes requests need to be authenticated/authorized before allowed
```
yifan-c: `authenticated/authorized` ==> `authenticated and authorized`

server/src/main/java/org/apache/cassandra/sidecar/config/CacheConfiguration.java
```diff
@@ -41,4 +41,14 @@ default boolean enabled()
     {
         return true;
     }
+
+    /**
+     * @return number of retries for cache warming
+     */
+    int warmingRetries();
```
yifan-c: How about `warmUpRetries`? 

`warming` looks too similar to `warning`

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/AccessControlConfigurationImpl.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.cassandra.sidecar.config.AccessControlConfiguration;
+import org.apache.cassandra.sidecar.config.AuthenticatorsConfiguration;
+import org.apache.cassandra.sidecar.config.CacheConfiguration;
+
+/**
+ * Encapsulates configuration needed for creating authenticators in Sidecar.
```
yifan-c:
```suggestion
 * {@inheritDoc}
```

It is clearer to inherit the doc from interface, rather than writing a new version of it.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/AuthenticatorsConfigurationImpl.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package org.apache.cassandra.sidecar.config.yaml;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.cassandra.sidecar.config.AuthenticatorsConfiguration;
+import org.apache.cassandra.sidecar.config.MutualTlsAuthenticatorConfiguration;
+
+/**
+ * Encapsulates configuration needed to support multiple authenticators in Sidecar.
```
yifan-c:
```suggestion
 * {@inheritDoc}
```

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/MutualTlsAuthenticatorConfigurationImpl.java
```diff
@@ -0,0 +1,86 @@
...
+package org.apache.cassandra.sidecar.config.yaml;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.vertx.ext.auth.mtls.impl.CertificateValidatorImpl;
+import io.vertx.ext.auth.mtls.impl.SpiffeIdentityExtractor;
+import org.apache.cassandra.sidecar.config.MutualTlsAuthenticatorConfiguration;
+
+/**
+ * Encapsulates configuration needed for creating {@link io.vertx.ext.auth.mtls.impl.MutualTlsAuthenticationImpl}
+ * authentication provider.
```
yifan-c:
```suggestion
 * {@inheritDoc}
```

server/src/main/java/org/apache/cassandra/sidecar/db/SystemAuthDatabaseAccessor.java
```diff
@@ -0,0 +1,89 @@
...
+        ResultSet result = execute(statement);
+        return result.one().getString("role");
+    }
+
+    /**
+     * Queries Cassandra for all rows in identity_to_role table
+     *
+     * @return - {@code List<Row>} containing each row in the identity to roles table
+     */
+    public Map<String, String> findAllIdentityRoles()
```
yifan-c:
```suggestion
    public Map<String, String> findAllIdentitiesToRoles()
```
Adds the word "to" back. And use plural form consistently for both words.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
```diff
@@ -0,0 +1,107 @@
...
+
+        getAllRolesAndIdentities = prepare(getAllRolesAndIdentities,
+                                           session,
+                                           CqlLiterals.getAllRolesAndIdentities());
+    }
+
+    protected String tableName()
+    {
+        throw new UnsupportedOperationException("SystemAuthSchema supports reading information from multiple " +
+                                                "tables in system_auth keyspace");
```
yifan-c, line 66: There is only one table declared in the class. no?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
```diff
@@ -0,0 +1,107 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Schema for getting information stored in system_auth keyspace.
+ */
+public class SystemAuthSchema extends TableSchema
+{
+    private static final String IDENTITY_TO_ROLE_TABLE = "identity_to_role";
+    PreparedStatement selectRoleFromIdentity;
+    PreparedStatement getAllRolesAndIdentities;
```
yifan-c: `private final`?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
```diff
@@ -0,0 +1,107 @@
...
+                                                "tables in system_auth keyspace");
+    }
+
+    @Override
+    protected boolean exists(@NotNull Metadata metadata)
+    {
+        // check tables exists before preparing
+        KeyspaceMetadata keyspaceMetadata = metadata.getKeyspace(keyspaceName());
+        // identity_to_role table exists in Cassandra versions starting 5.x
+        return keyspaceMetadata != null && keyspaceMetadata.getTable(IDENTITY_TO_ROLE_TABLE) != null;
```
yifan-c: note that if the method returns false, it will try to create the table, which is not desirable.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Override
+    protected String createSchemaStatement()
+    {
+        return "";
+    }
```
yifan-c: It should never attempt to create system table.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
```diff
@@ -0,0 +1,107 @@
...
+import com.datastax.driver.core.KeyspaceMetadata;
+import com.datastax.driver.core.Metadata;
+import com.datastax.driver.core.PreparedStatement;
+import com.datastax.driver.core.Session;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Schema for getting information stored in system_auth keyspace.
+ */
+public class SystemAuthSchema extends TableSchema
```
yifan-c: Can you use `org.apache.cassandra.sidecar.adapters.base.db.schema.ConnectedClientsSchema` as an example? It is similar to the use case here.

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/CertificateIdentityExtractor.java
```diff
@@ -40,8 +42,8 @@ public interface CertificateIdentityExtractor
      * </ul>
      *
      * @param certificateCredentials certificate chain of user that is already verified
-     * @return {@code String} identity string extracted from certificate, uniquely represents client
+     * @return list of valid identities extracted from certificate.
      * @throws CredentialValidationException when a valid identity cannot be extracted from certificate chain.
      */
-    String validIdentity(CertificateCredentials certificateCredentials) throws CredentialValidationException;
+    List<String> validIdentities(CertificateCredentials certificateCredentials) throws CredentialValidationException;
```
yifan-c, line 48: What is the reason for the interface change?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    protected String tableName()
+    {
+        throw new UnsupportedOperationException("SystemAuthSchema supports reading information from multiple " +
+                                                "tables in system_auth keyspace");
+    }
+
+    @Override
+    protected boolean exists(@NotNull Metadata metadata)
```
yifan-c: Is there integration test for the scenario when connecting to cassandra version, say 4.0, and the table does not exist.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCacheService.java
sarankk: It is for maintaining auth caches together, we warm caches. It will be useful for invalidate too. Right now we dont have any use case for cache retrieval, hence didnt add getter methods

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/authentication/MutualTlsAuthenticationHandler.java
sarankk: The reason we hand off to executor pool, is because, we use `identityRoleCache`, during loads, since database calls are involved, it blocks thread. I can move hand off to executor pool inside `IdentityRoleCache`

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
sarankk: When we add authorization, there will be one more table `role_permissions`

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
sarankk: Makes sense, I left it empty, so that no schema related statement will be executed. We can throw too, we don't call that method during initialize

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
sarankk: I think you mean for `createSchemaStatement` ? updated it to return null. `ConnectedClientsSchema` does not check table existence before preparing statemtent.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
sarankk: test ` testAssociatingRoleWithNonAdminIdentity` in `MutualTlsAuthenticationIntegrationTest` tests this. in 4.x cluster we expect a failure if the identity is not an admin identity (when table does not exists, it allows only admin identities). In 5.x cluster, if the identity is part of the table and not an admin identity we allow it through.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
sarankk: `system_auth` keyspace doesn't allow table creation, it is only internal Cassandra tables. Do we still need to worry about this?

vertx-auth-mtls/src/main/java/io/vertx/ext/auth/mtls/CertificateIdentityExtractor.java
sarankk: Sometimes we could extract multiple identities from a certificate. If there are multiple spiffe(s) listed in SAN, or just from multiple SAN entries we could extract different identities. Hence we need to return a list rather than a single identity from cert.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/IdentityRoleCache.java
sarankk: I think so, Cassandra doesn't allow empty string

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCacheService.java
sarankk: We could add retrieval to `AuthCacheService` that way we dont have to provide each cache in `MainModule`

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/authentication/MutualTlsAuthenticationHandler.java
sarankk: Regarding 3. since we use `ChainAuthHandler` calling next, allows the next authentication handler to be called. `ChainAuthHandler` finally returns `UNAUTHORIZED` if all handlers fail.

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/AuthCache.java
```diff
@@ -0,0 +1,157 @@
...
+        if (!config.enabled())
+        {
+            return bulkLoadFunction.get();
+        }
+        return Collections.unmodifiableMap(cache.asMap());
+    }
+
+    private LoadingCache<K, V> initCache()
+    {
+        if (!config.enabled())
```
yifan-c: nit: the condition can be combined with line#71 in the file.

server-common/src/main/java/org/apache/cassandra/sidecar/db/schema/CassandraSystemTableSchema.java
```diff
@@ -0,0 +1,52 @@
...
+    {
+        prepareStatements(session);
+        return true;
+    }
+
+    /**
+     * Should not create new tables, since schema is for existing tables.
+     */
+    @Override
+
```
yifan-c: nit: empty line

server/src/main/java/org/apache/cassandra/sidecar/accesscontrol/authentication/MutualTlsAuthenticationHandler.java
```diff
@@ -0,0 +1,58 @@
...
+        {
+            ctx.response().setStatusCode(HttpResponseStatus.BAD_REQUEST.code()).end();
+            return;
+        }
+
+        CertificateCredentials certificateCredentials = CertificateCredentials.fromHttpRequest(ctx.request());
+
+        authProvider.authenticate(certificateCredentials)
+                    .onSuccess(user -> handler.handle(Future.succeededFuture(user)))
+                    .onFailure(cause -> handler.handle(Future.failedFuture(new HttpException(HttpResponseStatus.UNAUTHORIZED.code(), cause))));
```
yifan-c: nit: alternative that is less wordy

```suggestion
        authProvider.authenticate(certificateCredentials)
                    .recover(cause -> { // converts any exception to unauthorized http exception
                        throw new HttpException(HttpResponseStatus.UNAUTHORIZED.code(), cause);
                    })
                    .andThen(handler);
```

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public PreparedStatement selectRoleFromIdentity()
+    {
+        return selectRoleFromIdentity;
+    }
+
+    public PreparedStatement getAllRolesAndIdentities()
+    {
+        return getAllRolesAndIdentities;
+    }
```
yifan-c: nit: add `@Nullable` for the return value

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -182,9 +194,63 @@ public Vertx vertx(SidecarConfiguration sidecarConfiguration, MetricRegistryFact
...
+                CertificateIdentityExtractor certificateIdentityExtractor;
+                if (mTLSAuthenticatorConfig.certificateIdentityExtractor().equalsIgnoreCase(CassandraIdentityExtractor.class.getName()))
+                {
+                    certificateIdentityExtractor
+                    = new CassandraIdentityExtractor(identityToRoleCache, sidecarConfiguration.accessControlConfiguration().adminIdentities());
+                }
+                else
+                {
+                    certificateIdentityExtractor = new SpiffeIdentityExtractor();
+                }
```
yifan-c: nit: `CertificateIdentityExtractor` can be created the same way as `CertificateValidator`
Not a big concern. Feel free to ignore.

server/src/main/java/org/apache/cassandra/sidecar/db/SystemAuthDatabaseAccessor.java
```diff
@@ -0,0 +1,89 @@
...
+        {
+            results.put(row.getString("identity"), row.getString("role"));
+        }
+        return results;
+    }
+
+    private boolean tableSchemaPrepared()
+    {
+        return tableSchema.selectRoleFromIdentity() != null && tableSchema.getAllRolesAndIdentities() != null;
+    }
```
yifan-c: How about replacing with this method and update the callsites?
```suggestion
    private void ensureIdentityToRoleTableAccess()
    {
        if (tableSchema.selectRoleFromIdentity() == null || tableSchema.getAllRolesAndIdentities() == null)
        {
            throw new IllegalStateException("SystemAuthSchema was not prepared, values can not be retrieved from table");
        }
    }
```

server/src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestModule.java
```diff
@@ -47,13 +58,26 @@
  */
 public class IntegrationTestModule extends AbstractModule
 {
+    public static final String ADMIN_IDENTITY = "spiffe://cassandra/sidecar/admin";
```
yifan-c, line 61: nit: same string declared in multiple locations. maybe consolidate.

yifan-c (commented):
final set of comments

server/src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestModule.java
sarankk: It is declared once in `IntegrationTestModule` and once in `MutualTlsAuthenticationHandlerTest`. Since both are different packages, integration test and unit test, there is no repetition.

spotbugs-exclude.xml
```diff
@@ -83,6 +83,7 @@
             <Class name="org.apache.cassandra.sidecar.routes.StreamSSTableComponentHandlerTest" />
             <Class name="org.apache.cassandra.sidecar.db.RestoreSliceDatabaseAccessor" />
             <Class name="org.apache.cassandra.sidecar.restore.StorageClient" />
+            <Class name="org.apache.cassandra.sidecar.db.SystemAuthDatabaseAccessor" />
```
frankgh, line 86: we probably shouldn't have this here?

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * Caches information needed for authenticating sidecar users.
+ *
+ * @param <K> Key type
+ * @param <V> Value type
+ */
+public abstract class AuthCache<K, V>
+{
+    protected static final Logger LOGGER = LoggerFactory.getLogger(AuthCache.class);
```
yifan-c:
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(AuthCache.class);
```

Let's avoid populating the loggers in the concrete implementations.

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -0,0 +1,164 @@
...
+    protected final String name;
+    protected final Vertx vertx;
+    protected final Function<K, V> loadFunction;
+    protected final Supplier<Map<K, V>> bulkLoadFunction;
+    protected final CacheConfiguration config;
+    private final TaskExecutorPool internalPool;
+
+    // cache is null when AuthCache is disabled
+    protected volatile LoadingCache<K, V> cache;
+
```
yifan-c: Actually none of them are needed for the subclass. They are exposed already via the methods, which is the preferred in terms of encapsulation. 

Can you change all the `protected` variables to `private`? 

Regarding `cache`, can you create a `protected` method to access the field?

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -0,0 +1,164 @@
...
+        EventBus eventBus = vertx.eventBus();
+        eventBus.localConsumer(ON_ALL_CASSANDRA_CQL_READY.address(),
+                               message -> internalPool.executeBlocking(() -> {
+                                   warm(config.warmupRetries());
+                                   return null;
+                               }));
+    }
+
+    @VisibleForTesting
+    protected void warm(int availableRetries)
```
yifan-c:
```suggestion
    protected void warmUp(int availableRetries)
```

to avoid misreading as `warn`

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -0,0 +1,164 @@
...
+        {
+            cache.putAll(bulkLoadFunction.get());
+        }
+        catch (Exception e)
+        {
+            LOGGER.warn("Unexpected error encountered during pre-warming of cache={} ", name, e);
+            vertx.setTimer(config.warmupRetryIntervalMillis(), t -> internalPool.executeBlocking(() -> {
+                warm(availableRetries - 1);
+                return null;
+            }));
```
yifan-c: The async warmup can be extracted out. The same function is called in `configureSidecarServerEventListener`.

```java
    private void warmUpAsync(int availableRetries)
    {
        internalPool.runBlocking(() -> warmUp(availableRetries));
    }
```

Then, have this in the catch block and `configureSidecarServerEventListener` respectively 

```java
internalPool.setTimer(config.warmupRetryIntervalMillis(), t -> warmUpAsync(availableRetries - 1));

eventBus.localConsumer(ON_ALL_CASSANDRA_CQL_READY.address(), message -> warmUpAsync(config.warmupRetries()));
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/AuthenticationProviderFactory.java
```diff
@@ -0,0 +1,38 @@
...
+package org.apache.cassandra.sidecar.acl.authentication;
+
+import java.util.Map;
+
+import io.vertx.core.Vertx;
+import io.vertx.ext.web.handler.impl.AuthenticationHandlerInternal;
+import org.apache.cassandra.sidecar.config.AccessControlConfiguration;
+import org.apache.cassandra.sidecar.exceptions.ConfigurationException;
+
+public interface AuthenticationProviderFactory
```
yifan-c: javadoc for the class and methods

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/AuthenticationProviderFactory.java
```diff
@@ -0,0 +1,38 @@
...
+{
+    default void validate(Map<String, String> parameters) throws ConfigurationException
+    {
+        // no validation
+    }
+
+    AuthenticationHandlerInternal initialize(Vertx vertx,
```
yifan-c: 1. add an explanation about why `AuthenticationHandlerInternal` instead of `AuthenticationHandler`. (vertx auth framework flaw).
2. the name should be `create`. The method does not initialize the factory, but create auth handlers.

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/AuthenticationProviderFactoryRegistry.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.acl.authentication;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class AuthenticationProviderFactoryRegistry
```
yifan-c: javadoc

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/CassandraIdentityExtractor.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.cassandra.sidecar.acl.IdentityToRoleCache;
+
+/**
+ * {@link CassandraIdentityExtractor} verifies {@code SPIFFE} identities extracted from certificate are mapped
+ * to a valid role in Cassandra or a pre-configured administrative identity.
+ */
+public class CassandraIdentityExtractor extends SpiffeIdentityExtractor
+{
+    protected final IdentityToRoleCache identityToRoleCache;
+    protected final Set<String> adminIdentities;
```
yifan-c: prefer private

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/MutualTLSAuthenticationProviderFactory.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.cassandra.sidecar.acl.IdentityToRoleCache;
+import org.apache.cassandra.sidecar.config.AccessControlConfiguration;
+import org.apache.cassandra.sidecar.exceptions.ConfigurationException;
+
+@Singleton
+public class MutualTLSAuthenticationProviderFactory implements AuthenticationProviderFactory
+{
+    static final String CERTIFICATE_VALIDATOR_PARAM_KEY = "certificate_validator";
+    static final String CERTIFICATE_IDENTITY_EXTRACTOR_PARAM_KEY = "certificate_identity_extractor";
+    private final IdentityToRoleCache identityToRoleCache;
```
yifan-c: private

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/MutualTLSAuthenticationProviderFactory.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        if (!parameters.containsKey(CERTIFICATE_VALIDATOR_PARAM_KEY))
+        {
+            throw new ConfigurationException("message");
+        }
+
+        if (!parameters.containsKey(CERTIFICATE_IDENTITY_EXTRACTOR_PARAM_KEY))
+        {
+            throw new ConfigurationException("message");
```
yifan-c: update "message"

server/src/main/java/org/apache/cassandra/sidecar/config/AuthenticatorsConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+// * Configuration used to support multiple authenticators in Sidecar.
+// */
+//public interface AuthenticatorsConfiguration
+//{
+//    /**
+//     * @return configuration needed for mTLS authentication in Sidecar. With mTLS, both server and client exchange
+//     * certificates and verify each other
+//     */
+//    MutualTlsAuthenticatorConfiguration mTlsAuthenticatorConfiguration();
+//}
```
yifan-c: remove it?

server/src/main/java/org/apache/cassandra/sidecar/config/ParameterizedClassConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+
+package org.apache.cassandra.sidecar.config;
+
+import java.util.Map;
+
+import org.jetbrains.annotations.Nullable;
+
+public interface ParameterizedClassConfiguration
+{
+    String className();
```
yifan-c, line 34: nit: `classCanonicalName`

server/src/main/java/org/apache/cassandra/sidecar/config/ParameterizedClassConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.config;
+
+import java.util.Map;
+
+import org.jetbrains.annotations.Nullable;
+
+public interface ParameterizedClassConfiguration
```
yifan-c, line 29: javadoc

server/src/main/java/org/apache/cassandra/sidecar/config/ParameterizedClassConfiguration.java
```diff
@@ -0,0 +1,31 @@
...
+{
+    String className();
+    
+    @Nullable
+    Map<String, String> parameters();
```
yifan-c: nit: `namedParameters` for better clarity about the map.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/AuthenticatorsConfigurationImpl.java
```diff
@@ -0,0 +1,54 @@
...
+//    /**
+//     * {@inheritDoc}
+//     */
+//    @Override
+//    @JsonProperty(value = "mtls_authenticator")
+//    public MutualTlsAuthenticatorConfiguration mTlsAuthenticatorConfiguration()
+//    {
+//        return mTlsAuthenticatorConfiguration;
+//    }
+//}
```
yifan-c: remove

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/MutualTlsAuthenticatorConfigurationImpl.java
```diff
@@ -0,0 +1,70 @@
...
+//    /**
+//     * {@inheritDoc}
+//     */
+//    @Override
+//    @JsonProperty(value = "certificate_identity_extractor")
+//    public String certificateIdentityExtractor()
+//    {
+//        return identityExtractor;
+//    }
+//}
```
yifan-c: remove

server/src/main/java/org/apache/cassandra/sidecar/exceptions/ConfigurationException.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.exceptions;
+
+public class ConfigurationException extends RuntimeException
```
yifan-c, line 25: Maybe `InvalidConfigurationException`? And javadoc

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/AuthenticationProviderFactory.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.Map;
+
+import io.vertx.core.Vertx;
+import io.vertx.ext.web.handler.impl.AuthenticationHandlerInternal;
+import org.apache.cassandra.sidecar.config.AccessControlConfiguration;
+import org.apache.cassandra.sidecar.exceptions.ConfigurationException;
+
+public interface AuthenticationProviderFactory
+{
+    default void validate(Map<String, String> parameters) throws ConfigurationException
```
yifan-c: `validate` should be part of `initialize/create` method, instead of a standalone step. Since both methods take a map as input, there is no guarantee (from the perspective of the interface) that the map used for creating the handler, is the same as the map for validation. 
I would suggest removing the `validate` method. The implementations run validation on the parameters anyway.

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
sarankk: Changed everything except `Vertx` incase we use it in AuthCache implementations

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/AuthenticationProviderFactory.java
sarankk: Makes sense, was thinking the same.

server/src/main/java/org/apache/cassandra/sidecar/exceptions/ConfigurationException.java
sarankk: Added Javadoc, we could also use it for missing configuration, so left the name as is.

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -155,10 +163,7 @@ protected void warm(int availableRetries)
         catch (Exception e)
         {
             LOGGER.warn("Unexpected error encountered during pre-warming of cache={} ", name, e);
-            vertx.setTimer(config.warmupRetryIntervalMillis(), t -> internalPool.executeBlocking(() -> {
-                warm(availableRetries - 1);
-                return null;
-            }));
+            vertx.setTimer(config.warmupRetryIntervalMillis(), t -> warmUp(availableRetries - 1));
```
yifan-c: I think you want to call `warmUpAsync` to keep the behavior.

server/src/main/java/org/apache/cassandra/sidecar/acl/IdentityToRoleCache.java
```diff
@@ -47,11 +46,18 @@ public IdentityToRoleCache(Vertx vertx,
...
     }
 
     public boolean containsKey(String identity)
     {
-        return cache != null && get(identity) != null;
+        try
+        {
+            return cache() != null && get(identity) != null;
+        }
+        catch (IllegalStateException e)
```
yifan-c: nit: probably throw a dedicated exception, say `CqlPreparedStatementUnavailableException`, to avoid mute other ISE by accident.

sarankk: Named it `SchemaUnavailableException`

frankgh (approved):
+1 Looks good to me

PR #148, opened 2024-11-18
CASSSIDECAR-159: Add advanced driver settings to allow taking in credentials for Cassandra connection

sarankk:


server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -133,7 +133,7 @@ private LoadingCache<K, V> initCache()
     private void configureSidecarServerEventListener()
     {
         EventBus eventBus = vertx.eventBus();
-        eventBus.localConsumer(ON_ALL_CASSANDRA_CQL_READY.address(), message -> warmUpAsync(config.warmupRetries()));
+        eventBus.localConsumer(ON_SIDECAR_SCHEMA_INITIALIZED.address(), message -> warmUpAsync(config.warmupRetries()));
```
sarankk, line 140: Because we were not starting warm up on `ON_SIDECAR_SCHEMA_INITIALIZED`, first few tries at warming up at failing. Hence switched to starting on schema initialized

yifan-c: Interesting. Maybe add a line of comment to explain the dependency, because it is not obvious that warmup depends on `ON_SIDECAR_SCHEMA_INITIALIZED`

server/src/test/integration/org/apache/cassandra/sidecar/cluster/locator/CqlSessionProviderIntegrationTest.java
```diff
@@ -0,0 +1,163 @@
...
+                          return;
+                      }
+                  }
+                  if (checkSsl)
+                  {
+                      context.failNow("Did not see any SSL connection");
+                      return;
+                  }
+                  context.completeNow();
+              })));
```
yifan-c: `context.completeNow();` has to be outside of the `verify` block. Otherwise, any assertion failure causes the context not gets completed. 
Besides that, I moved the code around a bit to hopefully simplify the assertion. 

```suggestion
              .send(context.succeeding(response -> {
                  context.verify(() -> {
                      ConnectedClientStatsResponse clientStatsResponse = response.bodyAsJson(ConnectedClientStatsResponse.class);
                      assertThat(clientStatsResponse).isNotNull();

                      boolean seeSslConnection = false;
                      for (ClientConnectionEntry entry : clientStatsResponse.clientConnections())
                      {
                          assertThat(entry.username()).isEqualTo(expectedUsername);
                          if (checkSsl && entry.sslEnabled())
                          {
                              seeSslConnection = true;
                              break;
                          }
                      }
                      // We expect some connections to be non-SSL (i.e. for identity setup)
                      // and some connections to be SSL (Sidecar connecting to the cluster)
                      // so from the list of client connections we should see at least
                      // two (regular+control) connections.
                      assertThat(seeSslConnection)
                      .describedAs("Did not see any SSL connection")
                      .isTrue();
                  });
                  context.completeNow();
              }));
```

server/src/main/java/org/apache/cassandra/sidecar/cluster/CQLSessionProviderImpl.java
```diff
@@ -137,16 +170,30 @@ public synchronized Session get()
...
+                     .withoutMetrics()
+                     .withLoadBalancingPolicy(lbp)
+                     .withQueryOptions(queryOptions)
+                     // tests can create a lot of these Cluster objects, to avoid creating HWTs and
+                     // event thread pools for each we have the override
+                     .withNettyOptions(nettyOptions);
+
+            if (sslContext != null)
+            {
+                RemoteEndpointAwareJdkSSLOptions sslOptions
```
yifan-c: nit: use `RemoteEndpointAwareNettySSLOptions`

yifan-c (commented):
looks good in general. some minor comments

server/src/main/java/org/apache/cassandra/sidecar/cluster/CQLSessionProviderImpl.java
sarankk: Addressed this

server/src/test/integration/org/apache/cassandra/sidecar/cluster/locator/CqlSessionProviderIntegrationTest.java
sarankk: After moving around looks like it expects all tests to see a sslConnection. But we want to see only if checkSsl is true. Will update it 👍

server/src/main/dist/conf/sidecar.yaml
```diff
@@ -173,6 +167,18 @@ driver_parameters:
   contact_points:
     - "127.0.0.1:9042"
     - "127.0.0.2:9042"
+  username: cassandra
+  password: cassandra
```
bbotella, line 171: Super nit: `password: password` to keep consistency with the password fields for keystore and truststore?

server/src/main/java/org/apache/cassandra/sidecar/cluster/CQLSessionProviderImpl.java
```diff
@@ -137,16 +168,30 @@ public synchronized Session get()
...
+                     .addContactPointsWithPorts(contactPoints)
+                     .withReconnectionPolicy(reconnectionPolicy)
+                     .withoutMetrics()
+                     .withLoadBalancingPolicy(lbp)
+                     .withQueryOptions(queryOptions)
+                     // tests can create a lot of these Cluster objects, to avoid creating HWTs and
+                     // event thread pools for each we have the override
+                     .withNettyOptions(nettyOptions);
+
+            if (username != null && password != null)
```
bbotella, line 188: Should this be an `or` instead of an `and`?

server/src/main/java/org/apache/cassandra/sidecar/cluster/CQLSessionProviderImpl.java
```diff
@@ -203,4 +248,85 @@ public void close()
...
+        {
+            return new MutualTLSAuthenticator();
+        }
+
+        private static class MutualTLSAuthenticator implements Authenticator
+        {
+            @Override
+            public byte[] initialResponse()
+            {
+                return new byte[]{ 0, 0 };
```
bbotella: Just curious. Why is the initial response a `0, 0` here?

server/src/test/integration/org/apache/cassandra/sidecar/cluster/locator/CqlSessionProviderIntegrationTest.java
```diff
@@ -0,0 +1,172 @@
...
+
+import static org.apache.cassandra.sidecar.testing.IntegrationTestModule.ADMIN_IDENTITY;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assumptions.assumeThat;
+
+/**
+ * Test for authenticated {@link org.apache.cassandra.sidecar.cluster.CQLSessionProviderImpl}
+ */
+@ExtendWith(VertxExtension.class)
+class CqlSessionProviderIntegrationTest extends IntegrationTestBase
```
bbotella, line 48: Add a test with failed ssl configuration? Basically, with `createSslContext` method returning a `ConfigurationException`?

server/src/main/dist/conf/sidecar.yaml
sarankk: Since by default `cassandra` user is created in Cassandra with `cassandra` password. Better to keep this way

server/src/main/java/org/apache/cassandra/sidecar/cluster/CQLSessionProviderImpl.java
sarankk: Driver doesn't allow password to be null. So I think `and`

server/src/test/integration/org/apache/cassandra/sidecar/cluster/locator/CqlSessionProviderIntegrationTest.java
sarankk: We throw `ConfigurationException` in unexpected failures, not sure what failure scenario we want to recreate here.

bbotella: For example, an invalid path for keystore or truststore should return a 400, right?

frankgh (approved):
+1

server/src/test/integration/org/apache/cassandra/sidecar/cluster/locator/CqlSessionProviderIntegrationTest.java
```diff
@@ -0,0 +1,172 @@
...
+        waitForSchemaReady(30, TimeUnit.SECONDS);
+        // we enable only SSL and do not set any authenticator, hence username is "anonymous"
+        retrieveClientStats(context, "anonymous", true);
+    }
+
+    @CassandraIntegrationTest(buildCluster = false)
+    void testWithMTLS(VertxTestContext context, ConfigurableCassandraTestContext cassandraContext) throws Exception
+    {
+        // mTLS authentication was added in Cassandra starting 5.0 version
+        assumeThat(cassandraContext.version.major)
```
yifan-c, line 105: nice assertj feature

server/src/main/java/org/apache/cassandra/sidecar/cluster/CQLSessionProviderImpl.java
sarankk: We figured it wasn't necessary, even a empty array would work. But removed this `MtlsAuthProvider`

server/src/test/integration/org/apache/cassandra/sidecar/cluster/locator/CqlSessionProviderIntegrationTest.java
frankgh: not really, when a connection cannot be established we return a 503 - Service Unavailable. 400 is a client making a connection with bad params.

PR #152, opened 2024-12-03
CASSSIDECAR-162: Sidecar endpoint to GET sstable's preemptive interval value

skoppu22:
A new endpoint is developed in Sidecar, to retrieve value of C*'s sstable_preemptive_open_interval value. Currently end users have to invoke C*'s JMX call, which requires C* superuser role to invoke any JMX operation.  With this change, end users can just use this endpoint, which abstracts underlying C*'s version dependencies/changes. We can enable authorization checks for these endpoints once we have authorization developed in Sidecar

Circle CI run: https://app.circleci.com/pipelines/github/skoppu22/cassandra-sidecar?branch=jmxep1


server/src/main/java/org/apache/cassandra/sidecar/metrics/ServerMetrics.java
```diff
@@ -53,4 +53,9 @@ public interface ServerMetrics
      * @return metrics related to internal caches that are tracked.
      */
     CacheMetrics cacheMetrics();
+
+    /**
+     * @return metrics related to invocation of C* JMX operations
+     */
+    JmxOperationsMetrics jmxOperationsMetrics();
```
bbotella: Should these Jmx operations metrics be part of a different PR?

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
```diff
@@ -331,4 +334,34 @@ public static String extractHostAddressWithoutPort(String address)
...
+    protected StorageOperations getStorageOperations(String host)
+    {
+        CassandraAdapterDelegate delegate = this.metadataFetcher.delegate(host);
+        StorageOperations storageOperations = delegate == null ? null : delegate.storageOperations();
+        if (storageOperations == null)
+        {
+            throw cassandraServiceUnavailable();
+        }
+
+
```
bbotella: Extra space

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
```diff
@@ -331,4 +334,34 @@ public static String extractHostAddressWithoutPort(String address)
...
         return address;
     }
+
+    protected StorageOperations getStorageOperations(String host)
+    {
+        CassandraAdapterDelegate delegate = this.metadataFetcher.delegate(host);
+        StorageOperations storageOperations = delegate == null ? null : delegate.storageOperations();
+        if (storageOperations == null)
+        {
+            throw cassandraServiceUnavailable();
```
bbotella: Do we want to add some text to this error?

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
```diff
@@ -331,4 +334,34 @@ public static String extractHostAddressWithoutPort(String address)
...
+    }
+
+    protected <V> void updateJmxMetric(AsyncResult<V> result,
+                                       JmxOperationsMetrics jmxOperationsMetrics,
+                                       String operationName,
+                                       long startTime)
+    {
+        if (result.succeeded())
+        {
+            jmxOperationsMetrics.recordTimeTaken(operationName + "Succeeded",
```
bbotella: For the record time taken method, I think we could explore the possibility of splitting out the `Succeeded`/`Failed` into its own Status `parameter`.

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -118,6 +118,11 @@ public final class ApiEndpointsV1
 
     public static final String CONNECTED_CLIENT_STATS_ROUTE = API_V1 + CASSANDRA + "/stats/connected-clients";
 
+    // Endpoint to retrieve sstable's preemptiveOpenInterval value.
+    // Value returned is in MB, may return negative value when disabled
+    private static final String SSTABLE = "/sstable";
+    public static final String SSTABLE_PREEMPTIVE_OPEN_INTERVAL_ROUTE = API_V1 + CASSANDRA + SSTABLE +
```
bbotella, line 127: I know we are reusing the `getSsTablePreemtiveOpenIntervalInMB` JMX operation, but, from a pure API standpoint, we are omitting the `MB` part in the endpoint, but returning it as part of the response (`SSTablePreemptiveOpenIntervalInMB`). My suggestion for this would be to either:
- We add the `MB` part to the endpoint (`/preemptive-open-interval-in-mb`)
Or
- We just support an optional `unit` filtering parameter in the url, that can be extended in the future with other unit types if needed. That way, the url would end in `/preemptive-open-interval?unit=mb` (remember that the unit is optional with MB as a default, so no need to pass it from the client).

That would make us have to modify the response to something like:
```
{
  "SSTablePreemptiveOpenInterval": 1234,
  "unit": "MB"
}
```


What do you think?

server/src/test/java/org/apache/cassandra/sidecar/routes/cassandra/GetPreemptiveOpenIntervalHandlerTest.java
```diff
@@ -0,0 +1,110 @@
...
+        WebClient client = WebClient.create(vertx);
+        client.get(server.actualPort(), "127.0.0.1", testRoute)
+              .expect(ResponsePredicate.SC_SERVICE_UNAVAILABLE)
+              .send(context.succeeding(response -> {
+                  assertThat(response.statusCode()).isEqualTo(SERVICE_UNAVAILABLE.code());
+                  context.completeNow();
+              }));
+    }
+
+    private void verifyResponse(VertxTestContext context, HttpResponse<Buffer> response, String expectedValue)
```
bbotella: Nit: verifyValidResponse?

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
skoppu22: - We want to make endpoint generic, so underlying changes to C* shouldn't impact the endpoint usage for the end user, for example C* 6.0 returning value in different unit, while older versions continue to return the value in MB, the same endpoint should work regardless of the running c* version. Hence not including MB in the endpoint name. Whereas JMX operation name need to match exactly the JMX endpoint we are invoking. The same with passing option param like unit=MB, better we avoid supporting new unit param in the endpoint whenever C* changes for different versions.

- In the response we need to somehow tell the end user that value returned is in in MB. I preferred the name SSTablePreemptiveOpenIntervalInMB as it matches the JMX operation currently the end users are calling. Your suggestion of separating the unit into another field sounds good, but that may complicate callers's code, as they need to parse unit as a string and convert value into those units accordingly. C* latest code changed this yaml param to for example "60MiB", let me see which one is simpler for the calling code "60MiB" or {60, "MiB"} and will get back to you

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
skoppu22: If we split the status, the metric would be like (<jmx operation name>, <status>, <timetaken>), i.e, that will be triplet. And also we may need count of failures to trigger escalation if needed, i.e, count after filtering on first two in this triplet. I am not sure is it feasible with metric system and dashboards we have now (I am not very good in these).

skoppu22: I remember doing the same in C* as well, creating metric name on the fly based on the status, something like operationSucceeded or operationFailed and publishing the time which also gives us the failure count indirectly.

server/src/main/java/org/apache/cassandra/sidecar/metrics/ServerMetrics.java
skoppu22: This is the first Sidecar endpoint for a jmx operation, we need this, we will use the same for upcoming endpoints as well which invoke C* JMX operations. I understand there are existing calls as well in Sidecar invoking C* JMX, we will need to modify them to publish metrics. I prefer to have JmxOperationsMetrics accessible in JmxClient.java and record metrics there, that way each endpoint doesn't have to worry about publishing JMX calls metrics, instead JmxClient will take care of this for every JMX call being made, but I need to see the possibility of this.

On another thought, do we really need to track metrics of JMX calls? Does JVM (or something in Java world) provide this already? If so, we don't need this. 

Note: this is different from Sidecar endpoints invocation metrics, which are already provided by vertx.

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
skoppu22: cassandraServiceUnavailable currently says "Cassandra service is unavailable". The same message goes back in client's response. As this is an internal error, adding further info may leak internal info to the client. We use the same cassandraServiceUnavailable almost everywhere without adding further details, must be the same reason I am assuming. 

Any other kind of exception is caught in AbstractHandler's handle() and logged and updated in the response.

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
skoppu22: Removed metrics code from this PR, will open separate PR just for JMX calls metrics

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
skoppu22: I did explore this a little, I believe the one you suggested /preemptive-open-interval?unit=mb is more easy way to implement and also is in-line with current JMX call semantics. Other alternatives unnecessarily complicates conversions from human readable format to byte size at the client.

skoppu22: Added unit=MiB param, unit names matches with names C* is using.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -213,4 +216,17 @@ public void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tabl
         jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME)
                  .forceKeyspaceCleanup(concurrency, keyspace, table);
     }
+
+    @Override
+    public GetPreemptiveOpenIntervalResponse getSSTablePreemptiveOpenInterval(DataStorageUnit unit)
```
skoppu22, line 221: Removed 'inMB' from StorageOperations function names to make them generic for supporting other units later on.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/StorageJmxOperations.java
```diff
@@ -159,4 +159,10 @@ public interface StorageJmxOperations
      * @throws InterruptedException it does not really throw but declared in MBean
      */
     int forceKeyspaceCleanup(int jobs, String keyspaceName, String... tables) throws IOException, ExecutionException, InterruptedException;
+
+    /**
+     * Invokes C* StorageServiceMBean's JMX function getSSTablePreemptiveOpenIntervalInMB
+     * @return the same value returned by the JMX operation
+     */
+    int getSSTablePreemptiveOpenIntervalInMB();
```
skoppu22, line 167: This is the only place we use 'inMB', this should exactly match C* JMX function name

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
skoppu22: This code no longer exists, Sidecar code was refactored to check for these in delegate getters

server/src/test/java/org/apache/cassandra/sidecar/routes/cassandra/JmxCommonTest.java
```diff
@@ -0,0 +1,120 @@
+/*
```
skoppu22, line 1: This file is same as in other PR https://github.com/apache/cassandra-sidecar/pull/159/files. Either one of these PRs will merge this file

server/src/main/java/org/apache/cassandra/sidecar/routes/GetPreemptiveOpenIntervalHandler.java
```diff
@@ -0,0 +1,80 @@
...
+                                               CassandraInputValidator validator)
+    {
+        super(metadataFetcher, executorPools, validator);
+    }
+
+    @Override
+    protected DataStorageUnit extractParamsOrThrow(RoutingContext context)
+    {
+        // Only supported unit for preemptive open interval value is MB
+        String unit = RequestUtils.parseStringQueryParam(context.request(), "unit", "MiB");
```
bbotella, line 56: Super nit: Maybe extracting the string `MiB` to a static `DEFAULT_UNIT` string type variable helps with readability.

server/src/main/java/org/apache/cassandra/sidecar/utils/RequestUtils.java
```diff
@@ -61,4 +61,14 @@ public static Integer parseIntegerQueryParam(HttpServerRequest request, String p
         }
         return defaultValue;
     }
+
+    public static String parseStringQueryParam(HttpServerRequest request, String paramName, String defaultValue)
```
bbotella, line 65: nit: You are technically getting the param, not parsing it. Consider renaming the method

server/src/test/integration/org/apache/cassandra/sidecar/routes/cassandra/GetPreemptiveOpenIntervalHandlerIntegrationTest.java
```diff
@@ -0,0 +1,91 @@
...
+    {
+        client.get(server.actualPort(), "127.0.0.1", testRoute + "?unit=MiB")
+              .expect(ResponsePredicate.SC_OK)
+              .send(testContext.succeeding(response -> verifyValidResponse(testContext, response, 80)));
+    }
+
+    @CassandraIntegrationTest(yamlProps = "sstable_preemptive_open_interval_in_mb=90")
+    void testPreemptiveOpenIntervalInvalidUnit(VertxTestContext testContext)
+    {
+        client.get(server.actualPort(), "127.0.0.1", testRoute + "?unit=KiB")
```
bbotella, line 71: Nit: We may add support for KiB in the future, which would change the behavior of this test. Maybe just having a string that will never be a unit? Something like `foo`?

server/src/test/integration/org/apache/cassandra/testing/CassandraIntegrationTest.java
```diff
@@ -67,6 +67,8 @@
      */
     int numDataDirsPerInstance() default 1;
 
+    String yamlProps() default "";
```
bbotella, line 70: Why is this needed?

bbotella (commented):
Looks good overall. Just a couple of nits. I'm +1 on this PR (nb)

bbotella (approved):

PR #155, opened 2024-12-04
[CASSSIDECAR-167] Update Gradle Version

isaacreath:
This PR updates the Gradle version for cassandra-sidecar to the latest stable Gradle release (8.11). 

A couple of things to note:

First, the generated Windows start script (`cassandra-sidecar.bat`) does not work on trunk. It fails with the following error message:

```
> .\bin\cassandra-sidecar.bat
Error opening zip file or JAR manifest missing : APP_HOME_TO_REPLACE/agents/jolokia-jvm-1.6.0-agent.jar
Error occurred during initialization of VM
agent library failed to init: instrument
```

The changes I made in https://github.com/apache/cassandra-sidecar/compare/trunk...isaacreath:CASSSIDECAR-167?expand=1#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R129-R135 do not fix the Windows start script. I'm open to disabling the generation of this start script in this PR, but I think to get it actually working properly would be outside of the scope of this change.

Second, the use of `${APP_HOME}` as a default JVM option in the generated start script is discouraged in modern versions of Gradle and is not possible to do out of the box with Gradle 7.2+ (hence the hacks on https://github.com/apache/cassandra-sidecar/compare/trunk...isaacreath:CASSSIDECAR-167?expand=1#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R113-R135). My goal was to ensure as much compatibility with how the `bin/cassandra-sidecar` script was working today, but as a follow up I think we should probably reconsider using it in the default JVM options.


frankgh:
Thanks for the PR @isaacreath!

The Cassandra project does not support Windows, so I think we should exclude the bat file from the repo. What are your thoughts on that?

I will take a look at the PR sometime this week

isaacreath:
Yeah I agree, I'll update this PR to ensure that Gradle does not generate a bat file.

bbotella:
Thanks a lot for the patch!

I will take a look at it today as well (I'm not a committer, so we will still need @frankgh).

build.gradle
```diff
@@ -102,18 +105,35 @@ application {
...
     mainClassName = "org.apache.cassandra.sidecar.CassandraSidecarDaemon"
     applicationName = 'cassandra-sidecar'
-    // Config file location should be in file:/// format for local files,
-    def confFile = "file:" + File.separator + File.separator + "APP_HOME_TO_REPLACE/conf/sidecar.yaml"
-    applicationDefaultJvmArgs = ["-Dsidecar.logdir=./logs",
-                                 "-Dsidecar.config=" + confFile,
-                                 "-Dlogback.configurationFile=./conf/logback.xml",
-                                 "-Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory",
-                                 "-javaagent:APP_HOME_TO_REPLACE/agents/jolokia-jvm-1.6.0-agent.jar=port=7777,host=localhost"]
+    applicationDefaultJvmArgs = ["PLACEHOLDER"]
```
bbotella, line 100: Just for my education, what is this placeholder here and on line 134?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 #Fri May 01 11:14:42 PDT 2020
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.2.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.11-all.zip
```
bbotella: Now that we are on it, should we go with 8.11.1 newer version?

build.gradle
```diff
@@ -38,11 +38,11 @@ plugins {
     id 'application'
 
     // since we're using a specific version here, we delay applying the plugin till the all projects
-    id "com.github.spotbugs" version "3.0.0" apply false
+    id "com.github.spotbugs" version "5.0.14" apply false
```
frankgh: when I tried upgrading gradle, I was able to get to version `6.20.0` for spotbugs, but I needed a low number of changes on `checkstyle.xml`

build.gradle
```diff
@@ -173,14 +193,22 @@ tasks.named('idea').configure {
 
 // Lets copy the distributions from build/install directory to /bin and /lib
 // directories to be aligned with C* distribution format
-tasks.register('copyDist', Copy) {
-    from "$buildDir/install/$applicationName"
-    into "$projectDir"
+tasks.register('copyDist') {
```
frankgh: we no longer have a `copyDist` task, so this can be removed.

frankgh (commented):
Can you rebase from the latest? Two things that I think are missing are the updated `gradlew` script and the updated gradle jar under `gradle/wrapper/gradle-wrapper.jar`. The rest looks good. Can you also share a link to CI?

build.gradle
isaacreath: When the `startScripts` Gradle task runs, it will create a variable in the start script called `DEFAULT_JVM_OPTIONS` which is set to the values in `applicaitonDefaultJvmArgs`. This variable is purposefully created in a way such that the environment variables that we want to embed into the `DEFAULT_JVM_OPTIONS` will not be evaluated.

The way around this is to generate `DEFAULT_JVM_OPTIONS='PLACEHOLDER'` in the start script, then modify that value after the start script is generated and replace `'PLACEHOLDER'` with the actual default JVM options.

build.gradle
isaacreath: I'm guessing you mean `6.0.20` not `6.20.0`?

frankgh: yes `"6.0.20"` is what I meant. I think it's fine to move to `5.0.14`, so nothing actionable here

build.gradle
```diff
@@ -102,18 +97,35 @@ application {
...
+        //
+        // Replacing the placeholder after the script has been generated will preserve the semantics from Gradle < 7.2.
+        // See this Gradle issue for more context: https://github.com/gradle/gradle/issues/18170.
+
+        def defaultJvmOpts = ["-Dsidecar.logdir=./logs",
+                              "-Dsidecar.config=" + confFile,
+                              "-Dlogback.configurationFile=./conf/logback.xml",
+                              "-Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory",
+                              "-javaagent:\${APP_HOME}/agents/jolokia-jvm-1.6.0-agent.jar=port=7777,host=localhost"]
+        unixScript.text = unixScript.text.replace("DEFAULT_JVM_OPTS='\"PLACEHOLDER\"'",
```
frankgh, line 126: verified that this is generating something that looks like this:

```
DEFAULT_JVM_OPTS='"-Dsidecar.logdir=./logs" "-Dsidecar.config=file://${APP_HOME}/conf/sidecar.yaml" "-Dlogback.configurationFile=./conf/logback.xml" "-Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory" "-javaagent:${APP_HOME}/agents/jolokia-jvm-1.6.0-agent.jar=port=7777,host=localhost"'
```

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,7 @@
-#Fri May 01 11:14:42 PDT 2020
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.2.1-all.zip
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.9-bin.zip
```
frankgh, line 3: we might also want to downgrade the jar for these to be in sync and not sure if the `gradlew` needs downgrading as well.

frankgh (commented):
The PR looks good to me. Let me know when you have CI results ready so we can work on merging this.

bbotella (approved):
Thanks for the patch! +1 (nb)

isaacreath:
> The PR looks good to me. Let me know when you have CI results ready so we can work on merging this.

Working through an issue that seems to have popped up on CI right now:

https://app.circleci.com/pipelines/circleci/2pKxPuEmc4jySZHiGcJrJE/LTM5bKmo57v3FcPSxgh5mk/2/workflows/e26029fa-ab66-40d3-870d-18d59c721a59/jobs/19/parallel-runs/0/steps/0-107?invite=true#step-107-54412_0

isaacreath:
Got CI passing: https://app.circleci.com/pipelines/circleci/2pKxPuEmc4jySZHiGcJrJE/LTM5bKmo57v3FcPSxgh5mk/4/workflows/8f21d3e0-5548-4647-b82e-7e5970b4e519

yifan-c (approved):

isaacreath:
Thank you everyone for your review!

PR #156, opened 2024-12-05
CASSSIDECAR-170: Reprepare statements when sidecar reconnects to Cass…

yifan-c:
…andra

Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-170

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
@@ -35,6 +35,7 @@ import org.apache.cassandra.sidecar.exceptions.SidecarSchemaModificationExceptio
 import org.apache.cassandra.sidecar.metrics.SchemaMetrics;
 
 import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_CASSANDRA_CQL_READY;
+import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_CASSANDRA_DRIVER_CLOSED;
 import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_SERVER_STOP;
 import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_SIDECAR_SCHEMA_INITIALIZED;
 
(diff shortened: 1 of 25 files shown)


server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -69,21 +71,21 @@ public SidecarSchema(Vertx vertx,
...
-        }
-        else
+        configureSidecarServerEventListenersMaybe();
+    }
+
+    private void configureSidecarServerEventListenersMaybe()
+    {
+        if (!this.schemaKeyspaceConfiguration.isEnabled())
         {
             LOGGER.info("Sidecar schema is disabled!");
```
bbotella: Nit: Maybe being more descriptive with the logger message? I think we could answer the question: "so what?". In this case, what do you think about something like: `Sidecar schema is disabled. Skipping configuring the sidecar event listeners`?

bbotella:
+1 (nb)

server/src/test/integration/org/apache/cassandra/sidecar/routes/ConnectedClientStatsHandlerIntegrationTest.java
```diff
@@ -105,7 +107,8 @@ void retrieveClientStatsListConnectionsWithKeyspace(VertxTestContext context)
     void retrieveClientStatsMultipleConnections(VertxTestContext context)
     throws Exception
     {
-        // Creates an additional connection pair
+        // create an extra pair of connections
+        sidecarTestContext.buildNewCqlSessionProvider().get();
```
yifan-c, line 111: The test was incorrectly depending on building new cluster object when creating keyspace. It get fixed in this patch, and an explicit method to build new cluster object is added for clarity.

server/src/test/integration/org/apache/cassandra/sidecar/testing/CassandraSidecarTestContext.java
```diff
@@ -197,6 +197,7 @@ public InstancesConfig refreshInstancesConfig()
     {
         // clean-up any open sessions or client resources
         close();
+        closeSessionProvider();
```
yifan-c, line 200: `setInstancesConfig` create a new instance of session provider. Therefore, close the current one before creating a new one.

yifan-c:
Closing as no longer needed by #165

PR #159, opened 2024-12-10
CASSSIDECAR-173: Endpoint to retrieve C* gossip status

skoppu22:
https://issues.apache.org/jira/browse/CASSSIDECAR-173
Developing a new Sidecar endpoint to retrieve gossip running status of Cassandra, so the end user can use this endpoint without having to invoke C* JMX call. This endpoint will have authorization checks enabled once authorization feature is added to the Sidecar.

Circle CI run: https://app.circleci.com/pipelines/github/skoppu22/cassandra-sidecar?branch=gossips2

### DIFF
--- a/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
+++ b/adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
@@ -233,4 +233,13 @@ public class CassandraStorageOperations implements StorageOperations
         StorageJmxOperations ssProxy = jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME);
         ssProxy.decommission(force);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean isGossipRunning()
+    {
+        return jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME).isGossipRunning();
+    }
 }
--- a/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
+++ b/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
@@ -220,6 +220,20 @@ public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExt
         return executor.executeRequestAsync(requestBuilder().gossipInfoRequest().build());
     }
 
+
+    /**
+     * Executes the GET gossip health request using the default retry policy and configured selection policy
+     * @param instance the instance where the request will be executed
+     * @return a completable future with gossip health response
+     */
+    public CompletableFuture<HealthResponse> gossipHealth(SidecarInstance instance)
+    {
+        return executor.executeRequestAsync(requestBuilder()
+                                            .singleInstanceSelectionPolicy(instance)
+                                            .gossipHealthRequest()
+                                            .build());
+    }
+
     /**
      * Executes the time skew request using the default retry policy and configured selection policy
      *
--- a/server-common/src/main/java/org/apache/cassandra/sidecar/common/server/StorageOperations.java
+++ b/server-common/src/main/java/org/apache/cassandra/sidecar/common/server/StorageOperations.java
@@ -114,4 +114,9 @@ public interface StorageOperations
      * @param force force decommission, bypassing RF checks, when this flag is set
      */
     void decommission(boolean force);
+
+    /**
+     * @return returns true if gossip is running, false otherwise
+     */
+    boolean isGossipRunning();
 }
--- a/server/src/main/java/org/apache/cassandra/sidecar/routes/GossipHealthHandler.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/routes/GossipHealthHandler.java
@@ -58,9 +58,9 @@ public class GossipHealthHandler extends AbstractHandler<Void>
     {
         StorageOperations operations = metadataFetcher.delegate(host).storageOperations();
         executorPools.service()
-                         .executeBlocking(operations::isGossipRunning)
-                         .onSuccess(isGossipRunning -> context.json(isGossipRunning ? OK_STATUS : NOT_OK_STATUS))
-                         .onFailure(cause -> processFailure(cause, context, host, remoteAddress, request));
+                     .executeBlocking(operations::isGossipRunning)
+                     .onSuccess(isGossipRunning -> context.json(isGossipRunning ? OK_STATUS : NOT_OK_STATUS))
+                     .onFailure(cause -> processFailure(cause, context, host, remoteAddress, request));
     }
 
     /**
--- a/server/src/main/java/org/apache/cassandra/sidecar/routes/GossipHealthHandler.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/routes/GossipHealthHandler.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.routes;
+
+import com.google.inject.Inject;
+import io.vertx.core.http.HttpServerRequest;
+import io.vertx.core.net.SocketAddress;
+import io.vertx.ext.web.RoutingContext;
+import org.apache.cassandra.sidecar.common.server.StorageOperations;
+import org.apache.cassandra.sidecar.concurrent.ExecutorPools;
+import org.apache.cassandra.sidecar.utils.InstanceMetadataFetcher;
+
+import static org.apache.cassandra.sidecar.server.MainModule.NOT_OK_STATUS;
+import static org.apache.cassandra.sidecar.server.MainModule.OK_STATUS;
+
+
+/**
+ * Handler to retrieve gossip health
+ */
+public class GossipHealthHandler extends AbstractHandler<Void>
+{
+    /**
+     * Constructs a handler with the provided {@code metadataFetcher}
+     * @param metadataFetcher the metadata fetcher
+     * @param executorPools   executor pools for blocking executions
+     */
+    @Inject
+    protected GossipHealthHandler(InstanceMetadataFetcher metadataFetcher, ExecutorPools executorPools)
+    {
+        super(metadataFetcher, executorPools, null);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void handleInternal(RoutingContext context,
+                               HttpServerRequest httpRequest,
+                               String host,
+                               SocketAddress remoteAddress,
+                               Void request)
+    {
+        StorageOperations operations = metadataFetcher.delegate(host).storageOperations();
+        executorPools.service()
+                         .executeBlocking(operations::isGossipRunning)
+                         .onSuccess(isGossipRunning -> context.json(isGossipRunning ? OK_STATUS : NOT_OK_STATUS))
+                         .onFailure(cause -> processFailure(cause, context, host, remoteAddress, request));
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected Void extractParamsOrThrow(RoutingContext context)
+    {
+        return null;
+    }
+}
--- a/server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
@@ -113,6 +113,7 @@ import org.apache.cassandra.sidecar.routes.CassandraHealthHandler;
 import org.apache.cassandra.sidecar.routes.ConnectedClientStatsHandler;
 import org.apache.cassandra.sidecar.routes.DiskSpaceProtectionHandler;
 import org.apache.cassandra.sidecar.routes.FileStreamHandler;
+import org.apache.cassandra.sidecar.routes.GossipHealthHandler;
 import org.apache.cassandra.sidecar.routes.GossipInfoHandler;
 import org.apache.cassandra.sidecar.routes.JsonErrorHandler;
 import org.apache.cassandra.sidecar.routes.KeyspaceRingHandler;
@@ -319,6 +320,7 @@ public class MainModule extends AbstractModule
                               TokenRangeReplicaMapHandler tokenRangeHandler,
                               LoggerHandler loggerHandler,
                               GossipInfoHandler gossipInfoHandler,
+                              GossipHealthHandler gossipHealthHandler,
                               TimeSkewHandler timeSkewHandler,
                               NodeSettingsHandler nodeSettingsHandler,
                               SSTableUploadHandler ssTableUploadHandler,
@@ -387,6 +389,9 @@ public class MainModule extends AbstractModule
         // authenticated users can also read node settings information.
         router.get(ApiEndpointsV1.NODE_SETTINGS_ROUTE)
               .handler(nodeSettingsHandler);
+
+        router.get(ApiEndpointsV1.GOSSIP_HEALTH_ROUTE)
+              .handler(gossipHealthHandler);
 
         router.get(ApiEndpointsV1.TIME_SKEW_ROUTE)
               .handler(timeSkewHandler);
--- a/server/src/test/java/org/apache/cassandra/sidecar/routes/CommonTest.java
+++ b/server/src/test/java/org/apache/cassandra/sidecar/routes/CommonTest.java
@@ -90,9 +90,9 @@ public class CommonTest
         @Singleton
         public InstancesMetadata instanceConfig()
         {
-            final int instanceId = 100;
-            final String host = "127.0.0.1";
-            final InstanceMetadata instanceMetadata = mock(InstanceMetadata.class);
+            int instanceId = 100;
+            String host = "127.0.0.1";
+            InstanceMetadata instanceMetadata = mock(InstanceMetadata.class);
             when(instanceMetadata.host()).thenReturn(host);
             when(instanceMetadata.port()).thenReturn(9042);
             when(instanceMetadata.id()).thenReturn(instanceId);
--- a/server/src/test/java/org/apache/cassandra/sidecar/routes/CommonTest.java
+++ b/server/src/test/java/org/apache/cassandra/sidecar/routes/CommonTest.java
@@ -1,0 +1,110 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.routes;
+
+import java.util.Collections;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.inject.AbstractModule;
+import com.google.inject.Guice;
+import com.google.inject.Injector;
+import com.google.inject.Module;
+import com.google.inject.Provides;
+import com.google.inject.Singleton;
+import com.google.inject.util.Modules;
+import io.vertx.core.Vertx;
+import io.vertx.junit5.VertxTestContext;
+import org.apache.cassandra.sidecar.TestModule;
+import org.apache.cassandra.sidecar.cluster.CassandraAdapterDelegate;
+import org.apache.cassandra.sidecar.cluster.InstancesMetadata;
+import org.apache.cassandra.sidecar.cluster.instance.InstanceMetadata;
+import org.apache.cassandra.sidecar.server.MainModule;
+import org.apache.cassandra.sidecar.server.Server;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Common test code to setup/teardown the server and populate test instances metadata
+ */
+public class CommonTest
+{
+    static final Logger LOGGER = LoggerFactory.getLogger(CommonTest.class);
+    Vertx vertx;
+    Server server;
+    CassandraAdapterDelegate delegate = mock(CassandraAdapterDelegate.class);
+
+    @BeforeEach
+    void before() throws InterruptedException
+    {
+        Injector injector;
+        Module testOverride = Modules.override(new TestModule())
+                                     .with(new CommonTestModule());
+        injector = Guice.createInjector(Modules.override(new MainModule())
+                                               .with(testOverride));
+        vertx = injector.getInstance(Vertx.class);
+        server = injector.getInstance(Server.class);
+        VertxTestContext context = new VertxTestContext();
+        server.start()
+              .onSuccess(s -> context.completeNow())
+              .onFailure(context::failNow);
+        context.awaitCompletion(5, TimeUnit.SECONDS);
+    }
+
+    @AfterEach
+    void after() throws InterruptedException
+    {
+        CountDownLatch closeLatch = new CountDownLatch(1);
+        server.close().onSuccess(res -> closeLatch.countDown());
+        if (closeLatch.await(60, TimeUnit.SECONDS))
+            LOGGER.info("Close event received before timeout.");
+        else
+            LOGGER.error("Close event timed out.");
+    }
+
+    class CommonTestModule extends AbstractModule
+    {
+        @Provides
+        @Singleton
+        public InstancesMetadata instanceConfig()
+        {
+            final int instanceId = 100;
+            final String host = "127.0.0.1";
+            final InstanceMetadata instanceMetadata = mock(InstanceMetadata.class);
+            when(instanceMetadata.host()).thenReturn(host);
+            when(instanceMetadata.port()).thenReturn(9042);
+            when(instanceMetadata.id()).thenReturn(instanceId);
+            when(instanceMetadata.stagingDir()).thenReturn("");
+            when(instanceMetadata.delegate()).thenReturn(delegate);
+
+            InstancesMetadata mockInstancesMetadata = mock(InstancesMetadata.class);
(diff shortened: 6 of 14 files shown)


client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipStatusResponse.java
```diff
@@ -0,0 +1,46 @@
...
+        this.gossipRunning = gossipRunning;
+    }
+
+    @JsonProperty("gossipRunning")
+    public boolean gossipRunning()
+    {
+        return gossipRunning;
+    }
+
+    private final boolean gossipRunning;
```
bbotella: Define this at the beginning of the class

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -380,6 +382,9 @@ public Router vertxRouter(Vertx vertx,
         router.get(ApiEndpointsV1.GOSSIP_INFO_ROUTE)
               .handler(gossipInfoHandler);
 
+        router.get(ApiEndpointsV1.GOSSIP_STATUS_ROUTE)
```
bbotella: It is really unfortunate having to add an entire new API when we already have the GOSSIP_INFO_ROUTE endpoint? I even see that the GossipInfoResponse already has a deprecated STATUS field. 

I understand that we are trying to make this call more efficient by only checking the gossip status. I wonder, could we add a filtering parameter that helps us only executing the part of the handler that we want?

I'm thinking of a filtering pattern like:
`/api/v1/cassandra/gossip?filter=['status']` or something along those lines?

That way, and finding out why/where it was deprecated, maybe we can hijack it to use this `storageOperations::isGossipRunning` can save us from having to add this new API

skoppu22: ?filter=<> doesn't sound right in this case, as filter is usually used to chose some/partial fields of the output which was returned without using filter. In this case /gossip endpoint doesn't return 'status', so we can't `filter` status from it. Let me see what are alternatives and will get back to you.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipStatusResponse.java
skoppu22: Most of Sidecar request/response classes following this style. I am also used to the style of defining at the beginning of the class, but just trying to be inline with existing classes. May be we can start changing the style now, as easy to find all variables right at the beginning of the class.

frankgh: > Most of Sidecar request/response classes following this style

This is not true, I'm not sure which class you are looking at, but all classes under `client-common/src/main/java/org/apache/cassandra/sidecar/common/response` define fields at the beginning of the class. And we should follow the Cassandra code structure https://cassandra.apache.org/_/development/code_style.html under Code Structure section > Class Layout.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipStatusResponse.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    public GossipStatusResponse(@JsonProperty("gossipRunning") boolean gossipRunning)
+    {
+        this.gossipRunning = gossipRunning;
+    }
+
+    @JsonProperty("gossipRunning")
+    public boolean gossipRunning()
+    {
+        return gossipRunning;
```
frankgh: returning a boolean becomes inflexible if you need to support a use case where you have more than one value. I suggest we change this to a payload that looks like 

```json
{
   "status": "RUNNING"
}
```

and 

```json
{
   "status": "NOT_RUNNING"
}
```

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -215,6 +216,16 @@ public CompletableFuture<GossipInfoResponse> gossipInfo()
...
     }
 
+    /**
+     * Executes the GET gossip status request using the default retry policy and configured selection policy
+     *
+     * @return a completable future with gossip status response
+     */
+    public CompletableFuture<GossipStatusResponse> gossipStatus()
+    {
+        return executor.executeRequestAsync(requestBuilder().gossipStatusRequest().build());
```
frankgh: this request should be instance-specific

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/StorageOperations.java
```diff
@@ -102,4 +103,9 @@ default void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tab
     {
         outOfRangeDataCleanup(keyspace, table, 1);
     }
+
+    /**
+     * @return returns gossip status
+     */
+    GossipStatusResponse isGossipRunning();
```
frankgh: Maybe better aligned with the action we are trying to perform
```suggestion
    GossipStatusResponse gossipStatus();
```

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
```diff
@@ -331,4 +332,16 @@ public static String extractHostAddressWithoutPort(String address)
...
+    {
+        CassandraAdapterDelegate delegate = this.metadataFetcher.delegate(host);
+        StorageOperations storageOperations = delegate == null ? null : delegate.storageOperations();
+        if (storageOperations == null)
+        {
+            throw cassandraServiceUnavailable();
+        }
+
+        return storageOperations;
+    }
```
frankgh: we already have a method that does something similar in this class, we should just generalize that method. I suggest renaming `ifMetricsOpsAvailable` , and the code would look like this instead

```
    protected <P> void ifAvailableFromDelegate(RoutingContext context,
                                               String host,
                                               Function<CassandraAdapterDelegate, P> mapper,
                                               Consumer<P> ifAvailable)
    {
        CassandraAdapterDelegate delegate = metadataFetcher.delegate(host);
        if (delegate == null)
        {
            context.fail(cassandraServiceUnavailable());
            return;
        }
        P applied = mapper.apply(delegate);
        if (applied == null)
        {
            context.fail(cassandraServiceUnavailable());
            return;
        }
        ifAvailable.accept(applied);
    }
```

The existing callsite would change to this: 

```
ifAvailableFromDelegate(context, host, CassandraAdapterDelegate::metricsOperations, operations -> {
```

server/src/main/java/org/apache/cassandra/sidecar/routes/GossipStatusHandler.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    @Override
+    public void handleInternal(RoutingContext context,
+                               HttpServerRequest httpRequest,
+                               String host,
+                               SocketAddress remoteAddress,
+                               Void request)
+    {
+        StorageOperations storageOperations = getStorageOperations(host);
+        logger.debug("Retrieving gossip status, remoteAddress={}, instance={}",
```
frankgh: this log entry is redundant, we already log 

```
logger.debug("{} received request={}, remoteAddress={}, instance={}",
                         this.getClass().getSimpleName(), requestParams, remoteAddress, host);
```

for all requests in handlers of the `AbstractHandler` type.

server/src/test/integration/org/apache/cassandra/sidecar/routes/GossipStatusHandlerIntegrationTest.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Test GET gossip status endpoint with C* container
+ */
+@ExtendWith(VertxExtension.class)
+public class GossipStatusHandlerIntegrationTest extends IntegrationTestBase
+{
+    private static final String testRoute = "/api/v1/cassandra/gossip/status";
+
+    @CassandraIntegrationTest()
+    void testGossipDisabled(CassandraTestContext context, VertxTestContext testContext)
```
frankgh: can we combine both test cases into a single test case? these are expensive to spin up, so whenever possible we should reuse the in-jvm dtest apparatus available .

server/src/test/java/org/apache/cassandra/sidecar/routes/JmxCommonTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Common test code to setup/teardown the server, populate test instances metadata and mocks for JMX calls
+ */
+class JmxCommonTest
+{
+    static final Logger LOGGER = LoggerFactory.getLogger(JmxCommonTest.class);
```
frankgh: if we already have an in-jvm dtest this becomes less useful IMO. There is no need to do mocking if we have the actual server running

frankgh (commented):
Added some comments to the PR

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipStatusResponse.java
skoppu22: This one I followed https://github.com/apache/cassandra-sidecar/blob/trunk/client-common/src/main/java/org/apache/cassandra/sidecar/common/response/ConnectedClientStatsResponse.java

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipStatusResponse.java
skoppu22: Client need to rely on deterministic values to parse the response and take actions accordingly. String "RUNNING" or "NOT_RUNNING" can break the client even with any minor changes or extra spaces etc in the output. Instead, if it is possible to return an enum as rest API response, we can use an enum. Otherwise boolean value true/false would be deterministic for the caller to parse the output

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
skoppu22: As we discussed offline, we cannot infer gossip running status from STATUS (or STATUS_WITH_PORT) fields of gossip info. Hence we cannot use gossip/filter=status for gossip running status, as that filter can be mistaken for STATUS field of gossip info. Will continue with gossip/status endpoint (Other option could be gossip/runningStatus, but sounds too verbose)

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
skoppu22: Nice

server/src/test/integration/org/apache/cassandra/sidecar/routes/GossipStatusHandlerIntegrationTest.java
skoppu22: OK

server/src/test/java/org/apache/cassandra/sidecar/routes/JmxCommonTest.java
skoppu22: - We do more test scenarios using mock, especially related to failure cases as mock makes it easy to throw exceptions or return invalid values
- Can we get code coverage measured with in-jvm dtests? For example, someone using open source Sidecar and hooking to their build systems, can they get code coverage numbers with in-jvm dtests?

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
skoppu22: Thanks for noticing this

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipStatusResponse.java
bbotella: Well, both enum or boolean flags are technically strings that need to be parsed/interpreted by the client as well. I do like the idea @frankgh  is proposing of having the enum with two possible states, that allows us to easily expand the amount of information we can provide.

skoppu22: Why boolean is treated as a string? Client can prepare GossipStatusResponse from the json received in the response and can get boolean using GossipStatusResponse.gossipRunning()

bbotella: I meant that a response payload is a bunch of bytes that need to be parsed anyway, even if it is a json payload, it needs to be parsed. The client is in charge of parsing it and interpreting fields as booleans, strings, or anything else.

That said, I do agree with @frankgh in which the flexibility added by returning that enum is preferable in this case. On the client side, it is client's responsibility to parse the response correctly. In our current implementation, we control de java client for instance. We are also adding OpenAPI documentation that will help clients parse the given responses as part of this [PR](https://github.com/apache/cassandra-sidecar/pull/164). 

Long story short, I don't think returning a `"RUNNING"` response is making things harder for the client than returning a `true`, and it gives us flexibility to grow the API in the future.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipStatusResponse.java
frankgh: This is something we should fix

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipStatusResponse.java
skoppu22: Updated to String

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
skoppu22: Looks like this is no longer needed as you moved null checks into getters

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipStatusResponse.java
```diff
@@ -0,0 +1,46 @@
...
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class GossipStatusResponse
+{
+    private final String gossipRunningStatus;
+
+    /**
+     * Constructs a response with gossip status
+     * @param gossipRunningStatus running status of gossip
+     */
+    public GossipStatusResponse(@JsonProperty("gossipRunningStatus") String gossipRunningStatus)
```
frankgh: I think `status` should be sufficient. We don't have a precedent for qualifying the field name like this.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipStatusResponse.java
```diff
@@ -0,0 +1,46 @@
...
+    {
+        this.gossipRunningStatus = gossipRunningStatus;
+    }
+
+    @JsonProperty("gossipRunningStatus")
+    public String gossipRunningStatus()
```
frankgh: should we add a utility method here `isRunning` that returns `"RUNNING".equals(gossipRunningStatus)`. Similar to what you had in an earlier iteration of this PR.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/StorageOperations.java
```diff
@@ -102,4 +103,9 @@ default void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tab
     {
         outOfRangeDataCleanup(keyspace, table, 1);
     }
+
+    /**
+     * @return returns gossip status
```
frankgh:
```suggestion
     * @return the gossip status
```

server/src/main/java/org/apache/cassandra/sidecar/routes/GossipStatusHandler.java
```diff
@@ -0,0 +1,71 @@
...
+import io.vertx.core.http.HttpServerRequest;
+import io.vertx.core.net.SocketAddress;
+import io.vertx.ext.web.RoutingContext;
+import org.apache.cassandra.sidecar.common.server.StorageOperations;
+import org.apache.cassandra.sidecar.concurrent.ExecutorPools;
+import org.apache.cassandra.sidecar.utils.InstanceMetadataFetcher;
+
+
+/**
+ * Handler for to retrieve gossip status
```
frankgh:
```suggestion
 * Handles gossip status retrieval
```

server/src/test/integration/org/apache/cassandra/sidecar/routes/GossipStatusHandlerIntegrationTest.java
```diff
@@ -0,0 +1,74 @@
...
+                     .expect(ResponsePredicate.SC_OK)
+                     .send()
+                     .compose(response -> {
+                         assertThat(response.statusCode()).isEqualTo(OK.code());
+                         return Future.succeededFuture(response.bodyAsJson(GossipStatusResponse.class));
+                     });
+    }
+
+    @CassandraIntegrationTest()
+    void testGossipStatus(CassandraTestContext context, VertxTestContext testContext)
```
frankgh: nice!

server/src/test/java/org/apache/cassandra/sidecar/routes/GossipStatusHandlerTest.java
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * Tests for {@link GossipInfoHandler}
+ */
+@ExtendWith(VertxExtension.class)
+public class GossipStatusHandlerTest extends JmxCommonTest
+{
+    private static final String testRoute = "/api/v1/cassandra/gossip/status";
+
+    @Test
+    void testGossipRunning(VertxTestContext context)
```
frankgh: this test is redundant, not sure what we gain here.

server/src/test/java/org/apache/cassandra/sidecar/routes/GossipStatusHandlerTest.java
```diff
@@ -0,0 +1,104 @@
...
+        when(storageOperations.gossipStatus()).thenReturn(new GossipStatusResponse("RUNNING"));
+
+        WebClient client = WebClient.create(vertx);
+        client.get(server.actualPort(), "127.0.0.1", testRoute)
+              .expect(ResponsePredicate.SC_OK)
+              .send(context.succeeding(response -> verifyValidResponse(context, response, "RUNNING")));
+    }
+
+    @Test
+    void testGossipNotRunning(VertxTestContext context)
```
frankgh: this is also redundant not sure what exactly we are testing here.

server/src/test/java/org/apache/cassandra/sidecar/routes/GossipStatusHandlerTest.java
```diff
@@ -0,0 +1,104 @@
...
+        when(storageOperations.gossipStatus()).thenReturn(new GossipStatusResponse("NOT_RUNNING"));
+
+        WebClient client = WebClient.create(vertx);
+        client.get(server.actualPort(), "127.0.0.1", testRoute)
+              .expect(ResponsePredicate.SC_OK)
+              .send(context.succeeding(response -> verifyValidResponse(context, response, "NOT_RUNNING")));
+    }
+
+    @Test
+    void testWithInstanceId(VertxTestContext context)
```
frankgh: also, this test doesn't make a lot of sense to me. My problem with the test is that in the setup `JmxCommonTest` you share the `delegate` and `storageOperations` among 3 instances. This setup is problematic because in the real execution of sidecar each instance will have its own `delegate` and all the operations associated with a single delegate. I would drop this test, we already cover accessing different instances with the query parameter, so I would consider testing this outside of the scope of this test.

server/src/test/java/org/apache/cassandra/sidecar/routes/GossipStatusHandlerTest.java
```diff
@@ -0,0 +1,104 @@
...
+        WebClient client = WebClient.create(vertx);
+        client.get(server.actualPort(), "127.0.0.1", testRoute)
+              .expect(ResponsePredicate.SC_INTERNAL_SERVER_ERROR)
+              .send(context.succeeding(response -> {
+                  assertThat(response.statusCode()).isEqualTo(INTERNAL_SERVER_ERROR.code());
+                  context.completeNow();
+              }));
+    }
+
+    private void verifyValidResponse(VertxTestContext context, HttpResponse<Buffer> response, String expectedValue)
```
frankgh: maybe call this `verifyAndComplete`

server/src/test/java/org/apache/cassandra/sidecar/routes/JmxCommonTest.java
```diff
@@ -0,0 +1,123 @@
...
+            LOGGER.error("Close event timed out.");
+    }
+
+    class JmxTestModule extends AbstractModule
+    {
+        @Provides
+        @Singleton
+        public InstancesMetadata instanceConfig()
+        {
+            int instanceId1 = 100;
```
frankgh: I think we should probably model this correctly if we are meaning to use this in other tests.

frankgh (commented):
Looking good, I have a few additional comments.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipStatusResponse.java
yifan-c: Can it be enforced with code style? We want to layout the class in this order. 
```
// variables 
public static 
protected static
private static
public 
protected
private 
// methods 
// same ordering goes again
// visible for testing methods
// then, inner class
```

yifan-c (changes requested):
Thanks for the contribution.

First of all, please add an entry to `CHANGES.txt`.

Second, the patch is to add an API to check whether gossip is enabled or disabled in the companion Cassandra node. There are existing health APIs that check the status of corresponding components. See the following code block for example. 

```
GET /api/v1/cassandra/native/__health
{"status" : "OK | NOT_OK"}

GET /api/v1/cassandra/jmx/__health
{"status" : "OK | NOT_OK"}
```

I think we should be consistent with the existing pattern. In other words, the API should be this.

```
GET /api/v1/cassandra/gossip/__health
{"status" : "OK | NOT_OK"}
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -213,4 +214,11 @@ public void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tabl
         jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME)
                  .forceKeyspaceCleanup(concurrency, keyspace, table);
     }
+
+    @Override
+    public GossipStatusResponse gossipStatus()
+    {
+        return new GossipStatusResponse(jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME)
+                                                 .isGossipRunning() ? "RUNNING" : "NOT_RUNNING");
```
bbotella: What do you think about moving `"RUNNING"` and `"NOT_RUNNING"` to an enum?

frankgh:
> I think we should be consistent with the existing pattern. In other words, the API should be this.

I agree with Yifan, we should try to stick closer to the existing patterns. +1 on this:

```
GET /api/v1/cassandra/gossip/__health
{"status" : "OK | NOT_OK"}
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
bbotella: I actually agree with @yifan-c  suggestion. So, what about having the enum for the `"OK"` and `"NOT_OK"`?

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipStatusResponse.java
skoppu22: Just calling it 'status' can be mistaken with STATUS field of gossip info, as we discussed before regarding this field. Gossip running status just tells is gossip thread running in C*. Where as STATUS field has different meaning. So we need to be clear that this is gossip thread running status

skoppu22:
> Thanks for the contribution.
> 
> First of all, please add an entry to `CHANGES.txt`.
> 
> Second, the patch is to add an API to check whether gossip is enabled or disabled in the companion Cassandra node. There are existing health APIs that check the status of corresponding components. See the following code block for example.
> 
> ```
> GET /api/v1/cassandra/native/__health
> {"status" : "OK | NOT_OK"}
> 
> GET /api/v1/cassandra/jmx/__health
> {"status" : "OK | NOT_OK"}
> ```
> 
> I think we should be consistent with the existing pattern. In other words, the API should be this.
> 
> ```
> GET /api/v1/cassandra/gossip/__health
> {"status" : "OK | NOT_OK"}
> ```


This API is returning C* gossip thread running status (just tells is gossip thread running or not). Having thread running doesn't mean gossip health is OK. So I wouldn't prefer to call it 'gossip health'. Gossip thread may be running, but gossip state might be not healthy at C* side

server/src/test/java/org/apache/cassandra/sidecar/routes/GossipStatusHandlerTest.java
skoppu22: The gain depends on answer to https://github.com/apache/cassandra-sidecar/pull/159/files#r1886797448

Do we get code coverage with in-jvm tests? If yes, we can remove this test

server/src/test/java/org/apache/cassandra/sidecar/routes/JmxCommonTest.java
skoppu22: We can remove multi instance stuff from this. But I see almost all tests in this directory are having module extending AbstractModule and using mocks, I believe we can move common code to a common test framework file and reuse in most of tests

server/src/test/java/org/apache/cassandra/sidecar/routes/CommonTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.apache.cassandra.sidecar.server.MainModule;
+import org.apache.cassandra.sidecar.server.Server;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Common test code to setup/teardown the server and populate test instances metadata
+ */
+public class CommonTest
```
skoppu22, line 52: This class contains common test setup and teardown which repeated in almost all test files. For now I am using this common in GossipInfoHandlerTest and GossipHealthHandlerTest. Will open a separate PR to modify other test files to use this common test class wherever possible

server/src/main/java/org/apache/cassandra/sidecar/routes/GossipHealthHandler.java
```diff
@@ -0,0 +1,74 @@
...
+                               HttpServerRequest httpRequest,
+                               String host,
+                               SocketAddress remoteAddress,
+                               Void request)
+    {
+        StorageOperations operations = metadataFetcher.delegate(host).storageOperations();
+        executorPools.service()
+                         .executeBlocking(operations::isGossipRunning)
+                         .onSuccess(isGossipRunning -> context.json(isGossipRunning ? OK_STATUS : NOT_OK_STATUS))
+                         .onFailure(cause -> processFailure(cause, context, host, remoteAddress, request));
```
yifan-c: please align the dots. 
I thought it is already enforced by checkstyle, but not. I will make sure the rule is added.

server/src/test/java/org/apache/cassandra/sidecar/routes/CommonTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+    class CommonTestModule extends AbstractModule
+    {
+        @Provides
+        @Singleton
+        public InstancesMetadata instanceConfig()
+        {
+            final int instanceId = 100;
+            final String host = "127.0.0.1";
+            final InstanceMetadata instanceMetadata = mock(InstanceMetadata.class);
```
yifan-c: the `final` modifier in the method scope is usually omitted. See https://cassandra.apache.org/_/development/code_style.html#boilerplate

> never declare local variables and parameters final.

yifan-c (commented):
Looks good to me. There are a few comments on codestyle. Please update. 

Besides that, please add an entry to CHANGES.txt

server/src/test/java/org/apache/cassandra/sidecar/routes/CommonTest.java
skoppu22: This code is moved here from GossipInfoHandlerTest, don't know why were they marked final. Removed them

frankgh (approved):
+1

bbotella:
+1 (nb)

bbotella (approved):

yifan-c (approved):

PR #160, opened 2024-12-11
CASSSIDECAR-174: Mechanism to have a reduced number of Sidecar instan…

frankgh:
…ces run operations

Patch by Francisco Guerrero; Reviewed by TBD for CASSSIDECAR-174


server/src/test/java/org/apache/cassandra/sidecar/server/ServerTest.java
```diff
@@ -268,7 +268,8 @@ void unrecognizedAuthenticationHandlerSet()
     {
         assertThatThrownBy(() -> configureServer("config/sidecar_unrecognized_authenticator.yaml"))
         .hasCauseInstanceOf(RuntimeException.class)
-        .hasMessageContaining("Implementation for class org.apache.cassandra.sidecar.acl.authentication.UnrecognizedAuthenticationHandler has not been registered");
+        .hasMessageContaining("Implementation for class org.apache.cassandra.sidecar.acl.authentication." +
```
frankgh, line 307: this was causing checkstyle to either fail or produce a warning for me

server/src/test/java/org/apache/cassandra/sidecar/acl/authentication/MutualTlsAuthenticationHandlerFactoryTest.java
```diff
@@ -69,12 +69,14 @@ void testMissingParameters()
         {{
             put(CERTIFICATE_IDENTITY_EXTRACTOR_PARAM_KEY, "org.apache.cassandra.sidecar.acl.authentication.CassandraIdentityExtractor");
         }};
-        testConfigError(missingValidatorParams, String.format("Missing %s parameter for MutualTlsAuthenticationHandler creation", CERTIFICATE_VALIDATOR_PARAM_KEY));
+        testConfigError(missingValidatorParams, String.format("Missing %s parameter for MutualTlsAuthenticationHandler creation",
```
frankgh, line 72: these changes here because checkstyle was either failing or producing a warning for me

server/src/test/integration/org/apache/cassandra/testing/TestVersion.java
```diff
@@ -34,4 +34,10 @@ public String version()
     {
         return version;
     }
+
+    @Override
+    public String toString()
+    {
+        return version;
```
frankgh, line 41: to be able to display the version during the test

server/src/test/integration/org/apache/cassandra/testing/CassandraTestTemplate.java
```diff
@@ -311,7 +309,7 @@ public Object resolveParameter(ParameterContext parameterContext, ExtensionConte
         }
     }
 
-    public static UpgradeableCluster retriableStartCluster(UpgradeableCluster.Builder builder, int maxAttempts)
+    public static UpgradeableCluster retryableStartCluster(UpgradeableCluster.Builder builder, int maxAttempts)
```
frankgh: this is a typo I fixed

server/src/test/integration/org/apache/cassandra/sidecar/routes/tokenrange/TestTokenSupplier.java
```diff
@@ -86,7 +86,7 @@ private static List<String>[] allocateExistingNodeTokens(int numNodesPerDC,
         List<String>[] tokens = new List[(numNodesPerDC + newNodesPerDC) * numDcs];
         for (int i = 0; i < ((numNodesPerDC + newNodesPerDC) * numDcs); ++i)
         {
-            tokens[i] = new ArrayList(numTokensPerNode);
+            tokens[i] = new ArrayList<>(numTokensPerNode);
```
frankgh, line 89: I think checkstyle was complaining here maybe? or just IDE

server/src/test/integration/org/apache/cassandra/sidecar/cluster/locator/CqlSessionProviderIntegrationTest.java
```diff
@@ -106,21 +106,21 @@ void testWithMTLS(VertxTestContext context, ConfigurableCassandraTestContext cas
...
-                                                 .set("client_encryption_options.optional", "true")
-                                                 .set("client_encryption_options.require_client_auth", "true")
-                                                 .set("client_encryption_options.require_endpoint_verification", "false")
-                                                 .set("client_encryption_options.keystore", serverKeystorePath.toAbsolutePath().toString())
-                                                 .set("client_encryption_options.keystore_password", serverKeystorePassword)
-                                                 .set("client_encryption_options.truststore", truststorePath.toAbsolutePath().toString())
-                                                 .set("client_encryption_options.truststore_password", truststorePassword));
-        });
+        cassandraContext.configureAndStartCluster(builder -> builder.appendConfig(config -> {
+            config.set("authenticator.class_name", "org.apache.cassandra.auth.MutualTlsWithPasswordFallbackAuthenticator")
```
frankgh, line 110: checkstyle complaining with line length

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
```diff
@@ -80,14 +80,7 @@ public PreparedStatement getAllRolesAndIdentities()
...
-        static String selectRoleFromIdentity()
-        {
-            return "SELECT role FROM system_auth.identity_to_role WHERE identity = ?";
-        }
-
-        static String getAllRolesAndIdentities()
-        {
-            return "SELECT * FROM system_auth.identity_to_role;";
-        }
+        static final String SELECT_ROLE_FROM_IDENTITY = "SELECT role FROM system_auth.identity_to_role WHERE identity = ?";
```
frankgh, line 83: no need for static methods, so I just changed to static constant definitions

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/cluster/locator/TokenRangeReplicas.java
```diff
@@ -73,7 +73,7 @@ public static List<TokenRangeReplicas> generateTokenRangeReplicas(Token start,
             return unwrapRange(start, end, partitioner, replicaSet);
         }
 
-        LOGGER.info("Generating replica-map for range: {} - {} : Replicaset: {}", start, end, replicaSet);
+        LOGGER.debug("Generating replica-map for range: {} - {} : Replicaset: {}", start, end, replicaSet);
```
frankgh, line 76: this was showing up a log in my logs, so it really should go in debug

conf/sidecar.yaml
```diff
@@ -215,6 +222,7 @@ cassandra_input_validation:
     - system_auth
     - system_views
     - system_virtual_schema
+    - sidecar_internal
```
frankgh, line 234: even though we've had `sidecar_internal` in the default list https://github.com/apache/cassandra-sidecar/blob/trunk/server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CassandraInputValidationConfigurationImpl.java#L43 , we are actually missing it in the yaml. So added it here

build.gradle
```diff
@@ -118,7 +118,7 @@ startScripts {
 }
 
 run {
-    def confFile = "file:" + File.separator + File.separator + "$projectDir/conf/sidecar.yaml"
+    def confFile = System.getProperty("sidecar.config", "file:" + File.separator + File.separator + "$projectDir/conf/sidecar.yaml")
```
frankgh, line 133: we didn't honor the user-provided config , when I was testing locally I found out that something like this `./gradlew run -Dsidecar.config=file:///$PWD/conf/sidecar.yaml` is not working

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -210,7 +207,12 @@ public void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tabl
     {
         requireNonNull(keyspace, "keyspace must be non-null");
         requireNonNull(table, "table must be non-null");
-        jmxClient.proxy(StorageJmxOperations.class, STORAGE_SERVICE_OBJ_NAME)
-                 .forceKeyspaceCleanup(concurrency, keyspace, table);
+        initializeStorageOps().forceKeyspaceCleanup(concurrency, keyspace, table);
+    }
+
+    protected StorageJmxOperations initializeStorageOps()
```
JeetKunDoug: This is really `createStorageOps` since it returns a new one every time.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
```diff
@@ -210,7 +207,12 @@ public void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tabl
...
+    {
+        return new GossipDependentStorageJmxOperations(jmxClient.proxy(StorageJmxOperations.class,
```
JeetKunDoug: Does _every operation_ in here really need Gossip to be enabled? It doesn't seem like things like taking local snapshots would require gossip, for example.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/utils/StringUtils.java
```diff
@@ -0,0 +1,45 @@
...
+     * the Cassandra server. This is useful when we consume data representing an Inet address
+     * and port from JMX that has been stringyfied, and we need to perform string matching
+     * against those results from JMX.
+     *
+     * @param address the {@link InetSocketAddress address}
+     * @return the string representation of the address with the same signature as represented by
+     * the Cassandra server
+     */
+    @SuppressWarnings("UnstableApiUsage")
+    public static String hostAddress(InetSocketAddress address)
```
JeetKunDoug: NIT: maybe `cassandraFormattedHostAddress`? Since it's in a fairly generically-named StringUtils class, probably good to be more specific here.

scripts/build-dtest-jars.sh
```diff
@@ -19,9 +19,9 @@
 
 set -xe
 CANDIDATE_BRANCHES=(
-  "cassandra-4.0:1f79c8492528f01bcc5f88951a1cc9e0d7265c54"
-  "cassandra-4.1:725655dda2776fef35567496a6e331102eb7610d"
-  "cassandra-5.0:f19dd0bb1309c35535876e8f0f996ad2b76adda5"
+  "cassandra-4.0:659558c980c67a80287ca7ccdfc8a70b1a56b7e2"
+  "cassandra-4.1:c44008e9e4ac06ea25bb63384a6777c0c1b8b41a"
+  "cassandra-5.0:1655578ae8f9de00393d923edbefe6bab0414153"
   # note the trunk hash cannot be advanced beyond ae0842372ff6dd1437d026f82968a3749f555ff4 (TCM), which breaks integration test
```
JeetKunDoug, line 25: Does TCM still break integration tests? Has anyone had a chance to look into why?

server/src/main/java/org/apache/cassandra/sidecar/coordination/SingleInstanceExecutor.java
```diff
@@ -0,0 +1,53 @@
...
+package org.apache.cassandra.sidecar.coordination;
+
+/**
+ * Defines an interface to choose a single Sidecar instance that will run certain types of operations that need to
+ * run on a limited subset of Sidecar instances. In most cases there will be a single Sidecar instance chosen as
+ * the executor.
+ */
+public interface SingleInstanceExecutor
+{
+    SingleInstanceExecutor ALWAYS_SCHEDULE_EXECUTOR = new SingleInstanceExecutor()
```
JeetKunDoug: Can we move this into a class in the tests somewhere as it should really never be used in production code - maybe put this and the NEVER_SCHEDULE_EXECUTOR into a class called `TestSingleInstanceExecutors`?

server/src/main/java/org/apache/cassandra/sidecar/coordination/BestEffortSingleInstanceExecutor.java
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    /**
+     * Returns the configured host ID for Sidecar.
+     *
+     * @return the configured host ID for Sidecar
+     */
+    protected String owner()
+    {
+        return config.hostId();
```
JeetKunDoug: Given this class can't operate if it can't talk to Cassandra, what do you think about using `StorageService#getLocalHostId` via JMX from the first managed instance for this in stead of configuration? We can populate it in `config` still or via some other mechanism, but it should be constant and not just a random UUID on each startup?

server/src/main/java/org/apache/cassandra/sidecar/coordination/BestEffortSingleInstanceExecutor.java
```diff
@@ -0,0 +1,261 @@
...
+            else
+            {
+                // For the case where the current Sidecar was a lease-holder but the information was lost from
+                // the in-memory process (i.e. Sidecar restarted) but the information is still persisted
+                // in the database, so we recover the state from the database
+                return owner.equals(result.existingOwner);
+            }
+        }
+        return isCurrentLeaseHolder;
+    }
```
JeetKunDoug: There was quite a bit or repetition in `determineSingleInstanceExecutor` and it was somewhat hard to follow. Did some refactoring to make sure I understood what was going on and I think it's easier to follow. Sorry for the very long suggestion.

```suggestion
    @Override
    public void determineSingleInstanceExecutor(ElectorateMembership electorateMembership)
    {
        boolean shouldParticipate = electorateMembership.shouldParticipate();
        LOGGER.debug("Sidecar instance shouldParticipate={} in the selection", shouldParticipate);
        if (!shouldParticipate)
        {
            return;
        }

        boolean wasCurrentExecutor = isLocalSidecarSingleInstanceExecutor;
        Boolean isCurrentExecutor = null;

        String hostId = hostId();
        LOGGER.debug("Starting selection for hostId={}", hostId);
        if (wasCurrentExecutor)
        {
            isCurrentExecutor = executeLeaseAction("extend", hostId, accessor::extendLease, null);
        }

        if (isCurrentExecutor == null || !isCurrentExecutor)
        {
            isCurrentExecutor = executeLeaseAction("claim", hostId, accessor::claimLease, isCurrentExecutor);
            LOGGER.debug("Claim lease for hostId={} result={}", hostId, isCurrentExecutor);
        }

        maybeNotifyResults(isCurrentExecutor, hostId, wasCurrentExecutor);
    }

    private void maybeNotifyResults(Boolean isCurrentExecutor, String owner, boolean wasCurrentExecutor)
    {
        if (isCurrentExecutor == null)
        {
            LOGGER.debug("Unable to perform lease election for owner={}", owner);
//            if (wasCurrentExecutor)
//            {
            // TODO : do we give up lease after some period of time? ie TTL which is 10 minutes?
//            }
            return;
        }

        if (wasCurrentExecutor && !isCurrentExecutor)
        {
            LOGGER.info("Cluster-wide lease has been lost by owner={}", owner);
            isLocalSidecarSingleInstanceExecutor = false;
            // notify lease has been lost
            vertx.eventBus().publish(ON_SIDECAR_GLOBAL_LEASE_LOST.address(), owner);
        }

        if (!wasCurrentExecutor && isCurrentExecutor)
        {
            LOGGER.info("Cluster-wide lease has been claimed by owner={}", owner);
            isLocalSidecarSingleInstanceExecutor = true;
            // notify lease has been gained
            vertx.eventBus().publish(ON_SIDECAR_GLOBAL_LEASE_CLAIMED.address(), owner);
        }
    }

    private Boolean executeLeaseAction(String actionName,
                                       String owner,
                                       Function<String, SidecarLeaseDatabaseAccessor.LeaseClaimResult> claimFunction,
                                       Boolean isCurrentExecutor)
    {
        SidecarLeaseDatabaseAccessor.LeaseClaimResult result = null;
        try
        {
            LOGGER.debug("Attempting to " + actionName + " lease for owner={}", owner);
            result = claimFunction.apply(owner);
        }
        catch (CASWriteUnknownException | NoHostAvailableException e)
        {
            LOGGER.debug("Unable to claim lease for owner={}", owner, e);
        }
        catch (Exception e)
        {
            LOGGER.error("Unable to " + actionName + " lease for owner={}", owner, e);
        }

        isCurrentExecutor = determineIfIsCurrentLeaseHolder(isCurrentExecutor, result, owner);
        LOGGER.debug("{} lease for owner={} result={}", capitalize(actionName), owner, isCurrentExecutor);
        return isCurrentExecutor;
    }

    protected Boolean determineIfIsCurrentLeaseHolder(Boolean isCurrentLeaseHolder,
                                                   SidecarLeaseDatabaseAccessor.LeaseClaimResult result,
                                                   String hostId)
    {
        if (result != null && (result.leaseAcquired || hostId.equals(result.existingOwner)))
        {
            return true;
        }
        return isCurrentLeaseHolder;
    }
```

server/src/main/java/org/apache/cassandra/sidecar/coordination/ElectorateMembership.java
```diff
@@ -0,0 +1,32 @@
...
+ * An interface that defines if the local Sidecar instance should participate during the choosing of a single
+ * instance executor process.
+ */
+public interface ElectorateMembership
+{
+    /**
+     * @return {@code true} if the local Sidecar instance should participate in the process,
+     * {@code false} otherwise
+     */
+    boolean shouldParticipate();
```
JeetKunDoug: I was struggling with the name of this class/method for some reason... I think the method should be called `isMember` rather than `shouldParticipate` and it all makes much more sense. Comments should also reflect that this (and the TokenZeroElectorateMembership class) are just about membership in a particular electorate (which I'm still not sure of name-wise, but I haven't come up with anything better yet either).

JeetKunDoug (commented):
Submitting review so far - thanks for the contribution. Looks pretty good so far, just some refactoring/naming suggestions

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStorageOperations.java
frankgh: I think you are right these probably don't need gossip enabled, also it's an unrelated change so I will revert it

conf/sidecar.yaml
frankgh: I ran into issues while manually testing this feature, so I've added it

scripts/build-dtest-jars.sh
frankgh: yes, the reason it breaks is that our touch points for the byte buddy injections have diverged after TCM

server/src/main/java/org/apache/cassandra/sidecar/cluster/CQLSessionProviderImpl.java
```diff
@@ -70,7 +70,7 @@ public class CQLSessionProviderImpl implements CQLSessionProvider
 {
     private static final Logger logger = LoggerFactory.getLogger(CQLSessionProviderImpl.class);
     private final List<InetSocketAddress> contactPoints;
-    private final int numConnections;
+    private final int numAdditionalConnections;
```
frankgh, line 73: changing the name because when I was running tests I had issues, internally this is actually called numAdditionalConnections, so making the change here to reflect what it really signifies

server/src/main/java/org/apache/cassandra/sidecar/db/SidecarLeaseDatabaseAccessor.java
```diff
@@ -0,0 +1,89 @@
...
+        super(tableSchema, sessionProvider);
+    }
+
+    /**
+     * Attempts to obtain the lease, returning the result of the claim
+     *
+     * @param owner the owner performing the claim
+     * @return the results of performing the lease claim
+     */
+    public LeaseClaimResult claimLease(String owner)
```
JeetKunDoug: Most of the usages of `String owner` are really the hostId of the host attempting to _claim_ or _extend_ the lease, not the _owner_ of the lease - can we just use `hostId` in these instances, and reserve `owner` for cases where the host in question _is the owner_? See the suggestion in `BestEffortSingleInstanceExecutor` where I did this as well. I think the LeaseClaimResult usage of `owner` makes perfect sense, as an example of where owner makes sense vs. hostId.

server/src/main/java/org/apache/cassandra/sidecar/db/SidecarLeaseDatabaseAccessor.java
```diff
@@ -0,0 +1,89 @@
...
+        return LeaseClaimResult.from(resultSet, owner);
+    }
+
+    /**
+     * Captures the results of claiming lease for a given Sidecar owner
+     */
+    public static class LeaseClaimResult
+    {
+        public final boolean leaseAcquired;
+        public final String existingOwner;
```
JeetKunDoug: do we need `existing` here? either just `owner` or maybe `currentOwner` as I'm not sure what exactly `existingOwner` means - is it _previous_ or _current_?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarInternalKeyspace.java
```diff
@@ -33,20 +33,19 @@
  * {@link org.apache.cassandra.sidecar.db.RestoreJob} table and {@link org.apache.cassandra.sidecar.db.RestoreSlice}
  * table
  */
-@Singleton
 public class SidecarInternalKeyspace extends AbstractSchema
 {
     private final SchemaKeyspaceConfiguration keyspaceConfig;
     private final boolean isEnabled;
-    private final Set<TableSchema> tableSchemas = new HashSet<>();
+    private final Set<TableSchema> tableSchemas = ConcurrentHashMap.newKeySet();
```
JeetKunDoug, line 40: Was this done just to get rid of the `synchronized` on registerTableSchema for stylistic reasons (as I can't imaging it was a significant performance issue), or was there a bug this fixed?

Assuming we really need to make this change, I wonder if we want to be specific about the type of tableSchemas to show it really needs to be concurrent:
```suggestion
    private final ConcurrentHashMap.KeySetView<TableSchema, Boolean> tableSchemas = ConcurrentHashMap.newKeySet();
```

It's a bit ugly just because of the type `newKeyset` returns, but it makes it a bit clearer that this needs to be a concurrent set (now, why the Java implementers didn't just give us a ConcurrentSet to make this clear, no idea).

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarLeaseSchema.java
```diff
@@ -0,0 +1,113 @@
...
+     */
+    @Override
+    @VisibleForTesting
+    public String createSchemaStatement()
+    {
+        return String.format("CREATE TABLE IF NOT EXISTS %s.%s ("
+                             + "name text PRIMARY KEY,"
+                             + "owner text) "
+                             + "WITH default_time_to_live = %d",
+                             keyspaceName(), tableName(), TimeUnit.MINUTES.toSeconds(10));
```
JeetKunDoug: NIT: Make the TTL configurable and write the rows using `USING TTL ?` so it can be updated easily rather than depending on the default in the table?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -181,4 +186,23 @@ protected void reportSidecarSchemaInitialized()
...
+     * @return {@code true} if the schema should be created by this Sidecar instance, {@code false} otherwise
+     */
+    protected boolean shouldCreateSchema(@Nullable AbstractSchema schema)
+    {
+        if (singleInstanceExecutor != null && schema instanceof InitializeOnSingleInstanceExecutor)
+        {
+            return singleInstanceExecutor.isLocalSidecarSingleInstanceExecutor();
+        }
+        return true;
+    }
```
JeetKunDoug: I'm not sure if `singleInstanceExecutor` can actually be null (it doesn't look that way)... could we drop the null check here? If so, this method simplifies a bit:
```suggestion
    protected boolean shouldCreateSchema(@Nullable AbstractSchema schema)
    {
        return singleInstanceExecutor.isLocalSidecarSingleInstanceExecutor() || 
               !(schema instanceof InitializeOnSingleInstanceExecutor);
    }
```

JeetKunDoug (commented):
Some more suggestions/questions to go with yesterday's review.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
frankgh: you don't really need to inject an implementation, so I made it generic so you can omit this feature

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarLeaseSchema.java
frankgh: yeah, but there are risks with managing your TTLs, so I avoided doing that

server/src/main/java/org/apache/cassandra/sidecar/db/SidecarLeaseDatabaseAccessor.java
frankgh: I meant current, I will refactor

server/src/main/java/org/apache/cassandra/sidecar/coordination/BestEffortSingleInstanceExecutor.java
frankgh: I chose randomness to simplify, maybe we can think about a follow up for this?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarInternalKeyspace.java
frankgh: I think this is a class-internal detail, but I'd prefer to keep it as Set for now

frankgh: I think I wanted to avoid synchronization.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarLeaseSchema.java
frankgh: something to consider, but I'd prefer to let the db do the heavy lifting here. We can discuss more.

frankgh: ended up addressing this by making TTL configurable

server/src/test/integration/org/apache/cassandra/testing/CassandraTestTemplate.java
yifan-c: hmm.. I think it is not. see [retriable](https://en.wiktionary.org/wiki/retriable#:~:text=%2B%E2%80%8E%20%2Dable.-,Adjective,Capable%20of%20being%20retried.)
When adding suffix to words end with "y", "y" is changed to "i". See the [spell rule](https://www.languagecouncils.sg/goodenglish/resources/spelling-tips/adding-endings-to-words-that-end-in--y#:~:text=When%20adding%20endings%20to%20words,pretty%3A%20prettier%20%2F%20prettiest)

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/db/schema/ConnectedClientsSchema.java
```diff
@@ -63,7 +63,7 @@ public PreparedStatement connectionsByUser()
         return connectionsByUserStatement;
     }
 
-    private String statsStatement()
+    static String statsStatement()
```
yifan-c, line 67: On a side note, those changes should ideally be brought up during the review of that patch and addressed there.

server/src/test/integration/org/apache/cassandra/testing/CassandraTestTemplate.java
frankgh: it seems both are valid spellings . I will revert this change

server/src/main/java/org/apache/cassandra/sidecar/coordination/BestEffortSingleConditionalExecutor.java
```diff
@@ -0,0 +1,377 @@
...
+        }
+
+        boolean wasCurrentExecutor = determination.shouldExecuteOnLocalInstance();
+        Boolean isCurrentExecutor = null;
+
+        String sidecarHostId = sidecarHostId();
+        LOGGER.debug("Starting selection for sidecarHostId={}", sidecarHostId);
+        if (wasCurrentExecutor)
+        {
+            isCurrentExecutor = executeLeaseAction("extend", sidecarHostId, accessor::extendLease, isCurrentExecutor);
```
JeetKunDoug: NIT: I know _I_ wrote this, so this is my NIT: We should have an enum for LeaseAction rather than passing the strings in here... Something like:
```java
    public enum LeaseAction
    {
        CLAIM("claim");
        EXTEND("extend"),

        public final String description;

        LeaseAction(String description)
        {
            this.description = description;
        }
    }
```
It's slightly "forward-looking" but I can see this being used elsewhere and in general hate the whole "I just stuck a constant string in the middle of a method" now that I see it.

server/src/main/java/org/apache/cassandra/sidecar/coordination/BestEffortSingleConditionalExecutor.java
```diff
@@ -0,0 +1,377 @@
...
+            if (!wasCurrentExecutor || leaseExpired())
+            {
+                if (leaseExpired())
+                {
+                    LOGGER.info("Giving up lease for sidecarHostId={} leaseAcquired={} leaseExpired={}",
+                                sidecarHostId, leaseTime, leaseExpirationTime());
+                }
+
+                determination = ExecutionDetermination.INDETERMINATE;
+                leaseTime = null;
```
JeetKunDoug: It feels like we should notify the eventBus here as well?

server/src/main/java/org/apache/cassandra/sidecar/coordination/BestEffortSingleConditionalExecutor.java
```diff
@@ -0,0 +1,377 @@
...
+        }
+
+        if (result.leaseAcquired)
+        {
+            return true;
+        }
+
+        // For the case where the current Sidecar was a lease-holder but the information was lost from
+        // the in-memory process (i.e. Sidecar restarted) but the information is still persisted
+        // in the database, so we recover the state from the database
```
JeetKunDoug: Should document that this depends on a persistent `sidecarHostId`, which we don't have at this time but may be a future enhancement?

server/src/main/java/org/apache/cassandra/sidecar/coordination/TokenZeroElectorateMembership.java
```diff
@@ -0,0 +1,201 @@
...
+                    LOGGER.warn("Unable to determine local storage broadcast address for instance={}", instance);
+                    continue;
+                }
+
+                String localInstanceHostAndPort = StringUtils.cassandraFormattedHostAndPort(address);
+                if (tokenRangeReplicas == null)
+                {
+                    // Token range replicas should be the same across all instances assuming the view of the ring
+                    // is the same for all instances, so we just get it once, as this could be an expensive call
+                    tokenRangeReplicas = operations.tokenRangeReplicas(new Name(userKeyspace), nodeSettings.partitioner());
```
JeetKunDoug: 👍

server/src/main/java/org/apache/cassandra/sidecar/coordination/TokenZeroElectorateMembership.java
```diff
@@ -0,0 +1,201 @@
...
+                }
+
+                InetSocketAddress address = delegate.localStorageBroadcastAddress();
+                if (address == null)
+                {
+                    LOGGER.warn("Unable to determine local storage broadcast address for instance={}", instance);
+                    continue;
+                }
+
+                String localInstanceHostAndPort = StringUtils.cassandraFormattedHostAndPort(address);
```
JeetKunDoug: _SUPER_ NIT: move this declaration below the retrieval of tokenRangeReplicas so it's closer to its usage (we don't need it until line 111l)

JeetKunDoug (commented):
Submitting comments up to TokenZeroElectorateMembership for now.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -181,4 +186,23 @@ protected void reportSidecarSchemaInitialized()
     {
         vertx.eventBus().publish(ON_SIDECAR_SCHEMA_INITIALIZED.address(), "SidecarSchema initialized");
     }
+
+    /**
+     * Returns true when the schema should be created by this Sidecar instance. This currently
+     * depends on whether the election of single instance executor process is available, and
+     * the schemas are of type {@link InitializeOnSingleInstanceExecutor}, and the local Sidecar
+     * is a single instance executor. For all other types of schemas or if the election of leader
```
JeetKunDoug: Do we want to initialize the schema in every instance even if the we can't determine that we should be doing it? Could we defer the schema creation in that case, as we know (pre-TCM) this can cause significant issues with schema mismatches across instances?

server/src/test/java/org/apache/cassandra/sidecar/coordination/BestEffortSingleConditionalExecutorParametersTest.java
```diff
@@ -0,0 +1,113 @@
...
+    @Test
+    void testDefaults()
+    {
+        BestEffortSingleConditionalExecutor.Parameters params = BestEffortSingleConditionalExecutor.Parameters.from(Collections.emptyMap());
+        assertThat(params.isEnabled()).as("Enabled by default").isTrue();
+        assertThat(params.delayMillis()).as("Run every minute by default").isEqualTo(60_000L);
+        assertThat(params.initialDelayMillis()).as("Wait 1 second before starting the process by default").isEqualTo(1_000L);
+    }
+
+    @DisplayName("Ensures that is_enabled cannot be configured with an invalid string value")
```
JeetKunDoug:
```suggestion
    @DisplayName("Ensures that is_enabled returns the default value when an invalid string value is configured")
```

server/src/test/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutorTest.java
```diff
@@ -77,4 +80,71 @@ public void close()
...
+        SimulatedTask taskThatRunsOnNonLeader = new SimulatedTask(latch);
+        PeriodicTaskExecutor taskExecutorNonLeader = new PeriodicTaskExecutor(executorPools,
+                                                                              TestConditionalExecutors.NEVER_EXECUTE);
+
+        taskExecutorNonLeader.schedule(taskThatRunsOnNonLeader);
+        assertThat(Uninterruptibles.awaitUninterruptibly(latch, 30, TimeUnit.SECONDS)).isTrue();
+        assertThat(taskThatRunsOnNonLeader.executionCount.get()).isEqualTo(0);
+    }
+
+    @Test
```
JeetKunDoug, line 99: We need to add a test for an executor that returns `INDETERMINATE` at least once, and then `EXECUTE`, so we can test the retry of a task that requires a conditional executor.

JeetKunDoug (commented):
Final comments from me - great stuff @frankgh ... looking forward to getting it merged soon.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
frankgh: This is the status quo and not to be addressed in this PR. I think TCM will help here.

yifan-c (approved):
+1. My review feedback was sent offline. Waiting for green CI

PR #163, opened 2024-12-13
CASSSIDECAR-179: Added additional dirs required for live migration

nvharikrishna:

Adding commitlog, hints, saved caches and local system data file directories to InstanceMetadata so that they migrated as part of CEP-40.

For CASSSIDECAR-179


server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
```diff
@@ -92,9 +112,13 @@ public InstanceConfigurationImpl(int id,
         this.id = id;
         this.host = host;
         this.port = port;
-        this.dataDirs = Collections.unmodifiableList(dataDirs);
+        this.dataDirs = dataDirs == null ? null : Collections.unmodifiableList(dataDirs);
```
bbotella: This looks like an unrelated bug fix?

server/src/test/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImplTest.java
```diff
@@ -39,6 +39,10 @@ class InstanceMetadataImplTest
     private static final String DATA_DIR_2 = "test/data/data2";
     private static final String CDC_DIR = "cdc_dir";
     private static final String STAGING_DIR = "staging_dir";
+    private static final String COMMITLOG_DIR = "commitlog";
```
bbotella: Nit: Should we append `_dir` to keep consistency? (both cdc and staging append it)

bbotella (commented):
Just a couple of nits. Otherwise, LGTM.

Thanks for the patch!
+1 (nb)

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
nvharikrishna: Not trying to fix anything actually :). Trying to initialize the member variables in a single constructor. Earlier constructor initialized dataDirs with null, but this constructor is wrapping dataDirs with `Collections.unmodifiableList` which doesn't accept null. So checking for null before wrapping it.

server/src/test/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImplTest.java
nvharikrishna: I have used the names used in cassandra.yaml for newly introduced directories. Wanted to use default names used in Cassandra, but I don't mind changing the names if required.

nvharikrishna (commented):
@bbotella Thanks for the review. Addressed the comments.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
```diff
@@ -64,17 +76,21 @@ public class InstanceConfigurationImpl implements InstanceConfiguration
...
-        this.jmxSslEnabled = false;
-        this.jmxRole = null;
-        this.jmxRolePassword = null;
+        this(0,
+             null,
+             9042,
+             null,
+             null,
+             null,
+             null,
```
sarankk: For commit log, hints and saved caches, can we build it if it is not set. Like we do in Cassandra, by taking in CASSANDRA_HOME

frankgh: I agree that we should simplify Sidecar configuration, where we follow Cassandra's configuration. Ideally, we will only have to configure the `cassandra_home` configuration per instance. And then we resolve the directory files in Cassandra unless explicitly provided.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
frankgh: when would data dirs be null (or empty)? Maybe this needs to be validated instead of accepting null

server/src/test/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImplTest.java
frankgh: we should not, this is the default directory name for the commit log in Cassandra, so it would be incorrect to add `_dir`. See https://github.com/apache/cassandra/blob/trunk/conf/cassandra.yaml#L411

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
frankgh, line 1: Just a general comment. We should update the sidecar.yaml configuration file under `conf/sidecar.yaml` with configuration parameters. Ideally I think we want to only have to specify the `cassandra_home_dir` , and optionally specify Cassandra directories. The exception are data directories

frankgh (commented):
I have some suggestions to simplify configuration and make operability simpler

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
nvharikrishna: > when would data dirs be null (or empty)? Maybe this needs to be validated instead of accepting null

It is set to null if the data_dirs is not configured in sidecar.yaml. I was not sure why it was allowed to be null earlier. So did not change the behavior. I can make changes now though.

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
nvharikrishna: > Ideally I think we want to only have to specify the cassandra_home_dir , and optionally specify Cassandra directories. The exception are data directories

As per the documentation comment in [cassandra.yaml](https://github.com/apache/cassandra/blob/trunk/conf/cassandra.yaml#L398), the default data directory is`$CASSANDRA_HOME/data/data`. Is there any reason to take exception for data directories?

nvharikrishna: Followed the footsteps of existing code. I would be happy to adopt `cassandra_home_dir` as part of this pr.

server/src/test/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImplTest.java
nvharikrishna: >so it would be incorrect to add _dir. 

Wondering why CDC_DIR has "_dir" postfix. As per cassandra.yaml, cdc_raw_directory is `cdc_raw`.

frankgh: maybe because it's test code, but it feels we should align with Cassandra directories

nvharikrishna: I ended up with setting these values to non default values in order to test that InstanceMetaImpl is not using default values when these directories are set explicitly. Using "_dir" like in previous constants.

(I wrote this test class a few days back with the same intention, could not recollect it earlier, my apologies!)

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
```diff
@@ -21,80 +21,63 @@
...
-    @JsonProperty("jmx_ssl_enabled")
     protected final boolean jmxSslEnabled;
-
-    @JsonProperty("jmx_role")
     protected final String jmxRole;
-
-    @JsonProperty("jmx_role_password")
     protected final String jmxRolePassword;
 
-    public InstanceConfigurationImpl()
```
nvharikrishna, line 65: Removed this in order to reduce the confusion with what can be null and what cannot be null.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
```diff
@@ -21,80 +21,63 @@
...
 import org.apache.cassandra.sidecar.config.InstanceConfiguration;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Encapsulates the basic configuration needed to connect to a single Cassandra instance
  */
 public class InstanceConfigurationImpl implements InstanceConfiguration
 {
-    @JsonProperty("id")
```
nvharikrishna, line 32: Using constructor for this.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
```diff
@@ -21,80 +21,63 @@
...
-                                     String stagingDir,
-                                     String cdcDir,
-                                     String jmxHost,
-                                     int jmxPort,
-                                     boolean jmxSslEnabled,
-                                     String jmxRole,
-                                     String jmxRolePassword)
+    @JsonCreator
+    public InstanceConfigurationImpl(@JsonProperty("id") int id,
+                                     @NotNull @JsonProperty("host") String host,
```
nvharikrishna, line 54: @frankgh  I guessing a bit for `NotNull` and `Nullable` of these fields. Please suggest changes if my interpretation is incorrect.

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
```diff
@@ -52,15 +65,42 @@ protected InstanceMetadataImpl(Builder builder)
         id = builder.id;
         host = builder.host;
         port = builder.port;
+        cassandraHomeDir = FileUtils.maybeResolveHomeDirectory(builder.cassandraHomeDir);
+        Path cassandraHomeDirPath = cassandraHomeDir == null ?  null : Paths.get(cassandraHomeDir);
```
nvharikrishna: Okay with having cassandraHomeDir as a mandatory configuration? If yes, then I can push changes to make it mandatory. Assuming it is not mandatory for now.

frankgh: I think we should disallow the cassandra home dir to be null.

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
```diff
@@ -66,6 +72,26 @@ public interface InstanceMetadata
...
+
+    /**
+     * @return saved caches directory of the cassandra instance
+     */
+    @Nullable String savedCachesDir();
+
+    /**
+     * @return local system data file directory of the cassandra instance
+     */
+    @Nullable String localSystemDataFileDir();
```
frankgh: I don't really know what this maps to in the cassandra.yaml configuration. Did I miss something?

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
```diff
@@ -66,6 +72,26 @@ public interface InstanceMetadata
      */
     @NotNull CassandraAdapterDelegate delegate() throws CassandraUnavailableException;
 
+    /**
+     * @return commitlog directory of the cassandra instance
+     */
+    @Nullable String commitlogDir();
```
frankgh: I feel these should be not-null, but we can default to the Cassandra's defaults:
```
    /**
     * Returns the commit log directory of the Cassandra instance. If not set, the default directory is
     * {@link #cassandraHomeDir()}/data/commitlog
     *
     * @return commitlog directory of the cassandra instance
     */
    @NotNull
    default String commitlogDir()
    {
        return String.format("%s/data/commitlog", cassandraHomeDir());
    }
```

frankgh (commented):
I have some comments and a few questions. Please take a look and tell me if it makes sense

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
nvharikrishna: It maps to `local_system_data_file_directory` in [cassandra.yaml](https://github.com/apache/cassandra/blob/trunk/conf/cassandra.yaml#L407C3-L407C35).

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
nvharikrishna: Made the changes

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
nvharikrishna: It is needed only for CEP-40, so I thought of not making it mandatory. Now marking it @NotNull since we are making cassandra_home_dir mandatory. `String.format` hurts performance, so replaced with normal string concatenation.

nvharikrishna:
@frankgh @sarankk addressed the review comments.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
```diff
@@ -21,80 +21,63 @@
...
+                                     @JsonProperty("port") int port,
+                                     @NotNull @JsonProperty("cassandra_home_dir") String cassandraHomeDir,
+                                     @NotNull @JsonProperty("data_dirs") List<String> dataDirs,
```
frankgh: we should allow null for data dirs if we are going to allow configuring the `cassandra_home_dir` only.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
```diff
@@ -21,80 +21,63 @@
...
+                                     @NotNull @JsonProperty("jmx_host") String jmxHost,
+                                     @JsonProperty("jmx_port") int jmxPort,
+                                     @JsonProperty("jmx_ssl_enabled") boolean jmxSslEnabled,
+                                     @Nullable @JsonProperty("jmx_role") String jmxRole,
+                                     @Nullable @JsonProperty("jmx_role_password") String jmxRolePassword)
     {
         this.id = id;
         this.host = host;
         this.port = port;
+        this.cassandraHomeDir = cassandraHomeDir;
```
frankgh: `dataDirs` should handle for the null case in line 74

frankgh (commented):
Looking closer at the Cassandra code, I think we need to base the configuration on the `cassandra.storagedir` configuration instead. I have a patch that changes this and have provided it offline. This change aligns better with Cassandra. Also, in the Cassandra process `storage_dir` can be omitted, and you can specify all the required configuration directories `hints`, `saved_caches`, `data_directories`, etc. This is also addressed in the patch I shared offline

frankgh (approved):
+1 looks good to me

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
```diff
@@ -48,24 +49,48 @@ public interface InstanceMetadata
...
+    /**
+     * @return saved caches directory of the Cassandra instance
+     */
+    @NotNull
+    String savedCachesDir();
+
+    /**
+     * @return local system data file directory of the cassandra instance
+     */
+    @Nullable String localSystemDataFileDir();
```
yifan-c: Please have a consistent placement of the annotation.

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
```diff
@@ -56,13 +69,19 @@ protected InstanceMetadataImpl(Builder builder)
...
         port = builder.port;
-        dataDirs = builder.dataDirs.stream()
-                                   .map(FileUtils::maybeResolveHomeDirectory)
-                                   .collect(Collectors.collectingAndThen(Collectors.toList(), Collections::unmodifiableList));
-        stagingDir = FileUtils.maybeResolveHomeDirectory(builder.stagingDir);
-        cdcDir = FileUtils.maybeResolveHomeDirectory(builder.cdcDir);
         delegate = builder.delegate;
         metrics = builder.metrics;
+
+        // Sidecar-managed directories
```
yifan-c, line 75: 👍 to the comments

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
```diff
@@ -48,24 +49,48 @@ public interface InstanceMetadata
...
      * @return a {@link CassandraAdapterDelegate} specific for the instance, or throws when the delegate is unavailable
      * @throws CassandraUnavailableException when the Cassandra service is unavailable
      */
     @NotNull CassandraAdapterDelegate delegate() throws CassandraUnavailableException;
 
+    /**
+     * @return CDC directory of the cassandra instance, it can be null when CDC is not enabled for the Cassandra
+     * instance
+     */
+    @Nullable String cdcDir();
```
yifan-c: nit: I am a bit concerned about the javadoc. It could mislead the reader that this method can be used to check whether cdc is enabled

conf/sidecar.yaml
```diff
@@ -23,6 +23,10 @@ cassandra_instances:
   - id: 1
     host: localhost1
     port: 9042
+    # The instance's storage directory as defined per the cassandra.storagedir property
+    # which defaults to the $CASSANDRA_HOME/data directory, but can be configured to any
+    # directory.
```
yifan-c: How about also mentioning the evaluation order? The configured `storage_dir` will be ignored, if the concrete directories of different kinds are configured. For example, `data_dirs` if configured, the configured value takes the precedence when evaluating the data directories list.

yifan-c (commented):
looks good overall. some minor comments

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
nvharikrishna: Updated the comment to:

"CDC directory of the cassandra instance, it can be **configured** as null when CDC is not enabled for the Cassandra."

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadata.java
nvharikrishna: updated

conf/sidecar.yaml
nvharikrishna: Updated.

nvharikrishna:
@yifan-c addressed the review comments.

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
nvharikrishna: '👍' goes to @frankgh :)

PR #164, opened 2024-12-13
CASSSIDECAR-176: OpenAPI API description

bbotella:

PR #166, opened 2024-12-20
CASSSIDECAR-180 - Changes to add endpoint for stream stats

arjunashok:
These changes are part of the effort to introduce bespoke Sidecar APIs to support key operational functionality currently managed through nodetool commands. 

These changes specifically return the stream-stats data in a sidecar endpoint.

### Changes
- Incorporates stream stats API - streaming progress subset of the `nodetool netstats` dataset 
- Aggregates the streaming data across sessions
- Unit tests and integration tests
- SidecarClient changes and tests for the new API


adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
```diff
@@ -176,6 +177,15 @@ public MetricsOperations metricsOperations()
         return new CassandraMetricsOperations(cqlSessionProvider);
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public StreamManagerOperations streamManagerOperations()
```
frankgh: I think this and `MetricsOperations` are essentially  the same. We should fold stats into a single interface. Conceptually abstracts access to functionality behind Cassandra. The abstraction here is access to statistics from Cassandra. I think `MetricsOperations` does not really capture what we are trying to abstract, and adding another interface here doesn't make sense.

I would avoid trying to have a one-to-one mapping from interface to C* bean here because in C*'s code base the functionality was exposed through a small set of beans. In recent contributions to the C* codebase I've seen that we are making a better effort to separate bean access into something that makes sense logically. For that reason we should really think about some abstraction that makes sense from a point of view of what we are trying to access, and maybe not focus too much on how things are exposed in Cassandra

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/GossipDependentStorageJmxOperations.java
```diff
@@ -146,6 +146,11 @@ public int forceKeyspaceCleanup(int jobs, String keyspaceName, String... tables)
         return delegate.forceKeyspaceCleanup(jobs, keyspaceName, tables);
     }
 
+    public String getOperationMode()
```
frankgh, line 150: I see a few `@Override` annotations. Can you please add them where they are missing in your PR?

server/src/test/java/org/apache/cassandra/sidecar/routes/StreamStatsHandlerTest.java
```diff
@@ -0,0 +1,183 @@
...
+            LOGGER.error("Close event timed out.");
+    }
+
+    @Test
+    void testStreamingStatsHandler(VertxTestContext context)
+    {
+        StreamingStatsTestModule.streamingStatsSupplier = () -> {
+            StreamStatsResponse response = new StreamStatsResponse("NORMAL",
+                                                                   new StreamProgressStats(7, 7, 1024, 1024, 0, 0, 0, 0));
+            return response;
```
frankgh: NIT 
```suggestion
            return new StreamStatsResponse("NORMAL",
                                                                   new StreamProgressStats(7, 7, 1024, 1024, 0, 0, 0, 0));
```

server/src/test/java/org/apache/cassandra/sidecar/routes/StreamStatsHandlerTest.java
```diff
@@ -0,0 +1,183 @@
...
+                  assertThat(statsResponse).isNotNull();
+                  assertThat(statsResponse.operationMode()).isEqualTo("NORMAL");
+                  context.completeNow();
+              }));
+    }
+
+
+    static class StreamingStatsTestModule extends AbstractModule
+    {
+        static Supplier<StreamStatsResponse> streamingStatsSupplier;
```
frankgh: hmm, why not directly inject this in the constructor?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStreamManagerOperations.java
```diff
@@ -0,0 +1,99 @@
...
+        Set<CompositeData> streamData = jmxClient.proxy(StreamManagerJmxOperations.class, STREAM_MANAGER_OBJ_NAME)
+                                                 .getCurrentStreams();
+
+        List<StreamState> streamStates = streamData.stream().map(StreamState::new).collect(Collectors.toList());
+        return computeStats(streamStates);
+    }
+
+    private StreamProgressStats computeStats(List<StreamState> streamStates)
+    {
+        List<SessionInfo> sessions = streamStates.stream().map(s -> s.sessions()).flatMap(Collection::stream).collect(Collectors.toList());
```
bbotella: I understand this is by no means as hot path as the ones the [dev thread](https://lists.apache.org/thread/65glsjzkmpktzmns6j9wvr4nczvskx36) was referring to, but I wonder if we should keep that choice in sidecar of avoiding streams in non test code.

Same for the other streams on the PR

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/ProgressInfo.java
```diff
@@ -0,0 +1,55 @@
...
+    public final String peer;
+    public final int sessionIndex;
+    public final String fileName;
+    public final String direction;
+    public final long currentBytes;
+    public final long totalBytes;
+
+    public ProgressInfo(CompositeData data)
+    {
+
```
bbotella: Remove extra line

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/data/StreamProgressStats.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @JsonProperty("totalBytesSent")
+    public long totalBytesSent()
+    {
+        return totalBytesSent;
+    }
+
+    /* Across all sessions
+         * private final long sendProgress; from "sent to"
```
bbotella:
```suggestion
         private final long sendProgress; from "sent to"
```

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
```diff
@@ -1573,6 +1575,36 @@ public void onError(Throwable throwable)
         assertThat(new String(baos.toByteArray(), StandardCharsets.UTF_8)).isEqualTo("Test Content");
     }
 
+    @Test
+    public void testStreamsStats() throws Exception
+    {
+        String streamStatsResponseAsString = "{\"operationMode\":\"NORMAL\"," +
+                                             "\"streamProgressStats\":{\"totalFilesToReceive\":7," +
+                                             "\"totalFilesReceived\":7,\"totalBytesToReceive\":15088," +
```
bbotella: These numbers make me wonder if it would be worth to consider adding [Quick Theories](https://github.com/quicktheories/QuickTheories) testing library to sidecar like we have in Cassandra. I could see the case of leaving that out of this PR, but it would definitely increase coverage.

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
```diff
@@ -1573,6 +1575,36 @@ public void onError(Throwable throwable)
...
+        assertThat(progressStats).isNotNull();
+        assertThat(progressStats.totalBytesReceived()).isNotNull();
+        assertThat(progressStats.totalBytesSent()).isNotNull();
+        assertThat(progressStats.totalBytesToReceive()).isNotNull().isEqualTo(progressStats.totalBytesReceived());
+        assertThat(progressStats.totalBytesToSend()).isNotNull();
+        assertThat(progressStats.totalFilesToReceive()).isNotNull();
+        assertThat(progressStats.totalFilesToSend()).isNotNull();
+        assertThat(progressStats.totalFilesReceived()).isNotNull().isEqualTo(progressStats.totalFilesToReceive());
+        assertThat(progressStats.totalFilesSent()).isNotNull();
+
```
bbotella:
```suggestion
```

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -372,6 +374,9 @@ public Router vertxRouter(Vertx vertx,
         router.get(ApiEndpointsV1.LIST_OPERATIONAL_JOBS_ROUTE)
               .handler(listOperationalJobsHandler);
 
+        router.get(ApiEndpointsV1.STREAM_STATS_ROUTE)
```
bbotella: If (hopefully when :-P ) [this PR](https://github.com/apache/cassandra-sidecar/pull/164) gets merged, I would love to see the new API definition updated there as well. Let's work together to make that happen.

server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
```diff
@@ -0,0 +1,202 @@
...
+                              AtomicBoolean hasStats, AtomicBoolean dataReceived)
+    {
+        StreamStatsResponse streamStatsResponse = response.bodyAsJson(StreamStatsResponse.class);
+        assertThat(streamStatsResponse).isNotNull();
+        StreamProgressStats streamProgress = streamStatsResponse.streamProgressStats();
+        assertThat(streamProgress).isNotNull();
+        if (streamProgress.totalFilesToReceive() > 0)
+        {
+            hasStats.set(true);
+            if (streamProgress.totalFilesToReceive() == streamProgress.totalFilesReceived() &&
```
bbotella, line 126: Do we need an else here?

server/src/test/java/org/apache/cassandra/sidecar/routes/StreamStatsHandlerTest.java
```diff
@@ -0,0 +1,183 @@
...
+        else
+            LOGGER.error("Close event timed out.");
+    }
+
+    @Test
+    void testStreamingStatsHandler(VertxTestContext context)
+    {
+        StreamingStatsTestModule.streamingStatsSupplier = () -> {
+            StreamStatsResponse response = new StreamStatsResponse("NORMAL",
+                                                                   new StreamProgressStats(7, 7, 1024, 1024, 0, 0, 0, 0));
```
bbotella: Same comment with quick theories testing library.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStreamManagerOperations.java
arjunashok: Thanks for sharing. This requires a broader discussion around which of these requirements apply to the sidecar in general. One way of looking at it is for sidecar to inherit the model that Cassandra adopts (although, seems like the thread did not arrive at an action item). 

The alternative, which I am inclined to, is to look closer into the specific sidecar behavior/functionality we are optimizing for. For control plane use-cases, which involve APIs that we query to determine node states, we should be fine with the current usage, with readability and maintainability in mind. 

For bulk-analytics, CDC and resource sensitive/large-dataset use-cases, we should rely on perf benchmarks and review the streams usage (on similar lines to the hot-path discussion in C*).

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/ProgressInfo.java
```diff
@@ -0,0 +1,55 @@
...
+package org.apache.cassandra.sidecar.adapters.base.data;
+
+import javax.management.openmbean.CompositeData;
+
+/**
+ * Representation of the stream progress info
+ */
+public class ProgressInfo
+{
+
```
frankgh:
```suggestion
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/ProgressInfo.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * @return true if transfer is completed
+     */
+    public boolean isCompleted()
+    {
+        return currentBytes >= totalBytes;
+    }
+
```
frankgh:
```suggestion
```

server/src/main/java/org/apache/cassandra/sidecar/routes/StreamStatsHandler.java
```diff
@@ -0,0 +1,95 @@
...
+            return;
+        }
+
+        StorageOperations storageOperations = delegate.storageOperations();
+        StreamManagerOperations streamMgrOperations = delegate.streamManagerOperations();
+        if (storageOperations == null || streamMgrOperations == null)
+        {
+            context.fail(cassandraServiceUnavailable());
+            return;
+        }
```
frankgh: this changed in https://issues.apache.org/jira/browse/CASSSIDECAR-182, and should be simplified to not do any null checks.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/StorageOperations.java
```diff
@@ -102,4 +102,6 @@ default void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tab
     {
         outOfRangeDataCleanup(keyspace, table, 1);
     }
+
+    String getOperationMode();
```
frankgh: should be `operationMode` and there are no javadocs here
```suggestion
    String operationMode();
```

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/StreamManagerOperations.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package org.apache.cassandra.sidecar.common.server;
+
+import org.apache.cassandra.sidecar.common.response.data.StreamProgressStats;
+
+/**
+ * An interface that defines interactions with the stream manager system in Cassandra.
+ */
+public interface StreamManagerOperations
```
frankgh: we should combine this into a `Statistics` interface that combines both the `MetricsOperations` and `StreamManagerOperations` interfaces

server/src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -410,6 +411,10 @@ public MetricsOperations metricsOperations()
         return fromAdapter(ICassandraAdapter::metricsOperations);
     }
 
+    public StreamManagerOperations streamManagerOperations()
```
frankgh: should go away

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
```diff
@@ -121,6 +122,19 @@ protected void ifMetricsOpsAvailable(RoutingContext context,
...
+                                           Consumer<StreamManagerOperations> ifAvailable)
+    {
+        StreamManagerOperations operations = delegate.streamManagerOperations();
+        if (operations == null)
+        {
+            context.fail(cassandraServiceUnavailable());
+            return;
+        }
+        ifAvailable.accept(operations);
+    }
```
frankgh: this is no longer necessary after https://issues.apache.org/jira/browse/CASSSIDECAR-182, please remove

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/StreamStatsResponse.java
```diff
@@ -0,0 +1,61 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.cassandra.sidecar.common.response.data.StreamProgressStats;
+
+/**
+ * Class response for the StreamStats API
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class StreamStatsResponse
+{
+   @JsonProperty("operationMode")
```
frankgh: there are some formatting issues in this class. Also methods should go below the constructor. Please refer to the Code Structure -> Class Layout section in the Cassandra [code style](https://cassandra.apache.org/_/development/code_style.html)

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/StreamStatsResponse.java
```diff
@@ -0,0 +1,61 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.cassandra.sidecar.common.response.data.StreamProgressStats;
+
+/**
+ * Class response for the StreamStats API
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class StreamStatsResponse
```
frankgh, line 30: I think we need to understand why the operation mode is required for this endpoint. maybe elaborate somewhere why we want the operation mode as part of this response.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/data/StreamProgressStats.java
```diff
@@ -0,0 +1,113 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * A class representing stats summarizing the progress of streamed bytes and files on the node
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class StreamProgressStats
+{
+    @JsonProperty("totalFilesToReceive")
+    public long totalFilesToReceive()
```
frankgh: class layout is wrong, can you please follow the code style guidelines.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/SessionInfo.java
```diff
@@ -0,0 +1,163 @@
...
+        this.sessionIndex = (int) data.get("sessionIndex");
+        this.connecting = (String) data.get("connecting");
+        this.receivingSummaries = parseSummaries((CompositeData[]) data.get("receivingSummaries"));
+        this.sendingSummaries = parseSummaries((CompositeData[]) data.get("sendingSummaries"));
+        this.state = (String) data.get("state");
+        this.receivingFiles = parseFiles((CompositeData[]) data.get("receivingFiles"));
+        this.sendingFiles = parseFiles((CompositeData[]) data.get("sendingFiles"));
+    }
+
+    private Collection<StreamSummary> parseSummaries(CompositeData[] summaries)
```
frankgh: these helper methods should go at the end.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/SessionInfo.java
```diff
@@ -0,0 +1,163 @@
...
+        long total = 0;
+        for (StreamSummary summary : summaries)
+            total += summary.files;
+        return total;
+    }
+
+    /**
+     * @return total size(in bytes) already received.
+     */
+    public long getTotalSizeReceived()
```
frankgh:
```suggestion
    public long totalSizeReceived()
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/SessionInfo.java
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return getTotalSizeInProgress(receivingFiles);
+    }
+
+    /**
+     * @return total size(in bytes) already sent.
+     */
+    public long getTotalSizeSent()
```
frankgh:
```suggestion
    public long totalSizeSent()
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/SessionInfo.java
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return getTotalSizeInProgress(sendingFiles);
+    }
+
+    /**
+     * @return total number of files to receive in the session
+     */
+    public long getTotalFilesToReceive()
```
frankgh:
```suggestion
    public long totalFilesToReceive()
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/SessionInfo.java
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return getTotalFiles(receivingSummaries);
+    }
+
+    /**
+     * @return total number of files to send in the session
+     */
+    public long getTotalFilesToSend()
```
frankgh:
```suggestion
    public long totalFilesToSend()
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/SessionInfo.java
```diff
@@ -0,0 +1,163 @@
...
+        long total = 0;
+        for (StreamSummary summary : summaries)
+            total += summary.totalSize;
+        return total;
+    }
+
+    /**
+     * @return total size(in bytes) to receive in the session
+     */
+    public long getTotalSizeToReceive()
```
frankgh:
```suggestion
    public long totalSizeToReceive()
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/SessionInfo.java
```diff
@@ -0,0 +1,163 @@
...
+        long total = 0;
+        for (ProgressInfo stream : streams)
+            total += stream.currentBytes;
+        return total;
+    }
+
+    /**
+     * @return total size(in bytes) to send in the session
+     */
+    public long getTotalSizeToSend()
```
frankgh:
```suggestion
    public long totalSizeToSend()
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/SessionInfo.java
```diff
@@ -0,0 +1,163 @@
...
+    private long getTotalFilesCompleted(Collection<ProgressInfo> files)
+    {
+        Iterable<ProgressInfo> completed = Iterables.filter(files, input -> input.isCompleted());
+        return Iterables.size(completed);
+    }
+
+    /**
+     * @return total number of files already received.
+     */
+    public long getTotalFilesReceived()
```
frankgh:
```suggestion
    public long totalFilesReceived()
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/SessionInfo.java
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return getTotalFilesCompleted(receivingFiles);
+    }
+
+    /**
+     * @return total number of files already sent.
+     */
+    public long getTotalFilesSent()
```
frankgh:
```suggestion
    public long totalFilesSent()
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/SessionInfo.java
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return getTotalFilesCompleted(sendingFiles);
+    }
+
+
```
frankgh:
```suggestion
```

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -125,6 +125,7 @@ public final class ApiEndpointsV1
     public static final String PER_OPERATIONAL_JOB = OPERATIONAL_JOBS + '/' + OPERATIONAL_JOB_ID_PATH_PARAM;
     public static final String LIST_OPERATIONAL_JOBS_ROUTE = API_V1 + CASSANDRA + OPERATIONAL_JOBS;
     public static final String OPERATIONAL_JOB_ROUTE = API_V1 + CASSANDRA + PER_OPERATIONAL_JOB;
+    public static final String STREAM_STATS_ROUTE = API_V1 + CASSANDRA + "/stats/stream";
```
frankgh: Should be plural here if you are retrieving all the streams
```suggestion
    public static final String STREAM_STATS_ROUTE = API_V1 + CASSANDRA + "/stats/streams";
```

server/src/main/java/org/apache/cassandra/sidecar/routes/StreamStatsHandler.java
```diff
@@ -0,0 +1,95 @@
...
+                         String mode = storageOperations.getOperationMode();
+                         StreamProgressStats stats = streamMgrOperations.getStreamProgressStats();
+                         return new StreamStatsResponse(mode, stats);
+                     })
+                     .onSuccess(context::json)
+                     .onFailure(cause -> processFailure(cause, context, host, remoteAddress, request));
+
+    }
+
+    protected Void extractParamsOrThrow(RoutingContext context)
```
frankgh, line 90: let's add the `@Override` annotation here.

server/src/main/java/org/apache/cassandra/sidecar/routes/StreamStatsHandler.java
```diff
@@ -0,0 +1,95 @@
...
+import org.apache.cassandra.sidecar.utils.InstanceMetadataFetcher;
+
+import static org.apache.cassandra.sidecar.utils.HttpExceptions.cassandraServiceUnavailable;
+
+/**
+ * Handler for retrieving node streams stats
+ */
+public class StreamStatsHandler extends AbstractHandler<Void>
+{
+
```
frankgh:
```suggestion
```

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -664,6 +665,18 @@ public CompletableFuture<ListOperationalJobsResponse> listOperationalJobs(Sideca
                                             .build());
     }
 
+    /**
+     * Executes the stream stats request using the default retry policy and configured selection policy
```
frankgh:
```suggestion
     * Executes the streams stats request using the default retry policy and configured selection policy
```

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -664,6 +665,18 @@ public CompletableFuture<ListOperationalJobsResponse> listOperationalJobs(Sideca
...
+     *
+     * @return a completable future of the connected client stats
+     */
+    public CompletableFuture<StreamStatsResponse> streamsStats()
```
frankgh: this request should be directed at a single instance I assume?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/StreamSummary.java
```diff
@@ -0,0 +1,50 @@
...
+    public final int files;
+    public final long totalSize;
+
+
+    public StreamSummary(String tableId, int files, long totalSize)
+    {
+        this.tableId = tableId;
+        this.files = files;
+        this.totalSize = totalSize;
+    }
```
frankgh: this constructor is unused. do we need it?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/StreamSummary.java
```diff
@@ -0,0 +1,50 @@
...
+public class StreamSummary
+{
+    public final String tableId;
+
+    /**
+     * Number of files to transfer. Can be 0 if nothing to transfer for some streaming request.
+     */
+    public final int files;
+    public final long totalSize;
+
```
frankgh, line 35:
```suggestion
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/StreamState.java
```diff
@@ -0,0 +1,48 @@
...
+public class StreamState
+{
+    Collection<SessionInfo> sessions;
+
+    public StreamState(CompositeData data)
+    {
+        this.sessions = parseSessions((CompositeData[]) data.get("sessions"));
+    }
+
+    private Collection<SessionInfo> parseSessions(CompositeData[] sessions)
```
frankgh: let's move helper methods to the end of the class

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/StreamState.java
```diff
@@ -0,0 +1,48 @@
...
+    {
+        return Arrays.stream(sessions).map(SessionInfo::new).collect(Collectors.toList());
```
frankgh: also agree with Bernardo that streams should be used carefully. This is less readable to me than a good old for loop.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/StorageJmxOperations.java
```diff
@@ -159,4 +159,10 @@ public interface StorageJmxOperations
      * @throws InterruptedException it does not really throw but declared in MBean
      */
     int forceKeyspaceCleanup(int jobs, String keyspaceName, String... tables) throws IOException, ExecutionException, InterruptedException;
+
+    /**
+     * Fetch the operation-mode of the node
+     * @return string representation of theoperation-mode
```
frankgh:
```suggestion
     * @return string representation of the operation-mode
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraStreamManagerOperations.java
```diff
@@ -0,0 +1,99 @@
...
+            totalBytesReceived += s.getTotalSizeReceived();
+            totalFilesToReceive += s.getTotalFilesToReceive();
+            totalFilesReceived += s.getTotalFilesReceived();
+            totalBytesToSend += s.getTotalSizeToSend();
+            totalBytesSent += s.getTotalSizeSent();
+            totalFilesToSend += s.getTotalFilesToSend();
+            totalFilesSent += s.getTotalFilesSent();
+
+        }
+        LOGGER.info("Progress Stats: {}, {}, {}, {}", totalBytesToReceive, totalBytesReceived, totalBytesToSend, totalBytesSent);
```
frankgh: hmm, should we move this to debug instead?

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
```diff
@@ -1573,6 +1575,36 @@ public void onError(Throwable throwable)
...
+        MockResponse response = new MockResponse().setResponseCode(OK.code()).setBody(streamStatsResponseAsString);
+        enqueue(response);
+        StreamStatsResponse result = client.streamsStats().get();
+
+        assertThat(result).isNotNull();
+        assertThat(result.operationMode()).isNotNull().isEqualTo("NORMAL");
+        StreamProgressStats progressStats = result.streamProgressStats();
+        assertThat(progressStats).isNotNull();
+        assertThat(progressStats.totalBytesReceived()).isNotNull();
```
frankgh: no validation of the actual values?

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
```diff
@@ -1573,6 +1575,36 @@ public void onError(Throwable throwable)
         assertThat(new String(baos.toByteArray(), StandardCharsets.UTF_8)).isEqualTo("Test Content");
     }
 
+    @Test
+    public void testStreamsStats() throws Exception
+    {
+        String streamStatsResponseAsString = "{\"operationMode\":\"NORMAL\"," +
+                                             "\"streamProgressStats\":{\"totalFilesToReceive\":7," +
```
frankgh:
```suggestion
                                             "\"streamsProgressStats\":{\"totalFilesToReceive\":7," +
```

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
```diff
@@ -1573,6 +1575,36 @@ public void onError(Throwable throwable)
...
+                                             "\"totalFilesReceived\":7,\"totalBytesToReceive\":15088," +
+                                             "\"totalBytesReceived\":15088,\"totalFilesToSend\":0,\"totalFilesSent\":0," +
+                                             "\"totalBytesToSend\":0,\"totalBytesSent\":0}}";
```
frankgh: maybe let's try with non-zero values, since 0 is the default

server/src/test/java/org/apache/cassandra/sidecar/routes/StreamStatsHandlerTest.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+
+    static class StreamingStatsTestModule extends AbstractModule
+    {
+        static Supplier<StreamStatsResponse> streamingStatsSupplier;
+
+        @Provides
+        @Singleton
+        public InstancesConfig instanceConfig() throws IOException
```
frankgh:
```suggestion
        public InstancesConfig instanceConfig()
```

server/src/test/java/org/apache/cassandra/sidecar/routes/StreamStatsHandlerTest.java
```diff
@@ -0,0 +1,183 @@
...
+                  assertThat(response.statusCode()).isEqualTo(OK.code());
+                  StreamStatsResponse statsResponse = response.bodyAsJson(StreamStatsResponse.class);
+                  assertThat(statsResponse).isNotNull();
+                  assertThat(statsResponse.operationMode()).isEqualTo("NORMAL");
+                  context.completeNow();
+              }));
+    }
+
+    @Test
+    void testStreamingStatsHandlerFailure(VertxTestContext context)
```
frankgh: this test is exactly the same as the `testStreamingStatsHandler` test. Are we really testing failure here?

spotbugs-exclude.xml
```diff
@@ -62,7 +62,10 @@
 
     <Match>
         <Bug pattern="ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD" />
-        <Class name="org.apache.cassandra.sidecar.routes.RingHandlerTest" />
+        <Or>
+            <Class name="org.apache.cassandra.sidecar.routes.RingHandlerTest" />
+            <Class name="org.apache.cassandra.sidecar.routes.StreamStatsHandlerTest" />
```
frankgh: let's not statically set the fields

server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
```diff
@@ -0,0 +1,202 @@
...
+/**
+ * Tests the stream stats endpoint with cassandra container.
+ */
+@ExtendWith(VertxExtension.class)
+public class StreamStatsIntegrationTest extends IntegrationTestBase
+{
+    @CassandraIntegrationTest(numDataDirsPerInstance = 4, nodesPerDc = 5, network = true, buildCluster = false)
+    void streamStatsTest(VertxTestContext context, ConfigurableCassandraTestContext cassandraTestContext) throws InterruptedException
+    {
+
```
frankgh:
```suggestion
```

server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
```diff
@@ -0,0 +1,202 @@
...
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests the stream stats endpoint with cassandra container.
+ */
+@ExtendWith(VertxExtension.class)
+public class StreamStatsIntegrationTest extends IntegrationTestBase
+{
+    @CassandraIntegrationTest(numDataDirsPerInstance = 4, nodesPerDc = 5, network = true, buildCluster = false)
```
frankgh: do we need 5 nodes? We are trying to save time on test runtime and carefully choosing the cluster size is a goal we have to reduce the runtime and resources needed to run tests.

server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
```diff
@@ -0,0 +1,202 @@
...
+        testWithClient(context, client -> {
+            BBHelperDecommissioningNode.transientStateEnd.countDown();
+            client.get(server.actualPort(), "127.0.0.1", testRoute)
+                  .send(context.succeeding(response -> {
+                       assertRingResponseOK(response, sidecarTestContext, hasStats, dataReceived);
+                  }));
+        });
+    }
+
+    void assertRingResponseOK(HttpResponse<Buffer> response, CassandraSidecarTestContext cassandraTestContext,
```
frankgh: this is not really a ring response? Also cassandraTestContext is unused
```suggestion
    void assertRingResponseOK(HttpResponse<Buffer> response,
```

server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
```diff
@@ -0,0 +1,202 @@
...
+        AtomicBoolean hasStats = new AtomicBoolean(false);
+        AtomicBoolean dataReceived = new AtomicBoolean(false);
+
+        // Wait until nodes have reached expected state
+        awaitLatchOrThrow(BBHelperDecommissioningNode.transientStateStart, 2, TimeUnit.MINUTES, "transientStateStart");
+
+        ClusterUtils.awaitRingState(seed, node, "Leaving");
+        BBHelperDecommissioningNode.transientStateEnd.countDown();
+
+        for (int i = 0; i < 20; i++)
```
frankgh: why do it 20 times?

frankgh (changes requested):

frankgh (changes requested):
There are some newer commits that will simplify a few aspects of the handler. Can you please rebase.

server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
arjunashok: This was an optimal no. of attempts to poll for stats with the provided delay to deterministically capture streaming stats after a node was decommissioned. Will add this as a comment.

server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
arjunashok: Since the integration test is polling the stats API until we actually have some data streamed to validate, we only validate bytes received in such cases. However, the test does validate that there were stats captured and data received at the end of all the attempts.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/StreamStatsResponse.java
arjunashok: Good question. It seems like the mode is relevant to the streaming stats, as the stats are only meaningful when a cluster has nodes that are not in a terminal state i.e. `NORMAL` or `DECOMMISSIONED`. Also, since we are drawing parallels to streaming-stats from nodetool netstats, the mode is part of the netstats results as well.

server/src/test/java/org/apache/cassandra/sidecar/routes/StreamStatsHandlerTest.java
arjunashok: Ah, missed this. will update

server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
arjunashok: Switched to 2 node cluster

spotbugs-exclude.xml
arjunashok: Removed

server/src/test/java/org/apache/cassandra/sidecar/routes/StreamStatsHandlerTest.java
arjunashok: This goes away with the recent refactoring

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
arjunashok: Yep, this is the sidecar specific interface for data access, so can be part of `MetricsOperations`.

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
arjunashok: It's a valid point. Although, this API (as it stands today) is a simple aggregation (sum) of the stats, so we would need to evaluate the value prop testing would provide. We should track that in a separate ticket regardless.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/StreamSummary.java
arjunashok: Removed

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/StreamState.java
arjunashok: I would argue that this is more readable given it's a simple transformation to a class type (each transformation could be on a newline if needed, but that's a styling aspect). Also I see several references across sidecar codebase of similar usage. 

If we want to make a concerted effort move away from stream usage in sidecar, we should have a broader discussion with consensus if it makes sense for sidecar, which I feel is outside the scope of this PR.

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
arjunashok: Added

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraMetricsOperations.java
```diff
@@ -70,6 +88,52 @@ public ConnectedClientStatsResponse connectedClientDetails()
...
+        Set<CompositeData> streamData = jmxClient.proxy(StreamManagerJmxOperations.class, STREAM_MANAGER_OBJ_NAME)
+                                                 .getCurrentStreams();
+
+        List<StreamState> streamStates = streamData.stream().map(StreamState::new).collect(Collectors.toList());
+        return computeStats(streamStates);
+    }
+
+    private StreamsProgressStats computeStats(List<StreamState> streamStates)
+    {
+        List<SessionInfo> sessions = streamStates.stream().map(s -> s.sessions()).flatMap(Collection::stream).collect(Collectors.toList());
```
yifan-c: Nit: Since `streamStates` is turned into stream immediately, you can avoid materialize the stream into list at line#100. Just pass the `Stream<StreamState>` to `computeStats`. 

Here is my suggestion to avoid materializing the lists at all. 

```java
        return computeStats(streamData.stream().map(StreamState::new));
    }

    private StreamsProgressStats computeStats(Stream<StreamState> streamStates)
    {
        Iterator<SessionInfo> sessions = streamStates.map(StreamState::sessions).flatMap(Collection::stream).iterator();

        long totalFilesToReceive = 0;
        long totalFilesReceived = 0;
        long totalBytesToReceive = 0;
        long totalBytesReceived = 0;

        long totalFilesToSend = 0;
        long totalFilesSent = 0;
        long totalBytesToSend = 0;
        long totalBytesSent = 0;

        while (sessions.hasNext())
        {
            SessionInfo sessionInfo = sessions.next();
            totalBytesToReceive += sessionInfo.totalSizeToReceive();
            totalBytesReceived += sessionInfo.totalSizeReceived();
            totalFilesToReceive += sessionInfo.totalFilesToReceive();
            totalFilesReceived += sessionInfo.totalFilesReceived();
            totalBytesToSend += sessionInfo.totalSizeToSend();
            totalBytesSent += sessionInfo.totalSizeSent();
            totalFilesToSend += sessionInfo.totalFilesToSend();
            totalFilesSent += sessionInfo.totalFilesSent();
        }
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/StreamManagerJmxOperations.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.management.openmbean.CompositeData;
+
+/**
+ * An interface that pulls methods from the Cassandra Stream manager Proxy
+ */
+public interface StreamManagerJmxOperations
+{
+
+    String STREAM_MANAGER_OBJ_NAME = "org.apache.cassandra.net:type=StreamManager";
+    Set<CompositeData> getCurrentStreams();
```
yifan-c, line 36: add java doc.

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/ProgressInfo.java
```diff
@@ -0,0 +1,52 @@
...
+    public final long totalBytes;
+
+    public ProgressInfo(CompositeData data)
+    {
+        this.peer = (String) data.get("peer");
+        this.sessionIndex = (int) data.get("sessionIndex");
+        this.fileName = (String) data.get("fileName");
+        this.direction = (String) data.get("direction");
+        this.currentBytes = (long) data.get("currentBytes");
+        this.totalBytes = (long) data.get("totalBytes");
```
yifan-c: How about having this utility method to provide richer info when value is absent or type mismatches. 

```java
    public static <T> T extractValue(CompositeData data, String key)
    {
        Object value = data.get(key);
        if (value == null)
        {
            throw new NoSuchElementException("No value is present for key: " + key);
        }

        try
        {
            return (T) value;
        }
        catch (ClassCastException cce)
        {
            throw new RuntimeException("Value type mismatched of key: " + key, cce);
        }
    }
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/StreamState.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Collection;
+import java.util.stream.Collectors;
+import javax.management.openmbean.CompositeData;
+
+/**
+ * Representation of the stream state data
+ */
+public class StreamState
+{
+    Collection<SessionInfo> sessions;
```
yifan-c: 1. `private final`?
2. It does not buy anything with using `Collection`. How about changing it to `List`?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/StreamState.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * @return the session info for the sessions in the stream stats data
+     */
+    public Collection<SessionInfo> sessions()
+    {
+        return sessions;
+    }
+
+    private Collection<SessionInfo> parseSessions(CompositeData[] sessions)
```
yifan-c: Return `List` in both methods.

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -129,6 +129,7 @@ public final class ApiEndpointsV1
     public static final String LIST_OPERATIONAL_JOBS_ROUTE = API_V1 + CASSANDRA + OPERATIONAL_JOBS;
     public static final String OPERATIONAL_JOB_ROUTE = API_V1 + CASSANDRA + PER_OPERATIONAL_JOB;
     public static final String NODE_DECOMMISSION_ROUTE = API_V1 + CASSANDRA + "/operations/decommission";
+    public static final String STREAM_STATS_ROUTE = API_V1 + CASSANDRA + "/stats/streams";
```
yifan-c, line 132: should `streams` be `stream` instead? why it is in plural form? 

I am reading it as getting the `stream (stat)` from the `stats` (in plural) resource.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/StreamStatsRequest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import io.netty.handler.codec.http.HttpMethod;
+import org.apache.cassandra.sidecar.common.ApiEndpointsV1;
+import org.apache.cassandra.sidecar.common.response.StreamStatsResponse;
+
+/**
+ * Class response for the StreamsStats API
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
yifan-c: This is not json pojo. The annotation can be removed.

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
```diff
@@ -1622,6 +1624,42 @@ public void onError(Throwable throwable)
...
+                                                        .isEqualTo(2);
+            assertThat(progressStats.totalBytesToSend()).isNotNull()
+                                                      .isEqualTo(progressStats.totalBytesSent())
+                                                      .isEqualTo(1024);
+            assertThat(progressStats.totalBytesToReceive()).isNotNull()
+                                                           .isEqualTo(progressStats.totalBytesReceived())
+                                                           .isEqualTo(15088);
+            assertThat(progressStats.totalFilesToReceive()).isNotNull()
+                                                           .isEqualTo(progressStats.totalFilesReceived())
+                                                           .isEqualTo(7);
```
yifan-c: nit: create `StreamsProgressStats` object and serialize it into string, instead of declaring  `streamStatsResponseAsString`. For assertion, now it can check the received `StreamsProgressStats` has the same values as the original `StreamsProgressStats`

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/MetricsOperations.java
```diff
@@ -32,4 +33,9 @@ public interface MetricsOperations
      */
     ConnectedClientStatsResponse connectedClientStats(boolean summaryOnly);
 
+    /**
+     * Retrieve the stream progress stats metrics from the cluster
```
yifan-c: nit: `stats` == `metrics`. We can remove `metrics`

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/MetricsOperations.java
```diff
@@ -32,4 +33,9 @@ public interface MetricsOperations
...
+     * @return the requested stream progress stats, in full or summary
```
yifan-c: since `in full or summary` is mentioned, can you explain what is full and what is summary in the javadoc.

server/src/main/java/org/apache/cassandra/sidecar/routes/StreamStatsHandler.java
```diff
@@ -0,0 +1,99 @@
...
+                               HttpServerRequest httpRequest,
+                               String host,
+                               SocketAddress remoteAddress,
+                               Void request)
+    {
+
+        CassandraAdapterDelegate delegate = metadataFetcher.delegate(host);
+
+        StorageOperations storageOperations = delegate.storageOperations();
+        MetricsOperations metricsOperations = delegate.metricsOperations();
```
yifan-c: nit: move those 2 lines into the lambda of `executeBlocking`

server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
yifan-c: can we use `loopAssert` here?

server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
```diff
@@ -0,0 +1,208 @@
...
+        {
+            if (nodeNumber == 2)
+            {
+                TypePool typePool = TypePool.Default.of(cl);
+                TypeDescription description = typePool.describe("org.apache.cassandra.streaming.StreamCoordinator")
+                                                      .resolve();
+                new ByteBuddy().rebase(description, ClassFileLocator.ForClassLoader.of(cl))
+                               .method(named("connectAllStreamSessions"))
+//                               .method(named("onInitializationComplete"))
+//                               .method(named("start"))
```
yifan-c: remove them?

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
arjunashok: @frankgh had a comment to make it plural in previous revision.  I believe the reasoning was, since we are aggregating stats across multiple streams

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/StreamStatsRequest.java
arjunashok: This is a request pojo. I believe what you're saying is we're not serializing it, makes sense.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/MetricsOperations.java
arjunashok: not applicable. will remove

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
```diff
@@ -65,5 +65,5 @@ public class BasicPermissions
     public static final Permission READ_TOPOLOGY = new DomainAwarePermission("TOPOLOGY:READ");
 
     // cassandra stats permissions
-    public static final Permission READ_STATS = new DomainAwarePermission("STATS:READ");
+    public static final Permission STATS = new StandardPermission("STATS");
```
yifan-c: yeah. It make sense to omit "READ", as read is the only action applicable here plausibly

yifan-c (approved):

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraMetricsOperations.java
```diff
@@ -18,33 +18,52 @@
...
 import org.jetbrains.annotations.NotNull;
 
+import static org.apache.cassandra.sidecar.adapters.base.StreamManagerJmxOperations.STREAM_MANAGER_OBJ_NAME;
+
 /**
  * Default implementation that pulls methods from the Cassandra Metrics Proxy
  */
 public class CassandraMetricsOperations implements MetricsOperations
 {
+    private static final Logger LOGGER = LoggerFactory.getLogger(CassandraStorageOperations.class);
```
frankgh: is this on purpose?
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(CassandraMetricsOperations.class);
```

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/GossipDependentStorageJmxOperations.java
frankgh: I guess we have this method available as of https://github.com/apache/cassandra-sidecar/commit/f45bbacc42990e5196c927ed6169d82cd9faf5f6. Maybe restore this file since the change is no longer necessary?

adapters/base/src/main/java/org/apache/cassandra/sidecar/adapters/base/data/SessionInfo.java
```diff
@@ -0,0 +1,160 @@
...
+        long total = 0;
+        for (StreamSummary summary : summaries)
+            total += summary.totalSize;
+        return total;
+    }
+
+    private long totalFilesCompleted(List<ProgressInfo> files)
+    {
+        Iterable<ProgressInfo> completed = Iterables.filter(files, input -> input.isCompleted());
+        return Iterables.size(completed);
```
frankgh: maybe simpler to do ? Any reason why we decide to use Iterables here?
```suggestion
                return files.stream()
                    .filter(ProgressInfo::isCompleted)
                    .count();
```

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/StreamStatsRequest.java
frankgh: yeah, this should not be included here. Can we remove it ?

client/src/main/java/org/apache/cassandra/sidecar/client/RequestContext.java
```diff
@@ -537,6 +538,17 @@ public Builder nodeDecommissionRequest()
...
 
+        /**
+         * Sets the {@code request} to be a {@link StreamStatsRequest} and returns a reference to this Builder
+         * enabling method chaining.
+         *
+         * @return a reference to this Builder
+         */
+        public Builder streamsStatsRequest()
+        {
+            return request(new StreamStatsRequest());
```
frankgh: minor NIT: Can be created once, since it won't ever change for the client. ie.
```
protected static final StreamStatsRequest STREAM_STATS_REQUEST = new StreamStatsRequest();
```

Obviously there are drawbacks with this approach, so that's why it's a minor NIT

frankgh (commented):
Looking good. I left a few additional comments, and there's some comments from @yifan-c that need to be addressed.

client/src/main/java/org/apache/cassandra/sidecar/client/RequestContext.java
arjunashok: Will update to make it consistent with existing code.

frankgh (approved):
+1

PR #168, opened 2024-12-22
CASSSIDECAR-182: Refactor access to delegate methods to simplify

frankgh:
Patch by Francisco Guerrero; Reviewed by Yifan Cai for CASSSIDECAR-182


yifan-c (commented):
I am feeling neutral about the patch. If I were to refactor the code, I would rather ensure the getters of `CassandraAdapterDelegate` return non-null result or throws proper exception. Returning null only transfers the handling to the callsites, and most of the callsites handle `null` case the same way by throwing exception. For certain scenarios that `null` is expected, the delegate can expose `xxxOrNull` methods (which are the same as the current ones).

There are 2 problems with the new methods introduced in this patch, (e.g. `ifAvailableFromDelegate`). First, they are complex. Second, they cannot be easily discovered, since one has to pair those helpers with the concrete getters.

frankgh:
> I am feeling neutral about the patch. If I were to refactor the code, I would rather ensure the getters of `CassandraAdapterDelegate` return non-null result or throws proper exception. Returning null only transfers the handling to the callsites, and most of the callsites handle `null` case the same way by throwing exception. For certain scenarios that `null` is expected, the delegate can expose `xxxOrNull` methods (which are the same as the current ones).
> 
> There are 2 problems with the new methods introduced in this patch, (e.g. `ifAvailableFromDelegate`). First, they are complex. Second, they cannot be easily discovered, since one has to pair those helpers with the concrete getters.

I agree with you . I tried with the direction you suggested and applied the changes.

yifan-c (commented):
It is getting better with the removal of nullable methods. There are still other parts to be updated. I will prepare a patch.

PR #172, opened 2025-01-07
CASSSIDECAR-186: Standardize configuration for duration units

frankgh:
Patch by Francisco Guerrero; reviewed by TBD for CASSSIDECAR-186


conf/sidecar.yaml
```diff
@@ -60,70 +60,74 @@ cassandra_instances:
...
     stream_requests_per_sec: 5000
-    timeout_sec: 10
+    timeout: 10s
   traffic_shaping:
     inbound_global_bandwidth_bps: 0               # 0 implies unthrottled, the inbound bandwidth in bytes per second
     outbound_global_bandwidth_bps: 0              # 0 implies unthrottled, the outbound bandwidth in bytes per second
     peak_outbound_global_bandwidth_bps: 419430400 # the peak outbound bandwidth in bytes per second. The default is 400 mebibytes per second
-    max_delay_to_wait_millis: 15000               # 15 seconds
-    check_interval_for_stats_millis: 1000         # 1 second
+    max_delay_to_wait: 15s                        # 15 seconds
```
bbotella, line 75: Do we need the `15 seconds` comment anymore? Same for the rest of the commented time units.

server/src/main/java/org/apache/cassandra/sidecar/config/DurationSpec.java
```diff
@@ -0,0 +1,120 @@
...
+
+import static java.util.concurrent.TimeUnit.DAYS;
+import static java.util.concurrent.TimeUnit.HOURS;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Represents a positive time duration. Wrapper interface for Cassandra Sidecar duration configuration parameters,
+ * providing to the users the opportunity to be able to provide configuration values with a unit of their choice
```
bbotella:
```suggestion
 * allowing the users the opportunity to be able to provide configuration values with a unit of their choice
```

server/src/main/java/org/apache/cassandra/sidecar/config/DurationSpec.java
```diff
@@ -0,0 +1,120 @@
...
+            case HOURS:
+                return "h";
+            case MINUTES:
+                return "m";
+            case SECONDS:
+                return "s";
+            case MILLISECONDS:
+                return "ms";
+        }
+        throw new AssertionError();
```
bbotella: Maybe adding some text to the error?

server/src/main/java/org/apache/cassandra/sidecar/config/MillisecondBoundConfiguration.java
```diff
@@ -0,0 +1,66 @@
...
+ * Represents a duration used for Sidecar configuration. The bound is [0, Long.MAX_VALUE) in milliseconds.
+ * If the user sets a different unit - we still validate that converted to milliseconds the quantity will not exceed
+ * that upper bound.
+ */
+public interface MillisecondBoundConfiguration extends DurationSpec
+{
+    /**
+     * Represents a 0-millisecond configuration
+     */
+    MillisecondBoundConfiguration ZERO = MillisecondBoundConfiguration.parse("0ms");
```
bbotella: Nit: wouldn't it be faster to have a constructor instead of having to parse the string?

server/src/main/java/org/apache/cassandra/sidecar/config/ServiceConfiguration.java
```diff
@@ -92,7 +92,7 @@ default List<SocketAddress> listenSocketAddresses()
     /**
      * @return the maximum time skew allowed between the server and the client
      */
-    int allowableSkewInMinutes();
+    SecondBoundConfiguration allowableTimeSkew();
```
bbotella: Is it ok to be using `SecondBoundConfiguration` instead of `MinuteBoundConfiguration` for this one?

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/AccessControlConfigurationImpl.java
```diff
@@ -36,7 +36,8 @@ public class AccessControlConfigurationImpl implements AccessControlConfiguratio
     private static final boolean DEFAULT_ENABLED = false;
     private static final List<ParameterizedClassConfiguration> DEFAULT_AUTHENTICATORS_CONFIGURATION = Collections.emptyList();
     private static final Set<String> DEFAULT_ADMIN_IDENTITIES = Collections.emptySet();
-    private static final CacheConfiguration DEFAULT_PERMISSION_CACHE_CONFIGURATION = new CacheConfigurationImpl(TimeUnit.HOURS.toMillis(2), 1_000);
+    private static final CacheConfiguration DEFAULT_PERMISSION_CACHE_CONFIGURATION =
+    new CacheConfigurationImpl(MillisecondBoundConfiguration.parse("2h"), 1_000);
```
bbotella, line 40: Really nice

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CacheConfigurationImpl.java
```diff
@@ -41,34 +46,50 @@ public class CacheConfigurationImpl implements CacheConfiguration
...
+
+    @JsonProperty("expire_after_access")
+    public void setExpireAfterAccess(MillisecondBoundConfiguration expireAfterAccess)
+    {
+        this.expireAfterAccess = expireAfterAccess;
+    }
+
     @JsonProperty("expire_after_access_millis")
-    public long expireAfterAccessMillis()
+    public void setExpireAfterAccessMillis(long expireAfterAccessMillis)
```
bbotella, line 96: Given the fact that we haven't made a release yet, wouldn't it be ok to just remove this?

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/DurationSpecImpl.java
```diff
@@ -0,0 +1,143 @@
...
+        return unit;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String toString()
+    {
+        return "'" + quantity + "' " + unit.toString().toLowerCase();
```
bbotella: Why the quotes around quantity? I think it should support to be parsed back.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/DurationSpecImpl.java
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        TimeUnit minUnit = this.unit.compareTo(that.unit()) < 0 ? this.unit : that.unit();
+        return Long.compare(this.to(minUnit), that.to(minUnit));
+    }
+
+    void validateMinUnit(TimeUnit unit, TimeUnit minUnit)
+    {
+        if (unit.compareTo(minUnit) < 0)
+            throw new IllegalArgumentException(String.format("Invalid duration: %s Accepted units:%s",
```
bbotella:
```suggestion
            throw new IllegalArgumentException(String.format("Invalid duration: %s Accepted units: %s",
```

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/PeriodicTaskConfigurationImpl.java
```diff
@@ -67,47 +71,74 @@ public boolean enabled()
...
+            LOGGER.warn("Invalid executeInterval configuration {}, the minimum value is 1ms", executeInterval);
+            this.executeInterval = ONE;
         }
     }
+
+    /**
+     * Legacy properties {@code execute_interval_millis}, {@code poll_freq_millis}, and {@code poll_interval_millis}
+     *
+     * @param executeIntervalMillis the interval in milliseconds
+     * @deprecated in favor of {@code execute_interval}
```
bbotella, line 140: If we finally decide to leave all the deprecated configs, I think we should have a Jira for removing them.

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -741,7 +741,8 @@ private static InstanceMetadata buildInstanceMetadata(Vertx vertx,
                                        .password(cassandraInstance.jmxRolePassword())
                                        .enableSsl(cassandraInstance.jmxSslEnabled())
                                        .connectionMaxRetries(jmxConfiguration.maxRetries())
-                                       .connectionRetryDelayMillis(jmxConfiguration.retryDelayMillis())
+                                       // TODO: connectionRetryDelayMillis should be a MillisecondBoundConfiguration
```
bbotella: Why postponing it?

server/src/test/java/org/apache/cassandra/sidecar/TestModule.java
```diff
@@ -133,12 +135,15 @@ protected SidecarConfigurationImpl abstractConfig(SslConfiguration sslConfigurat
                                 .build();
         RestoreJobConfiguration restoreJobConfiguration =
         RestoreJobConfigurationImpl.builder()
-                                   .restoreJobTablesTtlSeconds(TimeUnit.DAYS.toSeconds(14) + 1)
+                                   .restoreJobTablesTtl(SecondBoundConfiguration.parse((TimeUnit.DAYS.toSeconds(14) + 1) + "s"))
```
bbotella, line 138: It'd be great to be able to sum DurationSpecs directly. But probably overengineering.

bbotella (commented):
This is a really nice PR! Just a few comments.

conf/sidecar.yaml
frankgh: Yeah, I think it's fine to remove it, just left it to be explicit, but it's unnecessary now

server/src/main/java/org/apache/cassandra/sidecar/config/MillisecondBoundConfiguration.java
frankgh: yeah, it's negligible for a one-time configuration, but I can change it.

server/src/main/java/org/apache/cassandra/sidecar/config/ServiceConfiguration.java
frankgh: we don't have a `MinuteBoundConfiguration` class, and the reason I avoided adding one was because it feels to me the time skew resolution is too coarse. This resolution should have been at the second level to begin with IMO. But I am hoping that we think about why the resolution is that coarse and maybe reduce it to seconds.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CacheConfigurationImpl.java
frankgh: We can't assume no one is using Sidecar in the open source, so I think to be on the safe side and play well with open source, we should do the work to be backwards compatible.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/DurationSpecImpl.java
frankgh: good point, I think we should be able to go back and forth. Good suggestion.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/DurationSpecImpl.java
frankgh: I am changing this anyway, I will push changes soon

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/PeriodicTaskConfigurationImpl.java
frankgh: Yeah, I think that should be done after we have the first release though :)

server/src/test/java/org/apache/cassandra/sidecar/TestModule.java
frankgh: yeah, I don't think there's a real use case for production code. Maybe a nice to have for tests, which I don't think we should take on. In any case, it's outside the scope of this PR

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
frankgh: mostly because it'd require a refactor of the JmxClient class which lives in a different subproject. I will give it a go though, I think it makes sense to do the change here

server/src/main/java/org/apache/cassandra/sidecar/config/DurationSpec.java
frankgh: done

server/src/main/java/org/apache/cassandra/sidecar/config/MillisecondBoundConfiguration.java
frankgh: changed

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/DurationSpecImpl.java
frankgh: done!

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/DurationSpecImpl.java
frankgh: error messages have changed a bit, I made them very explicit on what the allowed values are

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
frankgh: done

conf/sidecar.yaml
frankgh: I left the comments in place. I think it's fine to be extra verbose for the configuration yaml.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/utils/DurationSpec.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * @param symbol the time unit symbol
+     * @return the time unit associated to the specified symbol
+     * @throws IllegalArgumentException when the {@code symbol} is unsupported
+     */
+    static TimeUnit fromSymbol(String symbol)
+    {
+        switch (symbol.toLowerCase())
```
bbotella, line 284: I think we are missing NANOSECONDS here

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/utils/DurationSpec.java
```diff
@@ -0,0 +1,132 @@
...
+     */
+    default long to(TimeUnit targetUnit)
+    {
+        return targetUnit.convert(quantity(), unit());
+    }
+
+    /**
+     * @return the duration in seconds
+     */
+    default long toSeconds()
```
bbotella: I understand this helper, but, isn't it a bit redundant? I think `to(TimeUnit.SECONDS)` is as readable as `toSeconds`.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/utils/DurationSpecImpl.java
```diff
@@ -0,0 +1,197 @@
...
+        return Long.compare(this.to(minUnit), that.to(minUnit));
+    }
+
+    void validateMinUnit(Object value, TimeUnit unit, TimeUnit minUnit)
+    {
+        if (unit.compareTo(minUnit) < 0)
+            throw iae(value);
+    }
+
+    void validateQuantity(Object value, long quantity, TimeUnit sourceUnit, TimeUnit minUnit, long max)
```
bbotella: Do we need `max` as a parameter? I think it is always `Long.MAX_VALUE` right?

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CacheConfigurationImpl.java
```diff
@@ -41,34 +46,57 @@ public class CacheConfigurationImpl implements CacheConfiguration
...
     protected final int warmupRetries;
 
-    @JsonProperty("warmup_retry_interval_millis")
-    protected final long warmupRetryIntervalMillis;
+    protected MillisecondBoundConfiguration warmupRetryInterval;
 
     public CacheConfigurationImpl()
     {
-        this(TimeUnit.HOURS.toMillis(1), 100, true, 5, 1000);
+        this(MillisecondBoundConfiguration.parse("1h"), 100, true, 5, MillisecondBoundConfiguration.parse("60s"));
```
bbotella: Is the `1000ms -> 60s` change intended?

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CacheConfigurationImpl.java
```diff
@@ -41,34 +46,57 @@ public class CacheConfigurationImpl implements CacheConfiguration
...
     }
 
     @VisibleForTesting
-    public CacheConfigurationImpl(long expireAfterAccessMillis, long maximumSize)
+    public CacheConfigurationImpl(MillisecondBoundConfiguration expireAfterAccess, long maximumSize)
     {
-        this(expireAfterAccessMillis, maximumSize, true, 5, 1000);
+        this(expireAfterAccess, maximumSize, true, 5, MillisecondBoundConfiguration.parse("60s"));
```
bbotella: Is the `1000ms -> 60s` change intended?

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/utils/DurationSpec.java
frankgh: it's a convenience method as this is used a lot throughout the codebase.

server/src/main/java/org/apache/cassandra/sidecar/config/ServiceConfiguration.java
JeetKunDoug: It should be minutes btw... this isn't a replacement for NTP, it's a guardrail to make sure that the Spark workers aren't so far skewed from C* that deletes and TTLs do weird things (since Spark writes the row timestamps in the Analytics library, where we consume this endpoint).  It is both possible and acceptable for there to be several minutes of skew in environments that don't have tightly controlled NTP sync, and specifying this in seconds really doesn't make any sense (is 65 seconds acceptable, but 66 not?)

That said, I don't really mind that it's _set_ in seconds, but the expectation that somehow the resolution of seconds is meaningful here isn't realistic.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/utils/DurationSpec.java
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public int compareTo(@NotNull DurationSpec that)
+    {
+        if (this.unit == that.unit())
+        {
```
JeetKunDoug, line 216: While I'm a big proponent of using curly braces all the time, given our coding style this shouldn't have curly braces. 😿

server/src/main/java/org/apache/cassandra/sidecar/config/ServiceConfiguration.java
frankgh: thanks for the context. I think we can have a minute bound configuration if that's the case.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/S3ClientConfigurationImpl.java
```diff
@@ -67,27 +72,27 @@ public S3ClientConfigurationImpl()
...
+                                     SecondBoundConfiguration threadKeepAlive,
                                      int rangeGetObjectBytesSize,
-                                     long apiCallTimeoutMillis,
+                                     MillisecondBoundConfiguration apiCallTimeout,
                                      S3ProxyConfiguration proxyConfig)
     {
-        Preconditions.checkArgument(apiCallTimeoutMillis > TimeUnit.SECONDS.toMillis(10),
-                                    "apiCallTimeout cannot be smaller than 10 seconds. Configured: " + apiCallTimeoutMillis + " ms");
+        Preconditions.checkArgument(apiCallTimeout.compareTo(MINIMUM_API_CALL_TIMEOUT) > 0,
+                                    "apiCallTimeout cannot be smaller than 10 seconds. Configured: " + apiCallTimeout);
```
JeetKunDoug: Can we use `MINIMUM_API_CALL_TIMEOUT` in the error message here rather than the hard coded `10 seconds`? That way if one ever changes the other will also be updated.

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -691,22 +698,31 @@ public ElectorateMembership electorateMembership(InstancesMetadata instancesMeta
...
+    public PeriodicTaskExecutor periodicTaskExecutor(Vertx vertx,
+                                                     ExecutorPools executorPools,
+                                                     ClusterLease clusterLease,
+                                                     ClusterLeaseClaimTask clusterLeaseClaimTask)
+    {
+        PeriodicTaskExecutor periodicTaskExecutor = new PeriodicTaskExecutor(executorPools, clusterLease);
         vertx.eventBus().localConsumer(ON_SIDECAR_SCHEMA_INITIALIZED.address(),
-                                       ignored -> periodicTaskExecutor.schedule(task));
-        return task;
+                                       ignored -> periodicTaskExecutor.schedule(clusterLeaseClaimTask));
```
JeetKunDoug, line 729: Maybe not for this PR, but I'm not really thrilled that the setup of the `clusterLeaseClaimTask` is in here... I don't think it was really any better before, but somehow starting a periodic task in a provider method like this seems inappropriate - don't really have a simple/sensical answer yet though.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/utils/MillisecondBoundConfiguration.java
```diff
@@ -0,0 +1,101 @@
...
+        return MILLISECONDS;
+    }
+
+    /**
+     * @return the duration in milliseconds
+     */
+    public long toMillis()
+    {
+        return to(TimeUnit.MILLISECONDS);
+    }
```
JeetKunDoug: We frequently have to convert a `DurationSpec` to milliseconds - thoughts on pulling this up to the parent class, and then replacing the `.to(TimeUnit.MILLISECONDS)` calls?

server/src/test/java/org/apache/cassandra/sidecar/cdc/CDCLogCacheTest.java
```diff
@@ -116,13 +118,18 @@ private File instance1CommitLogFile()
     private CdcLogCache cdcLogCache()
     {
         ExecutorPools executorPools = ExecutorPoolsHelper.createdSharedTestPool(Vertx.vertx());
-        return new CdcLogCache(executorPools, instancesMetadata, 100L);
+        SecondBoundConfiguration mockCacheExpiryConfig = mock(SecondBoundConfiguration.class);
```
JeetKunDoug, line 123: Can you please document why you're using a `mock` here rather than just instantiating a DurationSpec? Took me a moment to realize you're working around the SecondBoundConfiguration's min unit, if I understand correctly.

JeetKunDoug (commented):
I'm down to test code and only have a few minor comments so far... wanted to submit the feedback on the production code already... will finish up the test code later today.

server/src/main/java/org/apache/cassandra/sidecar/config/ServiceConfiguration.java
frankgh: addressed this

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CacheConfigurationImpl.java
frankgh: good catch

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CacheConfigurationImpl.java
frankgh: good catch

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/utils/DurationSpec.java
frankgh: is this a rule in Sidecar ? I think we enforce curly braces always with our code style rules

JeetKunDoug: This _file_ has both styles in it (see `equals` above). I'm perfectly happy if we want to always use curly braces but, if that's the case, let's be consistent and do it everywhere please.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/utils/DurationSpec.java
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public int hashCode()
+    {
+        // Milliseconds seems to be a reasonable tradeoff
+        return Objects.hash(unit.toMillis(quantity));
```
JeetKunDoug: NIT: Can now be replaced with a call to `toMillis`
```suggestion
        return Objects.hash(toMillis());
```

server/src/main/java/org/apache/cassandra/sidecar/config/ServiceConfiguration.java
```diff
@@ -94,7 +94,7 @@ default List<SocketAddress> listenSocketAddresses()
     /**
      * @return the maximum time skew allowed between the server and the client
      */
-    SecondBoundConfiguration allowableTimeSkew();
+    MinuteBoundConfiguration allowableTimeSkew();
```
JeetKunDoug, line 97: 👍

JeetKunDoug (approved):
:shipit: ! Thanks for the improvement in our configuration system - I love having human-readable durations like this.

bbotella (approved):
Awesome PR!! Thanks a lot for this!

PR #175, opened 2025-01-21
CASSSIDECAR-189: Fix SidecarSchema stuck at initialization due to ClusterLeaseTask scheduling

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-189

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/config/KeyStoreConfiguration.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/config/KeyStoreConfiguration.java
@@ -19,6 +19,7 @@
 package org.apache.cassandra.sidecar.config;
 
 import org.apache.cassandra.sidecar.common.server.utils.SecondBoundConfiguration;
+import org.jetbrains.annotations.NotNull;
 
 /**
  * Encapsulates key or trust store option configurations
@@ -44,10 +45,11 @@ public interface KeyStoreConfiguration
 
     /**
      * Returns the interval in which the key store will be checked for filesystem changes. Setting
-     * this value to 0 or negative will disable reloading the store.
+     * this value to {@link SecondBoundConfiguration#ZERO} will disable reloading the store.
      *
      * @return the interval in which the key store will be checked for changes in the filesystem
      */
+    @NotNull
     SecondBoundConfiguration checkInterval();
 
     /**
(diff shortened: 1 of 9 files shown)


server/src/main/java/org/apache/cassandra/sidecar/config/KeyStoreConfiguration.java
```diff
@@ -55,7 +55,7 @@ public interface KeyStoreConfiguration
      */
     default boolean reloadStore()
     {
-        return checkInterval().compareTo(SecondBoundConfiguration.ZERO) > 0;
+        return checkInterval() == SecondBoundConfiguration.ZERO;
```
frankgh: this doesn't look correct. We want to reload the store only if the configuration is greater than 0. Not if the configuration is zero.

yifan-c: 👍 true. it is inverted

server/src/main/java/org/apache/cassandra/sidecar/config/KeyStoreConfiguration.java
```diff
@@ -55,7 +55,7 @@ public interface KeyStoreConfiguration
      */
     default boolean reloadStore()
     {
-        return checkInterval().compareTo(SecondBoundConfiguration.ZERO) > 0;
+        return checkInterval() != SecondBoundConfiguration.ZERO;
```
frankgh: maybe compare by value, and not reference?
```suggestion
        return !checkInterval().equals(SecondBoundConfiguration.ZERO);
```

frankgh (commented):
Looks good. Waiting on green CI

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarInternalKeyspace.java
```diff
@@ -73,13 +73,14 @@ protected boolean initializeInternal(@NotNull Session session,
     {
         super.initializeInternal(session, shouldCreateSchema);
 
+        boolean initialized = true;
         for (AbstractSchema schema : tableSchemas)
         {
-            if (!schema.initialize(session, shouldCreateSchema))
-                return false;
+            // run initialize on all schema. Set the initialized to false if any of the schema initialization fails
```
frankgh:
```suggestion
            // attempt to initialize all schemas. Sets initialized to false if any of the schema initialization fails
```

frankgh (commented):
Looks good, thanks for the fix. I have one small NIT

PR #176, opened 2025-01-26
CASSSIDECAR-195: Fix missing field for INDEX_STATUS in GossipInfo

anderoo:
Jira: https://issues.apache.org/jira/browse/CASSSIDECAR-195

`INDEX_STATUS` was added as a field to Gossip information in 5.0. This patch adds the field to the GossipInfoResponse so Sidecar doesn't error out when the new field is present. INDEX_STATUS was added here: https://github.com/apache/cassandra/commit/cde91e56f09d9ebf315c79c9a81b89f70f4eb724 (https://issues.apache.org/jira/browse/CASSANDRA-18058)

### DIFF
--- a/client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipInfoResponse.java
+++ b/client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipInfoResponse.java
@@ -283,7 +283,8 @@ public class GossipInfoResponse extends HashMap<String, GossipInfoResponse.Gossi
          **/
         SSTABLE_VERSIONS,
         DISK_USAGE,
-        INDEX_STATUS;
+        INDEX_STATUS, // Introduced in Cassandra 5.0 for SAI 
+        ;
 
         static String read(GossipInfo gossipInfo, GossipField field)
         {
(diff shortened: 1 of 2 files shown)


frankgh:
@anderoo can you please file a JIRA for this work? Also would it be possible to get some tests ?

anderoo:
> @anderoo can you please file a JIRA for this work? Also would it be possible to get some tests ?

Sure, I can create a ticket and rename.

anderoo:
@frankgh Added JIRA and updated existing tests which covers both the field being present and not being present. LMK if there's any other specific tests you'd like.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/GossipInfoResponse.java
```diff
@@ -275,7 +282,8 @@ protected enum GossipField
          * as a comma-separated list.
          **/
         SSTABLE_VERSIONS,
-        DISK_USAGE;
+        DISK_USAGE,
+        INDEX_STATUS;
```
yifan-c: Add a comment to indicate since what Cassandra verison the gossip value is available.
And move `;` to a new line, to avoid showing the diff on the existing enum in the future. 

I can also update it on merging, if you agree with the suggestion, but do not feel like to push a new commit. 

```suggestion
        INDEX_STATUS, // Intorduced in Cassandra 5.0 for SAI 
        ;
```

anderoo: Fixed typo and updated 👍

frankgh (approved):
+1 thanks for the patch !

yifan-c (approved):

yifan-c:
Committed as https://github.com/apache/cassandra-sidecar/commit/104d8048aa8d5fa055aa64a9a64ba2e570b60f52

PR #178, opened 2025-01-31
Implemented Schema Reporter for Integration with DataHub

5:
In order to support the best practices of data governance, Cassandra Sidecar is now capable of periodically reporting the full cluster schema to an instance of DataHub in a reliable and efficient manner.


server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
```diff
@@ -0,0 +1,107 @@
...
+    public void execute(Promise<Void> promise)
+    {
+        try
+        {
+            reporter.process(session.get().getCluster());
+            promise.complete();
+        }
+        catch (Throwable throwable)
+        {
+            LOGGER.error("Failed to convert and report the current schema");
```
frankgh: maybe log the exception here?
```suggestion
            LOGGER.error("Failed to convert and report the current schema", throwable);
```

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
```diff
@@ -0,0 +1,107 @@
...
+            promise.fail(throwable);
+        }
+    }
+
+    @Override
+    public Void result()
+    {
+        return null;  // No result can be returned from a {@link PeriodicTask}
+    }
```
frankgh: NIT, we can omit the override here.

frankgh (commented):
Looks good overall. I left a couple minor comments.

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
5: @yifan-c had the same suggestion, but the base implementation throws an exception, which I don't want to happen ever, even if someone calls the method.

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
5: Good point, added.

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
frankgh: the PeriodicTaskExecutor class will handle this task, and it will not call the method (by design). So there's no need to do this.

5: Removed this override and amended the same commit.

yifan-c:
Please add the changes.txt

yifan-c:
Closing as it is committed. https://github.com/apache/cassandra-sidecar/commit/372d02d856dc37e4a4cb4454b7dfd51c56eea7b9

PR #179, opened 2025-01-31
CASSSIDECAR-193: Add feature level permissions to Sidecar

sarankk:
This patch introduces feature level permissions that combine basic permissions. With feature level permissions it will be easier to grant access to users. 
For e.g. for analytics bulk read feature, users needs access to multiple handlers such as, schema, create snapshot, stream SSTables etc. In order to grant bulk read access, operators need to grant all basic permissions such as SCHEMA:READ, SNAPSHOT:CREATE etc. Instead with feature level permissions, operators can grant ANALYTICS:READ_DIRECT to users.

Patch by Saranya Krishnkumar; Reviewed by 	
Francisco Guerrero, Yifan Cai for [CASSSIDECAR-193](https://issues.apache.org/jira/browse/CASSSIDECAR-193)


server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/StandardPermission.java
```diff
@@ -47,13 +48,33 @@ public String name()
...
+     * Sets resource scope for given permission.
+     *
+     * @param resourceScope scope of resource this permission checks or grants permission for.
+     * @return a reference to permission
+     */
+    public StandardPermission withScope(ResourceScope resourceScope)
+    {
+        this.resourceScope = resourceScope;
+        return this;
+    }
```
yifan-c: This method leaves the created `StandardPermission` objects in the mutable state. The resourceScope of the permissions defined in `BasicPermissions` can be altered w/o notice. 

I would suggest just change the constructor, since there are only 2 fields and remove the `withScope` method. Both fields now can be marked as `final`. 

```java
    protected final String name;
    protected final ResourceScope resourceScope;

    public StandardPermission(String name, ResourceScope resourceScope)
    {
        if (isNullOrEmpty(name))
        {
            throw new IllegalArgumentException("Permission name can not be null or empty");
        }
        this.name = name;
        this.resourceScope = resourceScope;
    }
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/StandardPermission.java
```diff
@@ -47,13 +48,33 @@ public String name()
...
+
     @Override
     public Authorization toAuthorization(String resource)
     {
         PermissionBasedAuthorization authorization = PermissionBasedAuthorization.create(name);
         if (isNotEmpty(resource))
         {
-            authorization.setResource(resource);
+            ResourceScope resourceScope = resourceScope();
+            String resolvedResource = resourceScope != null ? resourceScope.resolveWithResource(resource) : resource;
```
yifan-c: Looks like all permissions declared do not have `resourceScope == null`. Maybe we can define the contract that all permissions should have non-null `resourceScope`?

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
```diff
@@ -0,0 +1,163 @@
...
+    {
+        if (isNullOrEmpty(resource))
+        {
+            throw new IllegalArgumentException("Resource expected for resolving");
+        }
+
+        if (!resource.startsWith("data"))
+        {
+            return variableAwareResource();
+        }
```
yifan-c: Should it throw `IllegalArgumentException` too? 
For example, what if the `resource` in the database row is somehow defined as "foo". It seems to me that "foo" is an invalid definition. For data resource, can we enforce the format to be `data[/keyspace[/table]]`

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
```diff
@@ -0,0 +1,163 @@
...
+     */
+    public static DataResourceScope createWithKeyspaceScope()
+    {
+        return new DataResourceScope(true, false);
+    }
+
+    /**
+     * @return {@link DataResourceScope} scoped at table level within a keyspace
+     */
+    public static DataResourceScope createWithTableScope()
```
yifan-c: nit: maybe create the static instances directly, instead of expose static methods.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/FeaturePermission.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.apache.cassandra.sidecar.acl.authorization.CassandraPermissions.MODIFY;
+import static org.apache.cassandra.sidecar.acl.authorization.CassandraPermissions.SELECT;
+import static org.apache.cassandra.sidecar.acl.authorization.ResourceScopes.TABLE;
+
+/**
+ * Enumerates a list of feature level permissions that Sidecar recognizes and honors.
+ */
+public enum FeaturePermission
+{
+    BULK_READ_DIRECT("ANALYTICS:READ_DIRECT",
```
yifan-c: Can the enum match with the `name`? The enum is `BULK_READ_DIRECT`, but the name has `ANALYTICS`

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/FeaturePermission.java
```diff
@@ -0,0 +1,130 @@
...
+                         READ_TOPOLOGY,
+                         CREATE_RESTORE_JOB,
+                         READ_RESTORE_JOB,
+                         EDIT_RESTORE_JOB,
+                         DELETE_RESTORE_JOB),
+
+    CDC("CDC", BasicPermissions.CDC);
+
+    // TODO make forbidden feature permission configurable
+    private static final Set<String> FORBIDDEN_FEATURE_PERMISSION = Collections.singleton("*:*");
```
yifan-c: It is to be addressed in this patch or a follow-up?

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/FeaturePermission.java
```diff
@@ -0,0 +1,130 @@
...
+    {
+        if (FORBIDDEN_FEATURE_PERMISSION.contains(name))
+        {
+            return null;
+        }
+        List<Permission> combinedPermissions = new ArrayList<>();
+        WildcardPermissionBasedAuthorization requestedAuthorization = new WildcardPermissionBasedAuthorizationImpl(name);
+        for (FeaturePermission featurePermission : FEATURE_PERMISSIONS)
+        {
+            if (requestedAuthorization.verify(featurePermission.nameAuthorization))
```
yifan-c: Is it to support this scenario, for example, `name == "Analytics:*"`?

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
```diff
@@ -0,0 +1,163 @@
...
+    // TABLE_WILDCARD as a variable. This hack allows to read resource level permissions that could be set for all
+    // tables through data/<keyspace_name>/*. Bug should be fixed in 4.5.12
+    // Note: DATA_WITH_KEYSPACE_ALL_TABLES authorizes for all tables under the keyspace excluding the keyspace itself
+    private static final String DATA_WITH_KEYSPACE_ALL_TABLES = String.format("data/{%s}/{TABLE_WILDCARD}", KEYSPACE);
+
+    private static final String DATA_WITH_KEYSPACE_TABLE = String.format("data/{%s}/{%s}", KEYSPACE, TABLE);
+
+    private final boolean keyspaceScoped;
+    private final boolean tableScoped;
+    private final Set<String> expandedResources;
```
yifan-c, line 78: ok. I guess order does not matter here.

server/src/main/java/org/apache/cassandra/sidecar/routes/cdc/ListCdcDirHandler.java
```diff
@@ -94,6 +93,14 @@ protected void handleInternal(RoutingContext context,
...
     {
         String cdcDir = metadataFetcher.instance(host).cdcDir();
+        if (isNullOrEmpty(cdcDir))
+        {
+            context.response()
+                   .setStatusCode(HttpResponseStatus.SERVICE_UNAVAILABLE.code())
+                   .setStatusMessage("CDC not turned on for cluster")
+                   .end();
+            return;
+        }
```
yifan-c: I expressed the concern a few days ago at https://github.com/apache/cassandra-sidecar/pull/163#discussion_r1936764719. And it now became true :D

So, we should not rely `null` value of `cdcDir` to decide cdc is not enabled. The `null` value only means the `cdcDir` is not configured. Server should only reply with "CDC directory is not configured in Sidecar".

server/src/main/java/org/apache/cassandra/sidecar/routes/cdc/ListCdcDirHandler.java
```diff
@@ -102,7 +109,7 @@ protected void handleInternal(RoutingContext context,
             LOGGER.warn("Error listing the CDC commit log segments", cause);
             context.response()
                    .setStatusCode(HttpResponseStatus.INTERNAL_SERVER_ERROR.code())
-                   .setStatusMessage(cause.getMessage())
+                   .setStatusMessage(cause != null ? cause.getMessage() : "Error while listing CDC segments")
```
yifan-c: Do you see response message with 'null'?

yifan-c (commented):
Looks good overall. I like the introduction of `ResourceScope`

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/FeaturePermission.java
sarankk: Yes to support feature level permission naming in format `domain:action`

server/src/main/java/org/apache/cassandra/sidecar/routes/cdc/ListCdcDirHandler.java
sarankk: Yes I was seeing NPE

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/FeaturePermission.java
sarankk: Will add it in this patch

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/FeaturePermission.java
sarankk: No, feature level permission name will be `ANALYTICS:READ_DIRECT` like other sidecar basic permissions, we can do following `ANALYTICS:READ_DIRECT,WRITE_DIRECT` or `ANALYTICS:*`. `BULK_READ_DIRECT` enum name, we currently don't use it. We can use it for name match, if we choose to go with `BULK_READ_DIRECT` for feature permission name.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
sarankk: and make constructor public or keep it private? I wanted to add these since creating `new DataResourceScope(false, true)` doesn't make sense.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
sarankk: I thought about this too, it will be hard to enforce, because, for feature level permissions we have multiple resources involved , for e.g with bulk read feature, we have `data/<ks>/<tbl>` and `cluster` resource involved. While setting feature permission we only set `data/<ks>/<tbl>`, hence while resolving for cluster resource, we get `data/<ks>/<tbl>` and cluster resource resolves to the scope it understands without throwing.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
yifan-c: Nope. This is what I meant. In this class file, having this instead.

```java
public static DataResourceScope DATA = new DataResourceScope(false, false);
// so on for the other 2 scopes. 
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/FeaturePermission.java
yifan-c: I was asking if it make sense to have this

```suggestion
    ANALYTICS_READ_DIRECT("ANALYTICS:READ_DIRECT",
```

server/src/main/java/org/apache/cassandra/sidecar/routes/cdc/ListCdcDirHandler.java
yifan-c: How can the `cause` be null in the failure handler? I guess it is `cause.getMessage() == null` instead? If so, the change here does not make a difference. Probably get a better understanding of the issue first.
It is also unrelated with the patch. Given the change is tiny, I am fine.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
yifan-c: I do not follow. The string returned from line#119 is either `DATA` or `DATA_WITH_KEYSPACE` or `DATA_WITH_KEYSPACE_TABLE`. No? I do not think it will ever return `cluster` or `operation`

server/src/main/java/org/apache/cassandra/sidecar/routes/cdc/ListCdcDirHandler.java
sarankk: Sorry meant to do this `cause.getMessage() == null ? "Error while listing CDC segments" : cause.getMessage()` `cause.getMessage()` could be null and `setStatusMessage` does not allow setting null, throws NPE

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/FeaturePermission.java
sarankk: On other hand, I think we wont be marking any other feature permission as forbidden other than `*:*`, leaving it as is and removing comment.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/StandardPermission.java
sarankk: For cassandra permissions I think we don't want to set scope upfront in `CassandraPermissions`. Since we could grant `CREATE` keyspace scoped or table scoped. Instead we can use those permissions as baseline and recreate with their name `new StandardPermission(SELECT.name(), TABLE_SCOPE)` when required, wdyt

sarankk: When we read permissions from database we wont know the scope of permission, we will only know exact resource. Hence we should allow resource scope to be null, since we use same permission class on both verification and retrieval side. Also for `CompositePermission` I think scope should be `null`

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
sarankk: @yifan-c I did this, but this doesn't allow having all resource scopes in 1 place `ResourceScopes` shall I keep it as is.

yifan-c: In `ResourceScopes`, assign `DATA_SCOPE = DataResourceScope.DATA`

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/StandardPermission.java
yifan-c: According to the analysis, the scope is specific to `DomainAwarePermission`. In this case, can you move `resourceScope` to `DomainAwarePermission` and use the `ResourceScope` to produce authorization from `DomainAwarePermission` only. Does it make sense?

server/src/main/java/org/apache/cassandra/sidecar/routes/cdc/ListCdcDirHandler.java
yifan-c: Sounds good. Can you have this instead? It is a nit.

```suggestion
                   .setStatusMessage(Objects.requireNonNullElse(cause.getMessage(), "Error while listing CDC segments"))
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/StandardPermission.java
yifan-c: Scope is not specific to `DomainAwarePermission`

sarankk: Yes, we set scope for `SELECT` and `MODIFY` standard permissions too

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
sarankk: Discussed offline and addressed to validate data resource format

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
```diff
@@ -0,0 +1,153 @@
...
+import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.TABLE;
+import static org.apache.cassandra.sidecar.common.utils.StringUtils.isNullOrEmpty;
+
+/**
+ * Signifies scope of Cassandra data resource. {@code keyspaceScoped} can be set to true to create data scope
+ * restricted to a keyspace. {@code tableScoped} can be set to true to create data scope restricted to a table.
+ */
+public class DataResourceScope implements ResourceScope
+{
+    public static final Pattern pattern = Pattern.compile("data(/[^/]+(/[^/]+)?)?");
```
yifan-c: It is a good idea of using regular expression here. The pattern has some minor flaw though. 

Before suggesting a new pattern, can you rename the variable to `DATA_RESOURCE_PATTERN`? 

You can improve it by having the following pattern. It will match string starts with `data` and at most 3 segments. As a bonus, the captured values (i.e. keyspace and table) are ready to use.

```
^data(?:/([^/]+)?)(?:/([^/]+)?)$
```

IIRC Cassandra allows only alphabetic, digits and underscore for keyspace and table names. Please double check. In that case, the pattern is 

```
^data(?:/([a-z|A-Z|0-9|_]+)?)(?:/([a-z|A-Z|0-9|_]+)?)$
```

Try it out at https://regex101.com/r/B9wZvH/1

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
```diff
@@ -0,0 +1,153 @@
...
+    {
+        validate(resource);
+        if (tableScoped)
+        {
+            return resource;
+        }
+        else if (keyspaceScoped)
+        {
+            String[] parts = resource.split("/");
+            return parts.length == 3 ? "data/" + parts[1] : resource;
```
yifan-c: With the pattern, you can extract the variables out directly.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DomainAwarePermission.java
```diff
@@ -45,7 +46,18 @@ public class DomainAwarePermission extends StandardPermission
 
     public DomainAwarePermission(String name)
     {
-        super(name);
+        this(name, SCOPE_LESS);
```
yifan-c: I think scopeless is a single word. `-less` is the suffix.

yifan-c (commented):
Overall looks good to me. Some feedback on the pattern usage

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
sarankk: thanks @yifan-c, yes grouping and extracting makes sense, will update it. I would prefer not to have pattern check for characters `[a-z|A-Z|0-9|_]+`, because the rules for allowed keyspace and table name sometimes changes, which causes some errors and fixes. Since we do exact resource matching at handler side before authorizing a user. I think pattern match for keyspace and table can be avoided here. Wdyt?

sarankk: did you mean this `^data(?:/([^/]+))?(?:/([^/]+))?$` the above pattern allows `data//`

yifan-c: keyspace and table extraction is used at `org.apache.cassandra.sidecar.acl.authorization.DataResourceScope#resolveWithResource`. 

Currently, it again split the string and combine for keyspace scoped. Instead, you can have `"data/" + keyspace`.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DomainAwarePermission.java
sarankk: changing to `NO_SCOPE` instead

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
sarankk: because we want to return `data` as is, even if it is keyspace scoped

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/FeaturePermission.java
```diff
@@ -0,0 +1,91 @@
...
+    ANALYTICS_WRITE_DIRECT("ANALYTICS:WRITE_DIRECT",
+                           READ_SCHEMA_KEYSPACE_SCOPED,
+                           READ_TOPOLOGY,
+                           UPLOAD_STAGED_SSTABLE,
+                           IMPORT_STAGED_SSTABLE,
+                           DELETE_STAGED_SSTABLE,
+                           new StandardPermission(MODIFY.name(), TABLE_SCOPE)),
+
+    ANALYTICS_WRITE_S3_COMPAT("ANALYTICS:WRITE_S3_COMPAT",
+                              READ_SCHEMA,
```
frankgh: Does this need to be `READ_SCHEMA_KEYSPACE_SCOPED` ?

frankgh (commented):
Looks great! I have one minor comment

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/FeaturePermission.java
sarankk: Good catch, updated it

frankgh (approved):
+1 looks good to me

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/PermissionFactoryImpl.java
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public CompositePermission createFeaturePermission(String name)
+    {
+        if (name.startsWith(FORBIDDEN_FEATURE_PERMISSION_PREFIX))
+        {
+            return null;
```
yifan-c: Throw `IllegalArgumentException` instead of returning null. `*:` is not a permission that can be permitted by `DomainAwarePermission` neither. Let's fail fast.

yifan-c (commented):
The patch itself look good. I left a final nit. 

However, the direction is concerning after putting more thoughts on it. The feature permissions are declared statically to include all permissions that are needed by the _current version_ of Cassandra Analytics. It unfortunately establishes a strong coupling with Analytics and Sidecar. Here is a scenario, in the newer version of Analytics, it needs to access to an additional API for bulk reading, it would require a new release of Sidecar to update the statically declared feature permissions in code. 
That said, it might not be an issue for the foreseeable future, as the features are relatively stable. 
We need to think about it, and address it when the coupling is becoming problematic. IMO, there should be an API to allow _admins_ to declare feature permission, instead of being hard-coded.

sarankk:
The Final goal is to add these mappings into Cassandra database itself. So that we can grant these permissions with 

`GRANT BULK READ ON KEYSPACE university TO cassandra_role;`

When users start using a newer version of analytics library with updated list of APIs, they can update list of basic permissions feature permission maps to for their respective cluster. Having an admin level API in Sidecar to update these mappings make sense to me. Currently we store the mapping in code, so that it can't be modified without notice. 

> However, the direction is concerning after putting more thoughts on it. The feature permissions are declared statically to include all permissions that are needed by the _current version_ of Cassandra Analytics. It unfortunately establishes a strong coupling with Analytics and Sidecar. Here is a scenario, in the newer version of Analytics, it needs to access to an additional API for bulk reading, it would require a new release of Sidecar to update the statically declared feature permissions in code. That said, it might not be an issue for the foreseeable future, as the features are relatively stable. We need to think about it, and address it when the coupling is becoming problematic. IMO, there should be an API to allow _admins_ to declare feature permission, instead of being hard-coded.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/PermissionFactoryImpl.java
sarankk: @yifan-c I returned null, because we try to create a feature permission if not fallback to creating a basic permission. Not throwing an exception, allows someone to override `DomainAwarePermission` to allow for `*:` prefix in their permission, without overriding `createFeaturePermission` implementation. When we have fallback in place, didn't want to throw prematurely.

yifan-c: Can you show me, in this patch, how to return the domain aware permission subclass w/o having a new factory implementation? 
If someone wants to modifies the behavior in domain aware permission validation, they also need to have a new factory implementation. TBH, the imaginary scenario is out of the scope.

sarankk: Moved the exception throwing to `createPermission` itself, that way it is not a feature permission specific exception.

sarankk: they would need to override just `createPermission` they can still keep `createFeaturePermission` as is.

PR #180, opened 2025-01-31
CASSSIDECAR-192: Fix CassandraInstance not found error

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-192

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -63,7 +63,21 @@ def codeCheckTasks = task("codeCheckTasks")
 allprojects {
     apply plugin: 'jacoco'
     apply plugin: 'checkstyle'
-    apply plugin: "com.github.spotbugs"
+
+    // do not run spot-bug on test-specific sub-projects
+    if (!it.name.startsWith('integration-') && !it.name.startsWith("test-"))
+    {
+        apply plugin: "com.github.spotbugs"
+        spotbugs {
+            toolVersion = '4.2.3'
+            excludeFilter = file("${project.rootDir}/spotbugs-exclude.xml")
+        }
+        codeCheckTasks.dependsOn(tasks.withType(SpotBugsTask))
+        tasks.withType(Test) {
+            // define the order of the tasks, if SpotBugsTask runs
+            shouldRunAfter(tasks.withType(SpotBugsTask))
+        }
+    }
 
     repositories {
         mavenCentral()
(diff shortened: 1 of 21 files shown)


build.gradle
```diff
@@ -314,7 +325,7 @@ rat {
 
 installDist.dependsOn copyJolokia
 check.dependsOn codeCheckTasks
-build.dependsOn copyJolokia, copyDocs
+build.dependsOn codeCheckTasks, copyJolokia, copyDocs
```
yifan-c: run codeCheckTask in CI.

server/src/main/java/org/apache/cassandra/sidecar/cluster/InstancesMetadataImpl.java
```diff
@@ -64,37 +74,70 @@ public InstancesMetadataImpl(List<InstanceMetadata> instanceMetadataList, DnsRes
...
+        {
+            updateIpToInstanceMetadata();
+            instanceMetadata = ipToInstanceMetadata.get(hostOrIpAddress);
+        }
+
         if (instanceMetadata == null)
         {
+            throw new NoSuchCassandraInstanceException("Instance with host address '" + hostOrIpAddress + "' not found");
+        }
+        return instanceMetadata;
```
frankgh: Can we return early here?
```suggestion
        // if the host string is hostname string, resolve the ip address and loop up again
        InstanceMetadata instanceMetadata = hostNameToInstanceMetadata.get(hostOrIpAddress);
        if (instanceMetadata != null)
        {
            return instanceMetadata;
        }

        // if the 'host' string is IP address string
        instanceMetadata = ipToInstanceMetadata.get(hostOrIpAddress);
        if (instanceMetadata != null)
        {
            return instanceMetadata;
        }

        updateIpToInstanceMetadata();
        instanceMetadata = ipToInstanceMetadata.get(hostOrIpAddress);
        if (instanceMetadata == null)
        {
            throw new NoSuchCassandraInstanceException("Instance with host address '" + hostOrIpAddress + "' not found");
        }
        return instanceMetadata;
```

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
```diff
@@ -240,14 +262,40 @@ public Builder id(int id)
...
+        public Builder host(String host, DnsResolver dnsResolver)
+        {
+            String ipAddress;
+            try
+            {
+                ipAddress = dnsResolver.resolve(host);
+            }
+            catch (UnknownHostException e)
+            {
+                throw new ConfigurationException("Failed to resolve IP address from host: " + host, e);
```
frankgh: this is a big behavior change that I'm not comfortable with. You can prevent startup of the service when there's some transitive DNS issue. I don't think we should fail here

server/src/main/java/org/apache/cassandra/sidecar/cluster/InstancesMetadataImpl.java
```diff
@@ -64,37 +74,70 @@ public InstancesMetadataImpl(List<InstanceMetadata> instanceMetadataList, DnsRes
...
+    }
+
+    private void updateIpToInstanceMetadata()
+    {
+        long now = System.currentTimeMillis();
```
frankgh, line 142: this is not really thread safe, so I'm not sure if the throttling is handled correctly

server/src/main/java/org/apache/cassandra/sidecar/cluster/InstancesMetadataImpl.java
```diff
@@ -64,37 +74,70 @@ public InstancesMetadataImpl(List<InstanceMetadata> instanceMetadataList, DnsRes
...
+        }
+
         if (instanceMetadata == null)
         {
+            throw new NoSuchCassandraInstanceException("Instance with host address '" + hostOrIpAddress + "' not found");
+        }
+        return instanceMetadata;
+    }
+
+    private void updateIpToInstanceMetadata()
```
frankgh: maybe add a comment to mention what this is supposed to be doing?

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
yifan-c: Fair. I will change to lazy evaluation

frankgh (approved):
+1 looks good to me

PR #183, opened 2025-02-03
CASSSIDECAR-121: Sidecar should default to JMX on port 7199

frankgh:
Patch by Francisco Guerrero; reviewed by TBD for CASSSIDECAR-121


conf/sidecar-ccm.yaml
```diff
@@ -0,0 +1,373 @@
+#
```
yifan-c: How about relocating this file out of the `conf` folder?

yifan-c:
Can you rename directory name from `sample` to `examples`? 
`Sample` is to represent a group. `Example` is to illustrate ideas, to explain things. 

Let's also have the plural form, since we are likely to add more examples.

yifan-c (approved):

PR #185, opened 2025-02-04
CASSSIDECAR-196 Fixes StreamStatsIntegrationTest flaky test

arjunashok:
Test failure logs indicate failure in validating the received data. 

Since the validations for matching files received and bytes received (with those transmitted) are stringent and out of scope for this feature, the changes here are to - 

1) remove checks for "all files received" and "all bytes received". 
2) bump-up the time the assertion loop takes until a success from 10s to 15s, to allow for more polls before giving up.

```
java.lang.AssertionError: 
Expecting AtomicBoolean(false) to have value:
  true
but did not.
	at org.apache.cassandra.sidecar.routes.StreamStatsIntegrationTest.lambda$streamStatsTest$2(StreamStatsIntegrationTest.java:88)
```

### DIFF
--- a/server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
+++ b/server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
@@ -82,7 +82,7 @@ public class StreamStatsIntegrationTest extends IntegrationTestBase
         awaitLatchOrThrow(BBHelperDecommissioningNode.transientStateStart, 2, TimeUnit.MINUTES, "transientStateStart");
 
         // optimal no. of attempts to poll for stats to capture streaming stats during node decommissioning
-        loopAssert(10, 200, () -> {
+        loopAssert(15, 200, () -> {
             streamStats(hasStats, dataReceived);
             assertThat(hasStats).isTrue();
             assertThat(dataReceived).isTrue();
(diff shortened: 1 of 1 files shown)


server/src/test/integration/org/apache/cassandra/sidecar/routes/StreamStatsIntegrationTest.java
```diff
@@ -123,11 +123,9 @@ void assertStreamStatsResponseOK(HttpResponse<Buffer> response, AtomicBoolean ha
         if (streamProgress.totalFilesToReceive() > 0)
         {
             hasStats.set(true);
-            if (streamProgress.totalFilesToReceive() == streamProgress.totalFilesReceived() &&
-                streamProgress.totalFilesReceived() > 0)
```
yifan-c, line 127: Is there a bug in cassandra that files-to-receive does not match files-received?

yifan-c: Talked offline. It is a timing issue. When client making the request, the stream has just finished, cassandra responds with `totalFilesReceived == 0`, which violates the second condition.

arjunashok: As discussed offline, this is not a bug in Cassandra, but the bytes/files metrics reset to 0 once the session is completed. The failure is because the files received is 0 at this point, so it's purely a timing issue.

yifan-c (approved):

frankgh (approved):
+1 thanks for addressing the flakiness

PR #186, opened 2025-02-04
CASSSIDECAR-201: Make sidecar operations resilient to down Cassandra nodes

yifan-c:
This patch introduces InstanceMetadataFetcher#callOnFirstAvailableInstance to retry operations the next local Cassandra node until it is successful or exhausted. It also makes sure that host extracted from Host header is non-null to solidify the implementations.

Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-201

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/AuthorizationParameterValidateHandler.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/AuthorizationParameterValidateHandler.java
@@ -31,6 +31,7 @@ import org.apache.cassandra.sidecar.routes.RoutingContextUtils;
 import org.apache.cassandra.sidecar.snapshots.SnapshotPathBuilder;
 import org.apache.cassandra.sidecar.utils.CassandraInputValidator;
 import org.apache.cassandra.sidecar.utils.InstanceMetadataFetcher;
+import org.jetbrains.annotations.NotNull;
 
 /**
  * Simple handler that extracts authorization parameters keyspace/table parameters from the path, validates them,
@@ -61,7 +62,7 @@ public class AuthorizationParameterValidateHandler extends AbstractHandler<Quali
     @Override
     protected void handleInternal(RoutingContext context,
                                   HttpServerRequest httpRequest,
-                                  String host,
+                                  @NotNull String host,
                                   SocketAddress remoteAddress,
                                   QualifiedTableName qualifiedTableName)
     {
--- a/server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
@@ -32,6 +32,7 @@ import org.apache.cassandra.sidecar.adapters.base.exception.OperationUnavailable
 import org.apache.cassandra.sidecar.common.server.data.Name;
 import org.apache.cassandra.sidecar.common.server.data.QualifiedTableName;
 import org.apache.cassandra.sidecar.common.server.exceptions.JmxAuthenticationException;
+import org.apache.cassandra.sidecar.common.utils.Preconditions;
 import org.apache.cassandra.sidecar.concurrent.ExecutorPools;
 import org.apache.cassandra.sidecar.exceptions.NoSuchCassandraInstanceException;
 import org.apache.cassandra.sidecar.utils.CassandraInputValidator;
@@ -141,7 +142,7 @@ public abstract class AbstractHandler<T> implements Handler<RoutingContext>
             try
             {
                 int instanceId = Integer.parseInt(instanceIdParam);
-                return  metadataFetcher.instance(instanceId).host();
+                return metadataFetcher.instance(instanceId).host();
             }
             catch (NumberFormatException ex)
             {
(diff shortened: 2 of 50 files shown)


server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/AuthorizationParameterValidateHandler.java
```diff
@@ -61,7 +62,7 @@ protected AuthorizationParameterValidateHandler(InstanceMetadataFetcher metadata
     @Override
     protected void handleInternal(RoutingContext context,
                                   HttpServerRequest httpRequest,
-                                  String host,
+                                  @NotNull String host,
```
yifan-c, line 65: `host` is always non-null. Adding the annotation to clarify

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
```diff
@@ -290,22 +292,31 @@ private Name tableName(RoutingContext context, boolean required)
      * Given a combined host address like 127.0.0.1:9042 or [2001:db8:0:0:0:ff00:42:8329]:9042, this method
      * removes port information and returns 127.0.0.1 or 2001:db8:0:0:0:ff00:42:8329.
      *
-     * @param address ip address
+     * @param request http server request
      * @return host address without port information
+     * @throws NoSuchCassandraInstanceException thrown when input address is null
      */
-    public static String extractHostAddressWithoutPort(String address)
```
yifan-c, line 296: The original input `address` is nullable. But it is not handled. So the existing behavior is throwing NPE, leading to 503 internal server error. It is unexpected and no client side logic relies on this behavior. 

In this patch, `null` value is handled, and server returns 404 not found.

Regardless of the handling change, the `handlers` are only going to see non-null `host` value, in both cases. The fact is clarified with adding the `@NotNull` annotation.

server/src/main/java/org/apache/cassandra/sidecar/utils/InstanceMetadataFetcher.java
```diff
@@ -77,60 +80,40 @@ public InstanceMetadata instance(int instanceId) throws NoSuchCassandraInstanceE
...
-     * is {@code null}
-     * @throws NoSuchCassandraInstanceException when the Cassandra instance with {@code host} does not exist
-     * @throws CassandraUnavailableException  when Cassandra is not yet connected
+     * @param function function applies to {@link CassandraAdapterDelegate}
+     * @return function eval result. Null can be returned when all local instances are exhausted
+     * @param <T> type of the result
+     * @throws CassandraUnavailableException when all local instances are exhausted.
      */
     @NotNull
-    public CassandraAdapterDelegate delegate(@Nullable String host) throws NoSuchCassandraInstanceException, CassandraUnavailableException
```
frankgh, line 90: :(

server/src/test/java/org/apache/cassandra/sidecar/coordination/ClusterLeaseClaimTaskTest.java
```diff
@@ -93,6 +93,22 @@ void testShouldNotSkipForMemberOfTheElectorate()
         verify(mockElectorateMembership, times(1)).isMember();
     }
 
+    @Test
+    void testShouldRescheduleWhenDatabaseAccessorIsUnavailable()
+    {
+        ServiceConfiguration serviceConfiguration = mockConfiguration(true, true);
```
frankgh, line 101: nice test

frankgh (commented):
Looks good, but I think having the `delegate` method in the `InstanceMetadataFetcher` class was convenient. We are adding a lot more code by having to get the `instance(host)` first and then `delegate`.

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
```diff
@@ -290,22 +292,31 @@ private Name tableName(RoutingContext context, boolean required)
...
+    @NotNull
+    public static String extractHostAddressWithoutPort(HttpServerRequest request) throws NoSuchCassandraInstanceException
     {
-        if (address.contains(":"))
+        String host = request.host();
+        if (host == null)
+        {
+            throw new NoSuchCassandraInstanceException("No such Cassandra instance when Host header is absent");
```
sarankk: Nit: If host is not set in request, should we send back `Bad Request: Missing host header` instead of `404`

server/src/main/java/org/apache/cassandra/sidecar/utils/InstanceMetadataFetcher.java
sarankk: +1 since we seem to access `CassandraAdaptorDelegate` in many handlers shall we keep `delegate()`?

server/src/main/java/org/apache/cassandra/sidecar/routes/AbstractHandler.java
yifan-c: +10000
Very good point

server/src/main/java/org/apache/cassandra/sidecar/utils/InstanceMetadataFetcher.java
yifan-c: ok. I will add this specific method back, since both of you want it back. 

In my opinion, retrieving delegate is not the responsibility of `InstanceMetadataFetcher`. It does not add a lot of character to retrieve from instanceMetadata. However, the methods have been there before this patch, probably could be exempted.

sarankk (approved):
+1 Looks good thanks (nb)

frankgh (approved):
+1

PR #187, opened 2025-02-05
CASSSIDECAR-200: Sidecar schema initialization can be executed on multiple thread

yifan-c:
The patch defines a SidecarSchemaInitializer periodic task, which is guaranteed to run single-threaded. By doing so, the synchronizations and state checks can be all removed.

Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-200

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
@@ -18,8 +18,6 @@
 
 package org.apache.cassandra.sidecar.db.schema;
 
-import java.util.concurrent.atomic.AtomicBoolean;
-
 import com.google.common.annotations.VisibleForTesting;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
--- a/server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
@@ -40,6 +40,7 @@ import org.apache.cassandra.sidecar.exceptions.SidecarSchemaModificationExceptio
 import org.apache.cassandra.sidecar.metrics.SchemaMetrics;
 import org.apache.cassandra.sidecar.tasks.PeriodicTask;
 import org.apache.cassandra.sidecar.tasks.PeriodicTaskExecutor;
+import org.apache.cassandra.sidecar.tasks.ScheduleDecision;
 
 import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_CASSANDRA_CQL_READY;
 import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_SIDECAR_SCHEMA_INITIALIZED;
@@ -151,6 +152,17 @@ public class SidecarSchema
     private class SidecarSchemaInitializer implements PeriodicTask
     {
         @Override
+        public ScheduleDecision scheduleDecision()
+        {
+            if (cqlSessionProvider.getIfConnected() == null)
+            {
+                LOGGER.debug("CQL connection is not yet established. Skip this run of initialization.");
+                return ScheduleDecision.SKIP;
+            }
+            return ScheduleDecision.EXECUTE;
+        }
+
+        @Override
         public DurationSpec delay()
         {
             return INITIALIZATION_LOOP_DELAY;
@@ -176,7 +188,6 @@ public class SidecarSchema
                 LOGGER.warn("Failed to initialize schema. Retry in {}", delay(), ex);
                 if (ex instanceof CassandraUnavailableException)
                 {
-                    LOGGER.debug("Cql session is not yet available. Skip initializing...");
                     return; // do not count Cassandra unavailable as failure
                 }
                 else if (ex instanceof SidecarSchemaModificationException)
--- a/server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
@@ -97,14 +97,14 @@ public class SidecarSchema
     private void configureSidecarServerEventListeners()
     {
         EventBus eventBus = vertx.eventBus();
-        eventBus.localConsumer(ON_CASSANDRA_CQL_READY.address(), message -> startSidecarSchemaInitializerMaybe());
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -47,70 +52,68 @@
...
-
-        eventBus.localConsumer(ON_CASSANDRA_CQL_READY.address(), message -> startSidecarSchemaInitializer());
-        eventBus.localConsumer(ON_SERVER_STOP.address(), message -> cancelTimer(initializationTimerId.get()));
+        eventBus.localConsumer(ON_CASSANDRA_CQL_READY.address(), message -> startSidecarSchemaInitializerMaybe());
+        eventBus.localConsumer(ON_SERVER_STOP.address(), message -> periodicTaskExecutor.unschedule(sidecarSchemaInitializer));
     }
 
     @VisibleForTesting
-    public void startSidecarSchemaInitializer()
+    public void startSidecarSchemaInitializerMaybe()
```
bbotella: nit: `maybeStartSidecarSchemaInitializer`?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -203,4 +151,48 @@ protected boolean shouldCreateSchema(AbstractSchema schema)
...
+                if (isInitialized())
+                {
+                    LOGGER.info("Sidecar schema is initialized");
+                    periodicTaskExecutor.unschedule(this);
+                    reportSidecarSchemaInitialized();
+                }
+            }
+            catch (CassandraUnavailableException ignored)
+            {
+                LOGGER.debug("Cql session is not yet available. Skip initializing...");
```
bbotella: What about ending with a `"Retrying in " + INITIALIZATION_LOOP_DELAY + "..."`?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -203,4 +151,48 @@ protected boolean shouldCreateSchema(AbstractSchema schema)
...
+        public void execute(Promise<Void> promise)
+        {
+            try
+            {
+                Session session = cqlSessionProvider.get();
+                isInitialized = sidecarInternalKeyspace.initialize(session, SidecarSchema.this::shouldCreateSchema);
+
+                if (isInitialized())
+                {
+                    LOGGER.info("Sidecar schema is initialized");
```
bbotella: `"Sidecar schema is initialized. Stopping SchemaSidecarInitializer"`?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -203,4 +151,48 @@ protected boolean shouldCreateSchema(AbstractSchema schema)
         }
         return true;
     }
+
+    /**
+     * Initializer that retries until sidecar schema is initialized. Until then, it un-schedules itself.
```
bbotella:
```suggestion
     * Initializer that retries until sidecar schema is initialized. Once initialized, it un-schedules itself.
```

bbotella (approved):
Mostly nits.

+1(nb) on my end

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -47,70 +52,68 @@
...
         this.metrics = metrics;
         this.clusterLease = clusterLease;
         if (this.schemaKeyspaceConfiguration.isEnabled())
         {
+            this.sidecarSchemaInitializer = new SidecarSchemaInitializer();
             configureSidecarServerEventListeners();
         }
         else
         {
+            this.sidecarSchemaInitializer = null;
```
sarankk: Nit: should we add `schemaKeyspaceConfiguration.isEnabled()` check to `execute` method of `SidecarSchemaInitializer`, that way we can avoid null checks in other places.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -203,4 +151,52 @@ protected boolean shouldCreateSchema(AbstractSchema schema)
...
+                LOGGER.warn("Failed to initialize schema", ex);
+                if (ex instanceof SidecarSchemaModificationException)
+                {
+                    LOGGER.warn("Failed to modify schema", ex);
+                    metrics.failedModifications.metric.update(1);
+                }
+                metrics.failedInitializations.metric.update(1);
+            }
+
+            if (isInitialized())
```
sarankk: Should we add this block under finally? to ensure `unschedule` is always called 

```
finally {
        if (isInitialized())
        {
            LOGGER.info("Sidecar schema is initialized. Stopping SchemaSidecarInitializer");
            periodicTaskExecutor.unschedule(this); 
            reportSidecarSchemaInitialized();
        } 
       else
       {
            LOGGER.debug("Sidecar schema is not initialized. Retry in {}", delay());
       }
       promise.tryComplete();
   }
```

sarankk (approved):
+1 looks good (nb), some comments

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
yifan-c: I removed `sidecarSchemaInitializer`. It is not necessary.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -203,4 +144,49 @@ protected boolean shouldCreateSchema(AbstractSchema schema)
...
+                    return; // do not count Cassandra unavailable as failure
+                }
+                else if (ex instanceof SidecarSchemaModificationException)
+                {
+                    LOGGER.warn("Failed to modify schema", ex);
+                    metrics.failedModifications.metric.update(1);
+                }
+                metrics.failedInitializations.metric.update(1);
+            }
+            promise.tryComplete();
```
frankgh, line 194: finally block for tryComplete here?

frankgh (approved):
+1 added one small comment, otherwise looks good. Thanks for the fix

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
yifan-c: I will let the executor to catch and mark the promise as failed.

PR #189, opened 2025-02-07
CASSSIDECAR-206: Add Peer Health Monitor

bbotella:
CASSSIDECAR-206: Add Peer Health Monitor


conf/sidecar.yaml
```diff
@@ -236,6 +236,36 @@ healthcheck:
   initial_delay: 0ms
   execute_interval: 30s
 
+# Sidecar Peer Health Monitor settings
+# Enables a periodic task checking for the health of adjacent Sidecar peers in the token ring
+sidecar_peer_health:
+  enabled: true
```
yifan-c: can it be off by default? It is only needed for CDC publishing, which is an optional feature

conf/sidecar.yaml
```diff
@@ -236,6 +236,36 @@ healthcheck:
...
+  use_ssl: true
+  request_timeout: 1s
+  request_idle_timeout: 1s
+  connection_pool_max_size: 10
+  connection_pool_clearing_period: 10s
+  connection_pool_event_loop_size: 10
+  connection_pool_max_wait_queue_size: 10
+  max_retries: 3
+  retry_delay: 1s
+  max_retry_delay: 2s
```
yifan-c: I would be easier for operators to parse this content if you add descriptions to each configuration.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/health/SidecarPeerDownDetectorIntegrationTest.java
```diff
@@ -0,0 +1,201 @@
...
+    protected void startSidecar(ICluster<? extends IInstance> cluster) throws InterruptedException
+    {
+        Supplier<List<TestSidecarHostInfo>> supplier = () -> sidecarServerList;
+        PeersModule peersModule = new PeersModule(supplier);
+        for (IInstance cassandraInstance : cluster)
+        {
+            // Provider de una lista de Sidecar servers
+            LOGGER.info("Starting Sidecar instance for Cassandra instance {}",
+                        cassandraInstance.config().num());
+            Server server = startSidecarWithInstances(List.of(cassandraInstance), peersModule);
```
yifan-c, line 80: Although we run each integration test in its own JVM, can we still ensure the server is properly closed?

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/health/SidecarPeerDownDetectorIntegrationTest.java
```diff
@@ -0,0 +1,201 @@
...
+                                                          1,
+                                                          new MillisecondBoundConfiguration(500, TimeUnit.MILLISECONDS));
+        }
+    }
+
+    void stopSidecarInstanceForTest(int instanceId) throws Exception
+    {
+        assertThat(sidecarServerList).isNotEmpty();
+        Server server = sidecarServerList.get(instanceId).getServer();
+        server.stop(serverDeploymentIds.get(server)).toCompletionStage().toCompletableFuture().get(30, TimeUnit.SECONDS);
```
yifan-c: nit: you can use the `getBlocking` utility, which block the future and provides richer context on failure.
```suggestion
        String deploymentId = serverDeploymentIds.get(server);
        getBlocking(server.stop(serverDeploymentIds.get(server)), 30, TimeUnit.SECONDS, "Stopping server " + deploymentId);
```

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/health/SidecarPeerDownDetectorIntegrationTest.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    @Override
+    protected void startSidecar(ICluster<? extends IInstance> cluster) throws InterruptedException
+    {
+        Supplier<List<TestSidecarHostInfo>> supplier = () -> sidecarServerList;
+        PeersModule peersModule = new PeersModule(supplier);
+        for (IInstance cassandraInstance : cluster)
+        {
+            // Provider de una lista de Sidecar servers
```
yifan-c: Please translate it to English. I kind of guessed what it means, but there is not precedent of using Spanish in this repo :p
Can you also double check elsewhere in the patch, in case I missed.

server/src/main/java/org/apache/cassandra/sidecar/config/SidecarPeerHealthConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+{
+    /**
+     * @return the number of maximum retries to be performed during a Sidecar peer health check
+     */
+    int healthCheckRetries();
+
+    /**
+     * @return the delay between Sidecar peer health checks retries
+     */
+    MillisecondBoundConfiguration healthCheckRetryDelay();
```
yifan-c: Those properties are consumed by the sidecar client. Can you rename the configurations? The current names confused me and made me thought that they were used for periodic task scheduling. 

My suggestion is `sidecarClientHealthCheckRetries` and `sidecarClientHealthCheckRetryDelay`. 

(It might be worthy to introduce a configuration for the client retry policy, but it is out of the scope of this patch. Let's do _not_ address this). We can always refactor if it becomes useful in the future.)

server/src/main/java/org/apache/cassandra/sidecar/config/SidecarPeerHealthConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+package org.apache.cassandra.sidecar.config;
+
+import org.apache.cassandra.sidecar.common.server.utils.MillisecondBoundConfiguration;
+
+/**
+ * Configuration for Sidecar peers health checks
+ */
+public interface SidecarPeerHealthConfiguration extends PeriodicTaskConfiguration
```
yifan-c, line 26: Is `Sidecar` needed in the name? It seems to be redundant. `PeerHealthConfiguration` is good enough

server/src/main/java/org/apache/cassandra/sidecar/codecs/CommonCodecs.java
```diff
@@ -0,0 +1,37 @@
...
+import io.vertx.core.eventbus.impl.codecs.BooleanMessageCodec;
+import io.vertx.core.eventbus.impl.codecs.ByteArrayMessageCodec;
+import io.vertx.core.eventbus.impl.codecs.IntMessageCodec;
+import io.vertx.core.eventbus.impl.codecs.ShortMessageCodec;
+import io.vertx.core.eventbus.impl.codecs.StringMessageCodec;
+
+/**
+ * Codecs common to Sidecar
+ */
+public class CommonCodecs
```
yifan-c: The comment and the class name read vague to me. I cannot understand where the codec is going to be used. It cannot be "common". 

There does not seem to be a good value to have such a "common" codecs. Can you delete it and declare constant in `SidecarInstanceCodecs`?

server/src/main/java/org/apache/cassandra/sidecar/codecs/SidecarInstanceCodecs.java
```diff
@@ -0,0 +1,65 @@
...
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.eventbus.MessageCodec;
+import org.apache.cassandra.sidecar.common.client.SidecarInstance;
+import org.apache.cassandra.sidecar.common.client.SidecarInstanceImpl;
+
+/**
+ * Codecs for Sidecar instances
+ */
+@Singleton
+public class SidecarInstanceCodecs implements MessageCodec<SidecarInstance, SidecarInstance>
```
yifan-c: It is one codec, so not in the plural form. Rename it to `SidecarInstanceCodec`

server/src/main/java/org/apache/cassandra/sidecar/codecs/SidecarInstanceCodecs.java
```diff
@@ -0,0 +1,65 @@
...
+        CommonCodecs.STRING.encodeToWire(buf, instance.hostname());
+    }
+
+    @Override
+    public SidecarInstance decodeFromWire(int pos, Buffer buf)
+    {
+        final int port = buf.getInt(pos);
+        pos += 4; // advance 4 bytes after reading int
+        return new SidecarInstanceImpl(CommonCodecs.STRING.decodeFromWire(pos, buf), port);
+    }
```
yifan-c: Just declare the string codec in the scope of this class. As commented above, there is no need to have a common codecs group. 

```suggestion
    private static final StringMessageCodec STRING_MESSAGE_CODEC = new StringMessageCodec();

    @Override
    public void encodeToWire(Buffer buf, SidecarInstance instance)
    {
        buf.appendInt(instance.port());
        STRING_MESSAGE_CODEC.encodeToWire(buf, instance.hostname());
    }

    @Override
    public SidecarInstance decodeFromWire(int pos, Buffer buf)
    {
        int port = buf.getInt(pos);
        pos += 4; // advance 4 bytes after reading int
        return new SidecarInstanceImpl(STRING_MESSAGE_CODEC.decodeFromWire(pos, buf), port);
    }
```

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/SidecarPeerHealthConfigurationImpl.java
```diff
@@ -0,0 +1,83 @@
...
+public class SidecarPeerHealthConfigurationImpl extends PeriodicTaskConfigurationImpl implements SidecarPeerHealthConfiguration
+{
+    public static final boolean DEFAULT_ENABLED = true;
+    public static final MillisecondBoundConfiguration DEFAULT_FREQUENCY = new MillisecondBoundConfiguration(30, TimeUnit.SECONDS);
+    public static final int DEFAULT_HEALTH_CHECK_RETRIES = 5;
+    public static final MillisecondBoundConfiguration DEFAULT_HEALTH_CHECK_RETRY_DELAY = new MillisecondBoundConfiguration(10, TimeUnit.SECONDS);;
+
+    @JsonProperty(value = "health_check_retries")
+    private final int healthCheckRetries;
+    @JsonProperty(value = "health_check_retry_delay")
```
yifan-c: Update the json key value accordingly, once the configuration is renamed.

server/src/main/java/org/apache/cassandra/sidecar/coordination/InnerDcTokenAdjacentPeerProvider.java
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        Metadata metadata = localInstances.values().stream().findFirst()
+                                          .map(in -> in.delegate().metadata())
+                                          .orElse(null);
+        if (metadata == null)
+        {
+            LOGGER.debug("Not yet connect to Cassandra cluster");
+            return Set.of();
+        }
```
yifan-c: `in.delegate().metadata()` throws `CassandraUnavailableException` when there is no connection. 

In such scenario, the chain won't be evaluated to `null`, but throws. 

Here is the code snippet of my suggestion. 

```java
        // Replaces `InstancesMetadata instancesMetadata`
        InstanceMetadataFetcher instanceFetcher;

        Metadata metadata;
        try
        {
            metadata = instanceFetcher.callOnFirstAvailableInstance(instance -> instance.delegate().metadata());
        }
        catch (Throwable cause)
        {
            LOGGER.debug("Unable to retrieve metadata", cause);
            return Set.of();
        }
```

server/src/main/java/org/apache/cassandra/sidecar/coordination/InnerDcTokenAdjacentPeerProvider.java
```diff
@@ -0,0 +1,270 @@
...
+        Map<Integer, InstanceMetadata> localInstances = instancesMetadata
+                                                        .instances()
+                                                        .stream()
+                                                        .collect(Collectors.toMap(InstanceMetadata::id, Function.identity()));
+
+        if (localInstances.isEmpty())
+        {
+            LOGGER.debug("No local instances found");
+            return Set.of();
+        }
```
yifan-c: With `InstanceMetadataFetcher`, this `if` statement can be removed too. It is covered by the suggestion for line#99

server/src/main/java/org/apache/cassandra/sidecar/coordination/InnerDcTokenAdjacentPeerProvider.java
```diff
@@ -0,0 +1,270 @@
...
+        Metadata metadata = localInstances.values().stream().findFirst()
+                                          .map(in -> in.delegate().metadata())
+                                          .orElse(null);
+        if (metadata == null)
+        {
+            LOGGER.debug("Not yet connect to Cassandra cluster");
+            return Set.of();
+        }
+
+        final List<KeyspaceMetadata> keyspaces = metadata.getKeyspaces()
```
yifan-c: remove `final` and such `final`s elsewhere in the patch.

server/src/main/java/org/apache/cassandra/sidecar/utils/SidecarClientProvider.java
```diff
@@ -0,0 +1,150 @@
...
+        this.vertx = vertx;
+        this.clientConfig = sidecarConfiguration.sidecarClientConfiguration();
+        this.sidecarVersionProvider = sidecarVersionProvider;
+        this.client = initializeSidecarClient();
+    }
+
+    @Override
+    public SidecarClient get()
+    {
+        return client;
```
yifan-c, line 80: The client is not closed. Can you add close to this class, and call it when server is closing?

server/src/main/java/org/apache/cassandra/sidecar/coordination/SidecarPeerHealthProvider.java
```diff
@@ -0,0 +1,43 @@
...
+ * Interface to provide health information for other Sidecar instances
+ */
+public interface SidecarPeerHealthProvider
+{
+    /**
+     * Possible Health states
+     */
+    enum Health
+    {
+        OK, DOWN, UNKNOWN
```
yifan-c: - nit: either `UP` and `DOWN`, or `OK` and `NOT_OK`. But, let's not mix them.
- not a nit: Add javadoc to explain when a peer is considered in such health status

server/src/main/java/org/apache/cassandra/sidecar/coordination/SidecarPeerHealthMonitorTask.java
```diff
@@ -0,0 +1,191 @@
...
+        this.config = sidecarConfiguration.sidecarPeerHealthConfiguration();
+        this.sidecarPeerProvider = sidecarPeerProvider;
+        this.healthProvider = healthProvider;
+        vertx.eventBus().registerDefaultCodec(SidecarInstance.class, sidecarInstanceCodecs);
+    }
+
+    @NotNull
+    public SidecarPeerHealthProvider.Health status(InstanceMetadata instance)
+    {
+        return status.getOrDefault(instance, SidecarPeerHealthProvider.Health.UNKNOWN);
```
yifan-c: It looks wrong to look up `InstanceMetadata` from the map `Map<SidecarInstance, SidecarPeerHealthProvider.Health> status`. The key type does not match.

server/src/main/java/org/apache/cassandra/sidecar/coordination/SidecarPeerHealthMonitorTask.java
```diff
@@ -0,0 +1,191 @@
...
+    {
+        this.vertx = vertx;
+        this.config = sidecarConfiguration.sidecarPeerHealthConfiguration();
+        this.sidecarPeerProvider = sidecarPeerProvider;
+        this.healthProvider = healthProvider;
+        vertx.eventBus().registerDefaultCodec(SidecarInstance.class, sidecarInstanceCodecs);
+    }
+
+    @NotNull
+    public SidecarPeerHealthProvider.Health status(InstanceMetadata instance)
```
yifan-c: The method is not used.

yifan-c (changes requested):
Incomplete review. Submitting the comments so far.

server/src/main/java/org/apache/cassandra/sidecar/coordination/CassandraClientTokenRingProvider.java
```diff
@@ -0,0 +1,315 @@
...
+                tokenRanges.put(instance, Range.openClosed(rangeStart, rangeEnd));
+            }
+        }
+        return tokenRanges;
+    }
+
+    /**
+     * Class to encapsule Cassandra instance data
+     */
+    private static class CassandraInstance
```
yifan-c: It is used by `protected` methods. To allow those method to remain use-able, you need to open the access to this class to at least `protected`

server/src/main/java/org/apache/cassandra/sidecar/coordination/SidecarHttpHealthProvider.java
```diff
@@ -0,0 +1,81 @@
...
+    {
+        try
+        {
+            SidecarClient client = clientProvider.get();
+            CompletableFuture<HealthResponse> healthRequest = client.executeRequestAsync(client.requestBuilder()
+                                                                                               .singleInstanceSelectionPolicy(
+                                                                                               new SidecarInstanceImpl(instance.hostname(),
+                                                                                                                       instance.port()))
+                                                                                               .retryPolicy(retryPolicy())
+                                                                                               .sidecarHealthRequest()
```
yifan-c: I think this block belongs to `SidecarClient.java`

server/src/main/java/org/apache/cassandra/sidecar/coordination/SidecarHttpHealthProvider.java
```diff
@@ -0,0 +1,81 @@
...
+                                                                                               .build());
+            return Future.fromCompletionStage(healthRequest)
+                         .map(healthResponse -> healthResponse.isOk()
+                                                ? Health.OK
+                                                : Health.DOWN);
+        }
+        catch (Exception e)
+        {
+            return Future.succeededFuture(Health.DOWN);
```
yifan-c: Looks like that `UNKNOWN` is not really used anywhere. Can we remove it? If it make sense to have `UNKNOWN`, can you use it?

server/src/main/java/org/apache/cassandra/sidecar/coordination/SidecarPeerHealthMonitorTask.java
```diff
@@ -0,0 +1,191 @@
...
+    {
+        return config.executeInterval();
+    }
+
+    @Override
+    public void execute(Promise<Void> promise)
+    {
+        try
+        {
+            run().onComplete(v -> promise.tryComplete()).wait(10000);
```
yifan-c: 1. `.wait(10000)` is not wanted. Can you remove it? You do not want to block the thread. 
2. `onComplete(v -> promise.tryComplete())` is missing the failure from `run()`.


```suggestion
            run()
            .onSuccess(v -> promise.tryComplete())
            .onFailure(promise::tryFail);
```

server/src/main/java/org/apache/cassandra/sidecar/coordination/SidecarPeerHealthMonitorTask.java
```diff
@@ -0,0 +1,191 @@
...
+                    .map(instance -> healthProvider.health(instance)
+                                                   .onSuccess(health -> {
+                                                       updateHealth(instance, health);
+                                                   })
+                                                   .onFailure(throwable -> {
+                                                       LOGGER.error("Failed to run health check, marking instance as DOWN host={} port={}",
+                                                                    instance.hostname(), instance.port(), throwable);
+                                                       markDown(instance);
+                                                   }))
+                    .collect(Collectors.toList());
```
yifan-c: Both `onSuccess` and `onFailure` are not guaranteed to be executed when `Future.all(...)` completes. There could be a race condition between the block from line#135 to line#144 and the lambda at line#125. 

To ensure the completion handling is invoked before the completion of per-peer future, update the code to the following. (Use `andThen`)

```suggestion
        sidecarPeers.stream()
                    .map(instance ->
                         healthProvider.health(instance)
                                       .andThen(ar -> {
                                           if (ar.succeeded())
                                           {
                                               updateHealth(instance, ar.result());
                                           }
                                           else
                                           {
                                               LOGGER.error("Failed to run health check, marking instance as DOWN host={} port={}",
                                                            instance.hostname(), instance.port(), ar.cause());
                                               markDown(instance);
                                           }
                                       }))
                    .collect(Collectors.toList());
```

server/src/main/java/org/apache/cassandra/sidecar/server/SidecarServerEvents.java
```diff
@@ -88,6 +88,25 @@ public enum SidecarServerEvents
...
+     * peer instance is marked as UP.
+     */
+    ON_SIDECAR_PEER_UP,
+
+    // LeadershipEvents
+    ON_TOKEN_RANGE_GAINED,
+    ON_TOKEN_RANGE_LOST,
+
+    // RangeManagerEvents
+    ON_TOKEN_RANGE_CHANGED,
```
yifan-c: They are not used. Are they needed by this patch? If not, please do not include the unrelated changes.

server/src/main/java/org/apache/cassandra/sidecar/server/SidecarServerEvents.java
```diff
@@ -88,6 +88,25 @@ public enum SidecarServerEvents
...
+     * The {@link io.vertx.core.eventbus.EventBus} address where events will be published when the health of a Sidecar
+     * peer instance is marked as DOWN.
+     */
+    ON_SIDECAR_PEER_DOWN,
+
+    /**
+     * The {@link io.vertx.core.eventbus.EventBus} address where events will be published when the health of a Sidecar
+     * peer instance is marked as UP.
+     */
+    ON_SIDECAR_PEER_UP,
```
yifan-c: Enrich the javadoc. What are the message payload format of those 2 events?

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/health/SidecarPeerDownDetectorIntegrationTest.java
bbotella: LOL. Good catch. Don't know where this came from

conf/sidecar.yaml
```diff
@@ -236,6 +236,57 @@ healthcheck:
...
+  execute_interval: 30s
+  # The amount of retries the client will attempt a request
+  sidecar_client_health_check_retries: 5
+  # The initial delay between the retries the client will attempt a request
+  sidecar_client_health_check_retry_delay: 10s
+
+# Sidecar client settings used to interact with other sidecars
+sidecar_client:
+  # Determines if the sidecar client should use and SSL connection
+  use_ssl: true
```
sarankk: We should disable `ssl` by default for client, since sidecar's ssl config is disabled by default. Instead of `use_ssl` can be add enable flag under `ssl:` ?

conf/sidecar.yaml
```diff
@@ -236,6 +236,57 @@ healthcheck:
...
+  # Time in which a request made by the sidecar client will time out
+  request_timeout: 1s
+  # How long the request can be idle
```
sarankk: Nit: to avoid extra lines, can we add comments on same line as setting ?, for e.g. here 

```
schema_reporting:                            # Schema Reporting configuration
  enabled:          false                    # Disabled by default
  initial_delay:    6h                       # Maximum delay before the first schema report (actual delay is randomized)
```

conf/sidecar.yaml
```diff
@@ -236,6 +236,57 @@ healthcheck:
...
+# Sidecar Peer Health Monitor settings
+# Enables a periodic task checking for the health of adjacent Sidecar peers in the token ring
+sidecar_peer_health:
+  # Determines if the peer health monitor periodic task is enabled or not
+  enabled: false
+  # Time between peer health checks
+  execute_interval: 30s
+  # The amount of retries the client will attempt a request
+  sidecar_client_health_check_retries: 5
+  # The initial delay between the retries the client will attempt a request
```
sarankk, line 248: Nit: shall we add comments on same line as setting?

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -114,6 +115,22 @@ public CompletableFuture<HealthResponse> sidecarHealth()
...
+     */
+    public CompletableFuture<HealthResponse> peerHealth(SidecarInstance instance)
+    {
+        return executor.executeRequestAsync(requestBuilder()
+                                                .singleInstanceSelectionPolicy(
+                                                new SidecarInstanceImpl(instance.hostname(),
+                                                                        instance.port()))
+                                                .retryPolicy(oncePerInstanceRetryPolicy)
+                                                .sidecarHealthRequest()
+                                                .build());
```
yifan-c: not aligned properly. 

```suggestion
        return executor.executeRequestAsync(requestBuilder()
                                            .singleInstanceSelectionPolicy(instance)
                                            .retryPolicy(oncePerInstanceRetryPolicy)
                                            .sidecarHealthRequest()
                                            .build());
```

conf/sidecar.yaml
```diff
@@ -236,6 +236,57 @@ healthcheck:
   initial_delay: 0ms
   execute_interval: 30s
 
+# Sidecar Peer Health Monitor settings
+# Enables a periodic task checking for the health of adjacent Sidecar peers in the token ring
+sidecar_peer_health:
```
sarankk, line 241: Nit: since `sidecar_peer_health` has only 4 settings, shall we move it under `sidecar:` section and remove `sidecar` prefix?. We also seem to have `PeerHealthConfiguration` class

server/src/main/java/org/apache/cassandra/sidecar/config/SidecarClientConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Configuration for sidecar client
+ */
+public interface SidecarClientConfiguration
+{
+    /**
+     * @return {@code true} if SSL should be used for Sidecar client connections
+     */
+    boolean useSsl();
```
sarankk: `useSsl` should go under `SslConfiguration`

server/src/main/java/org/apache/cassandra/sidecar/coordination/InnerDcTokenAdjacentPeerProvider.java
```diff
@@ -66,42 +63,33 @@ public class InnerDcTokenAdjacentPeerProvider implements SidecarPeerProvider
...
+        this.instanceFetcher = instanceFetcher;
         this.cassandraClientTokenRingProvider = cassandraClientTokenRingProvider;
         this.serviceConfiguration = serviceConfiguration;
         this.dnsResolver = dnsResolver;
     }
 
     public Set<SidecarInstance> get()
     {
-        Map<Integer, InstanceMetadata> localInstances = instancesMetadata
-                                                        .instances()
```
yifan-c: Please drop the `final`s in this method

server/src/main/java/org/apache/cassandra/sidecar/config/SidecarClientConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+     * Return the configured number of event-loop the pool use.
+     *
+     * <ul>
+     *   <li>when the size is {@code 0}, the client pool will use the current event-loop</li>
+     *   <li>otherwise the client will create and use its own event loop</li>
+     * </ul>
+     *
+     * @return the configured number of event-loop the pool use
+     */
+    int connectionPoolEventLoopSize();
```
sarankk, line 66: I think this setting allows the client to create its own event loop, should we avoid this? should client use server's current event loop to allow for sharing of resources between server and client?

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/PeerHealthConfigurationImpl.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.cassandra.sidecar.common.server.utils.MillisecondBoundConfiguration;
+import org.apache.cassandra.sidecar.config.PeerHealthConfiguration;
+
+
+/**
+ * Configuration for Peer Health checks
+ */
+public class PeerHealthConfigurationImpl extends PeriodicTaskConfigurationImpl implements PeerHealthConfiguration
+{
+    public static final boolean DEFAULT_ENABLED = true;
```
sarankk: In `sidecar.conf` enabled is set to `false` by default. Here it is still true.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/PeerHealthConfigurationImpl.java
```diff
@@ -0,0 +1,84 @@
...
+                                       int sidecarClientHealthCheckRetries,
+                                       MillisecondBoundConfiguration sidecarClientHealthCheckRetryDelay)
+    {
+        super(enabled, frequency, frequency);
+        this.sidecarClientHealthCheckRetries = sidecarClientHealthCheckRetries;
+        this.sidecarClientHealthCheckRetryDelay = sidecarClientHealthCheckRetryDelay;
+    }
+
+    /**
+     * @return the number of maximum retries to be performed during a Sidecar peer health check
```
sarankk: can we use ` {@inheritDoc}` to use same doc as the interface?

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/SidecarClientConfigurationImpl.java
```diff
@@ -0,0 +1,216 @@
...
+    {
+        return retryDelay;
+    }
+
+    @Override
+    public MillisecondBoundConfiguration maxRetryDelay()
+    {
+        return maxRetryDelay;
+    }
+
```
sarankk: Nit: extra line

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/SidecarConfigurationImpl.java
```diff
@@ -437,6 +465,17 @@ public Builder sslConfiguration(SslConfiguration sslConfiguration)
             return update(b -> b.sslConfiguration = sslConfiguration);
         }
 
+        /**
+         * Sets the {@code downDetectorConfiguration} and returns a reference to this Builder enabling method chaining.
+         *
+         * @param peerHealthConfiguration the {@code downDetectorConfiguration} to set
+         * @return a reference to this Builder
+         */
+        public Builder downDetectorConfiguration(PeerHealthConfiguration peerHealthConfiguration)
```
sarankk: This method seems to have been added by mistake.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -113,6 +115,22 @@ public CompletableFuture<HealthResponse> sidecarHealth()
...
+    /**
+     * Executes the Sidecar health request using the configured selection policy and with no retries
+     *
+     * @return a completable future of the Sidecar health response
+     */
+    public CompletableFuture<HealthResponse> peerHealth(SidecarInstance instance)
+    {
+        return executor.executeRequestAsync(requestBuilder()
+                                                .singleInstanceSelectionPolicy(
+                                                new SidecarInstanceImpl(instance.hostname(),
```
sarankk: Why not share `instance` here instead of creating a new one?

server/src/main/java/org/apache/cassandra/sidecar/coordination/SidecarPeerHealthMonitorTask.java
```diff
@@ -67,19 +65,13 @@ public SidecarPeerHealthMonitorTask(Vertx vertx,
...
         this.sidecarPeerProvider = sidecarPeerProvider;
         this.healthProvider = healthProvider;
-        vertx.eventBus().registerDefaultCodec(SidecarInstance.class, sidecarInstanceCodecs);
-    }
-
-    @NotNull
-    public SidecarPeerHealthProvider.Health status(InstanceMetadata instance)
-    {
-        return status.getOrDefault(instance, SidecarPeerHealthProvider.Health.UNKNOWN);
+        vertx.eventBus().registerDefaultCodec(SidecarInstance.class, sidecarInstanceCodec);
```
yifan-c, line 75: nit: it is not the best place to register codec for types. For now, it is fine. Maybe comment a todo above this line.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/PeerHealthConfigurationImpl.java
```diff
@@ -0,0 +1,84 @@
...
+     */
+    @Override
+    @JsonProperty(value = "sidecar_client_health_check_retries")
```
sarankk: Nit: are we creating Json from `PeerHealthConfiguration` object? if not we dont need these `@JsonProperty` tags

server/src/main/java/org/apache/cassandra/sidecar/utils/SidecarClientProvider.java
```diff
@@ -76,6 +77,21 @@ public SidecarClient get()
         return client;
     }
 
+    public void close(Promise<Void> completion)
+    {
+        LOGGER.info("Closing Sidecar Client...");
+        try
+        {
+            client.close();
+            Thread.sleep(100);
```
yifan-c: This (sleep) is interesting. Why?

server/src/main/java/org/apache/cassandra/sidecar/utils/SidecarClientProvider.java
```diff
@@ -76,6 +77,21 @@ public SidecarClient get()
...
+            completion.complete();
+        }
+        catch (Throwable throwable)
+        {
+            completion.fail(throwable);
+        }
+    }
```
yifan-c: This is not really async. Let's not declare promise input. See the suggestion below, and at the callsite, i.e. `Server#close`, have this instead. 
```
        Future.future(p -> {
            sidecarClientProvider.close();
            p.complete();
        });
```

```suggestion
     private final AtomicBoolean isClosing = new AtomicBoolean(false);
    public void close()
    {
        if (isClosing.compareAndSet(false, true))
        {
            LOGGER.info("Closing Sidecar Client...");
            ThrowableUtils.propagate(client::close);
        }
    }
```

yifan-c (commented):
A few more comments. Looks much better.

server/src/main/java/org/apache/cassandra/sidecar/utils/SidecarClientProvider.java
bbotella: @frankgh identified a bug in vertx itself, that completing promises so close together may cause problems.

server/src/main/java/org/apache/cassandra/sidecar/utils/SidecarClientProvider.java
```diff
@@ -0,0 +1,166 @@
...
+
+    private SidecarClient initializeSidecarClient()
+    {
+        WebClientOptions webClientOptions = webClientOptions();
+        HttpClient httpClient = vertx.createHttpClient(webClientOptions);
+        WebClient webClient = WebClient.wrap(httpClient, webClientOptions);
```
sarankk, line 96: `WebClient webClient = WebClient.create(vertx, webClientOptions)` we can do this in place of wrapping `HttpClient`

server/src/main/java/org/apache/cassandra/sidecar/utils/SidecarClientProvider.java
```diff
@@ -0,0 +1,166 @@
...
+                                 requestExecutor,
+                                 config,
+                                 defaultRetryPolicy);
+    }
+
+    private WebClientOptions webClientOptions()
+    {
+        WebClientOptions options = new WebClientOptions();
+        options.getPoolOptions()
+               .setCleanerPeriod(clientConfig.connectionPoolCleanerPeriod().toIntMillis())
```
sarankk, line 131: Shall we add these client settings to `HttpClientConfig`, that way we dont have to create `WebClientOptions separately`, also `VertxHttpClient` takes care of setting ssl config, we can use it

server/src/main/java/org/apache/cassandra/sidecar/utils/SidecarClientProvider.java
```diff
@@ -0,0 +1,166 @@
...
+        WebClient webClient = WebClient.wrap(httpClient, webClientOptions);
+
+        HttpClientConfig httpClientConfig = new HttpClientConfig.Builder<>()
+                                            .ssl(webClientOptions().isSsl())
+                                            .timeoutMillis(clientConfig.requestTimeout().toMillis())
+                                            .idleTimeoutMillis(clientConfig.requestIdleTimeout().toIntMillis())
+                                            .userAgent("cassandra-sidecar/" + sidecarVersionProvider.sidecarVersion())
+                                            .build();
+
+        VertxHttpClient vertxHttpClient = new VertxHttpClient(vertx, webClient, httpClientConfig);
```
sarankk, line 105: If we add missing settings in `HttpClientConfig` we dont have to create `WebClient` separately, we can directly call `VertxHttpClient vertxHttpClient = new VertxHttpClient(vertx, httpClientConfig)`

server/src/main/java/org/apache/cassandra/sidecar/utils/SidecarClientProvider.java
```diff
@@ -0,0 +1,166 @@
...
+        RetryPolicy defaultRetryPolicy = new ExponentialBackoffRetryPolicy(clientConfig.maxRetries(),
+                                                                           clientConfig.retryDelay().toMillis(),
+                                                                           clientConfig.retryDelay().toMillis());
+        SidecarClientVertxRequestExecutor requestExecutor = new SidecarClientVertxRequestExecutor(vertxHttpClient);
+        SidecarInstance instance = new SidecarInstanceImpl(webClientOptions.getDefaultHost(), webClientOptions.getDefaultPort());
```
sarankk, line 110: Since we are not setting default host and port, should we avoid getting those values?

server/src/main/java/org/apache/cassandra/sidecar/utils/SidecarClientProvider.java
```diff
@@ -0,0 +1,166 @@
...
+        ArrayList<SidecarInstance> instances = new ArrayList<>();
+        instances.add(instance);
```
sarankk, line 112: We seem to create `SidecarClient` with random host and port and call `peerhealth` with actual instance we want to check health for. Should we instead create `SidecarClient` with instance of interest and directly call `sidecarHealth` ? this way we dont need `peerHealth` method and need to pass default host and port. We can maintain `Map<SidecarInstance, SidecarClient>` to get `Sidecarclient` specific for an instance

server/src/main/java/org/apache/cassandra/sidecar/server/Server.java
```diff
@@ -189,6 +197,7 @@ public Future<CompositeFuture> close()
         });
 
         return Future.all(closingFutures)
+
```
sarankk: Nit: extra line

server/src/main/java/org/apache/cassandra/sidecar/coordination/SidecarPeerProvider.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.apache.cassandra.sidecar.coordination;
+
+import java.util.Set;
+
+import org.apache.cassandra.sidecar.common.client.SidecarInstance;
+
+/**
+ * Provides a set of Sidecar peers to monitor and failover when they go DOWN
```
sarankk: Since `SidecarPeerProvider` can be used for other tasks other than health monitor, should we have `Provides a set of peers of current Sidecar` in comment?

server/src/main/java/org/apache/cassandra/sidecar/coordination/SidecarPeerProvider.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public interface SidecarPeerProvider
+{
+    /**
+     * @return a set of Sidecar peers to monitor
```
sarankk: Nit: shall we have `@return a set of Sidecar peers`

server/src/main/java/org/apache/cassandra/sidecar/utils/SidecarClientProvider.java
bbotella: Synced offline. Leaving it as it is.

server/src/main/java/org/apache/cassandra/sidecar/coordination/SidecarPeerProvider.java
bbotella: Synced offline. Leaving it as it is as it may be confused with Cassandra specific terminology.

server/src/main/java/org/apache/cassandra/sidecar/utils/SidecarClientProvider.java
bbotella: Synced offline. Leaving it as it is.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/PeerHealthConfigurationImpl.java
yifan-c: It uses the jackson framework to read from yaml. The annotation is to locate the field in yaml.

server/src/main/java/org/apache/cassandra/sidecar/server/Server.java
```diff
@@ -167,9 +167,7 @@ public Future<CompositeFuture> close()
         periodicTaskExecutor.close(periodicTaskExecutorPromise);
         closingFutures.add(periodicTaskExecutorPromise.future());
 
-        Promise<Void> closeSidecarClientPromise = Promise.promise();
-        sidecarClientProvider.close(closeSidecarClientPromise);
-        closingFutures.add(closeSidecarClientPromise.future());
+        sidecarClientProvider.close();
```
yifan-c: Still need to add it in the future chain. 

```suggestion
        closingFutures.add(Future.future(p -> {
            sidecarClientProvider.close();
            p.complete();
        }));
```

yifan-c (approved):
My last comment / suggestion. Waiting for green CI.

yifan-c:
Please also rebase.

frankgh (approved):
+1

PR #190, opened 2025-02-11
CASSSIDECAR-207: Upgrade Netty to 4.1.118.Final and Vert.x to 4.5.13 Version

sarankk:
Netty released 4.1.118.Final https://netty.io/news/2025/02/10/4-1-118-Final.html with critical CVE fix and Vert.x released 4.5.13 with Netty upgrade. This patch upgrades Netty version to 4.1.118.Final and Vert.x version to 4.5.13 in Sidecar. With Vert.x 4.5.13 version, a bug fix for vert.x auth package was also released. We integrate the bug fix as well.

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
@@ -53,14 +53,8 @@ public class DataResourceScope implements ResourceScope
      */
     public static final String DATA_WITH_KEYSPACE = String.format("data/{%s}", KEYSPACE);
 
-    // TODO remove this hack once VariableAwareExpression bug is fixed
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/DataResourceScope.java
```diff
@@ -53,14 +53,8 @@ public class DataResourceScope implements ResourceScope
      */
     public static final String DATA_WITH_KEYSPACE = String.format("data/{%s}", KEYSPACE);
 
-    // TODO remove this hack once VariableAwareExpression bug is fixed
```
yifan-c, line 56: TODO is addressed :)

yifan-c (approved):

frankgh (approved):
+1 one note. The Netty version should only be used for tests. Vertx can bring a different Netty version as a transitive dependency, and that's the version we'll end up using.

PR #191, opened 2025-02-12
CASSSIDECAR-209: HealthCheckPeriodicTask leak in Sidecar

JeetKunDoug:
Also fixes a minor build script issue where dtest jars could not be found if using the CASSANDRA_DEP_DIR environment variable to centralize dtest jars across projects (the shell scripts didn't honor it but the gradle build tried to).

### DIFF
--- a/scripts/build-shaded-dtest-jar-local.sh
+++ b/scripts/build-shaded-dtest-jar-local.sh
@@ -26,7 +26,7 @@ CASSANDRA_VERSION=$(cat build.xml | grep 'property name="base.version"' | awk -F
 GIT_HASH=$(git rev-parse --short HEAD)
 DTEST_ARTIFACT_ID=${ARTIFACT_NAME}-local
 DTEST_JAR_DIR="$(dirname "${SCRIPT_DIR}/")/dtest-jars"
-
+DTEST_JAR_DIR=${CASSANDRA_DEP_DIR:-$DTEST_JAR_DIR}
 echo "${CASSANDRA_VERSION}"
 echo "${GIT_HASH}"
 echo "${DTEST_ARTIFACT_ID}"
--- a/server/src/main/java/org/apache/cassandra/sidecar/server/Server.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/server/Server.java
@@ -303,11 +303,10 @@ public class Server
      */
     protected Future<String> scheduleInternalPeriodicTasks(String deploymentId)
     {
-        HealthCheckPeriodicTask healthCheckPeriodicTask = new HealthCheckPeriodicTask(
-        sidecarConfiguration,
-                                                                   instancesMetadata,
-                                                                   executorPools,
-                                                                   metrics);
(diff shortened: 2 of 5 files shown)


server/src/main/java/org/apache/cassandra/sidecar/server/Server.java
```diff
@@ -302,11 +303,15 @@ protected void validate()
...
-        periodicTaskExecutor.schedule(new HealthCheckPeriodicTask(vertx,
-                                                                  sidecarConfiguration,
-                                                                  instancesMetadata,
-                                                                  executorPools,
-                                                                  metrics));
+        HealthCheckPeriodicTask healthCheckPeriodicTask = new HealthCheckPeriodicTask(
+        sidecarConfiguration,
+                                                                   instancesMetadata,
+                                                                   executorPools,
+                                                                   metrics);
```
frankgh: Can we fix the formatting here?
```suggestion
        HealthCheckPeriodicTask healthCheckPeriodicTask = new HealthCheckPeriodicTask(sidecarConfiguration,
                                                                                      instancesMetadata,
                                                                                      executorPools,
                                                                                      metrics);
```

frankgh (approved):
Looks good, I've left a minor comment to be addressed

scripts/build-shaded-dtest-jar-local.sh
```diff
@@ -26,7 +26,7 @@ CASSANDRA_VERSION=$(cat build.xml | grep 'property name="base.version"' | awk -F
 GIT_HASH=$(git rev-parse --short HEAD)
 DTEST_ARTIFACT_ID=${ARTIFACT_NAME}-local
 DTEST_JAR_DIR="$(dirname "${SCRIPT_DIR}/")/dtest-jars"
-
+DTEST_JAR_DIR=${CASSANDRA_DEP_DIR:-$DTEST_JAR_DIR}
```
yifan-c, line 29: I do not think the dtest jars should be placed in a central place and shared. The projects can depend on different SHAs. Unlike the local maven repo, where the jars can be distinguished by version. Sharing the dtest jars could cause confusion. 
It is unrelated to the patch, just to bring it up, do not need to address.

yifan-c (approved):
lgtm.

scripts/build-shaded-dtest-jar-local.sh
JeetKunDoug: This builds a jar with `dtest-M.m.p.jar` in the shared directory _if you set the environment variable_, and the local dir if not - either way, they files are versioned. If you happen to use the shared dependencies directory, you often don't have to rebuild C* versions. A previous patch started to _use_ the shared directory in _gradle_ but didn't include it here, so I made the two consistent. I happen to like not having several dozen copies of C* builds floating around my system personally, so I use it and have never had issues... just don't set the `CASSANDRA_DEP_DIR` environment variable if you don't want to use it.

server/src/main/java/org/apache/cassandra/sidecar/server/Server.java
JeetKunDoug: Done

PR #192, opened 2025-02-13
CASSSIDECAR-210: Fix PeriodicTaskExecutor double execution due to race from reschedule

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-210

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutor.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutor.java
@@ -154,7 +154,7 @@ public class PeriodicTaskExecutor implements Closeable
             return;
         }
         long startTime = System.nanoTime();
-        internalPool.<Result>executeBlocking(promise -> executeInternal(promise, key, execCount), false)
+        internalPool.<ScheduleDecision>executeBlocking(promise -> executeInternal(promise, key, execCount), false)
                     .onComplete(outcome -> {
                         LOGGER.debug("Task run finishes. task='{}' outcome={} execCount={}", key, outcome, execCount);
                         runPromise.complete(); // mark the completion, regardless of the result from last run
@@ -162,7 +162,7 @@ public class PeriodicTaskExecutor implements Closeable
 
                         DurationSpec delay;
                         long priorExecutionDurationMillis;
-                        if (outcome.result() == Result.RESCHEDULED)
+                        if (outcome.result() == ScheduleDecision.RESCHEDULE)
                         {
                             priorExecutionDurationMillis = 0;
                             delay = key.task.initialDelay();
@@ -248,37 +248,28 @@ public class PeriodicTaskExecutor implements Closeable
         }
     }
 
-    private void executeInternal(Promise<Result> promise, PeriodicTaskKey key, long execCount)
+    private void executeInternal(Promise<ScheduleDecision> promise, PeriodicTaskKey key, long execCount)
     {
         PeriodicTask periodicTask = key.task;
-        switch (consolidateScheduleDecision(periodicTask))
+        ScheduleDecision scheduleDecision = consolidateScheduleDecision(periodicTask);
+        LOGGER.debug("{} task. task='{}' execCount={}", scheduleDecision, key, execCount);
+        if (scheduleDecision == ScheduleDecision.EXECUTE)
         {
-            case SKIP:
-                LOGGER.trace("Skip executing task. task='{}' execCount={}", key, execCount);
-                promise.tryComplete();
-                return;
-
-            case EXECUTE:
-                break;
-
-            case RESCHEDULE:
-            default:
-                LOGGER.debug("Rescheduling the task. task='{}' execCount={}", key, execCount);
-                promise.tryComplete(Result.RESCHEDULED);
-                return;
+            Promise<Void> taskRunPromise = Promise.promise();
+            taskRunPromise.future().onSuccess(ignored -> promise.tryComplete(ScheduleDecision.EXECUTE));
+            try
+            {
+                periodicTask.execute(taskRunPromise);
+            }
+            catch (Throwable throwable)
+            {
+                LOGGER.warn("Periodic task failed to execute. task='{}' execCount={}", periodicTask.name(), execCount, throwable);
+                promise.tryFail(throwable);
+            }
         }
-
-        Promise<Void> taskRunPromise = Promise.promise();
-        taskRunPromise.future().onSuccess(ignored -> promise.tryComplete());
-        try
+        else
         {
-            LOGGER.debug("Executing task. task='{}' execCount={}", key, execCount);
-            periodicTask.execute(taskRunPromise);
-        }
-        catch (Throwable throwable)
-        {
-            LOGGER.warn("Periodic task failed to execute. task='{}' execCount={}", periodicTask.name(), execCount, throwable);
-            promise.tryFail(throwable);
+            promise.tryComplete(scheduleDecision);
         }
     }
 
@@ -341,15 +332,6 @@ public class PeriodicTaskExecutor implements Closeable
         {
             return fqcnAndName;
         }
-    }
-
-    enum Result
-    {
-        /**
-         * Tells the run completion handler whether task is rescheduled.
-         * If it is rescheduled, the delay is adjusted to the initial delay of the task
-         */
-        RESCHEDULED
     }
 
     @VisibleForTesting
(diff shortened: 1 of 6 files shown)


server/src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutor.java
```diff
@@ -240,31 +263,22 @@ private void executeInternal(Promise<Void> promise, PeriodicTaskKey key, long ex
 
             case RESCHEDULE:
             default:
-                LOGGER.debug("Reschedule the task. task='{}' execCount={}", key, execCount);
-                reschedule(periodicTask);
-                promise.tryComplete();
+                LOGGER.debug("Rescheduling the task. task='{}' execCount={}", key, execCount);
+                promise.tryComplete(Result.RESCHEDULED);
```
yifan-c: This is the fix. Do not reschedule, but just signal the decision and update the delay when scheduling a new run.

server/src/test/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutorTest.java
```diff
@@ -162,80 +150,61 @@ void testUnscheduleShouldStopExecution()
...
-            }
-            Uninterruptibles.awaitUninterruptibly(latch1);
-        });
+        testRescheduleDecision(1, 0);
+        testRescheduleDecision(0, 0);
+        testRescheduleDecision(1, 1);
+        testRescheduleDecision(0, 1);
+    }
+
+    void testRescheduleDecision(long initialDelay, long delay)
```
yifan-c, line 194: This is the test that reproduce the bug. 
Test fails with the earlier version of `PeriodicTaskExecutor`, as there are flee tasks that keep on running. 

Verify with the following 2 commands
```
git checkout upstream/trunk -- server/src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutor.java
./gradlew :server:test --tests org.apache.cassandra.sidecar.tasks.PeriodicTaskExecutorTest.testRescheduleDecision
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutor.java
```diff
@@ -329,15 +343,18 @@ public String toString()
...
-    @VisibleForTesting
-    Map<PeriodicTaskKey, Long> timerIds()
+    enum Result
     {
-        return timerIds;
+        /**
+         * Tells the run completion handler whether task is rescheduled.
+         * If it is rescheduled, the delay is adjusted to the initial delay of the task
+         */
+        RESCHEDULED
```
JeetKunDoug: Can we have a Result that _doesn't_ mean RESCHEDULED, like COMPLETED? The fact that we  call`Promise#tryComplete` with either RESCHEDULED or nothing at all (essentially, `null`) seems really odd, even if we never really check for COMPLETED.

server/src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutor.java
```diff
@@ -130,22 +147,33 @@ private void schedule(PeriodicTaskKey key, long priorExecDurationMillis, long de
...
+                        LOGGER.debug("Task run finishes. task='{}' outcome={} execCount={}", key, outcome, execCount);
+                        runPromise.complete(); // mark the completion, regardless of the result from last run
+                        activeRuns.remove(key);
+
+                        DurationSpec delay;
+                        long priorExecutionDurationMillis;
+                        if (outcome.result() == Result.RESCHEDULED)
+                        {
+                            priorExecutionDurationMillis = 0;
+                            delay = key.task.initialDelay();
```
JeetKunDoug, line 168: Why do we always use the `initialDelay` here? What if the task has run 20 times already - do we still want to use `initialDelay`? Can we use `execCount` here to decide which delay to use?

JeetKunDoug (commented):
I'll re-review this in the morning (out of time for now) but a few comments so far. The scheduling/rescheduling logic is getting a bit complicated but not sure if there's much to do about that at this point.

server/src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutor.java
yifan-c: `COMPLETED` is unnecessary. When the handler at `onComplete` is invoked, we know that it is completed.

server/src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutor.java
yifan-c: Yes. It is to implement the semantics of "reschedule". When it is rescheduled, the first run the rescheduled task take the `initialDelay`. It is consistent with the prior `reschedule` method. 

`execCount` is irrelevant here.

server/src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutor.java
yifan-c: We can probably rename the enum. It is not really the result of execution. Looking for suggestions.

JeetKunDoug: Yeah - the biggest thing to me was that this enum is used as the type of an AsyncResult<T> and we set the result to `RESCHEDULED` or nothing at all, which was just jarring when I was looking at the code.

What do you think of actually using the `ScheduleDecision` enum here, and just getting rid of the `Result` all together? Then each branch of the `executeInternal` method would actually be able to indicate which decision was made - we end up only _using_ `RESCHEDULE` right now but at least there isn't an odd mix of `promise.tryComplete()` and `promise.tryComplete(Result.RESCHEDULE)` which didn't make a lot of sense to me - if the promise is a `Promise<Result>` it should really be completed with _something_ other than null... using `Promise<ScheduleDecision>` makes more sense.

server/src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutor.java
```diff
@@ -130,22 +147,33 @@ private void schedule(PeriodicTaskKey key, long priorExecDurationMillis, long de
      */
     private void executeAndScheduleNext(PeriodicTaskKey key, long execCount)
     {
+        Promise<Void> runPromise = Promise.promise();
+        if (activeRuns.computeIfAbsent(key, k -> runPromise.future()) != runPromise)
```
JeetKunDoug, line 151: `activeRuns` no longer has any use for the _value_ of it's entries and can be converted to a `ConcurrentHashSet`, which simplifies this and `unschedule` quite a bit.

JeetKunDoug: Well, "quite a bit" is an overstatement but a bit.

server/src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutor.java
yifan-c: yep. `ScheduleDecision` makes better sense.

server/src/main/java/org/apache/cassandra/sidecar/tasks/PeriodicTaskExecutor.java
yifan-c: It still wants to grab the `future` of the active run, so that unschedule can chain "task.close()" after the run has finished, to avoid any interruption to the active run.

JeetKunDoug: But why do you need the `runPromise`? It's _never_ used for anything except to call `runPromise.complete()` on line 160 below - the future isn't ever actually used as far as I can tell. It _used_ to be used in `unschedule but now it's just ignored anyway - if you turn `activeRuns` into a `ConcurrentHashSet` and run all the tests, it just works (although that may be more of an indication that there are code paths that are uncovered in the current tests?)

yifan-c: Added test to exercise that code path.

JeetKunDoug (approved):
+1 - thanks for the fix!

PR #195, opened 2025-02-16
CASSSIDECAR-213: Release resources in unit tests

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-213


server/src/test/resources/config/sidecar_cdc.yaml
```diff
@@ -59,7 +59,7 @@ sidecar:
   cdc:
     segment_hardlink_cache_expiry: 1m # 1 minute
   jmx:
-    max_retries: 42
+    max_retries: 1
```
yifan-c, line 62: For reviewers, the 42 retries is just too much for testing. The retry blocks the closure of server, leading to the test being stuck for 30 minutes and killed by test runner. 
Changing it to 1 does not alter the tests, but allows test completion way faster (like, in milliseconds).

server/src/test/java/org/apache/cassandra/sidecar/config/SidecarConfigurationTest.java
```diff
@@ -112,7 +112,7 @@ void testReadingJmxConfiguration() throws IOException
         SidecarConfiguration config = SidecarConfigurationImpl.readYamlConfiguration(yamlPath);
         assertThat(config.serviceConfiguration().jmxConfiguration()).isNotNull();
         JmxConfiguration jmxConfiguration = config.serviceConfiguration().jmxConfiguration();
-        assertThat(jmxConfiguration.maxRetries()).isEqualTo(42);
+        assertThat(jmxConfiguration.maxRetries()).isEqualTo(1);
```
yifan-c, line 115: We all like the ultimate answer, but since it has been modified in the ymls, update the test assertion accordingly.

frankgh (approved):
+1 thanks! This fixes the issues I've had for a while on my local environment

PR #196, opened 2025-02-18
CASSSIDECAR-214: Speed up ConnectedClientStatsHandlerIntegrationTest

frankgh:
Patch by Francisco Guerrero; reviewed by TBD for CASSSIDECAR-214


integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/ConnectedClientStatsHandlerIntegrationTest.java
```diff
@@ -0,0 +1,175 @@
...
+     * Expects unrecognized params to be ignored and invalid value for the expected parameter to be defaulted to true
+     * to prevent heavyweight query in the bad request case.
+     */
+    @Test
+    void retrieveClientStatsInvalidParameterValue()
+    {
+        Map<String, Boolean> expectedParams = Map.of("summary", true);
+        String testRoute = "/api/v1/cassandra/stats/connected-clients?summary=123&bad-arg=xyz";
+        HttpResponse<Buffer> response = getBlocking(trustedClient().get(server.actualPort(), "localhost", testRoute)
+                                                                   .expect(ResponsePredicate.SC_OK)
```
bbotella, line 118: Not related to this PR. I wonder if this should be a 200 or a 400 response.

bbotella (approved):
This is really nice.

+1 (nb)

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/ConnectedClientStatsHandlerIntegrationTest.java
frankgh: This should still be a 200. We don't validate unexpected query parameters, and looking at the `org.apache.cassandra.sidecar.utils.RequestUtils#parseBooleanQueryParam` logic, we only allow `true` or `false` otherwise we use the default value.

PR #198, opened 2025-02-19
CASSSIDECAR-203: Created Endpoint that Triggers an Immediate Schema Report

5:
https://issues.apache.org/jira/browse/CASSSIDECAR-203

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 0.1.0
 -----
+ * Create Endpoint that Triggers an Immediate Schema Report (CASSSIDECAR-203)
  * Add JWT Authentication support in Sidecar (CASSSIDECAR-160)
  * Add Token Ring Peer Health monitor (CASSSIDECAR-206)
  * Config APIs for storing CDC/Kafka configs for CDC feature (CASSSIDECAR-211)
@@ -12,7 +13,7 @@
  * Sidecar schema initialization can be executed on multiple thread (CASSSIDECAR-200)
  * Make sidecar operations resilient to down Cassandra nodes (CASSSIDECAR-201)
  * Fix Cassandra instance not found error (CASSSIDECAR-192)
- * Implemented Schema Reporter for Integration with DataHub (CASSSIDECAR-191)
+ * Implement Schema Reporter for Integration with DataHub (CASSSIDECAR-191)
  * Add sidecar endpoint to retrieve stream stats (CASSSIDECAR-180)
  * Add sidecar endpoint to retrieve cassandra gossip health (CASSSIDECAR-173)
  * Fix SidecarSchema stuck at initialization due to ClusterLeaseTask scheduling (CASSSIDECAR-189)
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -13,7 +13,7 @@
  * Sidecar schema initialization can be executed on multiple thread (CASSSIDECAR-200)
  * Make sidecar operations resilient to down Cassandra nodes (CASSSIDECAR-201)
  * Fix Cassandra instance not found error (CASSSIDECAR-192)
- * Implement Schema Reporter for Integration with DataHub (CASSSIDECAR-191)
+ * Implemented Schema Reporter for Integration with DataHub (CASSSIDECAR-191)
  * Add sidecar endpoint to retrieve stream stats (CASSSIDECAR-180)
  * Add sidecar endpoint to retrieve cassandra gossip health (CASSSIDECAR-173)
  * Fix SidecarSchema stuck at initialization due to ClusterLeaseTask scheduling (CASSSIDECAR-189)
--- a/client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
+++ b/client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
@@ -122,7 +122,7 @@ public final class ApiEndpointsV1
     public static final String STREAM_CDC_SEGMENTS_ROUTE = LIST_CDC_SEGMENTS_ROUTE + "/" + SEGMENT_PATH_PARAM;
 
     // Schema Reporting
-    private static final String REPORT_SCHEMA = "/report-schema";
+    private static final String REPORT_SCHEMA = "/datahub/schemas";
     public static final String REPORT_SCHEMA_ROUTE = API_V1 + REPORT_SCHEMA;
 
     public static final String SERVICES_PATH = "/services";
--- a/client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
+++ b/client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
@@ -121,12 +121,15 @@ public final class ApiEndpointsV1
     public static final String LIST_CDC_SEGMENTS_ROUTE = API_V1 + CDC_PATH + "/segments";
     public static final String STREAM_CDC_SEGMENTS_ROUTE = LIST_CDC_SEGMENTS_ROUTE + "/" + SEGMENT_PATH_PARAM;
 
+    // Schema Reporting
+    private static final String REPORT_SCHEMA = "/report-schema";
+    public static final String REPORT_SCHEMA_ROUTE = API_V1 + REPORT_SCHEMA;
+
     public static final String SERVICES_PATH = "/services";
     public static final String SERVICE_PARAM = ":service";
     public static final String CONFIG = "/config";
     public static final String SERVICE_CONFIG_ROUTE = API_V1 + SERVICES_PATH + "/" + SERVICE_PARAM + CONFIG;
     public static final String SERVICES_CONFIG_ROUTE = API_V1 + SERVICES_PATH;
-
 
     public static final String CONNECTED_CLIENT_STATS_ROUTE = API_V1 + CASSANDRA + "/stats/connected-clients";
 
--- a/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
+++ b/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
@@ -539,9 +539,9 @@ public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExt
     public CompletableFuture<ListCdcSegmentsResponse> listCdcSegments(SidecarInstance sidecarInstance)
     {
         return executor.executeRequestAsync(requestBuilder()
-                       .singleInstanceSelectionPolicy(sidecarInstance)
-                       .request(new ListCdcSegmentsRequest())
-                       .build());
+                                            .singleInstanceSelectionPolicy(sidecarInstance)
+                                            .request(new ListCdcSegmentsRequest())
+                                            .build());
     }
 
     /**
@@ -561,9 +561,9 @@ public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExt
                                   StreamConsumer streamConsumer)
     {
         executor.streamRequest(requestBuilder()
-                .singleInstanceSelectionPolicy(sidecarInstance)
-                .request(new StreamCdcSegmentRequest(segment, range))
-                .build(), streamConsumer);
+                               .singleInstanceSelectionPolicy(sidecarInstance)
+                               .request(new StreamCdcSegmentRequest(segment, range))
+                               .build(), streamConsumer);
     }
 
     /**
@@ -576,13 +576,11 @@ public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExt
      */
     public <T> CompletableFuture<T> reportSchema(SidecarInstance instance)
     {
-        RequestContext context = requestBuilder()
-                .singleInstanceSelectionPolicy(instance)
-                .reportSchemaRequest()
-                .noRetryPolicy()  // {@link NoRetryPolicy} is the preferred behavior here
-                .build();
-
-        return executor.executeRequestAsync(context);
+        return executor.executeRequestAsync(requestBuilder()
+                                            .singleInstanceSelectionPolicy(instance)
+                                            .reportSchemaRequest()
+                                            .noRetryPolicy()  // {@link NoRetryPolicy} is the preferred behavior here
+                                            .build());
     }
 
     /**
@@ -594,8 +592,8 @@ public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExt
     public CompletableFuture<AllServicesConfigPayload> allServicesConfig()
     {
         return executor.executeRequestAsync(requestBuilder()
-                       .request(new AllServicesConfigRequest())
-                       .build());
+                                            .request(new AllServicesConfigRequest())
+                                            .build());
     }
 
     /**
@@ -608,8 +606,8 @@ public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExt
     public CompletableFuture<UpdateCdcServiceConfigPayload> updateCdcServiceConfig(Service service, Map<String, String> config)
     {
         return executor.executeRequestAsync(requestBuilder()
-                       .request(new UpdateServiceConfigRequest(service, new UpdateCdcServiceConfigPayload(config)))
-                       .build());
+                                            .request(new UpdateServiceConfigRequest(service, new UpdateCdcServiceConfigPayload(config)))
+                                            .build());
     }
 
     /**
@@ -620,8 +618,8 @@ public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExt
     public CompletableFuture<Void> deleteCdcServiceConfig(Service service)
     {
         return executor.executeRequestAsync(requestBuilder()
-                       .request(new DeleteServiceConfigRequest(service))
-                       .build());
+                                            .request(new DeleteServiceConfigRequest(service))
+                                            .build());
     }
 
     /**
--- a/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
+++ b/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
@@ -24,7 +24,6 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -565,6 +564,28 @@ public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExt
                 .singleInstanceSelectionPolicy(sidecarInstance)
                 .request(new StreamCdcSegmentRequest(segment, range))
                 .build(), streamConsumer);
+    }
+
+    /**
+     * Sends a request to trigger an immediate, synchronous schema
+     * conversion and report on the specified instance of the Sidecar
+     * regardless of the periodic task schedule or status
+     *
+     * @param instance the {@link SidecarInstance} to receive the request
+     * @return a {@link CompletableFuture} for the request
+     */
+    public CompletableFuture<Void> reportSchema(SidecarInstance instance)
+    {
+        // Create an instance of {@link RequestContext.Builder} using its
+        // constructor instead of the {@link this.requestBuilder()} method,
+        // since {@link NoRetryPolicy} is the preferred behavior here
+
+        RequestContext context = new RequestContext.Builder()
+                .singleInstanceSelectionPolicy(instance)
+                .reportSchemaRequest()
+                .build();
+
+        return executor.executeRequestAsync(context);
     }
 
     /**
--- a/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
+++ b/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
@@ -574,15 +574,12 @@ public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExt
      * @param instance the {@link SidecarInstance} to receive the request
      * @return a {@link CompletableFuture} for the request
      */
-    public CompletableFuture<Void> reportSchema(SidecarInstance instance)
+    public <T> CompletableFuture<T> reportSchema(SidecarInstance instance)
     {
-        // Create an instance of {@link RequestContext.Builder} using its
-        // constructor instead of the {@link this.requestBuilder()} method,
-        // since {@link NoRetryPolicy} is the preferred behavior here
-
-        RequestContext context = new RequestContext.Builder()
+        RequestContext context = requestBuilder()
                 .singleInstanceSelectionPolicy(instance)
                 .reportSchemaRequest()
+                .noRetryPolicy()  // {@link NoRetryPolicy} is the preferred behavior here
                 .build();
 
         return executor.executeRequestAsync(context);
--- a/client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
+++ b/client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
@@ -34,7 +34,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -114,6 +113,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatException;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
+import static org.assertj.core.api.Assertions.assertThatNoException;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.Assertions.fail;
 
@@ -1607,7 +1607,7 @@ abstract class SidecarClientTest
     public void testListCdcSegments() throws ExecutionException, InterruptedException, JsonProcessingException
     {
         List<CdcSegmentInfo> segments = Arrays.asList(new CdcSegmentInfo("commit-log1", 100, 100, true, 1732148713725L),
-                new CdcSegmentInfo("commit-log2", 100, 10, false, 1732148713725L));
+                                                      new CdcSegmentInfo("commit-log2", 100, 10, false, 1732148713725L));
         ListCdcSegmentsResponse listSegmentsResponse = new ListCdcSegmentsResponse("localhost", 9043, segments);
         ObjectMapper mapper = new ObjectMapper();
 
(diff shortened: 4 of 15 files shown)


CHANGES.txt
```diff
@@ -8,7 +9,7 @@
  * Sidecar schema initialization can be executed on multiple thread (CASSSIDECAR-200)
  * Make sidecar operations resilient to down Cassandra nodes (CASSSIDECAR-201)
  * Fix Cassandra instance not found error (CASSSIDECAR-192)
- * Implemented Schema Reporter for Integration with DataHub (CASSSIDECAR-191)
+ * Implement Schema Reporter for Integration with DataHub (CASSSIDECAR-191)
```
yifan-c: Do not change the existing entries in the `CHANGES.txt`

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
```diff
@@ -63,6 +63,9 @@ public class BasicPermissions
     public static final Permission READ_OPERATIONAL_JOB = new DomainAwarePermission("OPERATIONAL_JOB:READ", OPERATION_SCOPE);
     public static final Permission DECOMMISSION_NODE = new DomainAwarePermission("NODE:DECOMMISSION", OPERATION_SCOPE);
 
+    // Permissions related to Schema Reporting
+    public static final Permission REPORT_SCHEMA = new DomainAwarePermission("SCHEMA:REPORT", CLUSTER_SCOPE);
```
yifan-c: I do not think an extra permission is required. 
The permission needed in order to publish to DataHub is `SCHEMA:READ`, which already exists. cc: @sarankk

server/src/main/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandler.java
```diff
@@ -0,0 +1,101 @@
...
+    @NotNull
+    private final SchemaReporter schemaReporter;
+
+    /**
+     * Constructs a new instance of {@link ReportSchemaHandler} using the provided instances
+     * of {@link InstanceMetadataFetcher}, {@link ExecutorPools}, and {@link SchemaReporter}
+     *
+     * @param metadata the metadata fetcher
+     * @param executor executor pools for blocking executions
+     * @param reporter executor pools for blocking executions
```
yifan-c: copy-paste error

server/src/main/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandler.java
```diff
@@ -0,0 +1,101 @@
...
+    protected void handleInternal(@NotNull RoutingContext context,
+                                  @NotNull HttpServerRequest http,
+                                  @NotNull String host,
+                                  @NotNull SocketAddress address,
+                                  @Nullable Void request)
+    {
+        executorPools.service()
+                     .runBlocking(() -> metadataFetcher.runOnFirstAvailableInstance(instance ->
+                            schemaReporter.process(instance.delegate().metadata())))
+                     .onSuccess(context::json)
```
yifan-c: The previous block does not return any value. What does the response json look like?

server/src/main/java/org/apache/cassandra/sidecar/utils/InstanceMetadataFetcher.java
```diff
@@ -97,13 +97,29 @@ public CassandraAdapterDelegate delegate(@NotNull String host) throws NoSuchCass
...
+     * @throws CassandraUnavailableException if all local instances were exhausted
+     */
+    public void runOnFirstAvailableInstance(Consumer<InstanceMetadata> consumer) throws CassandraUnavailableException
+    {
+        callOnFirstAvailableInstance(metadata ->
+        {
+            consumer.accept(metadata);
+            return null;
+        });
+    }
```
yifan-c: Can you remove this method? It is unnecessary. It does not retrieve anything from instance metadata fetcher, who is supposed to "fetch something". I have suggested a different implementation in the new handler w/o using this method.

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -902,7 +912,8 @@ public IdentifiersProvider identifiersProvider(@NotNull InstanceMetadataFetcher
             @NotNull
             protected String initialize()
             {
-                return fetcher.callOnFirstAvailableInstance(i -> i.delegate().storageOperations().clusterName());
+                return fetcher.callOnFirstAvailableInstance(instance ->
+                        instance.delegate().storageOperations().clusterName());
```
yifan-c: Please revert the unrelated change.

server/src/main/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandler.java
```diff
@@ -0,0 +1,101 @@
...
+                     .onFailure(throwable -> processFailure(throwable, context, host, address, request));
+    }
```
yifan-c:
```suggestion
        Metadata metadata = metadataFetcher.callOnFirstAvailableInstance(instance -> instance.delegate().metadata());
        
        executorPools.service()
                     .runBlocking(() -> {
                         schemaReporter.process(metadata);
                     })
                     .onSuccess(v -> context.json(OK_STATUS))
                     .onFailure(cause -> processFailure(cause, context, host, address, request));
```

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
```diff
@@ -0,0 +1,199 @@
...
+        VertxTestContext context = new VertxTestContext();
+        server.start()
+              .onSuccess(result -> context.completeNow())
+              .onFailure(context::failNow);
+        context.awaitCompletion(TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
+    }
+
+    @AfterEach
+    @SuppressWarnings("ResultOfMethodCallIgnored")
+    @SuppressFBWarnings("RV_RETURN_VALUE_IGNORED")
```
yifan-c: remove the unnecessary annotations.

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
```diff
@@ -0,0 +1,199 @@
...
+    void after() throws InterruptedException
+    {
+        CountDownLatch latch = new CountDownLatch(1);
+        server.close()
+              .onSuccess(future -> latch.countDown());
+        latch.await(TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
+    }
+
+    @Test
+    @SuppressWarnings("deprecation")
```
yifan-c: remove

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
```diff
@@ -0,0 +1,199 @@
...
+                  assertThat(response.statusCode())
+                          .isEqualTo(HttpResponseStatus.OK.code());
+                  assertThat(emitter.content())
+                          .isEqualTo(expected);
+                  context.completeNow();
+              }));
+    }
+
+    @Test
+    @SuppressWarnings("deprecation")
```
yifan-c: remove

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+    @AfterEach
+    @SuppressWarnings("ResultOfMethodCallIgnored")
+    @SuppressFBWarnings("RV_RETURN_VALUE_IGNORED")
+    void after() throws InterruptedException
+    {
+        CountDownLatch latch = new CountDownLatch(1);
+        server.close()
+              .onSuccess(future -> latch.countDown());
+        latch.await(TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
```
yifan-c: close the client too.

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    @Test
+    @SuppressWarnings("deprecation")
+    void testSuccess(@NotNull VertxTestContext context) throws IOException
+    {
+        String expected = IOUtils.readFully("/datahub/empty_cluster.json");
+        emitter = new JsonEmitter();
+
+        client.get(server.actualPort(), LOCALHOST, ENDPOINT)
```
yifan-c: add an assertion that `emitter.content` is empty, before making the http request.

server/src/test/java/org/apache/cassandra/sidecar/utils/InstanceMetadataFetcherTest.java
```diff
@@ -21,7 +21,7 @@
 import java.nio.file.Path;
 import java.util.Arrays;
 import java.util.List;
-
+import com.google.common.collect.ImmutableList;
```
yifan-c: Revert the changes in this file once you remove `runOnFirstAvailableInstance`

yifan-c (changes requested):

yifan-c:
Let's keep the patch succinct, i.e. no unrelated changes and add code only when they are definitely required.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
frankgh: The permission is not for someone to read the schema, but for someone to trigger the schema report on demand. So I think the permission is different. Would love to hear input from @sarankk

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
```diff
@@ -598,6 +600,14 @@ public Router vertxRouter(Vertx vertx,
                                     .handler(streamCdcSegmentHandler)
                                     .build();
 
+        // Schema Reporting
+        protectedRouteBuilderFactory.get()
+                                    .router(router)
+                                    .method(HttpMethod.GET)
```
frankgh: The verb should not be GET. This should either be PUT or POST:

- PUT is more suitable for idempotent operations
- POST is more suitable for operations that are not idempotent

Also, this endpoint might be suitable for the operations framework. Something to consider

frankgh (commented):
I left a few comments. I think we also need to add support on the client side to be able to run this new endpoint

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
yifan-c: The sentiment here is to have restraint in adding new verbs. Ideally, it should be a fixed set of verb to avoid operational pain. 
The reason that READ should work here is that the reporter is reading the cassandra schema. When it publishes (i.e. sends requests to DataHub), the authorization should be enforced by the server (DataHub), not the client (Sidecar).

sarankk: I feel having 2 different permissions is better in general, with 1 granting read permission automatically grants them report permission. But in this particular case, since we already have a periodic task to report schema, irrespective of this endpoint, we can use the `enable` flag for schema reporting to control whether we want to allow reporting or not without adding a separate permission for reporting control?

5: I was planning to get to this tomorrow, but see there's confusion about this one.

Changing permissions to the existing `SCHEMA:READ` will be logically equivalent to the following statement:

> "Everyone who is allowed to see cluster schema is also allowed to perform DoS attacks on Sidecar."

Is that actually true?

frankgh: I think we need a separate permission to allow a user trigger a schema reporting. We should not conflate the `SCHEMA:READ` permission with the ability to report schemas

CHANGES.txt
5: Sure.

Is there a specific reason you'd like me to keep this typo I myself accidentally introduced in there forever?

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
yifan-c: Let's also think about the future. We do not want to go down on the route of introducing new verbs for new actions. And this is triggering my concern on it. 

I am fine with letting `PUBLISH` pass. But we should really think about it and be mindful with introducing new verbs.

CHANGES.txt
yifan-c: It is mainly avoid amending the changes log. It is more important to have the git history of each line linked to the correct commit. 
Btw, it is not really a typo. "Implemented something" described the change as clear. 
If you look at the log, there are places where "Adds" and "Add", as well as "Adding" are used. There is no strict rule on the verb's form.

5: Got it, thanks for explaining, reverted the second chage.

(I was suspecting that maybe there's some kind of automation somewhere that depends on `blame` or something…)

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
5: Changed the value to `SCHEMA:PUBLISH`, thanks.

server/src/main/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandler.java
5: Fixed.

server/src/main/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandler.java
5: I changed this to `context.end()`, no difference in behavior.

An empty response with status code 200 is fine here, in my opinion.

server/src/main/java/org/apache/cassandra/sidecar/utils/InstanceMetadataFetcher.java
5: Removed the method, tests, stylistic changes.

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
5: Reverted.

server/src/main/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandler.java
5: Very helpful idea fetching `Metadata` before scheduling on the pool, thanks.

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
5: Removed the first one, the second one has to stay as it's treated as `ERROR` by SpotBugs otherwise.

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
5: Removed.

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
5: Done.

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
5: Added instruction to close the client, changed order to improve performance.

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
5: Added a check (not empty per se, just an opening bracket).

server/src/test/java/org/apache/cassandra/sidecar/utils/InstanceMetadataFetcherTest.java
5: Done.

server/src/main/java/org/apache/cassandra/sidecar/server/MainModule.java
5: Good point, changed everywhere to `PUT`.

server/src/main/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandler.java
```diff
@@ -0,0 +1,103 @@
...
+                                  @NotNull HttpServerRequest http,
+                                  @NotNull String host,
+                                  @NotNull SocketAddress address,
+                                  @Nullable Void request)
+    {
+        Metadata metadata = metadataFetcher.callOnFirstAvailableInstance(instance -> instance.delegate().metadata());
+
+        executorPools.service()
+                     .runBlocking(() -> schemaReporter.process(metadata))
+                     .onSuccess(ignored -> context.end())
```
frankgh: can we respond with a JSON response here? The expectation is that all endpoints return a JSON payload (except for streaming endpoints). I think we can do OK_STATUS
```suggestion
                     .onSuccess(ignored -> context.json(OK_STATUS))
```

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -121,13 +121,16 @@ public final class ApiEndpointsV1
     public static final String LIST_CDC_SEGMENTS_ROUTE = API_V1 + CDC_PATH + "/segments";
     public static final String STREAM_CDC_SEGMENTS_ROUTE = LIST_CDC_SEGMENTS_ROUTE + "/" + SEGMENT_PATH_PARAM;
 
+    // Schema Reporting
+    private static final String REPORT_SCHEMA = "/report-schema";
+    public static final String REPORT_SCHEMA_ROUTE = API_V1 + REPORT_SCHEMA;
```
frankgh, line 126: I'm not 100% convinced about this route. What do you think about this?
```suggestion
    public static final String REPORT_SCHEMA_ROUTE = API_V1 + "/services/schema-reporting/ REPORT_SCHEMA;
```

or something that scopes this feature to the schema reporting service.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -567,6 +566,28 @@ public void streamCdcSegments(SidecarInstance sidecarInstance,
...
+        // Create an instance of {@link RequestContext.Builder} using its
+        // constructor instead of the {@link this.requestBuilder()} method,
+        // since {@link NoRetryPolicy} is the preferred behavior here
+
+        RequestContext context = new RequestContext.Builder()
+                .singleInstanceSelectionPolicy(instance)
+                .reportSchemaRequest()
+                .build();
+
+        return executor.executeRequestAsync(context);
```
frankgh: you can always override the policy like this:
```suggestion
        return executor.executeRequestAsync(requestBuilder()
                                            .singleInstanceSelectionPolicy(instance)
                                            .retryPolicy(new NoRetryPolicy())
                                            .reportSchemaRequest()
                                            .build());
```

server/src/test/java/org/apache/cassandra/sidecar/utils/InstanceMetadataFetcherTest.java
frankgh, line 1: this entire class should be reverted before merge. Changes are unrelated to the PR.

CHANGES.txt
frankgh, line 1: Looks like there is a merge conflict here.

frankgh (commented):
looks good in general, I left some comments

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
yifan-c: I agree. The API should not contain verb. Given the API is to interact with DataHub, I would suggest the following to explicitly refer to DataHub

```
PUT /api/v1/datahub/schemas
```

Meaning, update the schema in datahub. 

The API on the datahub side is `/openapi/entities/v1`. Maybe `entities` also work if we want to define a similar API. However, I think `schemas` provides better clarity in the context of Sidecar.

frankgh: `schema-reporting` would map directly to our configuration in `sidecar.yaml`, so I'd prefer to have some consistency between what's user facing (sidecar.yaml) and the endpoints . WDYT?

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
```diff
@@ -1673,6 +1677,42 @@ public void onError(Throwable throwable)
...
+
+        SidecarInstance instance = instances.get(0);
+
+        CompletableFuture<Void> future = client.reportSchema(instance);
+
+        assertThrows(CompletionException.class, () ->
+                future.whenComplete((ignored, throwable) -> assertNotNull(throwable))
+                      .join());
+    }
+
```
yifan-c: formatting in this new tests

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
```diff
@@ -113,13 +113,23 @@ protected SchemaReporter(@NotNull IdentifiersProvider identifiersProvider,
...
+     *
+     * @param metadata the {@link Metadata} to extract Cassandra schema from
+     */
+    public void process(@NotNull Metadata metadata)
     {
         try (Emitter emitter = emitterFactory.emitter())
         {
-            stream(cluster.getMetadata())
+            stream(metadata)
                     .forEach(ThrowableUtils.consumer(emitter::emit));
```
yifan-c:
```suggestion
            stream(metadata).forEach(ThrowableUtils.consumer(emitter::emit));
```

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
```diff
@@ -1673,6 +1677,42 @@ public void onError(Throwable throwable)
...
+        enqueue(response);
+
+        SidecarInstance instance = instances.get(0);
+
+        CompletableFuture<Void> future = client.reportSchema(instance);
+
+        assertDoesNotThrow(() ->
+                future.whenComplete((ignored, throwable) -> assertNull(throwable))
+                      .join());
```
yifan-c: use assertJ

```suggestion
        SidecarInstance instance = instances.get(0);
        assertThatNoException().isThrownBy(() -> client.reportSchema(instance).get());
```

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
```diff
@@ -1673,6 +1677,42 @@ public void onError(Throwable throwable)
...
+
+        enqueue(response);
+
+        SidecarInstance instance = instances.get(0);
+
+        CompletableFuture<Void> future = client.reportSchema(instance);
+
+        assertThrows(CompletionException.class, () ->
+                future.whenComplete((ignored, throwable) -> assertNotNull(throwable))
+                      .join());
```
yifan-c: Use assertJ and enrich the assertion

```suggestion
        SidecarInstance instance = instances.get(0);
        assertThatThrownBy(() -> client.reportSchema(instance).get())
        .isExactlyInstanceOf(ExecutionException.class)
        .hasCauseInstanceOf(RetriesExhaustedException.class)
        .hasMessageContaining("statusCode=500, statusMessage='Server Error'");
```

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
```diff
@@ -0,0 +1,204 @@
...
+    @SuppressFBWarnings("RV_RETURN_VALUE_IGNORED")
+    void after() throws InterruptedException
+    {
+        CountDownLatch latch = new CountDownLatch(1);
+        server.close()
+              .onSuccess(future -> latch.countDown());
+
+        client.close();
+
+        latch.await(TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
```
yifan-c:
```suggestion
        TestResourceReaper.create().with(server).with(client).close();
```

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
```diff
@@ -0,0 +1,204 @@
...
+        client.put(server.actualPort(), LOCALHOST, ENDPOINT)
+              .expect(ResponsePredicate.SC_OK)
+              .send(context.succeeding(response ->
+              {
+                  assertThat(response.statusCode())
+                          .isEqualTo(HttpResponseStatus.OK.code());
+                  assertThat(emitter.content())
+                          .isEqualTo(expected);
+                  context.completeNow();
+              }));
```
yifan-c: Formatting.

Can you please run `./gradlew idea` to apply the codestyle settings and reformat on all the updated files in this patch? The `CONTINUATION_INDENT_SIZE` is 0.

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
yifan-c: The `schema_reporting` in sidecar.yml is to configure the scheduled task, which is different from the API. 

IMO, `PUT .../schemas` means publishing to the `schemas` resource already. The `publishing` portion is redundant. Having the same term in the configuration and API is not a goal and might be impractical to achieve, since the API scheme and ymal are unrelated.

server/src/main/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandler.java
5: Done.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
5: Added a convenience method for policy overrides following the existing pattern.

CHANGES.txt
5: Rebased.

server/src/test/java/org/apache/cassandra/sidecar/utils/InstanceMetadataFetcherTest.java
5: That was my intention, reverted the overlooked changes.

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
5: Replaced, thanks.

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
5: I would agree that `PUT .../schema` (singular, not plural) is an appropriate name if we provide the schema we intent to persist. But we're not putting schema in. We're triggering an event that causes the schema to be internally produced and reported. Using `PUT .../schema-report` or `PUT .../schema-report-request` would probably follow the general idea of REST better. What do you think?

yifan-c: In a RESTful design, the resource in the plural form is a more popular convention. The API is to publishing/inserting a schema entity in the Datahub's `entities` resource. The intent of publishing is expressed by the HTTP verb (`PUT`. TBH, it should be `POST`, but I am fine with `PUT`), and the receiver is the resource collection, i.e. `schemas`. `datahub` in the API is to clarify the external service.

As mentioned in my previous comment, in DataHub's term, they are `entities`. But in the consideration of the sidecar clients, it is clearer with calling them `schemas`.

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
5: Done.

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
5: I get your point, can you specify the exact string you want me put in there, so there's no confusion.

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
5: I've changed these a lot, let me know if the formatting still needs change.

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
5: Done.

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
5: Running `./gradlew idea` doesn't seem to update any files.

Just to confirm, you'd like me to apply automatic formatting to all files touched in this PR?

This will go against your desire to _"keep the patch succinct, i.e. no unrelated changes"_, as the formatting currently seems to be pretty inconsistent…

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
5: It's way easier to read, discuss, *and* put breakpoints in this way, can we keep it as is, if not violating any explicit convention?

frankgh (approved):
+1

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
yifan-c: Just to clarify.

1. run `./gradlew idea` to _apply_ the codestyle settings. It does not reformat, but update your project settings. 
2. reformat on _all the updated files in this patch_

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
yifan-c: If in 2 lines, the extra indentation at the second line should be removed. 

```suggestion 
            stream(metadata)
            .forEach(ThrowableUtils.consumer(emitter::emit));
```

server/src/test/java/org/apache/cassandra/sidecar/routes/ReportSchemaHandlerTest.java
5: Applied the required identations to all files affected by this PR.

PR #200, opened 2025-02-20
CASSSIDECAR-208: Guice modularization

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-208


client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -124,7 +124,7 @@ public final class ApiEndpointsV1
 
     public static final String CONNECTED_CLIENT_STATS_ROUTE = API_V1 + CASSANDRA + "/stats/connected-clients";
 
-    public static final String OPERATIONAL_JOBS = "/operational-jobs";
+    private static final String OPERATIONAL_JOBS = "/operational-jobs";
```
yifan-c, line 136: this is not the route. It is the intermediate string to define routes that share the same prefix. Such intermediate string does not need to be exposed.

integration-framework/build.gradle
```diff
@@ -76,11 +76,7 @@ dependencies {
     // Exposed to the test project
     api("io.vertx:vertx-junit5:${project.vertxVersion}")
     // The server itself
-    api(project(path: ":server")) {
-        exclude(group: 'org.apache.logging.log4j')
-        exclude(group: 'org.slf4j')
-        exclude(group: 'ch.qos.logback')
-    }
```
yifan-c, line 83: Excluding the dependencies loses the test log of sidecar in the pacakge

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraSchemaRouteIntegrationTest.java
```diff
@@ -43,9 +43,10 @@ protected void initializeSchemaForTest()
     void testListKeyspaces()
     {
         String testRoute = "/api/v1/schema/keyspaces";
-        SchemaResponse response = getBlocking(trustedClient().get(server.actualPort(), "localhost", testRoute)
-                                                             .expect(ResponsePredicate.SC_OK)
-                                                             .send())
+        SchemaResponse response = getBlocking(trustedClient()
+                                              .get(server.actualPort(), "localhost", testRoute)
+                                              .send()
+                                              .expecting(HttpResponseExpectation.SC_OK))
```
yifan-c, line 49: `expecting` is replacing the deprecated `expect`

server-common/src/main/java/org/apache/cassandra/sidecar/db/schema/TableSchema.java
```diff
@@ -37,4 +37,10 @@ protected boolean exists(@NotNull Metadata metadata)
             return false;
         return ksMetadata.getTable(tableName()) != null;
     }
+
+    @Override
+    public String toString()
+    {
+        return keyspaceName() + '.' + tableName();
```
yifan-c, line 44: The string value is used when registering table schema fails

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -117,81 +69,13 @@ public void ensureInitialized()
...
-        {
-            return clusterLease.isClaimedByLocalSidecar();
-        }
-        return true;
-    }
-
-    /**
-     * Initializer that retries until sidecar schema is initialized. Once initialized, it un-schedules itself.
-     */
-    private class SidecarSchemaInitializer implements PeriodicTask
```
yifan-c, line 146: `SidecarSchemaInitializer` is moved to its own class file.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemAuthSchema.java
```diff
@@ -106,4 +104,10 @@ protected void ensureSchemaAvailable() throws SchemaUnavailableException
             throw new SchemaUnavailableException(keyspaceName(), IDENTITY_TO_ROLE_TABLE);
         }
     }
+
+    @Override
+    public String toString()
+    {
+        return keyspaceName();
```
yifan-c, line 111: This guy has multiple tables. Therefore showing the keyspace name only.

server/src/main/java/org/apache/cassandra/sidecar/handlers/RouteBuilder.java
```diff
@@ -127,47 +95,59 @@ public AccessProtectedRouteBuilder setBodyHandler(Boolean setBodyHandler)
...
+    }
+
     /**
      * Builds an authorized route. Adds {@link io.vertx.ext.web.handler.AuthorizationHandler} at top of handler
      * chain if access control is enabled.
      */
-    public void build()
+    public VertxRoute build()
     {
-        Objects.requireNonNull(router, "Router must be set");
```
yifan-c: router must present now in the `VertxRoute` lambda

CONTRIBUTING.md
```diff
@@ -128,57 +159,94 @@ The `service` worker pool has the name `sidecar-worker-pool`.
...
+threads that are managed internal by vertx. For example, the below code snippet runs `action()` after `delayMillis`.
 
 ```java
-logger.debug("Retrying streaming after {} millis", millis);
-executorPools.service().setTimer(millis, t -> acquireAndSend(context, filename, fileLength, range, startTime));
+executorPools.service().setTimer(delayMillis, timerId -> action());
 ```
 
-### <a href="guice"></a>Dependency Injection
+Similarly, a simple periodic task can be schedule with the following code. The `action()` is scheduled to run after
```
frankgh:
```suggestion
Similarly, a simple periodic task can be scheduled with the following code. The `action()` is scheduled to run after 
```

CONTRIBUTING.md
```diff
@@ -128,57 +159,94 @@ The `service` worker pool has the name `sidecar-worker-pool`.
...
+executorPools.internal().setPeriodic(initialDelayMillis, delayMillis, timerId -> action());
+```
+Note that such periodic task is triggered whenever the scheduled time has arrived, consider the equivalent 
+`ScheduledExecutorService#scheduleAtFixedRate`. It is possible to have concurrent runs, depending on the delay parameters. 
+It might not be the desired scheduling behavior for the use case. If serial execution sequence is wanted, check out the 
+scheduling mechanism described in [Advanced periodic task scheduling](#advanced-periodic-scheduling)
 
-When a different implementation can be provided in the future, prefer using an interface and a default implementation
-that can be later switched transparently without affecting the classes that depend on it.
+#### <a name="advanced-periodic-scheduling">Advanced periodic task scheduling
```
frankgh: NIT: uppercase first letter for the title
```suggestion
#### <a name="advanced-periodic-scheduling">Advanced Periodic Task Scheduling
```

CONTRIBUTING.md
```diff
@@ -128,57 +159,94 @@ The `service` worker pool has the name `sidecar-worker-pool`.
...
-Prefer creating concrete classes over utility methods. The concrete classes can be managed as a
-[Singleton](https://en.wikipedia.org/wiki/Singleton_pattern) if they do not have external dependencies that might
-change their behavior based on the configuration.
+`PeriodicTaskExecutor` provides the scheduling behavior that is similar to the hybrid of `ScheduledExecutorService#scheduleAtFixedRate`
+and `ScheduledExecutorService#scheduleAtFixedDelay`. The unit of execution is `PeriodicTask`.
 
-Here's an example of a class being managed by Guice.
+A `PeriodicTask` is guaranteed to be executed in serial by `PeriodicTaskExecutor`, as if such task is executed from 
+a single thread. Memory consistency is ensured, i.e. the effect of write from the last run is visible to the current run.
+The interval between the consecutive runs is adjusted based on the duration taken by the last run. This way, its behavior
```
frankgh, line 190: I think we should not here that the duration is determined by the PeriodicTask, if the future returns immediately, but there are background tasks scheduled within the PeriodicTask, the duration of the actual task is not known to the scheduler

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
```diff
@@ -423,8 +423,8 @@ public InstanceMetadataImpl build()
 
             if (storageDir == null)
             {
-                Objects.requireNonNull(dataDirs, "dataDirs are required when storageDir is not configured");
-                Preconditions.checkArgument(!dataDirs.isEmpty(), "dataDirs are required when storageDir is not configured");
+                Preconditions.checkArgument(dataDirs != null && !dataDirs.isEmpty(),
```
frankgh: I think it's better to have an NPE for null input and IllegalArgumentException for empty. Also this change is unrelated to this PR, can you please revert?

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreSliceDatabaseAccessor.java
```diff
@@ -42,10 +42,9 @@ public class RestoreSliceDatabaseAccessor extends DatabaseAccessor<RestoreSlices
 
     @Inject
     protected RestoreSliceDatabaseAccessor(SidecarSchema sidecarSchema,
-                                           RestoreSlicesSchema restoreSlicesSchema,
                                            CQLSessionProvider cqlSessionProvider)
     {
-        super(restoreSlicesSchema, cqlSessionProvider);
+        super(sidecarSchema.tableSchema(RestoreSlicesSchema.class), cqlSessionProvider);
```
sarankk, line 47: Nit: should we call this method `tableSchemaByClass` ?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchemaInitializer.java
```diff
@@ -0,0 +1,158 @@
...
+        {
+            LOGGER.warn("Failed to initialize schema. Retry in {}", delay(), ex);
+            if (ex instanceof CassandraUnavailableException) // not quite expected here according to the schedule decision, but still check for it
+            {
+                return; // do not count Cassandra unavailable as failure
+            }
+            else if (ex instanceof SidecarSchemaModificationException)
+            {
+                LOGGER.warn("Failed to modify schema", ex);
+                schemaMetrics.failedModifications.metric.update(1);
```
sarankk, line 129: Since we are initializing, we should use `failedInitializations` here?

server/src/main/java/org/apache/cassandra/sidecar/handlers/CassandraHealthHandler.java
```diff
@@ -50,7 +50,7 @@ public class CassandraHealthHandler extends AbstractHandler<Void>
      * @param validator       a validator instance to validate Cassandra-specific input
      */
     @Inject
-    protected CassandraHealthHandler(InstanceMetadataFetcher metadataFetcher,
+    public CassandraHealthHandler(InstanceMetadataFetcher metadataFetcher,
```
sarankk, line 53: Nit: Incorrect indentation

server/src/main/java/org/apache/cassandra/sidecar/handlers/RingHandler.java
```diff
@@ -16,7 +16,7 @@
  * limitations under the License.
  */
 
-package org.apache.cassandra.sidecar.routes;
+package org.apache.cassandra.sidecar.handlers;
```
sarankk, line 19: My only concern with this package name change is, currently we have just API route specific handlers here, other speciality handlers in their respective packages, which keeps handlers more organized. Naming it generic, might lead to all handlers being put here. How about moving `handlers` package under `routes` package?

server/src/main/java/org/apache/cassandra/sidecar/routes/VertxRoute.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package org.apache.cassandra.sidecar.routes;
+
+import java.util.function.Consumer;
+
+import io.vertx.ext.web.Router;
+
+/**
+ * Defines a vert route.
```
sarankk: Missing spelling. `Defines a Vert.x route`

server/src/main/java/org/apache/cassandra/sidecar/modules/ApiModule.java
```diff
@@ -0,0 +1,166 @@
...
+    {
+        return SidecarLoggerHandler.create(LoggerHandler.create());
+    }
+
+    @ProvidesIntoMap
+    @KeyClassMapKey(VertxRouteMapKeys.GlobalUtilityHandlerKey.class)
+    VertxRoute globalUtilityHandler(SidecarConfiguration sidecarConfiguration,
+                                    LoggerHandler loggerHandler)
+    {
+        return VertxRoute.create(router -> {
```
sarankk, line 136: Curious, why are we not building this route with `RouterBuilder.Factory` ?

server/src/main/java/org/apache/cassandra/sidecar/modules/multibindings/RouteClassKey.java
```diff
@@ -0,0 +1,66 @@
...
+        try
+        {
+            return (HttpMethod) classKey.getDeclaredField(HTTP_METHOD_FIELD_NAME).get(null);
+        }
+        catch (IllegalAccessException | NoSuchFieldException | ClassCastException e)
+        {
+            throw new RuntimeException(e);
+        }
+    }
+
```
sarankk, line 47: Nit: missing javadoc

server/src/main/java/org/apache/cassandra/sidecar/modules/multibindings/RouteClassKey.java
```diff
@@ -0,0 +1,66 @@
...
+     * Reads the static {@link HttpMethod} defined in the class
+     *
+     * @param classKey class to read
+     * @return http method
+     */
+    static HttpMethod httpMethod(Class<? extends RouteClassKey> classKey)
+    {
+        try
+        {
+            return (HttpMethod) classKey.getDeclaredField(HTTP_METHOD_FIELD_NAME).get(null);
```
sarankk: Instead of static members can we have methods for `httpMethod` and `routeURI` in `RouteClassKey`? to make sure all `RouteClassKey` will have these 2 set. Currently `RouteClassKey.httpMethod` can throw `NoSuchFieldException` if we forget to set `HTTP_METHOD`

server/src/main/java/org/apache/cassandra/sidecar/modules/ApiModule.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    @ProvidesIntoMap
+    @KeyClassMapKey(VertxRouteMapKeys.GlobalErrorHandlerKey.class)
+    VertxRoute globalErrorHandler(ErrorHandler errorHandler)
+    {
+        return VertxRoute.create(router -> {
+            router.route()
+                  .path(ApiEndpointsV1.API + "/*")
+                  .order(RoutingOrder.HIGHEST.order)
```
sarankk: Currently global error handler is lower in priority than `ChainAuthHandler`, Having `ApiModule` before `AuthModule` changes this. Also should we rearrange this to have `AuthModule` before `ApiModule`? (we can have one more `VertxSetupModule` for vertx and router setup and then `AuthModule` and then `ApiModule`?)

server/src/main/java/org/apache/cassandra/sidecar/modules/CassandraOperationsModule.java
```diff
@@ -0,0 +1,189 @@
...
+                      .handler(nodeSettingsHandler)
+                      .build();
+    }
+
+    @ProvidesIntoMap
+    @KeyClassMapKey(VertxRouteMapKeys.AllKeyspacesSchemaRouteKey.class)
+    VertxRoute cassandraSchemaRoute(RouteBuilder.Factory factory,
+                                    SchemaHandler schemaHandler)
+    {
+        return factory.builderForRoute()
```
sarankk: Nit: Getting schema information doesn't seem like a operation related. We can leave it as is for now, but later we can move it to `CassandraModule` or `CassandraInfoModule`?. Same with ring, token, gossip endpoint

server/src/main/java/org/apache/cassandra/sidecar/modules/HealthCheckModule.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    @Deprecated
+    @ProvidesIntoMap
+    @KeyClassMapKey(VertxRouteMapKeys.CassandraHealthRouteKey.class)
+    VertxRoute cassandraHealthRoute(RouteBuilder.Factory factory,
+                                    CassandraHealthHandler cassandraHealthHandler)
+    {
+        // Backwards compatibility for the Cassandra health endpoint
+        return factory.builderForUnprotectedRoute()
```
sarankk: Nit: the phrase `Unprotected` reads like it is unauthenticated and unauthorized.  Whereas these endpoints are only unauthorized, they are still authenticated. Should we rename to `builderForUnauthorizedRoute` ?

server/src/main/java/org/apache/cassandra/sidecar/modules/SidecarModules.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public static List<Module> all(@Nullable Path confPath)
+    {
+        // To prevent unexpected circular dependency chains in your code, we recommend that you disable Guice's circular proxy feature.
+        return List.of(Modules.disableCircularProxiesModule(),
+                       new ApiModule(),
+                       new AuthModule(),
+                       new CassandraOperationsModule(),
+                       new CdcModule(),
+                       new ConfigurationModule(confPath),
```
sarankk, line 52: Nit: If we have `VertxSetupModule` we can move `ConfigurationModule` after it for easier reading

server/src/main/java/org/apache/cassandra/sidecar/modules/multibindings/KeyClassMapKey.java
```diff
@@ -0,0 +1,42 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import com.google.inject.multibindings.MapKey;
+import com.google.inject.multibindings.ProvidesIntoMap;
+
+/**
+ * Allows {@literal @}{@link ProvidesIntoMap} to specify a {@link ClassKey} class as map key.
+ * The class must extends from {@link ClassKey}
+ * <p>The annotation name {@link KeyClassMapKey} is read as "map key type is of {@link ClassKey} class"
```
sarankk, line 33: Nit: Shall we name this class to `ClassKeyMapKey` ? that way name matches this comment

server/src/main/java/org/apache/cassandra/sidecar/modules/RestoreJobModule.java
```diff
@@ -0,0 +1,195 @@
...
+public class RestoreJobModule extends AbstractModule
+{
+    @Override
+    protected void configure()
+    {
+        // Leverage TypeLiteral to preserve the lower bound type Key at the runtime
+        TypeLiteral<Class<? extends ClassKey>> keyType = new TypeLiteral<>() {};
+        TypeLiteral<PeriodicTask> valueType = new TypeLiteral<>() {};
+        MapBinder<Class<? extends ClassKey>, PeriodicTask> periodicTaskMapBinder = MapBinder.newMapBinder(binder(), keyType, valueType);
+        periodicTaskMapBinder.addBinding(PeriodicTaskMapKeys.RestoreJobDiscovererKey.class).to(RestoreJobDiscoverer.class);
```
sarankk, line 68: Curious, why do we inject for these periodic tasks and bind them this way. Where as for `HealthCheckPeriodicTask` we provide with `ProvidesIntoMap`

sarankk (commented):
Thanks Yifan, modules look neat and well organized. Left some comments

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreSliceDatabaseAccessor.java
yifan-c: it is to get the instance using the class to look up.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchemaInitializer.java
yifan-c: The original implementation is reporting as `failedModifications`. I would like to keep it as-is.

server/src/main/java/org/apache/cassandra/sidecar/handlers/CassandraHealthHandler.java
yifan-c: Done

server/src/main/java/org/apache/cassandra/sidecar/routes/VertxRoute.java
yifan-c: Done

server/src/main/java/org/apache/cassandra/sidecar/modules/ApiModule.java
yifan-c: The router builder is used to define the APIs (route with endpoint and verb). For this case, it only needs to mount the handler to the root route. The handlers does not target any specific API. I, therefore, named them global utility handlers.

server/src/main/java/org/apache/cassandra/sidecar/modules/multibindings/RouteClassKey.java
yifan-c: They can be missed, since there is not enforcement from compiler. However, it is the tradeoff made regarding the design.
Having instance methods means those keys are instantiated, which should be avoided, since it just uses the type/Class as keys. 
If any of the fields are missed, the test should fail. Not at the compile time, but it should be caught before a patch can be merged.

server/src/main/java/org/apache/cassandra/sidecar/modules/ApiModule.java
yifan-c: Not sure if I understand. The error handler is on the route `/api/v1/*`. The auth chain handler is on the root route. They are mounted on _different_ routes. 
Regarding the evaluation order, the auth chain handler is evaluated the first, then the request handlers, then the failure handler if failure happens.

server/src/main/java/org/apache/cassandra/sidecar/modules/CassandraOperationsModule.java
yifan-c: Cassandra related operations, i.e. trigger task or retrieving information, are grouped in the same module. If splitting the module up helps with readability, we can do that in the future.

server/src/main/java/org/apache/cassandra/sidecar/modules/HealthCheckModule.java
yifan-c: sounds good! I did not put much thought on the name.

server/src/main/java/org/apache/cassandra/sidecar/modules/SidecarModules.java
yifan-c: For this PR, I would like to stick to the idea of group bindings by feature. VertxSetup is not a feature. 

Btw, the order of the modules in the list does not affect bindings. They are just sorted alphabetically to look neat.

server/src/main/java/org/apache/cassandra/sidecar/modules/multibindings/KeyClassMapKey.java
yifan-c: The full name was `ClassKey Class MapKey`, map key using the class of `ClassKey`. There are 2 "Class" in the name, so I dropped one of them. It does not use the `ClassKey` instance, but its class. Therefore, the later "Class` is reserved.

server/src/main/java/org/apache/cassandra/sidecar/modules/RestoreJobModule.java
yifan-c: They are equivalent. I wanted to demo that there are different ways to add bindings. Plus, binding to a class (that is annotated with `@Inject`) has simpler code.

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/ReloadingJwtAuthenticationHandler.java
```diff
@@ -174,6 +173,7 @@ public void execute(Promise<Void> promise)
             if (!jwtParameters.enabled())
             {
                 delegateHandler.set(null);
+                promise.complete();
```
frankgh, line 176: good catch

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
```diff
@@ -61,6 +63,13 @@ public SchemaReportingTask(@NotNull SidecarConfiguration configuration,
         this.reporter = reporter;
     }
 
+    @Override
+    public void deploy(Vertx vertx, PeriodicTaskExecutor executor)
+    {
+        // TODO: react on ON_CASSANDRA_CQL_READY instead? When any CQL connection is ready, cluster metadata should be available from session
```
frankgh, line 69: this actually sounds like a bug. If your Sidecar instance manages multiple Cassandra instances, and your sidecar instance happens to be the cluster leaseholder, this task will never schedule and you will never perform the schema reporting unless another Sidecar instance takes over the cluster leadership.

CHANGES.txt
```diff
@@ -1,5 +1,6 @@
 0.1.0
 -----
+ * Guice modularization (CASSSIDECAR-208)
```
frankgh, line 3: this feature should go under 0.2.0 instead

server/src/main/java/org/apache/cassandra/sidecar/modules/guice-best-practice.md
```diff
@@ -0,0 +1,117 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+# Guice Best Practices in Sidecar
+
+First of all, refer to [Guice Wiki](https://github.com/google/guice/wiki/BestPractices) for the best practices in general.
+
+There are certain project-specific best practices we advocate for Apache Sidecar. This article capture those items in the below.
```
frankgh:
```suggestion
There are certain project-specific best practices we advocate for Apache Cassandra Sidecar. This article capture those items in the below. 
```

server/src/main/java/org/apache/cassandra/sidecar/modules/guice-best-practice.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Multi-bindings for Plugin Extension
+
+[Multi-bindings](https://github.com/google/guice/wiki/Multibindings) is a Guice feature to support the plugin-type architecture,
+where multiple modules contribute components to plug into the system.
+
+The applications of multi-bindings in Sidecar are, for example,
+1. Routes definition
+2. Periodic tasks deployment
+3. Sidecar internal schemas creation
```
frankgh: I don't think this is scoped to sidecar internal schemas only, for example system_auth is also managed by multi-bindings, as a read-only table

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobManagerGroup.java
```diff
@@ -57,18 +56,12 @@ public class RestoreJobManagerGroup
...
-                                  RestoreProcessor restoreProcessor,
-                                  RestoreJobDiscoverer jobDiscoverer,
-                                  RingTopologyRefresher ringTopologyRefresher)
+                                  RestoreProcessor restoreProcessor)
     {
         this.restoreJobConfig = configuration.restoreJobConfiguration();
         this.restoreProcessor = restoreProcessor;
         this.executorPools = executorPools;
         initializeManagers(instancesMetadata);
-        periodicTaskExecutor.schedule(jobDiscoverer);
```
frankgh, line 69: nice 👍

server/src/test/java/org/apache/cassandra/sidecar/routes/sstableuploads/BaseUploadsHandlerTest.java
frankgh, line 1: all the tests under the `org.apache.cassandra.sidecar.routes` package should move to the `org.apache.cassandra.sidecar.handlers` package

frankgh (commented):
What an improvement! thanks for working on this, this is very well-thought. I appreciate the effort

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -49,64 +35,30 @@
...
-        this.periodicTaskExecutor = periodicTaskExecutor;
         this.schemaKeyspaceConfiguration = config.serviceConfiguration().schemaKeyspaceConfiguration();
         this.sidecarInternalKeyspace = sidecarInternalKeyspace;
-        this.cqlSessionProvider = cqlSessionProvider;
-        this.metrics = metrics;
-        this.clusterLease = clusterLease;
         if (this.schemaKeyspaceConfiguration.isEnabled())
         {
-            configureSidecarServerEventListeners();
+            vertx.eventBus().localConsumer(ON_SIDECAR_SCHEMA_INITIALIZED.address(), ignored -> isInitialized.set(true));
```
frankgh: should we use `EventBusUtils.onceLocalConsumer` here?

server/src/test/java/org/apache/cassandra/sidecar/server/ServerTest.java
```diff
@@ -229,34 +232,15 @@ void updateTrafficShapingOptionsWithNonZeroListenPort()
...
-                      context.completeNow();
-                  }
-                  else
-                  {
-                      context.failNow("Expected server close to take less than 10000 millis, " +
-                                      "but it took " + elapsedTimeMillis + " millis");
-                  }
-              }));
+              .onComplete(context.failing(result -> serverStartFailed.countDown()));
+        Uninterruptibles.awaitUninterruptibly(serverStartFailed);
```
frankgh: should we wait indefinitely here? can we upperbound this?

CONTRIBUTING.md
yifan-c: The duration is the time taken between task start and the promise is completed. It is generally not advised to schedule async actions (during a task execution) that do not take the `promise` into consideration. It can break the memory consistency. On the other hand, if the `promise` is completed at the end of the last async action, (which is the correct way), the duration is known.

server/src/main/java/org/apache/cassandra/sidecar/modules/guice-best-practice.md
yifan-c: I updated the list. Please take another look if it sounds better.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
yifan-c: Yep. Good catch

server/src/test/integration/org/apache/cassandra/sidecar/testing/IntegrationTestModule.java
```diff
@@ -174,6 +179,27 @@ public ScheduleDecision scheduleDecision()
         };
     }
 
+    /**
+     * This is the example of replacing a bound objet. In the resolver below, the ClusterLeaseClaimTask from production code is removed and replaced by
```
frankgh:
```suggestion
     * This is the example of replacing a bound object. In the resolver below, the ClusterLeaseClaimTask from production code is removed and replaced by
```

frankgh (approved):
+1

PR #201, opened 2025-02-25
CASSSIDECAR-160 Add JWT Authentication support in Sidecar

sarankk:


conf/sidecar.yaml
```diff
@@ -175,6 +175,10 @@ access_control:
   enabled: false
   # Supports setting multiple authenticators, request is authenticated if it is authenticated by any of the
   # configured authenticators
+  # Out of the box, Cassandra Sidecar provides following authenticator provider factories
```
yifan-c: Can you also add a paragraph regarding the evaluation order for authentication? According to `ChainAuthHandlerImpl`, it is the first authenticator that passes.

conf/sidecar.yaml
```diff
@@ -193,6 +197,17 @@ access_control:
...
+      # supports for OpenID discovery.
+    - class_name: org.apache.cassandra.sidecar.acl.authentication.JwtAuthenticationHandlerFactory
+      parameters:
+        # Site for sidecar to dynamically retrieve the configuration information of an OpenID provider, without
+        # having to manually configure settings like issuer etc.
+        site: https://authorization.com
+        # Client Id is a unique identifier assigned by OpenID provider. It is used to identity applications/users
+        # trying to connect.
+        client_id: recognized_client_id
+        config_discover_interval: 1m
```
frankgh: I think this is running too frequently. I suggest a default of 1 hour instead. What do you think?
```suggestion
        config_discover_interval: 1h
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/JwtParameterExtractor.java
```diff
@@ -0,0 +1,142 @@
...
+            throw new IllegalArgumentException("JWT parameters can not be null");
+        }
+
+        validateParameterPresence(parameters, SITE_PARAM_KEY);
+        validateParameterPresence(parameters, CLIENT_ID_PARAM_KEY);
+    }
+
+    private void validateParameterPresence(Map<String, String> parameters, String paramKey)
+    {
+        if (!parameters.containsKey(paramKey) || isNullOrEmpty(parameters.get(paramKey)))
```
frankgh: minor NIT, no need to check if it doesn't contain. doing a get should be sufficient
```suggestion
        if (isNullOrEmpty(parameters.get(paramKey)))
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/ReloadingJwtAuthenticationHandler.java
```diff
@@ -0,0 +1,179 @@
...
+                                 delegateHandler.set(handler);
+                                 promise.complete();
+                             })
+                             .onFailure(cause -> {
+                                 LOGGER.error("Error encountered during OpenID discovery", cause);
+                                 promise.fail(cause);
+                             });
+        }
+    }
+}
```
yifan-c: I am a bit worried about the tight coupling between the refresh task and the authenticator. Also such periodic task is created per each `ReloadingJwtAuthenticationHandler` instance, which I do not think it is desired. Although `ReloadingJwtAuthenticationHandler` is effectively singleton, there is no strong guarantee on it. 

Can you move the `OAuth2AuthHandlerGenerateTask` to a dedicated class file and make it a singleton? It can register the "discover" listener/`ReloadingJwtAuthenticationHandler` and update the reference on success.

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/MutualTlsAuthenticationHandler.java
```diff
@@ -54,6 +68,39 @@ public void authenticate(RoutingContext ctx, Handler<AsyncResult<User>> handler)
...
+                            return;
+                        }
+
+                        List<String> identities = extractIdentities(authN.result());
+
+                        if (identities.isEmpty())
+                        {
+                            handler.handle(Future.failedFuture("Missing client identities"));
+                            return;
+                        }
```
yifan-c: I think this case is not tested. Can you add a test case? 
Btw, should it fail with `HttpException` too?

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/ReloadingJwtAuthenticationHandler.java
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+
+    /**
+     * Periodic task to generate {@link OAuth2AuthHandlerImpl} with refreshed configuration from
+     * {@link OpenIDConnectAuth} discover.
+     */
+    private class OAuth2AuthHandlerGenerateTask implements PeriodicTask
+    {
+        public DurationSpec delay()
```
frankgh, line 160:
```suggestion
        @Override
        public DurationSpec delay()
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/ReloadingJwtAuthenticationHandler.java
```diff
@@ -0,0 +1,179 @@
...
+        {
+            return jwtParameterExtractor.configDiscoverInterval();
+        }
+
+        public void execute(Promise<Void> promise)
```
frankgh, line 172:
```suggestion
        @Override
        public void execute(Promise<Void> promise)
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/JwtParameterExtractor.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(JwtParameterExtractor.class);
+    private static final String SITE_PARAM_KEY = "site";
+    private static final String CLIENT_ID_PARAM_KEY = "client_id";
+    private static final String SCOPE_SEPARATOR_PARAM_KEY = "scope_separator";
+    private static final String DEFAULT_SCOPE_SEPARATOR = ",";
+    private static final String SCOPES_SUPPORTED_PARAM_KEY = "scopes_supported";
+    private static final String CONFIG_DISCOVER_INTERVAL_PARAM_KEY = "config_discover_interval";
+    private static final SecondBoundConfiguration DEFAULT_CONFIG_DISCOVER_INTERVAL
+    = SecondBoundConfiguration.parse("1m");
```
frankgh: Can this be 1 hour instead?

server/src/test/java/org/apache/cassandra/sidecar/acl/authentication/JWTAuthenticationHandlerFactoryTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.mock;
+
+/**
+ * Test for {@link JwtAuthenticationHandlerFactory}
+ */
+class JWTAuthenticationHandlerFactoryTest
+{
+    Vertx vertx = Vertx.vertx();
```
yifan-c: close the vertx object.

On a second thought. There is no need to instantiate `vertx` object. It is not required for the sole test in the class. Can you just return a mock instance?

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/JwtRoleProcessorImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    private static final String IDENTITY_KEY = "identity";
+    private static final String IDENTITIES_KEY = "identities";
+    private final IdentityToRoleCache identityToRoleCache;
+
+    public JwtRoleProcessorImpl(IdentityToRoleCache identityToRoleCache)
+    {
+        this.identityToRoleCache = identityToRoleCache;
+    }
+
+    public List<String> processRoles(JsonObject decodedToken)
```
frankgh, line 44:
```suggestion
    @Override
    public List<String> processRoles(JsonObject decodedToken)
```

server/src/test/java/org/apache/cassandra/sidecar/acl/authentication/JwtParameterExtractorTest.java
```diff
@@ -0,0 +1,83 @@
...
+    {
+        JwtParameterExtractor parameterExtractor = new JwtParameterExtractor(Map.of("site", "x" + SITE_SUFFIX,
+                                                                                    "client_id", "id"));
+        assertThat(parameterExtractor.site()).as("Site suffix should be removed").isEqualTo("x");
+    }
+
+    @Test
+    void testCustomScopeSeparator()
+    {
+        JwtParameterExtractor parameterExtractor = new JwtParameterExtractor(Map.of("site", "x.com",
```
yifan-c: hmm.. do we want to put `x.com` here? It is an existing site. Maybe `www.apache.org` instead.

server/src/test/java/org/apache/cassandra/sidecar/acl/authentication/ReloadingJwtAuthenticationHandlerTest.java
```diff
@@ -0,0 +1,60 @@
...
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Test for {@link ReloadingJwtAuthenticationHandler}
+ */
+class ReloadingJwtAuthenticationHandlerTest
+{
+    Vertx vertx = Vertx.vertx();
```
yifan-c: close the vertx object

yifan-c (commented):
Looks good overall.

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/JwtRoleProcessorImpl.java
```diff
@@ -0,0 +1,79 @@
...
+        List<String> roles = new ArrayList<>();
+        for (String identity : identities)
+        {
+            String roleFromIdentity = identityToRoleCache.get(identity);
+            if (roleFromIdentity != null)
+            {
+                roles.add(roleFromIdentity);
+            }
+        }
+        return roles;
```
frankgh:
```suggestion
                String role;
        if ((role = decodedToken.getString(IDENTITY_KEY)) != null)
        {
            return List.of(role);
        }
        if ((role = decodedToken.getString(SUB_KEY)) != null)
        {
            return List.of(role);
        }

        JsonArray identityKeyArray = decodedToken.getJsonArray(IDENTITIES_KEY);
        List<String> roles = new ArrayList<>();
        if (identityKeyArray != null)
        {
            for (String identity : (List<String>) identityKeyArray.getList())
            {
                String roleFromIdentity = identityToRoleCache.get(identity);
                if (roleFromIdentity != null)
                {
                    roles.add(roleFromIdentity);
                }
            }
        }
        return List.copyOf(roles);
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/MutualTlsAuthenticationHandler.java
```diff
@@ -54,6 +68,39 @@ public void authenticate(RoutingContext ctx, Handler<AsyncResult<User>> handler)
...
+                        {
+                            handler.handle(Future.failedFuture(new HttpException(UNAUTHORIZED.code(), authN.cause())));
+                            return;
+                        }
+
+                        List<String> identities = extractIdentities(authN.result());
+
+                        if (identities.isEmpty())
+                        {
+                            handler.handle(Future.failedFuture("Missing client identities"));
```
frankgh: maybe preserve the prior status code?
```suggestion
                            handler.handle(Future.failedFuture(wrapHttpException(HttpResponseStatus.FORBIDDEN, "Client identities are missing")));
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/RoleBasedAuthorizationProvider.java
```diff
@@ -64,29 +59,28 @@ public void getAuthorizations(User user, Handler<AsyncResult<Void>> handler)
...
-        if (identities.isEmpty())
+        if (roles.isEmpty())
         {
-            return Future.failedFuture("Missing client identities");
+            return Future.failedFuture("No cassandra roles found associated with the user");
         }
 
-        Set<Authorization> authorizations = new HashSet<>();
-        for (String identity : identities)
+        for (String role : roles)
```
frankgh, line 69: can role be null? I'm referring to the check in line 71.  I guess it's fine to have the check, but it feels that it would be an error to return null

server/src/test/java/org/apache/cassandra/sidecar/acl/authentication/ReloadingJwtAuthenticationHandlerTest.java
frankgh: close the resource after the test completes.

frankgh (commented):
I have a few comments

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/JwtRoleProcessorImpl.java
sarankk: Thanks modified this a bit, we want to extract role from identity than share identity itself.

server/src/test/java/org/apache/cassandra/sidecar/acl/authentication/ReloadingJwtAuthenticationHandlerTest.java
sarankk: Updated it to mock

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/RoleBasedAuthorizationProvider.java
sarankk: We shouldn't get a null entry in List from roles extracted from user attributes, we discussed to keep this check for safer side

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/ReloadingJwtAuthenticationHandler.java
frankgh: I don't think `ReloadingJwtAuthenticationHandler` can be considered a singleton. You could configure multiple OpenID connect services in theory, and that would be valid. And you will need to independently refresh each of them

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/RoleBasedAuthorizationProvider.java
frankgh: discussed offline and decided to keep the null check to be on the safe side

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/ReloadingJwtAuthenticationHandler.java
sarankk: Discussed offline (We need the task per handler, since the site and client id could be different for different handlers), to have the task as is. But adding OpenID site and client id information to task name, to have 1 task per OpenID provider

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/MutualTlsAuthenticationHandler.java
sarankk: Removed this check, since IdentityExtractor throws if identity cant be extracted

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/AuthorizationWithAdminBypassHandler.java
```diff
@@ -55,13 +53,9 @@ public void handle(RoutingContext ctx)
         }
 
         List<String> identities = extractIdentities(ctx.user());
-        if (identities.isEmpty())
```
sarankk, line 58: With sidecar supporting multiple authenticators, we shouldn't mandate identities for authorization

server/src/test/java/org/apache/cassandra/sidecar/acl/authorization/AuthorizationHandlerTest.java
```diff
@@ -1,51 +0,0 @@
...
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
-
-/**
- * Test for {@link AuthorizationWithAdminBypassHandler}
- */
-class AuthorizationHandlerTest
-{
-    @Test
-    void testMissingIdentity()
```
sarankk, line 38: With multiple authenticators, sidecar should not throw when identities are missing

server/src/main/java/org/apache/cassandra/sidecar/utils/AuthUtils.java
```diff
@@ -62,4 +66,21 @@ public static List<String> extractIdentities(User user)
...
+    /**
+     * Extracts a list of cassandra roles associated with given user.
+     *
+     * @param user User representation in Vertx
+     * @return {@code List} of cassandra roles associated with user
+     */
+    public static List<String> extractCassandraRoles(User user)
+    {
+        // noinspection unchecked
+        JsonArray roles =  user.attributes().getJsonArray(CASSANDRA_ROLES_ATTRIBUTE_NAME, EMPTY_JSON_ARRAY);
```
frankgh: NIT
```suggestion
        JsonArray roles = user.attributes().getJsonArray(CASSANDRA_ROLES_ATTRIBUTE_NAME, EMPTY_JSON_ARRAY);
```

server/src/main/java/org/apache/cassandra/sidecar/utils/AuthUtils.java
```diff
@@ -62,4 +66,21 @@ public static List<String> extractIdentities(User user)
...
+        if (roles == null)
+        {
+            return List.of();
+        }
+        return roles.getList();
```
frankgh: can `roles` be `null`? can we change this to this instead if that's the case?
```suggestion
        return Objects.requireNonNullElse(user.attributes().getJsonArray(CASSANDRA_ROLES_ATTRIBUTE_NAME), EMPTY_JSON_ARRAY).getList();
```

sarankk: It can be `null` if someone stores null for `CASSANDRA_ROLES_ATTRIBUTE_NAME` key

frankgh (approved):
+1 Thanks for the patch

PR #203, opened 2025-02-26
CASSSIDECAR-220: Add CdcRawDirectorySpaceCleaner for tracking and cleaning up the cdc_…

jberragan:
…raw directories

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/config/CdcConfiguration.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/config/CdcConfiguration.java
@@ -18,6 +18,9 @@
 package org.apache.cassandra.sidecar.config;
 
 import org.apache.cassandra.sidecar.common.server.utils.MillisecondBoundConfiguration;
+
+import java.time.Duration;
+
 import org.apache.cassandra.sidecar.common.server.utils.SecondBoundConfiguration;
 
 /**
@@ -31,14 +34,49 @@ public interface CdcConfiguration
     SecondBoundConfiguration segmentHardLinkCacheExpiry();
 
     /**
-     *
      * @return true if cdc feature is enabled
      */
     boolean isEnabled();
 
     /**
-     *
      * @return how frequently CDC configs are to be refreshed
      */
     MillisecondBoundConfiguration cdcConfigRefreshTime();
+
+    /* CdcRawDirectorySpaceCleaner Configuration */
+
+    /**
+     * @return the cadence at which the CdcRawDirectorySpaceCleaner period task should run to check and clean-up old `cdc_raw` log segments.
+     */
+    SecondBoundConfiguration cdcRawDirectorySpaceCleanerFrequency();
+
+    /**
+     * @return `true` if CdcRawDirectorySpaceCleaner should monitor the `cdc_raw` directory and clean up the oldest commit log segments.
+     */
+    boolean enableCdcRawDirectoryRoutineCleanUp();
+
+    /**
+     * @return fallback value for maximum directory size in bytes for the `cdc_raw` directory when can't be read from `system_views.settings` table
+     */
+    long fallbackCdcRawDirectoryMaxSizeBytes();
+
+    /**
+     * @return max percent usage of the cdc_raw directory before CdcRawDirectorySpaceCleaner starts removing the oldest segments.
+     */
+    float cdcRawDirectoryMaxPercentUsage();
+
+    /**
+     * @return the critical time period in seconds that indicates the `cdc_raw` directory is not large enough to buffer this time-window of mutations.
+     */
+    Duration cdcRawDirectoryCriticalBufferWindow();
+
+    /**
+     * @return the low time period in seconds that indicates the `cdc_raw` directory is not large enough to buffer this time-window of mutations.
+     */
+    Duration cdcRawDirectoryLowBufferWindow();
+
+    /**
+     * @return the time period which the CdcRawDirectorySpaceCleaner should cache the cdc_total_space before refreshing.
+     */
+    Duration cacheMaxUsage();
 }
--- a/server/src/main/java/org/apache/cassandra/sidecar/config/CdcConfiguration.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/config/CdcConfiguration.java
@@ -19,8 +19,6 @@ package org.apache.cassandra.sidecar.config;
 
 import org.apache.cassandra.sidecar.common.server.utils.MillisecondBoundConfiguration;
 
-import java.time.Duration;
-
 import org.apache.cassandra.sidecar.common.server.utils.SecondBoundConfiguration;
 
 /**
@@ -56,7 +54,7 @@ public interface CdcConfiguration
     boolean enableCdcRawDirectoryRoutineCleanUp();
 
     /**
-     * @return fallback value for maximum directory size in bytes for the `cdc_raw` directory when can't be read from `system_views.settings` table
+     * @return fallback value for maximum directory size in bytes for the `cdc_raw` directory when can't be read from `system_views.settings` table.
      */
     long fallbackCdcRawDirectoryMaxSizeBytes();
 
@@ -68,15 +66,15 @@ public interface CdcConfiguration
     /**
      * @return the critical time period in seconds that indicates the `cdc_raw` directory is not large enough to buffer this time-window of mutations.
      */
-    Duration cdcRawDirectoryCriticalBufferWindow();
+    SecondBoundConfiguration cdcRawDirectoryCriticalBufferWindow();
 
     /**
      * @return the low time period in seconds that indicates the `cdc_raw` directory is not large enough to buffer this time-window of mutations.
      */
-    Duration cdcRawDirectoryLowBufferWindow();
+    SecondBoundConfiguration cdcRawDirectoryLowBufferWindow();
 
     /**
      * @return the time period which the CdcRawDirectorySpaceCleaner should cache the cdc_total_space before refreshing.
      */
-    Duration cacheMaxUsage();
+    SecondBoundConfiguration cacheMaxUsage();
 }
--- a/server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CdcConfigurationImpl.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CdcConfigurationImpl.java
@@ -70,19 +70,19 @@ public class CdcConfigurationImpl implements CdcConfiguration
     @JsonProperty(value = SEGMENT_HARD_LINK_CACHE_EXPIRY_PROPERTY)
     private SecondBoundConfiguration segmentHardLinkCacheExpiry;
     @JsonProperty(value = CDC_RAW_CLEANER_FREQUENCY_PROPERTY)
-    protected SecondBoundConfiguration cdcRawCleanerFrequency;
+    private SecondBoundConfiguration cdcRawCleanerFrequency;
     @JsonProperty(value = ENABLE_CDC_RAW_CLEANER_PROPERTY)
-    protected boolean enableCdcRawCleaner;
+    private boolean enableCdcRawCleaner;
     @JsonProperty(value = FALLBACK_CDC_RAW_MAX_DIRECTORY_SIZE_BYTES)
-    protected long fallbackCdcRawMaxDirectorySize;
+    private long fallbackCdcRawMaxDirectorySize;
     @JsonProperty(value = CDC_RAW_MAX_DIRECTORY_MAX_PERCENT)
-    protected float cdcRawMaxPercent;
+    private float cdcRawMaxPercent;
     @JsonProperty(value = CDC_RAW_MAX_CRITICAL_BUFFER_WINDOW)
-    protected SecondBoundConfiguration cdcRawCriticalBufferWindow;
+    private SecondBoundConfiguration cdcRawCriticalBufferWindow;
     @JsonProperty(value = CDC_RAW_MAX_LOW_BUFFER_WINDOW)
-    protected SecondBoundConfiguration cdcRawLowBufferWindow;
+    private SecondBoundConfiguration cdcRawLowBufferWindow;
     @JsonProperty(value = CDC_CACHE_MAX_USAGE_DURATION)
-    protected SecondBoundConfiguration cacheMaxUsage;
+    private SecondBoundConfiguration cacheMaxUsage;
 
     public CdcConfigurationImpl()
     {
--- a/server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CdcConfigurationImpl.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CdcConfigurationImpl.java
@@ -17,8 +17,6 @@
  */
 package org.apache.cassandra.sidecar.config.yaml;
 
-import java.time.Duration;
-import java.time.temporal.ChronoUnit;
 import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
@@ -57,13 +55,13 @@ public class CdcConfigurationImpl implements CdcConfiguration
     public static final float DEFAULT_CDC_RAW_MAX_DIRECTORY_MAX_PERCENT = 1.0f;
 
     public static final String CDC_RAW_MAX_CRITICAL_BUFFER_WINDOW = "cdc_raw_critical_buffer_window";
-    public static final Duration DEFAULT_CDC_RAW_MAX_CRITICAL_BUFFER_WINDOW = Duration.of(15, ChronoUnit.MINUTES);
+    public static final SecondBoundConfiguration DEFAULT_CDC_RAW_MAX_CRITICAL_BUFFER_WINDOW = SecondBoundConfiguration.parse("15m");
 
     public static final String CDC_RAW_MAX_LOW_BUFFER_WINDOW = "cdc_raw_low_buffer_window";
-    public static final Duration DEFAULT_CDC_RAW_MAX_LOW_BUFFER_WINDOW = Duration.of(60, ChronoUnit.MINUTES);
+    public static final SecondBoundConfiguration DEFAULT_CDC_RAW_MAX_LOW_BUFFER_WINDOW = SecondBoundConfiguration.parse("60m");
 
     public static final String CDC_CACHE_MAX_USAGE_DURATION = "cdc_raw_cache_max_usage_duration";
-    public static final Duration DEFAULT_CDC_CACHE_MAX_USAGE_DURATION = Duration.of(15, ChronoUnit.MINUTES);
+    public static final SecondBoundConfiguration DEFAULT_CDC_CACHE_MAX_USAGE_DURATION = SecondBoundConfiguration.parse("15m");
 
     @JsonProperty(value = IS_ENABLED_PROPERTY)
     private final boolean isEnabled;
@@ -80,11 +78,11 @@ public class CdcConfigurationImpl implements CdcConfiguration
     @JsonProperty(value = CDC_RAW_MAX_DIRECTORY_MAX_PERCENT)
     protected float cdcRawMaxPercent;
     @JsonProperty(value = CDC_RAW_MAX_CRITICAL_BUFFER_WINDOW)
-    protected Duration cdcRawCriticalBufferWindow;
+    protected SecondBoundConfiguration cdcRawCriticalBufferWindow;
     @JsonProperty(value = CDC_RAW_MAX_LOW_BUFFER_WINDOW)
-    protected Duration cdcRawLowBufferWindow;
+    protected SecondBoundConfiguration cdcRawLowBufferWindow;
     @JsonProperty(value = CDC_CACHE_MAX_USAGE_DURATION)
-    protected Duration cacheMaxUsage;
+    protected SecondBoundConfiguration cacheMaxUsage;
 
     public CdcConfigurationImpl()
     {
@@ -125,9 +123,9 @@ public class CdcConfigurationImpl implements CdcConfiguration
                                 boolean enableCdcRawCleaner,
                                 long cdcRawMaxDirectorySize,
                                 float cdcRawMaxPercent,
-                                Duration cdcRawCriticalBufferWindow,
-                                Duration cdcRawLowBufferWindow,
-                                Duration cacheMaxUsage)
+                                SecondBoundConfiguration cdcRawCriticalBufferWindow,
+                                SecondBoundConfiguration cdcRawLowBufferWindow,
+                                SecondBoundConfiguration cacheMaxUsage)
     {
         this.isEnabled = isEnabled;
         this.cdcConfigRefreshTime = cdcConfigRefreshTime;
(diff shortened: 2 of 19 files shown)


server/src/main/java/org/apache/cassandra/sidecar/config/CdcConfiguration.java
```diff
@@ -28,4 +30,41 @@ public interface CdcConfiguration
...
+     */
+    SecondBoundConfiguration cdcRawDirectorySpaceCleanerFrequency();
+
+    /**
+     * @return `true` if CdcRawDirectorySpaceCleaner should monitor the `cdc_raw` directory and clean up the oldest commit log segments.
+     */
+    boolean enableCdcRawDirectoryRoutineCleanUp();
+
+    /**
+     * @return fallback value for maximum directory size in bytes for the `cdc_raw` directory when can't be read from `system_views.settings` table
```
bbotella: nit: final `.` (dot)

server/src/main/java/org/apache/cassandra/sidecar/db/SystemViewsDatabaseAccessor.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @Nullable
+    public Long getCdcTotalSpaceSetting() throws SchemaUnavailableException
+    {
+        // attempt to parse old 'cdc_total_space_in_mb' prop
+        String cdcTotalSpaceInMb = getSetting(YAML_PROP_PREVIOUS);
+        if (cdcTotalSpaceInMb != null)
+        {
+            return FileUtils.mbStringToBytes(cdcTotalSpaceInMb);
```
bbotella, line 60: What do we mean by previous? Is this being deprecated? Maybe we should throw a deprecation warn if that's the case?

server/src/main/java/org/apache/cassandra/sidecar/server/Server.java
```diff
@@ -307,7 +312,9 @@ protected Future<String> scheduleInternalPeriodicTasks(String deploymentId)
                                                                                       instancesMetadata,
                                                                                       executorPools,
                                                                                       metrics);
+
         periodicTaskExecutor.schedule(healthCheckPeriodicTask);
+        periodicTaskExecutor.schedule(cdcRawDirectorySpaceCleaner);
         vertx.eventBus().localConsumer(ON_SERVER_STOP.address(), message ->
```
bbotella: Please also unschedule the new `cdcRawDirectorySpaceCleaner` periodic task

server/src/main/java/org/apache/cassandra/sidecar/config/CdcConfiguration.java
```diff
@@ -28,4 +30,41 @@ public interface CdcConfiguration
...
+
+    /**
+     * @return max percent usage of the cdc_raw directory before CdcRawDirectorySpaceCleaner starts removing the oldest segments.
+     */
+    float cdcRawDirectoryMaxPercentUsage();
+
+    /**
+     * @return the critical time period in seconds that indicates the `cdc_raw` directory is not large enough to buffer this time-window of mutations.
+     */
+    Duration cdcRawDirectoryCriticalBufferWindow();
```
bbotella: Why using Duration instead of `SecondBoundConfiguration`? Same for `cdcRawDirectoryLowBufferWindow` and `cacheMaxUsage`

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -0,0 +1,437 @@
...
+        return cdcConfiguration.fallbackCdcRawDirectoryMaxSizeBytes();
+    }
+
+    protected void routineCleanUp()
+    {
+        if (!cdcConfiguration.enableCdcRawDirectoryRoutineCleanUp())
+        {
+            LOGGER.debug("Skipping CdcRawDirectorySpaceCleaner: feature is disabled ");
+            return;
+        }
```
bbotella: Instead of having this check here, why not just prevent the task from being scheduled by overriding `scheduleDecision` method? Something like:
```
@Override
public ScheduleDecision scheduleDecision()
{
    return cdcConfiguration.enableCdcRawDirectoryRoutineCleanUp() ? ScheduleDecision.EXECUTE : ScheduleDecision.SKIP;
}
```

server/src/main/java/org/apache/cassandra/sidecar/db/SystemViewsDatabaseAccessor.java
jberragan: `cdc_total_space_in_mb` was removed in Cassandra and replaced with `cdc_total_space` in v5.0+

bbotella (approved):
+1 

Please rebase the conflict in Server

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CdcConfigurationImpl.java
```diff
@@ -41,29 +41,102 @@ public class CdcConfigurationImpl implements CdcConfiguration
...
 
     @JsonProperty(value = IS_ENABLED_PROPERTY)
     private final boolean isEnabled;
     @JsonProperty(value = CONFIGURATION_REFRESH_TIME_PROPERTY)
     private final MillisecondBoundConfiguration cdcConfigRefreshTime;
     @JsonProperty(value = SEGMENT_HARD_LINK_CACHE_EXPIRY_PROPERTY)
     private SecondBoundConfiguration segmentHardLinkCacheExpiry;
-
+    @JsonProperty(value = CDC_RAW_CLEANER_FREQUENCY_PROPERTY)
+    protected SecondBoundConfiguration cdcRawCleanerFrequency;
```
jyothsnakonisa: Can all the properties be private and final like the others in the class if possible

server/src/main/java/org/apache/cassandra/sidecar/db/SystemViewsDatabaseAccessor.java
```diff
@@ -0,0 +1,82 @@
...
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Database Accessor that queries cassandra to get information maintained under system_auth keyspace.
+ */
+@Singleton
+public class SystemViewsDatabaseAccessor extends DatabaseAccessor<SystemViewsSchema>
+{
+    static final String YAML_PROP_PREVIOUS = "cdc_total_space_in_mb";
+    static final String YAML_PROP_CURRENT = "cdc_total_space";
```
jyothsnakonisa: May be these can be private

server/src/main/java/org/apache/cassandra/sidecar/db/SystemViewsDatabaseAccessor.java
jyothsnakonisa: Instead of previous and current, it might be more clear if we use version number `YAML_PROP_40` `YAML_PROP_50`

jyothsnakonisa (approved):
Looks good to me!

server/src/main/java/org/apache/cassandra/sidecar/db/SystemViewsDatabaseAccessor.java
```diff
@@ -0,0 +1,82 @@
...
+import org.apache.cassandra.sidecar.db.schema.SystemViewsSchema;
+import org.apache.cassandra.sidecar.exceptions.SchemaUnavailableException;
+import org.apache.cassandra.sidecar.utils.FileUtils;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Database Accessor that queries cassandra to get information maintained under system_auth keyspace.
+ */
+@Singleton
+public class SystemViewsDatabaseAccessor extends DatabaseAccessor<SystemViewsSchema>
```
yifan-c, line 40: no integration test for this class.

server/src/main/java/org/apache/cassandra/sidecar/db/SystemViewsDatabaseAccessor.java
```diff
@@ -0,0 +1,82 @@
...
+{
+    private static final String YAML_PROP_40 = "cdc_total_space_in_mb";
+    private static final String YAML_PROP_50 = "cdc_total_space";
```
yifan-c: I do not think the 40/50 suffix is true. `cdc_total_space` is available since cassandra 4.1, and both are available in 5.0. 
It is another reason for adding a integration test.

server/src/main/java/org/apache/cassandra/sidecar/db/SystemViewsDatabaseAccessor.java
```diff
@@ -0,0 +1,82 @@
...
+    {
+        // attempt to parse Cassandra v4.0 'cdc_total_space_in_mb' yaml prop
+        String cdcTotalSpaceInMb = getSetting(YAML_PROP_40);
+        if (cdcTotalSpaceInMb != null)
+        {
+            return FileUtils.mbStringToBytes(cdcTotalSpaceInMb);
+        }
+
+        // otherwise parse current (v5.0+) 'cdc_total_space' yaml prop
+        String storageStringToBytes = getSetting(YAML_PROP_50);
```
yifan-c: It queries cassandra twice. Both values should be able to retrieve in a single query.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/SystemViewsSchema.java
```diff
@@ -0,0 +1,68 @@
...
+import com.datastax.driver.core.Session;
+import com.google.inject.Singleton;
+import org.apache.cassandra.sidecar.exceptions.SchemaUnavailableException;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Schema for getting information stored in system_views keyspace.
+ */
+@Singleton
+public class SystemViewsSchema extends CassandraSystemTableSchema
```
yifan-c, line 31: It needs to be exposed/declared in the `CdcModule` and add it into table schema mapBinder, so that the statement can be prepared. Taking the auth table for example,

```java
    @ProvidesIntoMap
    @KeyClassMapKey(TableSchemaMapKeys.SystemAuthSchemaKey.class)
    TableSchema systemAuthSchema() // Note that it returns TableSchema, not CassandraSystemTableSchema, in order to locate the same mapBinder
    {
        return new SystemAuthSchema();
    }
```

server/src/main/java/org/apache/cassandra/sidecar/metrics/CdcMetrics.java
```diff
@@ -0,0 +1,66 @@
...
+    {
+        this.metricRegistry = Objects.requireNonNull(metricRegistry, "Metric registry can not be null");
+        this.cdcRawCleanerFailed = createMetric("CdcRawCleanerFailed", name -> metricRegistry.gauge(name, DeltaGauge::new));
+        this.totalConsumedCdcBytes = createMetric("CdcRawTotalConsumedCdcBytes", name -> metricRegistry.gauge(name, () -> new DefaultSettableGauge<>(0L)));
+        this.totalCdcSpaceUsed = createMetric("CdcRawTotalCdcSpaceUsed", name -> metricRegistry.gauge(name, () -> new DefaultSettableGauge<>(0L)));
+        this.orphanedIdx = createMetric("CdcRawOrphanedIdxFile", name -> metricRegistry.gauge(name, DeltaGauge::new));
+        this.deletedSegment = createMetric("CdcRawDeletedSegment", name -> metricRegistry.gauge(name, DeltaGauge::new));
+        this.oldestSegmentAge = createMetric("CdcRawOldestSegmentAgeSeconds", name -> metricRegistry.gauge(name, () -> new DefaultSettableGauge<>(0)));
+        this.lowCdcRawSpace = createMetric("CdcRawLowSpace", name -> metricRegistry.gauge(name, DeltaGauge::new));
+        this.criticalCdcRawSpace = createMetric("CdcRawCriticalSpace", name -> metricRegistry.gauge(name, DeltaGauge::new));
```
yifan-c: How about removing the `CdcRaw` prefix and `Cdc` in the name? It is already defined in the domain part (i.e. `.Cdc`). For example, `CdcRawTotalCdcSpaceUsed` is now just `TotalSpaceUsed`. The full kpi is `xx.Cdc.TotalSpaceUsed`, which is still clear.

server/src/main/java/org/apache/cassandra/sidecar/metrics/CdcMetrics.java
```diff
@@ -0,0 +1,66 @@
...
+ * Tracks metrics related to cdc functionality provided by Sidecar.
+ */
+public class CdcMetrics
+{
+    public static final String DOMAIN = SERVER_PREFIX + ".Cdc";
+    protected final MetricRegistry metricRegistry;
+    public final NamedMetric<DeltaGauge> cdcRawCleanerFailed;
+    public final NamedMetric<DeltaGauge> orphanedIdx;
+    public final NamedMetric<DefaultSettableGauge<Integer>> oldestSegmentAge;
+    public final NamedMetric<DefaultSettableGauge<Long>> totalConsumedCdcBytes;
```
yifan-c: If my understanding is correct, cdc consumption can happen multiple times between the 2 stats publishing internal. In this case, `totalConsumedCdcBytes` should be using `DeltaGauge`. 
The other 2 usages of `DefaultSettableGauge` look valid.

server/src/main/java/org/apache/cassandra/sidecar/modules/CdcModule.java
```diff
@@ -63,6 +69,17 @@ PeriodicTask sidecarPeerHealthMonitorTask(SidecarPeerHealthMonitorTask task)
...
+    @ProvidesIntoMap
+    @KeyClassMapKey(PeriodicTaskMapKeys.CdcRawDirectorySpaceCleanerTaskKey.class)
+    PeriodicTask cdcRawDirectorySpaceCleanercPeriodicTask(TimeProvider timeProvider,
+                                                          SystemViewsDatabaseAccessor systemViewsDatabaseAccessor,
+                                                          CdcConfiguration cdcConfiguration,
+                                                          InstanceMetadata instanceMetadata,
+                                                          SidecarMetrics metrics)
+    {
+        return new CdcRawDirectorySpaceCleaner(timeProvider, systemViewsDatabaseAccessor, cdcConfiguration, instanceMetadata, metrics);
+    }
```
yifan-c: It is probably like a hack. But since `CdcRawDirectorySpaceCleaner` is annotated with `@Singleton` and `@Inject`, we can simply wire the singleton to the mapBinder. It is actually necessary to change to avoid breaking the `singleton` requirement, as `new`ing instance does not honor the annotations. 

```suggestion
    @ProvidesIntoMap
    @KeyClassMapKey(PeriodicTaskMapKeys.CdcRawDirectorySpaceCleanerTaskKey.class)
    PeriodicTask cdcRawDirectorySpaceCleanercPeriodicTask(CdcRawDirectorySpaceCleaner cleanerTask)
    {
        return cleanerTask;
    }
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -0,0 +1,467 @@
...
+    private final CdcConfiguration cdcConfiguration;
+    private final InstanceMetadata instanceMetadata;
+    private final CdcMetrics cdcMetrics;
+
+    @Nullable
+    private volatile Long maxUsageBytes = null;
+    // lazily loaded from system_views.settings if available
+    private volatile Long maxUsageLastReadNanos = null;
+    // cdc file -> file size in bytes. It memorizes the file set of the last time the checker runs.
+    private volatile Map<CdcRawSegmentFile, Long> priorCdcFiles = new HashMap<>();
```
yifan-c: nit: all the fields do not need to be `volatile`. The periodic task executor ensures memory visibility. That said, it does not harm much with the keyword.

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -0,0 +1,467 @@
...
+        long upperLimitBytes =
+        (long) (maxUsage() * cdcConfiguration.cdcRawDirectoryMaxPercentUsage());
+        // Sort the files by segmentId to delete commit log segments in write order
+        // The latest file is the current active segment, but it could be created before the retention duration, e.g. slow data ingress
+        Collections.sort(segmentFiles);
+        long nowInMillis = timeProvider.currentTimeMillis();
+
+        // track the age of the oldest commit log segment to give indication of the time-window buffer available
+        cdcMetrics.oldestSegmentAge.metric.setValue(
+        (int) MILLISECONDS.toMinutes(nowInMillis - segmentFiles.get(0).lastModified()));
```
yifan-c: `oldestSegmentAge` has the name `CdcRawOldestSegmentAgeSeconds`. The unit is seconds, not minutes.

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -0,0 +1,467 @@
...
+            Long newValue = systemViewsDatabaseAccessor.getCdcTotalSpaceSetting();
+            if (newValue != null)
+            {
+                if (!newValue.equals(maxUsageBytes))
+                {
+                    LOGGER.info(
+                    "Change in cdc_total_space from system_views.settings prev={} latest={}",
+                    maxUsageBytes, newValue);
+                    this.maxUsageBytes = newValue;
+                    this.maxUsageLastReadNanos = System.nanoTime();
```
yifan-c: It is odd that this class has dependency on `TimeProvider`, but has 2 different time sources. 
Well, ultimately, the time source is `System`. But from the dependency's perspective, there are `System` and `TimeProvider`. Can we use `System` only?  Or enrich `TimeProvider` for nano precision? I would prefer the former option.

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -0,0 +1,467 @@
...
+                       .map(dir -> new File(dir, CDC_DIR_NAME))
+                       .forEach(this::cleanUpCdcRawDirectory);
+    }
+
+    protected void cleanUpCdcRawDirectory(File cdcRawDirectory)
+    {
+        if (!cdcRawDirectory.exists() || !cdcRawDirectory.isDirectory())
+        {
+            LOGGER.debug("Skipping CdcRawDirectorySpaceCleaner: CDC directory does not exist: " +
+                         cdcRawDirectory);
```
yifan-c:
```suggestion
            LOGGER.debug("Skipping CdcRawDirectorySpaceCleaner: CDC directory does not exist: {}", cdcRawDirectory);
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -0,0 +1,467 @@
...
+                                                                   .filter(
+                                                                   CdcRawSegmentFile::indexExists)
+                                                                   .collect(Collectors.toList())
+                                               )
+                                               .orElseGet(List::of);
+        publishCdcStats(segmentFiles);
+        if (segmentFiles.size() < 2)
+        {
+            LOGGER.debug(
+            "Skipping cdc data cleaner routine cleanup: No cdc data or only one single cdc segment is found.");
```
yifan-c: nit: One line.

```suggestion
            LOGGER.debug("Skipping cdc data cleaner routine cleanup: No cdc data or only one single cdc segment is found.");
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -0,0 +1,467 @@
...
+        {
+            LOGGER.debug("Could not read cdc_total_space from system_views.settings", e);
+        }
+        catch (Throwable t)
+        {
+            LOGGER.warn("Error reading cdc_total_space from system_views.settings", t);
+        }
+
+        LOGGER.warn(
+        "Could not read cdc_total_space from system_views.settings, falling back to props");
```
yifan-c: nit: one line

```suggestion 
        LOGGER.warn("Could not read cdc_total_space from system_views.settings, falling back to props");
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -0,0 +1,467 @@
...
+            {
+                LOGGER.error("Cdc segments sorted incorrectly {} before {}",
+                             segmentFiles.get(0).segmentId, segmentFiles.get(1).segmentId);
+            }
+
+            long criticalMillis = cdcConfiguration.cdcRawDirectoryCriticalBufferWindow().toMillis();
+            long lowMillis = cdcConfiguration.cdcRawDirectoryLowBufferWindow().toMillis();
+
+            int i = 0;
+            while (i < segmentFiles.size() - 1 && directorySize > upperLimitBytes)
```
yifan-c, line 243: Need to left a command why the last/active segment should not be deleted.

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
jberragan: I was unsure of how these PeriodicTasks are executed internally.

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
jberragan: `TimeProvider` for some reason does not implement `nanoTime`

jberragan: I mean what's the point in `TimeProvider` if you want to just use `System` everywhere.

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
jberragan: "a comment"?

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
yifan-c: `TimeProvider` was introduced when adding the time skew API and it is used only there. At the time, there is no need for nanoTime. I am fine with either enriching `TimeProvider` or just using `System` (if considering `TimeProvider` is private to  the time skew API).

jberragan:
Green CircleCI: https://app.circleci.com/pipelines/circleci/DSY6xjBM591VFJ6A5dNS1P/HTfNc3hkdX2WnB6jtaxyFN/3/workflows/ae47dfcc-6dbe-4466-bf68-77fb777073d7

jberragan:
Any outstanding issues blocking merge @yifan-c  @bbotella ?

yifan-c:
No more from me. I am +1. Let me also comment on JIRA.

PR #204, opened 2025-03-06
CASSSIDECAR-216: Capture Metrics for Schema Reporting

5:


server/src/main/java/org/apache/cassandra/sidecar/datahub/IdentifiersProvider.java
```diff
@@ -231,13 +251,19 @@ public boolean equals(@Nullable Object other)
...
+        return new ToStringBuilder(this, STYLE)
                 .append(this.organization())
                 .append(this.platform())
                 .append(this.environment())
                 .append(this.application())
                 .append(this.cluster())
                 .append(this.identifier())
-                .toString();
+                .toString()
+                .replaceAll("\\s", "");
```
yifan-c, line 262: The toString is relatively complex. Is there a test?

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
```diff
@@ -96,45 +110,73 @@ public SchemaReporter(@NotNull IdentifiersProvider identifiersProvider,
...
+     * Private method for converting and reporting the Cassandra schema
+     *
+     * @param metadata the {@link Metadata} to extract Cassandra schema from
+     * @param started the {@link DeltaGauge} for the metric counting invocations
+     */
+    private void process(@NotNull Metadata metadata,
+                         @NotNull DeltaGauge started)
+    {
+        String action = " reporting schema for cluster with identifiers " + identifiersProvider;
+        LOGGER.info("Started" + action);
```
yifan-c: Use the value placeholder of logger to eval lazily. Update the other places too. 

```suggestion
    LOGGER.info("Started {}", action);
```

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
```diff
@@ -96,45 +110,73 @@ public SchemaReporter(@NotNull IdentifiersProvider identifiersProvider,
...
+    /**
+     * Private method for converting and reporting the Cassandra schema
+     *
+     * @param metadata the {@link Metadata} to extract Cassandra schema from
+     * @param started the {@link DeltaGauge} for the metric counting invocations
+     */
+    private void process(@NotNull Metadata metadata,
+                         @NotNull DeltaGauge started)
+    {
+        String action = " reporting schema for cluster with identifiers " + identifiersProvider;
```
yifan-c: you can remove the heading space. It should make the error message in the RuntimeException look a bit nicer too.

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
```diff
@@ -0,0 +1,91 @@
...
+    protected static final String FAILURE = "Failure";
+    protected static final String ASPECTS = "Aspects";
+    protected static final String MILLISECONDS = "Milliseconds";
+
+    public final NamedMetric<DeltaGauge> startedRequest;
+    public final NamedMetric<DeltaGauge> startedSchedule;
+    public final NamedMetric<DeltaGauge> finishedSuccess;
+    public final NamedMetric<DeltaGauge> finishedFailure;
+    public final NamedMetric<Histogram> durationMilliseconds;
+    public final NamedMetric<Histogram> sizeAspects;
```
yifan-c: Please update the patch to follow the other Metrics class? Let them be consistent. The string constants that are _only used once_ can all be removed.

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
```diff
@@ -0,0 +1,91 @@
...
+                .withDomain(DOMAIN)
+                .withName(SIZE)
+                .addTag(UNIT, ASPECTS)
+                .build();
+
+        durationMilliseconds = NamedMetric.builder(registry::histogram)
+                .withDomain(DOMAIN)
+                .withName(DURATION)
+                .addTag(UNIT, MILLISECONDS)
+                .build();
```
yifan-c: Prefer self-describing names over tags. For example, "StartedBySchedule" provides better readability than "Trigger=Schedule.Started". Note that all the tags are combined into this `key=value` format at the end.

server/src/test/integration/org/apache/cassandra/sidecar/datahub/SchemaReporterIntegrationTest.java
```diff
@@ -97,21 +124,31 @@ void testSchemaConverter() throws IOException
...
+        // completed successfully, with thirteen aspects produced in zero or more milliseconds
+        SchemaReportingMetrics metrics = this.metrics.server().schemaReporting();
+        assertEquals(0L, metrics.startedRequest.metric.getValue());
+        assertEquals(1L, metrics.startedSchedule.metric.getValue());
+        assertEquals(1L, metrics.finishedSuccess.metric.getValue());
+        assertEquals(0L, metrics.finishedFailure.metric.getValue());
+        assertEquals(1L, metrics.sizeAspects.metric.getCount());
+        assertEquals(13L, metrics.sizeAspects.metric.getSnapshot().getValues()[0]);
+        assertEquals(1L, metrics.durationMilliseconds.metric.getCount());
+        assertTrue(0L <= metrics.durationMilliseconds.metric.getSnapshot().getValues()[0]);
```
yifan-c: Use assertJ

server/src/test/java/org/apache/cassandra/sidecar/datahub/SchemaReporterTest.java
```diff
@@ -107,20 +141,29 @@ void testEmptyTable() throws IOException
...
+
+        SchemaReportingMetrics metrics = this.metrics.server().schemaReporting();            // Validate captured metrics:
+        assertEquals(1L, metrics.startedRequest.metric.getValue());                          //  * one execution triggered by request
+        assertEquals(0L, metrics.startedSchedule.metric.getValue());                         //  * zero executions triggered by schedule
+        assertEquals(1L, metrics.finishedSuccess.metric.getValue());                         //  * one execution resulted in success
+        assertEquals(0L, metrics.finishedFailure.metric.getValue());                         //  * zero executions resulted in failure
+        assertEquals(1L, metrics.sizeAspects.metric.getCount());                             //  * single number of aspects,
+        assertEquals(13L, metrics.sizeAspects.metric.getSnapshot().getValues()[0]);          //    equal to thirteen
+        assertEquals(1L, metrics.durationMilliseconds.metric.getCount());                    //  * single duration of execution,
+        assertTrue(0L <= metrics.durationMilliseconds.metric.getSnapshot().getValues()[0]);  //    that is non-negative
```
yifan-c: use assertJ

server/src/test/java/org/apache/cassandra/sidecar/datahub/SchemaReporterTest.java
```diff
@@ -261,12 +311,21 @@ void testUserTypes() throws IOException
...
+
+        SchemaReportingMetrics metrics = this.metrics.server().schemaReporting();            // Validate captured metrics:
+        assertEquals(0L, metrics.startedRequest.metric.getValue());                          //  * zero executions triggered by request
+        assertEquals(1L, metrics.startedSchedule.metric.getValue());                         //  * one execution triggered by schedule
+        assertEquals(1L, metrics.finishedSuccess.metric.getValue());                         //  * one execution resulted in success
+        assertEquals(0L, metrics.finishedFailure.metric.getValue());                         //  * zero executions resulted in failure
+        assertEquals(1L, metrics.sizeAspects.metric.getCount());                             //  * single number of aspects,
+        assertEquals(13L, metrics.sizeAspects.metric.getSnapshot().getValues()[0]);          //    equal to thirteen
+        assertEquals(1L, metrics.durationMilliseconds.metric.getCount());                    //  * single duration of execution,
+        assertTrue(0L <= metrics.durationMilliseconds.metric.getSnapshot().getValues()[0]);  //    that is non-negative
```
yifan-c: Use assertJ

yifan-c (commented):
@sarankk , you should also review this patch

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
skoppu22: +1

server/src/test/java/org/apache/cassandra/sidecar/datahub/SchemaReporterTest.java
```diff
@@ -33,70 +36,101 @@
...
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 /**
  * Unit tests for {@link SchemaReporter}
  */
-@SuppressWarnings("try")
+@SuppressWarnings("resource")
```
skoppu22: Same as other PR, why do we need SuppressWarnings for resource, I guess that's to avoid resource leak warning right? Do we have that warning for this class?

server/src/main/java/org/apache/cassandra/sidecar/datahub/IdentifiersProvider.java
5: Added new unit test file covering all important behaviors of the `IdentifiersProvider`.

server/src/test/integration/org/apache/cassandra/sidecar/datahub/SchemaReporterIntegrationTest.java
5: Switched to the new syntax.

server/src/test/java/org/apache/cassandra/sidecar/datahub/SchemaReporterTest.java
5: Switched to the new syntax.

server/src/test/java/org/apache/cassandra/sidecar/datahub/SchemaReporterTest.java
5: Switched to the new syntax.

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
5: This did not look good and resulted in bad log messages, fixed.

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
5: This did not look good and resulted in bad log messages, fixed.

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
5: I personally find putting units in tags more readable, but see your point.

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
5: Done.

server/src/test/java/org/apache/cassandra/sidecar/datahub/SchemaReporterTest.java
5: The point is to suppress the compiler warning caused by the use of an instance of `JsonEmitter` outside of `try-with-resource` block, which should not normally happen, but does make sense for tests.

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
```diff
@@ -0,0 +1,72 @@
...
+                                     .withName("StartedBySchedule")
+                                     .build();
+        startedRequest = NamedMetric.builder(name -> registry.gauge(name, DeltaGauge::new))
+                                    .withDomain(DOMAIN)
+                                    .withName("StartedByRequest")
+                                    .build();
+
+        finishedSuccess = NamedMetric.builder(name -> registry.gauge(name, DeltaGauge::new))
+                                     .withDomain(DOMAIN)
+                                     .withName("FinishedWithSuccess")
```
yifan-c: nit: `Succeeded` for simplicity?

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
```diff
@@ -0,0 +1,72 @@
...
+                                     .build();
+        finishedFailure = NamedMetric.builder(name -> registry.gauge(name, DeltaGauge::new))
+                                     .withDomain(DOMAIN)
+                                     .withName("FinishedWithFailure")
```
yifan-c: nit: `Failed` for simplicity?

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
```diff
@@ -0,0 +1,72 @@
...
+
+        sizeAspects = NamedMetric.builder(registry::histogram)
+                                 .withDomain(DOMAIN)
+                                 .withName("SizeInAspects")
+                                 .build();
+
+        durationMilliseconds = NamedMetric.builder(registry::histogram)
+                                          .withDomain(DOMAIN)
+                                          .withName("DurationInMilliseconds")
+                                          .build();
```
yifan-c: I think it should be a `Timer`. And we usually name the duration/time related metrics as `XyzTime`. Here it would be "SchemaReportTime" or similar. 

The timer collects values in nanoseconds to ensure the max precision during the raw data collection.

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
```diff
@@ -96,44 +110,72 @@ public SchemaReporter(@NotNull IdentifiersProvider identifiersProvider,
...
+    private void process(@NotNull Metadata metadata,
+                         @NotNull DeltaGauge started)
     {
+        LOGGER.info("Starting to report schema for cluster, identifiers={}", identifiersProvider);
+        started.increment();
+
         try (Emitter emitter = emitterFactory.emitter())
         {
-            stream(metadata).forEach(ThrowableUtils.consumer(emitter::emit));
+            Stopwatch stopwatch = Stopwatch.createStarted();
```
yifan-c: nit: avoid it if we can.

server/src/main/java/org/apache/cassandra/sidecar/datahub/IdentifiersProvider.java
```diff
@@ -42,7 +43,26 @@ public abstract class IdentifiersProvider
     protected static final String DATA_PLATFORM_INSTANCE = "dataPlatformInstance";
     protected static final String CONTAINER = "container";
     protected static final String DATASET = "dataset";
-    protected static final String PROD = "PROD";  // DataHub requires this to be {@code PROD} regardless
+    protected static final String PROD = "PROD";  // DataHub requires this to be {@code PROD}
+
+    protected static final ToStringStyle STYLE = new ToStringStyle()
+    {{
```
sarankk, line 49: Minor nit: We can use method chaining here with `ToStringStyle.DEFAULT_STYLE.withUseShortClassName()`

server/src/main/java/org/apache/cassandra/sidecar/metrics/DeltaGauge.java
```diff
@@ -33,7 +33,15 @@ public class DeltaGauge implements Gauge<Long>, Metric
 
     public DeltaGauge()
     {
-        this.count = new AtomicLong();
+        this.count = new AtomicLong(0L);
```
sarankk, line 36: Any reason why we add `0L` here? default `new AtomicLong()` creates with `0`

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
sarankk: We can use the metric to record time instead of `Stopwatch`,  `reportContext.stop();` calls `update`

```
try (Emitter emitter = emitterFactory.emitter())
{
    try (Timer.Context reportContext = reportingMetrics.durationMilliseconds.metric.time())
    {
        long counter = stream(metadata).map(ThrowableUtils.function(emitter::emit))
                                       .count();
        reportContext.stop();
        reportingMetrics.sizeAspects.metric.update(counter);
        reportingMetrics.finishedSuccess.metric.increment();
        LOGGER.info("Successfully reported schema for cluster, identifiers={}", identifiersProvider);
    }
}
```

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
sarankk: Yes agree, timer is better here. Also we can record time taken directly with the timer.

server/src/test/integration/org/apache/cassandra/sidecar/datahub/SchemaReporterIntegrationTest.java
```diff
@@ -97,18 +122,28 @@ void testSchemaConverter() throws IOException
         JsonEmitter emitter = new JsonEmitter();
         try (Session session = maybeGetSession())
         {
-            new SchemaReporter(IDENTIFIERS, () -> emitter).process(session.getCluster());
+            new SchemaReporter(IDENTIFIERS, () -> emitter, metrics).processScheduled(session.getCluster());
         }
         String   actualJson = normalizeNames(emitter.content());
```
sarankk: Nit: extra space

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
```diff
@@ -96,44 +110,72 @@ public SchemaReporter(@NotNull IdentifiersProvider identifiersProvider,
...
+            long counter = stream(metadata).map(ThrowableUtils.function(emitter::emit))
```
sarankk: This count seems to ignore exceptions in emit, what will we be using this counter metric for?

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
```diff
@@ -0,0 +1,72 @@
...
+                                     .withName("FinishedWithSuccess")
+                                     .build();
+        finishedFailure = NamedMetric.builder(name -> registry.gauge(name, DeltaGauge::new))
+                                     .withDomain(DOMAIN)
+                                     .withName("FinishedWithFailure")
+                                     .build();
+
+        sizeAspects = NamedMetric.builder(registry::histogram)
+                                 .withDomain(DOMAIN)
+                                 .withName("SizeInAspects")
```
sarankk: What is the use case of `SizeInAspects` metric?

sarankk (commented):
Looks good to me, left some comments

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
5: Done.

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
5: Done.

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
5: Good point, switched to using `NamedMetric<Timer>` instead, thanks.

server/src/test/integration/org/apache/cassandra/sidecar/datahub/SchemaReporterIntegrationTest.java
5: Not sure what you mean here, did I accidentally address it already?

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReporter.java
5: Exception in emit will be wrapped into a `RuntimeException` and rethrown, aborting conversion, and resulting in the `Failed` metric being updated from the `catch`. I guess we don't care about the number of aspects produced before failure has happened…

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
5: Number of DataHub aspects in the schema produced for the current cluster, the actual value is expected to be `number_of_tables * 7 + number_of_keyspaces * 4 + 2`.

server/src/main/java/org/apache/cassandra/sidecar/metrics/DeltaGauge.java
5: Just wanted to make it more obvious how this works; caught myself control-clicking the constructor every time I look at this class…

server/src/main/java/org/apache/cassandra/sidecar/datahub/IdentifiersProvider.java
5: That is a great point, changed.

5: That would be convenient, but I can't locate the `withUseShortClassName`; am I using an older version of the library?

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/SchemaReportingMetrics.java
5: Done.

server/src/test/integration/org/apache/cassandra/sidecar/datahub/SchemaReporterIntegrationTest.java
5: Got it, fixed.

yifan-c (approved):
LGTM

sarankk (approved):
+1 Lgtm (nb)

frankgh (approved):
+1

PR #205, opened 2025-03-06
CASSSIDECAR-217: Retry Failed Schema Reports

5:


server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
```diff
@@ -43,6 +45,7 @@ public class SchemaReportingTask implements PeriodicTask, ExecuteOnClusterLeaseh
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(SchemaReportingTask.class);
     private static final ThreadLocalRandom RANDOM = ThreadLocalRandom.current();
+    private static final Duration MINUTE = Duration.ofMinutes(1L);
```
skoppu22: Can we make the duration as a config option?

server/src/test/java/org/apache/cassandra/sidecar/datahub/SchemaReporterTest.java
```diff
@@ -41,7 +41,7 @@
 /**
  * Unit tests for {@link SchemaReporter}
  */
-@SuppressWarnings("try")
+@SuppressWarnings("resource")
```
skoppu22: I would like to understand why are we having this SuppressWarnings for resource? If I understand correctly, this is to avoid resource leak warning.

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
5: Done.

server/src/test/java/org/apache/cassandra/sidecar/datahub/SchemaReporterTest.java
5: This suppresses the compiler warning caused by the use of auto-closeable `JsonEmitter` outside of a try-with-resources block, but it's what we need to do in tests…

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
```diff
@@ -94,17 +97,34 @@ public DurationSpec delay()
...
+        execute(promise, 1);
+    }
+
+    protected void execute(@NotNull Promise<Void> promise,
+                           int attempt)
     {
         try
         {
+            LOGGER.info("Schema report has been triggered by the schedule");
             reporter.process(session.get().getCluster());
```
frankgh, line 111: maybe a log entry indicating success as well after the reporting completes?

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
```diff
@@ -94,17 +97,34 @@ public DurationSpec delay()
...
             promise.complete();
         }
         catch (Throwable throwable)
         {
-            LOGGER.error("Failed to convert and report the current schema", throwable);
-            promise.fail(throwable);
+            if (attempt < configuration.retries())
+            {
+                LOGGER.warn("Schema report has failed and will be retried soon", throwable);
```
frankgh:
```suggestion
                LOGGER.warn("Schema report has failed and will be retried in {}", configuration.delay(), throwable);
```

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
```diff
@@ -94,17 +97,34 @@ public DurationSpec delay()
...
+                executor.setTimer(configuration.delay().toMillis(),
+                                  identifier -> execute(promise, attempt + 1));
+                // Retry will take care of either completing or failing the promise
+            }
+            else
+            {
+                LOGGER.error("Schema report is failing repeatedly and will not be retried", throwable);
```
frankgh:
```suggestion
                LOGGER.error("Schema report failed after {} retries and will not be retried", configuration.retries(), throwable);
```

conf/sidecar.yaml
```diff
@@ -148,6 +148,7 @@ schema_reporting:                            # Schema Reporting configuration
   endpoint:         http://localhost/schema  # Endpoint address for schema reporting
   method:           PUT                      # HTTP verb to use for schema reporting
   retries:          3                        # Number of times a failing schema report is retried
+  delay:            1m                       # Delay before a failing schema report is retried
```
frankgh: can we rename this to `retry_delay` for consistency with the existing retry delay configurations?

frankgh (commented):
Looks good, I have some minor suggestions.

conf/sidecar.yaml
5: Renamed `retries` into `max_retries` and `delay` into `retry_delay` everywhere for consistency with the rest of configuration.

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
5: Done.

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
5: Done.

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
5: These are great suggestions, fixed.

server/src/main/java/org/apache/cassandra/sidecar/datahub/SchemaReportingTask.java
```diff
@@ -53,38 +53,41 @@ public class SchemaReportingTask implements PeriodicTask, ExecuteOnClusterLeaseh
...
     @Override
-    public void deploy(Vertx vertx, PeriodicTaskExecutor executor)
+    public void deploy(@NotNull Vertx vertx,
+                       @NotNull PeriodicTaskExecutor executor)
     {
-        // TODO: react on ON_CASSANDRA_CQL_READY instead? When any CQL connection is ready, cluster metadata should be available from session
-        EventBusUtils.onceLocalConsumer(vertx.eventBus(), ON_ALL_CASSANDRA_CQL_READY.address(), ignored -> executor.schedule(this));
+        EventBusUtils.onceLocalConsumer(vertx.eventBus(),
+                                        SidecarServerEvents.ON_CASSANDRA_CQL_READY.address(),
+                                        message -> executor.schedule(this));
```
yifan-c, line 76: 👍

frankgh (approved):
+1 thanks for the improvement!

yifan-c (approved):

PR #207, opened 2025-03-13
CASSSIDECAR-235 Improvements to CDCConfig classes

jyothsnakonisa:
Following changes are included in the patch
1. CdcConfigImpl should use SidecarConfiguration instead of CdcConfigAccessor & KafkaConfigAccessor separately in constructor
2. Make `CdcConfigAccessor.service()` return String to make `ConfigAccessorImpl` more extendible.


frankgh (approved):
+1 looks good to me

PR #209, opened 2025-03-15
CASSSIDECAR-229: SidecarInstanceCodec is failing to find codec for type

frankgh:
Patch by Francisco Guerrero; reviewed by TBD for CASSSIDECAR-229


integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
```diff
@@ -562,6 +534,25 @@ public static Cluster createDriverCluster(ICluster<? extends IInstance> dtest)
...
 
+    /**
+     * Wraps the Sidecar server and keeps a reference to the injector to be able to dynamically retrieve
+     * objects for testing purposes
+     */
+    public static class ServerWrapper
+    {
+        public final Injector injector;
+        public final Server server;
+        public volatile int serverPort;
```
yifan-c, line 549: `public final`? Can this field be updated?

yifan-c (commented):
Not a full review. Just sharing a suggestion on `SidecarInstanceCodec` ---> https://github.com/yifan-c/cassandra-sidecar/commit/cdedf4f8236c38cb78b82f2f4df79a6b78ba0a75

integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
frankgh: yes, it will be updated to a new value when the sidecar server is stopped and then started again. https://github.com/apache/cassandra-sidecar/pull/209/files#diff-1d9d4d6c86b0f7269b52c45778bf1c8e696d79968140f02e742216459ec5bd66R205-R207

yifan-c: Why not expose a _method_ for port that is sourcing from `server`? Thus, updating the int field is avoided.

frankgh: The problem is that once the sidecar server is down, the port information is no longer available. For the `SidecarPeerDownDetectorIntegrationTest` this means that the information is lost when attempting to perform a health check on a down node. The test will do the following:

- bring up a 3 node cassandra cluster, each managed by a sidecar instance
- perform health checks
- stop sidecar on one of the instances
- ensure health check fails
- start sidecar on the instance above
- ensure health check succeeds

Now if you stop Sidecar, and use the `actualPort` information from the server, this information is not available, and your health check will fail, not because the peer health monitor will correctly detect the node as down, but instead with an `IllegalStateException` encountered when [retrieving the actualPort](https://github.com/apache/cassandra-sidecar/blob/trunk/server/src/main/java/org/apache/cassandra/sidecar/server/Server.java#L250). For that reason I decided to keep track of the port as a field, instead of dynamically asking the instance for the port.

integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
```diff
@@ -562,6 +538,30 @@ public static Cluster createDriverCluster(ICluster<? extends IInstance> dtest)
...
```
bbotella, line 549: Small nit: `port` instead of `serverPort`?

PR #211, opened 2025-03-19
CASSSIDECAR-232: Enhance the Cluster Lease Claim task feature

frankgh:


conf/sidecar.yaml
```diff
@@ -121,20 +121,37 @@ sidecar:
...
+      #                                                       electorate iff one of the Cassandra instances it manages
+      #                                                       owns token {@code 0} for the {@code sidecar_internal}
+      #                                                       keyspace.
+      electorate_membership_strategy: MostReplicatedKeyspaceTokenZeroElectorateMembership
       # Whether the process is enabled
       enabled: true
       # The initial delay for the first execution of the cluster lease claim process task after being
       # scheduled or rescheduled.
       # The minimum value for the initial delay is 0ms.
-      initial_delay: 1s
```
yifan-c: > The minimum value for the initial delay is 0ms.

Is it true? Should the minimum be 0? It will cause a lot of cas contention.

examples/conf/sidecar-ccm.yaml
```diff
@@ -277,6 +307,19 @@ access_control:
...
         certificate_identity_extractor: org.apache.cassandra.sidecar.acl.authentication.CassandraIdentityExtractor
+      # JwtAuthenticationHandlerFactory adds support to authenticate users with their JWT tokens. It also includes
+      # supports for OpenID discovery.
+    - class_name: org.apache.cassandra.sidecar.acl.authentication.JwtAuthenticationHandlerFactory
+      parameters:
+        # To selectively enable or disable JWT authentication
+        enabled: false
+        # Site for sidecar to dynamically retrieve the configuration information of an OpenID provider, without
+        # having to manually configure settings like issuer etc.
+        site: https://authorization.com
```
yifan-c, line 323: Thanks for adding them back. It is easy to forgot update this file. We should think of "copying" from `conf/sidecar.yaml` in the future.

conf/sidecar.yaml
```diff
@@ -121,20 +121,37 @@ sidecar:
...
+      #                                                       the highest replication factor, the keyspace to be used
+      #                                                       is decided by the keyspace with the name that sorts
+      #                                                       first in the lexicographic sort order. If no user
+      #                                                       keyspaces are created, the internal sidecar keyspace will
+      #                                                       be used.
+      # - SidecarInternalTokenZeroElectorateMembership        the current Sidecar will be determined to be part of the
+      #                                                       electorate iff one of the Cassandra instances it manages
+      #                                                       owns token {@code 0} for the {@code sidecar_internal}
+      #                                                       keyspace.
+      electorate_membership_strategy: MostReplicatedKeyspaceTokenZeroElectorateMembership
```
yifan-c, line 144: What is the default value if the new field `electorate_membership_strategy` is not configured?

server/src/main/java/org/apache/cassandra/sidecar/coordination/AbstractTokenZeroOfKeyspaceElectorateMembership.java
```diff
@@ -0,0 +1,149 @@
...
+        return result;
+    }
+
+    /**
+     * @param tokenRangeReplicas         the token range replicas for a keyspace
+     * @param localInstancesHostAndPorts local instance(s) IP(s) and port(s)
+     * @return {@code true} if any of the local instances is a replica of token zero for a single keyspace,
+     * {@code false} otherwise
+     */
+    boolean anyInstanceOwnsTokenZero(TokenRangeReplicasResponse tokenRangeReplicas, Set<String> localInstancesHostAndPorts)
```
yifan-c: curious, why not `protected` for this and for `collectLocalInstancesHostsAndPorts` and `replicaOwnsTokenZero`?

server/src/main/java/org/apache/cassandra/sidecar/modules/CoordinationModule.java
```diff
@@ -70,6 +72,18 @@ ElectorateMembership electorateMembership(InstanceMetadataFetcher instanceMetada
...
+                                       .electorateMembershipStrategy();
+        switch (strategy)
+        {
+            case "MostReplicatedKeyspaceTokenZeroElectorateMembership":
+                return new MostReplicatedKeyspaceTokenZeroElectorateMembership(instanceMetadataFetcher, cqlSessionProvider, configuration);
+            case "SidecarInternalTokenZeroElectorateMembership":
+                return new SidecarInternalTokenZeroElectorateMembership(instanceMetadataFetcher, configuration);
+            default:
+                throw new ConfigurationException("Invalid electorate membership strategy value '" + strategy + "'");
+        }
```
yifan-c: can we move this code to a factory class? Module code is hard to testable, and complex code should be tested.

server/src/main/java/org/apache/cassandra/sidecar/db/SidecarLeaseDatabaseAccessor.java
```diff
@@ -76,20 +78,22 @@ public LeaseClaimResult extendLease(String currentOwner)
...
         {
             return resultSet.wasApplied()
-                   ? new LeaseClaimResult(true, newOwner)
-                   : new LeaseClaimResult(false, resultSet.one().getString("owner"));
+                   ? new LeaseClaimResult(newOwner)
+                   // In some rare cases, the resultSet will not contain the owner information
+                   // even though the resultSet was not applied. This will translate into an
+                   // IllegalArgumentException being thrown when trying to retrieve the non-existing
+                   // owner string. This exception is left to be handled by the caller method
+                   : new LeaseClaimResult(resultSet.one().getString("owner"));
```
yifan-c, line 96: It is unclear to me what is the proper handling at the caller. Maybe handle exception here and set a new state, say "unknown". Caller can react on the "unknown" state, for example, query retry.

server/src/main/java/org/apache/cassandra/sidecar/coordination/ClusterLeaseClaimTask.java
```diff
@@ -144,7 +147,13 @@ public ScheduleDecision scheduleDecision()
...
     public DurationSpec initialDelay()
     {
-        return periodicTaskConfiguration.initialDelay();
+        // Return a randomized delay to introduce jitter among all the instances participating
+        // in the lease claim process
+        MillisecondBoundConfiguration initialDelay = periodicTaskConfiguration.initialDelay();
+        long millis = initialDelay.toMillis();
+        return millis <= 0
+               ? initialDelay
+               : new MillisecondBoundConfiguration(ThreadLocalRandom.current().nextLong(millis), TimeUnit.MILLISECONDS);
```
yifan-c: How can `initialDelay` be configured negative? 
How about adding a lower bound to the randomized delay? I am thinking of 1 second.

conf/sidecar.yaml
frankgh: good question, let me write a quick test . It should default to `MostReplicatedKeyspaceTokenZeroElectorateMembership`

frankgh: pushed the test

conf/sidecar.yaml
frankgh: yeah, I think that's fine. If you have a single sidecar instance there will be no contention. i.e using the default configuration values for the sidecar_internal keyspace and `SidecarInternalTokenZeroElectorateMembership`

server/src/main/java/org/apache/cassandra/sidecar/coordination/AbstractTokenZeroOfKeyspaceElectorateMembership.java
frankgh: good point! this is just copy paste error. This code originally lived in the `MostReplicatedKeyspaceTokenZeroElectorateMembership` class, and it has now been moved to `AbstractTokenZeroOfKeyspaceElectorateMembership`

server/src/test/integration/org/apache/cassandra/sidecar/coordination/MostReplicatedKeyspaceTokenZeroElectorateMembershipIntegrationTest.java
```diff
@@ -107,12 +109,17 @@ void test(TestVersion version) throws IOException
 
     private void runTestScenario(AbstractCluster<?> cluster)
     {
-        List<MostReplicatedKeyspaceTokenZeroElectorateMembership> memberships = buildElectorateMembershipPerCassandraInstance(cluster);
+        var pair = buildElectorateMembershipPerCassandraInstance(cluster);
```
frankgh: using var here, not sure if this is something we should discourage

server/src/test/resources/config/sidecar_coordination.yaml
frankgh, line 1: inlined this to a yaml string so we avoid having to add new files every time

server/src/main/java/org/apache/cassandra/sidecar/coordination/ClusterLeaseClaimTask.java
frankgh: it cannot be negative, it's restricted by `DurationSpec` however `ThreadLocalRandom.current().nextLong` throws when you provide it a value that is 0

server/src/main/java/org/apache/cassandra/sidecar/modules/CoordinationModule.java
frankgh: done

server/src/main/java/org/apache/cassandra/sidecar/coordination/ClusterLeaseClaimTask.java
frankgh: I could have done `==` but I think it's fine to leave a `<=` what do you think?

server/src/main/java/org/apache/cassandra/sidecar/db/SidecarLeaseDatabaseAccessor.java
yifan-c: my thinking was also unclear. Disregard it :p

server/src/test/integration/org/apache/cassandra/sidecar/coordination/MostReplicatedKeyspaceTokenZeroElectorateMembershipIntegrationTest.java
yifan-c: can we avoid using `var`?

frankgh: I used it for convenience in the test, otherwise the line was huge! I'd prefer to keep it here, but I also am wary of introducing var because other contributors will start using it and I think we should lean on discouraging its use.

server/src/main/java/org/apache/cassandra/sidecar/coordination/ClusterLeaseClaimTask.java
yifan-c: Thinking it more. I think the current `RAND(initialDelay)` changes the semantics of the initial delay. It is almost guaranteed that the actual value is not the `initialDelay`. It could cause surprise. 
Would it make sense to split it into 2 parts, `minInitialDelay` and `maxInitialDelay`? The actual value is randomly picked that in the value range.

frankgh: I think that's fine, but I would probably like to address it as part of a different effort. What do you think ?  In that case we'd also need to address the randomized initial delay in `org.apache.cassandra.sidecar.datahub.SchemaReportingTask#initialDelay`

server/src/test/integration/org/apache/cassandra/sidecar/coordination/MostReplicatedKeyspaceTokenZeroElectorateMembershipIntegrationTest.java
frankgh: ok, I'm no longer using var, I think that's better

server/src/main/java/org/apache/cassandra/sidecar/coordination/ClusterLeaseClaimTask.java
```diff
@@ -144,7 +146,16 @@ public ScheduleDecision scheduleDecision()
...
+        // Return a randomized delay to introduce jitter among all the instances participating
+        // in the lease claim process
+        long randomDeltaDelayMillis = configuration.randomDeltaDelayMillis();
+        if (randomDeltaDelayMillis == 0)
+        {
+            // no jitter
+            return configuration.initialDelay();
+        }
+        long initialDelayMillis = configuration.initialDelay().toMillis() + randomDeltaDelayMillis;
+        return new MillisecondBoundConfiguration(initialDelayMillis, TimeUnit.MILLISECONDS);
```
yifan-c, line 158: 👍

yifan-c (approved):

PR #212, opened 2025-04-08
CASSSIDECAR-222: Adding Live Migration list files handler.

nvharikrishna:

Adding API for listing files of Cassnadra instance which is required during live migration.

CASSSIDECAR-222

Sample URL for listing files: `https://<host:port>/api/v1/live-migration/files`

Sample response:
```
{
    "files": [
        {
            "fileUrl": "/api/v1/live-migration/files/data/0/system_schema/tables-afddfb9dbc1e30688056eed6c302ba09/nb-25-big-Filter.db",
            "size": 24,
            "fileType": "FILE",
            "lastModifiedTime": 1744296699193
        },
        {
            "fileUrl": "/api/v1/live-migration/files/data/0/system_schema/views-9786ac1cdd583201a7cdad556410c985",
            "size": -1,
            "fileType": "DIRECTORY",
            "lastModifiedTime": 1744012964644
        },
        {
            "fileUrl": "/api/v1/live-migration/files/data/0/system_schema/types-5a8b1ca866023f77a0459273d308917a",
            "size": -1,
            "fileType": "DIRECTORY",
            "lastModifiedTime": 1744012964611
        }
    ],
    "totalSize": 24
}
```

fileUrl in above payload is in the format: 

```
/api/v1/live-migration/files/<directory_type>/<directory_index>/<relative path from the directory>
```

fileUrl paths for different type of directories starts as follows:
```
data file dir: /api/v1/live-migration/files/data/0 
# index will increase if C* instance has more data directories
# For example: path prefix for second data file directory will look like this
#   /api/v1/live-migration/files/data/1

commitlog dir: /api/v1/live-migration/files/commitlog/0
hints dir: /api/v1/live-migration/files/hints/0
saved caches dir: /api/v1/live-migration/files/saved_caches/0
cdc raw dir: /api/v1/live-migration/files/cdc_raw/0
local system data file dir: /api/v1/live-migration/files/local_system_data/0
```

Will raise a separate PR for supporting file transfers: `/api/v1/live-migration/files/...`

Admin can exclude files or directories from Live Migration. Excluded files will not be listed by this files list API. Exclusions can be specified in the sidecar yaml configuration using glob or regex patterns. Admin can also use placeholders for different type of directories to avoid specifying full directory paths.

Here are the different placeholders supported:

```
COMMITLOG_DIR - commit log directory
HINTS_DIR - hints directory
SAVED_CACHES_DIR - saved caches directory
CDC_RAW_DIR - cdc raw directory
LOCAL_SYSTEM_DATA_FILE_DIR - local system data file data directory
DATA_FILE_DIR - all data directories
DATA_FILE_DIR_0 - first data directory
DATA_FILE_DIR_1 - second data directory
```

Sample placeholder:
```
# Pattern to exclude snapshots directories in data folder: 
    glob:${DATA_FILE_DIR}/*/*/snapshots/
```


server/src/main/java/org/apache/cassandra/sidecar/modules/LiveMigrationModule.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.cassandra.sidecar.routes.RouteBuilder;
+import org.apache.cassandra.sidecar.routes.VertxRoute;
+
+/**
+ * Module for supporting LiveMigration feature.
+ */
+public class LiveMigrationModule extends AbstractModule
+{
+
+    protected void configure()
```
frankgh: Add missing override annotation
```suggestion
    @Override
    protected void configure()
```

server/src/main/java/org/apache/cassandra/sidecar/modules/LiveMigrationModule.java
```diff
@@ -0,0 +1,56 @@
...
+    {
+        bind(LiveMigrationMap.class).to(LiveMigrationMapSidecarConfigImpl.class);
+        super.configure();
```
frankgh: I think we can avoid calling the super here

server/src/main/java/org/apache/cassandra/sidecar/modules/multibindings/VertxRouteMapKeys.java
```diff
@@ -248,4 +248,9 @@ interface UpdateServiceConfigurationRouteKey extends RouteClassKey
         HttpMethod HTTP_METHOD = HttpMethod.PUT;
         String ROUTE_URI = ApiEndpointsV1.SERVICE_CONFIG_ROUTE;
     }
+    interface LiveMigrationListInstanceFilesRouteKey extends RouteClassKey
```
frankgh: this list is alphabetically sorted. Can you move this to the correct position?

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFileInfo.java
```diff
@@ -0,0 +1,72 @@
...
+package org.apache.cassandra.sidecar.common.response;
+
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * Class for holding information of a downloadable file during Live Migration.
+ */
+public class InstanceFileInfo
```
frankgh, line 32: maybe annotate with ignore unknown, include non-null? In case the API adds more fields?

```
@JsonIgnoreProperties(ignoreUnknown = true)
@JsonInclude(JsonInclude.Include.NON_NULL)
```

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFileInfo.java
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Enum for file type
+     */
+    public enum FileType
+    {
+        FILE,
+        DIRECTORY
+    }
+
+    public boolean equals(Object o)
```
frankgh, line 64: Add override annotation

```suggestion
    @Override
    public boolean equals(Object o)
```

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFileInfo.java
```diff
@@ -0,0 +1,72 @@
...
+    {
+        if (o == null || getClass() != o.getClass()) return false;
+        InstanceFileInfo that = (InstanceFileInfo) o;
+        return size == that.size
+               && lastModifiedTime == that.lastModifiedTime
+               && Objects.equals(fileUrl, that.fileUrl)
+               && fileType == that.fileType;
+    }
+
+    public int hashCode()
```
frankgh, line 75: Include override annotation
```suggestion
    @Override
    public int hashCode()
```

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
```diff
@@ -0,0 +1,102 @@
...
+package org.apache.cassandra.sidecar.common.response;
+
+import java.util.List;
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Class representing response for
```
frankgh: maybe you meant to say the endpoint here?

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
```diff
@@ -0,0 +1,102 @@
...
+            if (file.fileType.equals(InstanceFileInfo.FileType.FILE))
+            {
+                long size = file.size;
+                sum += size;
+            }
+        }
+        totalSize = sum;
+    }
+
+    @SuppressWarnings("unused")
```
frankgh: maybe use json creator instead?
```suggestion
    @JsonCreator
```

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
```diff
@@ -0,0 +1,102 @@
...
+    {
+        return files;
+    }
+
+    public long getTotalSize()
+    {
+        return totalSize;
+    }
+
+    public boolean equals(Object o)
```
frankgh, line 95:
```suggestion
    @Override
    public boolean equals(Object o)
```

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
```diff
@@ -0,0 +1,102 @@
...
+    {
+        if (o == null || getClass() != o.getClass()) return false;
+        InstanceFilesListResponse that = (InstanceFilesListResponse) o;
+        return totalSize == that.totalSize && Objects.equals(files, that.files);
+    }
+
+    public int hashCode()
```
frankgh, line 103:
```suggestion
    @Override
    public int hashCode()
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationMap.java
```diff
@@ -0,0 +1,72 @@
...
+{
+    /**
+     * Tells whether given instance is configured as source or not.
+     *
+     * @param instanceMeta Cassandra instance metadata
+     * @return true if given instance is configured as source for live migration.
+     */
+    default boolean isSource(@NotNull final InstanceMetadata instanceMeta)
+    {
+        final Map<String, String> map = getMigrationMap();
```
frankgh: NIT
```suggestion
       Map<String, String> map = getMigrationMap();
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationMap.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Tells whether given instance is configured as destination or not.
+     *
+     * @param instanceMeta Cassandra instance metadata
+     * @return true if given instance is configured as destination for live migration.
+     */
+    default boolean isDestination(@NotNull final InstanceMetadata instanceMeta)
+    {
+        final Map<String, String> map = getMigrationMap();
```
frankgh: NIT
```suggestion
        Map<String, String> map = getMigrationMap();
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationListInstanceFilesHandler.java
```diff
@@ -0,0 +1,99 @@
...
+        try
+        {
+
+            context.json(new InstanceFilesListResponse(filesList.getFiles()));
+        }
+        catch (IOException e)
+        {
+            LOGGER.error("Could not fetch instance files information.", e);
+            context.fail(wrapHttpException(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage(), e));
+        }
```
frankgh: This needs to run in a blocking service thread
```suggestion
                executorPools.service().runBlocking(() -> {
            try
            {
                context.json(new InstanceFilesListResponse(filesList.getFiles()));
            }
            catch (IOException e)
            {
                LOGGER.error("Could not fetch instance files information.", e);
                context.fail(wrapHttpException(HttpResponseStatus.INTERNAL_SERVER_ERROR, e.getMessage(), e));
            }
        });
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationListInstanceFilesHandler.java
```diff
@@ -0,0 +1,99 @@
...
+    @Inject
+    public LiveMigrationListInstanceFilesHandler(InstanceMetadataFetcher metadataFetcher,
+                                                 ExecutorPools executorPools,
+                                                 CassandraInputValidator validator,
+                                                 SidecarConfiguration sidecarConfiguration,
+                                                 InstancesMetadata instancesMetadata)
+    {
+        super(metadataFetcher, executorPools, validator);
+        this.sidecarConfiguration = sidecarConfiguration;
+        this.instancesMetadata = instancesMetadata;
```
frankgh: intstancesMetadata is redundant, you can instead use the `metadataFetcher`

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationListInstanceFilesHandler.java
```diff
@@ -0,0 +1,99 @@
...
+
+    protected Void extractParamsOrThrow(RoutingContext context)
+    {
+        return null;
+    }
+
+    protected void handleInternal(RoutingContext context, HttpServerRequest httpRequest, @NotNull String host,
+                                  SocketAddress remoteAddress, Void request)
+    {
+        InstanceMetadata instanceMetadata = instancesMetadata.instanceFromHost(host);
```
frankgh:
```suggestion
        InstanceMetadata instanceMetadata = metadataFetcher.instance(host);
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationListInstanceFilesHandler.java
```diff
@@ -0,0 +1,99 @@
...
+                                                 CassandraInputValidator validator,
+                                                 SidecarConfiguration sidecarConfiguration,
+                                                 InstancesMetadata instancesMetadata)
+    {
+        super(metadataFetcher, executorPools, validator);
+        this.sidecarConfiguration = sidecarConfiguration;
+        this.instancesMetadata = instancesMetadata;
+    }
+
+    protected Void extractParamsOrThrow(RoutingContext context)
```
frankgh, line 68: can we add override annotations?

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Enables route if local instance is getting live migrated (source).
+     *
+     * @param rc - routing context
+     */
+    public void isSource(final RoutingContext rc)
+    {
+        final InstanceMetadata instanceMeta = getLocalInstanceMeta(rc);
```
frankgh:
```suggestion
       InstanceMetadata instanceMeta = getLocalInstanceMeta(rc);
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Enables route if local instance is the destination.
+     *
+     * @param rc - routing context
+     */
+    public void isDestination(final RoutingContext rc)
+    {
+        final InstanceMetadata instanceMeta = getLocalInstanceMeta(rc);
```
frankgh:
```suggestion
       InstanceMetadata instanceMeta = getLocalInstanceMeta(rc);
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * Enables route if local instance is either source or destination in live migration map.
+     *
+     * @param rc - routing context
+     */
+    public void isSourceOrDestination(final RoutingContext rc)
+    {
+        final InstanceMetadata instanceMeta = getLocalInstanceMeta(rc);
```
frankgh:
```suggestion
       InstanceMetadata instanceMeta = getLocalInstanceMeta(rc);
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
```diff
@@ -0,0 +1,107 @@
...
+            rc.fail(HttpResponseStatus.NOT_FOUND.code());
+        }
+    }
+
+    /**
+     * Enables route if local instance is either source or destination in live migration map.
+     *
+     * @param rc - routing context
+     */
+    public void isSourceOrDestination(final RoutingContext rc)
```
frankgh:
```suggestion
    public void isSourceOrDestination(RoutingContext rc)
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
```diff
@@ -0,0 +1,107 @@
...
+        this.liveMigrationMap = liveMigrationMap;
+        this.instancesMetadata = instancesMetadata;
+    }
+
+    /**
+     * Enables route if local instance is getting live migrated (source).
+     *
+     * @param rc - routing context
+     */
+    public void isSource(final RoutingContext rc)
```
frankgh:
```suggestion
    public void isSource(RoutingContext rc)
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+        else
+        {
+            rc.fail(HttpResponseStatus.NOT_FOUND.code());
+        }
+    }
+
+    private InstanceMetadata getLocalInstanceMeta(final RoutingContext rc)
+    {
+        final String host = extractHostAddressWithoutPort(rc.request());
```
frankgh:
```suggestion
       String host = extractHostAddressWithoutPort(rc.request());
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
```diff
@@ -0,0 +1,107 @@
...
+        {
+            rc.next();
+        }
+        else
+        {
+            rc.fail(HttpResponseStatus.NOT_FOUND.code());
+        }
+    }
+
+    private InstanceMetadata getLocalInstanceMeta(final RoutingContext rc)
```
frankgh:
```suggestion
    private InstanceMetadata getLocalInstanceMeta(RoutingContext rc)
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationMapSidecarConfigImpl.java
```diff
@@ -0,0 +1,58 @@
...
+{
+    private final LiveMigrationConfiguration liveMigrationConfiguration;
+
+    @Inject
+    public LiveMigrationMapSidecarConfigImpl(SidecarConfiguration sidecarConfiguration)
+    {
+        this.liveMigrationConfiguration = sidecarConfiguration.liveMigrationConfiguration();
+    }
+
+    public boolean isSource(@NotNull InstanceMetadata instanceMeta)
```
frankgh, line 45: can we add override annotations?

frankgh (commented):
Partial review of 17 / 33 files. Will continue the review tomorrow

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
```diff
@@ -0,0 +1,193 @@
...
+        return dirsToCopy;
+    }
+
+    /**
+     * Returns map of directory that can be copied and the index to use while constructing
+     * url for file transfer.
+     */
+    public static Map<String, String> dirPathPrefixMap(InstanceMetadata instanceMetadata)
+    {
+        Map<String, String> dirIndexMap = new Hashtable<>();
```
frankgh: Any reason why we would use a hashtable here instead of a HashMap?
```suggestion
        Map<String, String> dirIndexMap = new HashMap<>();
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+        return dirIndexMap;
+    }
+
+    /**
+     * Returns map of directory and set of placeholders that represent the directory.
+     */
+    public static Map<String, Set<String>> dirPlaceHoldersMap(InstanceMetadata instanceMetadata)
+    {
+        Map<String, Set<String>> placeholderMap = new Hashtable<>();
```
frankgh: Can we use a hashmap here?
```suggestion
        Map<String, Set<String>> placeholderMap = new HashMap<>();
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
```diff
@@ -0,0 +1,193 @@
...
+            placeholderMap.put(instanceMetadata.localSystemDataFileDir(),
+                               Collections.singleton(LOCAL_SYSTEM_DATA_FILE_DIR_PLACEHOLDER));
+        }
+
+        List<String> dataDirs = instanceMetadata.dataDirs();
+        for (int i = 0; i < dataDirs.size(); i++)
+        {
+            String dir = dataDirs.get(i);
+            Set<String> placeholders = Collections.unmodifiableSet(
+            new HashSet<>(Arrays.asList(DATA_FILE_DIR_PLACEHOLDER, DATA_FILE_DIR_PLACEHOLDER + "_" + i)));
```
frankgh: We can simplify like this:
```suggestion
            Set<String> placeholders = Set.of(DATA_FILE_DIR_PLACEHOLDER, DATA_FILE_DIR_PLACEHOLDER + "_" + i);
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
```diff
@@ -0,0 +1,193 @@
...
+    public static String localPath(@NotNull String fileUrl,
+                                   @NotNull InstanceMetadata metadata)
+    {
+
+        Map<String, String> urlToLocalDirMap = migrationUrlLocalDirMap(metadata);
+        for (Map.Entry<String, String> entry : urlToLocalDirMap.entrySet())
+        {
+            if (fileUrl.startsWith(entry.getKey()))
+            {
+                assert entry.getValue() != null : "No local path found for url " + fileUrl;
```
frankgh: this statement will only run when assertions are enabled. Do we instead want to throw a runtime exception?
```suggestion
                Objects.requireNonNull(entry.getValue(), () -> "No local path found for url "+ fileUrl);
```

server/src/test/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtilTest.java
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+
+    @Test
+    public void testLocalPath()
+    {
+        String cassandraHomeDir = tempDir.resolve("testLocalPath").toString();
+        InstanceMetadata instanceMetadata = getInstanceMetadata(cassandraHomeDir);
+
+        validateLocalPath(instanceMetadata.dataDirs().get(0) + "/" + FILE_NAME,
```
frankgh, line 214: We need to defend against directory traversal attacks. I'd like to see an input like `LIVE_MIGRATION_DATA_FILE_DIR_PATH + "/0/../../../../etc/passwd"` fail. However when I tried this, it is able to resolve these files.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFilesImpl.java
```diff
@@ -0,0 +1,145 @@
...
+                                       Set<String> filesToExclude, Set<String> dirsToExclude)
+    {
+        if (null == homeDir)
+        {
+            return Optional.empty();
+        }
+        Path homeDirPath = Paths.get(homeDir);
+        if (!Files.exists(homeDirPath))
+        {
+            return Optional.empty();
```
frankgh, line 122: can we log here that the directory does not exist?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFilesImpl.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        Objects.requireNonNull(placeholders);
+        Objects.requireNonNull(pathPrefix);
```
frankgh: NIT:
```suggestion
        Objects.requireNonNull(placeholders, "placeholders are required");
        Objects.requireNonNull(pathPrefix, "pathPrefix is required");
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationPlaceholderUtil.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+        if (placeholders.contains(placeholder))
+        {
+            return input.replace("${" + placeholder + "}", value);
+        }
+
+        return null;
+    }
+
+    public static boolean hasAnyPlaceholder(String input)
```
frankgh, line 77: can we add javadocs here?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationPlaceholderUtil.java
```diff
@@ -0,0 +1,91 @@
...
+    {
+        String placeholder = getPlaceholder(input);
+        return placeholder != null;
+    }
+
+    public static boolean hasPlaceholder(String input, Set<String> knownPlaceHolders)
```
frankgh: can we add javadocs here?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationPlaceholderUtil.java
```diff
@@ -0,0 +1,91 @@
...
+        String placeholder = getPlaceholder(input);
+        return placeholder != null && knownPlaceHolders.contains(placeholder);
+    }
+
+    private static @Nullable String getPlaceholder(String input)
+    {
+        Matcher matcher = PLACEHOLDER_PATTERN.matcher(input);
+        if (matcher.find())
+        {
+            return matcher.group(1);
```
frankgh, line 102: can there be more than one placeholder?

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFileInfo.java
nvharikrishna: done

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFileInfo.java
nvharikrishna: done

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
nvharikrishna: right, missed it!! Updated it.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
nvharikrishna: done

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
nvharikrishna: done

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
nvharikrishna: done

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFileInfo.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationApiEnableDisableHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationListInstanceFilesHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationListInstanceFilesHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationListInstanceFilesHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationListInstanceFilesHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationMap.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationMap.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationMapSidecarConfigImpl.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/modules/LiveMigrationModule.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/modules/LiveMigrationModule.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/modules/multibindings/VertxRouteMapKeys.java
nvharikrishna: Thanks! done.

nvharikrishna (commented):
@frankgh Addressed comments given so far. Will wait for further review.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/LiveMigrationConfigurationImpl.java
```diff
@@ -0,0 +1,71 @@
...
+    @JsonProperty("dirs_to_exclude")
+    public Set<String> directoriesToExclude()
+    {
+        return directoriesToExclude;
+    }
+
+    @JsonProperty("migration_map")
+    public Map<String, String> migrationMap()
+    {
+        return migrationMap;
```
frankgh, line 72: Can we make this return an unmodifiable map to prevent changes to the configuration?
```suggestion
        return Collections.unmodifiableMap(migrationMap);
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DirVisitor.java
```diff
@@ -0,0 +1,102 @@
...
+        return new InstanceFileInfo(fileUrl, size, fileType, lastModifiedTime);
+    }
+
+    /**
+     * Constructs live migration URL path for given file.
+     *
+     * @param path file for which URL needs to be constructed.
+     * @return URL path for given file
+     */
+    private String getInstanceFileUrl(@NotNull Path path)
```
frankgh: NIT
```suggestion
    private String buildInstanceFileUrl(@NotNull Path path)
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DirVisitor.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+        LOGGER.debug("Found {} instance files for homeDir {}", infos.size(), homeDir);
+        return infos;
+    }
+
+    private InstanceFileInfo toInstanceFileInfo(@NotNull Path path) throws IOException
+    {
+        long lastModifiedTime = Files.getLastModifiedTime(path).toMillis();
+        String fileUrl = getInstanceFileUrl(path);
+        InstanceFileInfo.FileType fileType = Files.isDirectory(path) ?
```
frankgh: we should avoid doing FS operations multiple times whenever possible. Can we only check for Files.isDirectory once and keep that information in a local variable? `boolean isDirectory = attrs.isDirectory();` . We could even simplify a bit further by doing `BasicFileAttributes attrs = Files.readAttributes(path, BasicFileAttributes.class);` and this information will give us both the last modified timestamp and whether the Path is a file or directory.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/MigrationFileVisitor.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        if (!dir.toAbsolutePath().toString().equals(homeDir))
+        {
+            validFiles.add(dir);
+        }
+        return FileVisitResult.CONTINUE;
+    }
+
+    public FileVisitResult visitFile(Path file, BasicFileAttributes attrs)
```
frankgh, line 86: can we add the missing @Override annotations to this class?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/MigrationFileVisitor.java
```diff
@@ -0,0 +1,120 @@
...
+    public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs)
+    {
+        if (shouldExcludeDir(dir))
+        {
+            return FileVisitResult.SKIP_SUBTREE;
+        }
+
+        if (!dir.toAbsolutePath().toString().equals(homeDir))
+        {
+            validFiles.add(dir);
```
frankgh, line 80: do we need to add it to the set here ? or adding files to the set in the visitFile sufficient?

frankgh (commented):
Just left a few more comments.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/LiveMigrationConfigurationImpl.java
nvharikrishna: Done while initializing member variable itself.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFilesImpl.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFilesImpl.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DirVisitor.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DirVisitor.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
nvharikrishna: HashMap allows null key and values. dirIndexMap should not have either null key or null value.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
nvharikrishna: HashMap allows null key and values. placeholderMap should not have either null key or null value.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
nvharikrishna: Thanks! Just realised that JDK 8 support is dropped for server. Done.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationPlaceholderUtil.java
nvharikrishna: input can have only one placeholder

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationPlaceholderUtil.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationPlaceholderUtil.java
nvharikrishna: Done. Also renamed the method to "hasAnyPlaceholder"

server/src/main/java/org/apache/cassandra/sidecar/livemigration/MigrationFileVisitor.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/MigrationFileVisitor.java
nvharikrishna: visitFile is only for files, want to add directory, so adding it in `preVisitDirectory`.

server/src/test/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtilTest.java
nvharikrishna: I am defending it in the file stream API (not part of this PR), but make sense to handle it as part of this utility too. Updated it. Added test cases too.

nvharikrishna (commented):
Addressed review comments.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
frankgh: in general use of Hashtable is discouraged. It's an old class and it is synchronized, the recommendation is the usage of HashMap. I don't see any null keys/values used in the scope of this method. Are you thinking the map will be updated outside of this method? And maybe if we don't expect changes to the map returned in this method we can return an unmodifiable map here?
```
return Collections.unmodifiableMap(dirIndexMap);
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
frankgh: same comment as above for the usage of Hashtable

frankgh (approved):
+1 Looks good to me

yifan-c:
> Making this change as part of [CEP-40](https://issues.apache.org/jira/browse/CASSSIDECAR-138).
> 
> Adding API for listing files of Cassnadra instance which is required during live migration.
> 
> [CASSSIDECAR-222](https://issues.apache.org/jira/browse/CASSSIDECAR-222)
> 
> Sample URL for listing files: `https://<host:port>/api/v1/live-migration/files-list`
> 
> Sample response:
> 
> ```
> {
>     "files": [
>         {
>             "fileUrl": "/api/v1/live-migration/files/data/0/system_schema/tables-afddfb9dbc1e30688056eed6c302ba09/nb-25-big-Filter.db",
>             "size": 24,
>             "fileType": "FILE",
>             "lastModifiedTime": 1744296699193
>         },
>         {
>             "fileUrl": "/api/v1/live-migration/files/data/0/system_schema/views-9786ac1cdd583201a7cdad556410c985",
>             "size": -1,
>             "fileType": "DIRECTORY",
>             "lastModifiedTime": 1744012964644
>         },
>         {
>             "fileUrl": "/api/v1/live-migration/files/data/0/system_schema/types-5a8b1ca866023f77a0459273d308917a",
>             "size": -1,
>             "fileType": "DIRECTORY",
>             "lastModifiedTime": 1744012964611
>         }
>     ],
>     "totalSize": 24
> }
> ```
> 
> fileUrl in above payload is in the format:
> 
> ```
> /api/v1/live-migration/files/<directory_type>/<directory_index>/<relative path from the directory>
> ```
> 
> fileUrl paths for different type of directories starts as follows:
> 
> ```
> data file dir: /api/v1/live-migration/files/data/0 
> # index will increase if C* instance has more data directories
> # For example: path prefix for second data file directory will look like this
> #   /api/v1/live-migration/files/data/1
> 
> commitlog dir: /api/v1/live-migration/files/commitlog/0
> hints dir: /api/v1/live-migration/files/hints/0
> saved caches dir: /api/v1/live-migration/files/saved_caches/0
> cdc raw dir: /api/v1/live-migration/files/cdc_raw/0
> local system data file dir: /api/v1/live-migration/files/local_system_data/0
> ```
> 
> Will raise a separate PR for supporting endpoint: `/api/v1/live-migration/files/`
> 
> Admin can exclude files or directories from Live Migration. Excluded files will not be listed by this files list API. Exclusions can be specified in the sidecar yaml configuration using glob or regex patterns. Admin can also use placeholders for different type of directories to avoid specifying full directory paths.
> 
> Here are the different placeholders supported:
> 
> ```
> COMMITLOG_DIR - commit log directory
> HINTS_DIR - hints directory
> SAVED_CACHES_DIR - saved caches directory
> CDC_RAW_DIR - cdc raw directory
> LOCAL_SYSTEM_DATA_FILE_DIR - local system data file data directory
> DATA_FILE_DIR - all data directories
> DATA_FILE_DIR_0 - first data directory
> DATA_FILE_DIR_1 - second data directory
> ```
> 
> Sample placeholder:
> 
> ```
> # Pattern to exclude snapshots directories in data folder: 
>     glob:${DATA_FILE_DIR}/*/*/snapshots/
> ```

Can you copy this API doc to the corresponding JIRA ticket for visibility purpose?

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -141,6 +141,27 @@ public final class ApiEndpointsV1
     public static final String STREAM_STATS_ROUTE = API_V1 + CASSANDRA + "/stats/streams";
     public static final String TABLE_STATS_ROUTE = API_V1 + CASSANDRA + PER_KEYSPACE + PER_TABLE + "/stats";
 
+
+    // Live Migration APIs
+    public static final String LIVE_MIGRATION_API_PREFIX = API_V1 + "/live-migration";
+
+    public static final String LIVE_MIGRATION_LIST_FILES_PATH = "/files-list";
```
yifan-c: To follow the REST api pattern in the project, the endpoint should be `GET /api/v1/.../files` instead of `/files-list`.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFileInfo.java
```diff
@@ -0,0 +1,78 @@
...
+    public final long size;
+    public final FileType fileType;
+
+    // The latest timestamp at which the file/dir was modified represented in milliseconds.
+    public final long lastModifiedTime;
+
+    public InstanceFileInfo(@JsonProperty("fileUrl") final String fileUrl,
+                            @JsonProperty("size") final long size,
+                            @JsonProperty("fileType") final FileType fileType,
+                            @JsonProperty("lastModifiedTime") final long lastModifiedTime)
```
yifan-c: nit: remove the unnecessary `final` modifiers

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFileInfo.java
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Class for holding information of a downloadable file during Live Migration.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class InstanceFileInfo
+{
+    public final String fileUrl;
+    // size is only relevant when the fileType is 'FILE'
+    public final long size;
```
yifan-c: nit: can you aslo document the value of `size` when the `fileType` is `DIRECTORY`?

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
```diff
@@ -0,0 +1,106 @@
...
+ *       "fileUrl": "/api/v1/live-migration/files/{type_of_directory}/{directory_index}/{relative_file_path}",
+ *       "size": -1,
+ *       "fileType": "DIRECTORY",
+ *       "lastModifiedTime": 1707798915999
+ *     }
+ *   ]
+ * }
+ * </pre>
+ */
+public class InstanceFilesListResponse
```
yifan-c: 👍

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
```diff
@@ -0,0 +1,106 @@
...
+                long size = file.size;
+                sum += size;
+            }
+        }
+        totalSize = sum;
+    }
+
+    @JsonCreator
+    public InstanceFilesListResponse(@JsonProperty("files") final List<InstanceFileInfo> files,
+                                     @JsonProperty("totalSize") final long totalSize)
```
yifan-c: nit: remove the `final`s for the arguments.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
```diff
@@ -0,0 +1,106 @@
...
+ *       "fileType": "DIRECTORY",
+ *       "lastModifiedTime": 1707798915999
+ *     }
+ *   ]
+ * }
+ * </pre>
+ */
+public class InstanceFilesListResponse
+{
+    private final List<InstanceFileInfo> files;
```
yifan-c, line 58: nit: consider making an immutable copy for the list.

conf/sidecar.yaml
```diff
@@ -362,3 +362,10 @@ s3_client:
 #    uri:
 #    username:
 #    password:
+
+live_migration:
+  special_files_to_exclude: []
+  special_dirs_to_exclude:
```
yifan-c: nit: "special" can be removed, as it does not add context to the setting names. `files_to_exclude` means the same as `special_files_to_exclude`

conf/sidecar.yaml
```diff
@@ -362,3 +362,10 @@ s3_client:
...
+    - ${DATA_FILE_DIR}/*/*/snapshots # Excludes snapshot directories to copy to destination
+  migration_map:
+    localhost1: localhost4
```
yifan-c: can you add document to explain what this map expects?

server/src/main/java/org/apache/cassandra/sidecar/config/LiveMigrationConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Directories to be excluded from Live Migration.
+     * @return set of directory exclusion patterns.
+     */
+    Set<String> directoriesToExclude();
+
+    /**
+     * Map of source and destination Cassandra instances to migrate.
+     *
+     * @return Map of strings where key is the source instance name and value is the destination instance name.
```
yifan-c: Do you mean `instance hostname` instead of `instance name`?

server/src/main/java/org/apache/cassandra/sidecar/config/LiveMigrationConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+     */
+    Map<String, String> migrationMap();
+}
```
yifan-c: It is not covered in CEP-40, but reading this patch makes me concerning about the operation complexity. If the live migration configuration is expressed in the form of yaml config, it would require rolling restart of sidecar to pick up the configurations. 
If my understanding is correct, live migration is a task, the task configuration should be per task, not per cluster (when it is stored in sidecar.yaml). If you agree with me, a more suitable place for storing the task configuration is in a table in sidecar internal keyspace.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/LiveMigrationConfigurationImpl.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public LiveMigrationConfigurationImpl()
+    {
+        this(Collections.emptySet(), Collections.emptySet(), Collections.emptyMap());
+    }
+
+    @JsonCreator
+    public LiveMigrationConfigurationImpl(@JsonProperty("files_to_exclude") Set<String> filesToExclude,
+                                          @JsonProperty("dirs_to_exclude") Set<String> directoriesToExclude,
+                                          @JsonProperty("migration_map") Map<String, String> migrationMap)
```
yifan-c, line 47: The code refers to the fields as `files_to_exclude` and `dirs_to_exclude` already. Is there a test to check the yaml file? Feel like not, otherwise, the test would have failed. 
That being said, I lean towards persisting the configuration for live migration tasks in a table.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationMap.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public interface LiveMigrationMap
+{
+    /**
+     * Tells whether given instance is configured as source or not.
+     *
+     * @param instanceMeta Cassandra instance metadata
+     * @return true if given instance is configured as source for live migration.
+     */
+    default boolean isSource(@NotNull final InstanceMetadata instanceMeta)
```
yifan-c: nit: please remove `final`

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationMap.java
```diff
@@ -0,0 +1,72 @@
...
+    {
+        return isSource(instanceMeta) || isDestination(instanceMeta);
+    }
+
+    /**
+     * Returns map of source and destination
+     *
+     * @return map of source and destination
+     */
+    Map<String, String> getMigrationMap();
```
yifan-c, line 71: can this ever return null? if so, please annotate `@Nullable`. I would suggest avoid `null` values at the best, and _optionally_ annotate `@NotNull`. It would simplify the callsites, e.g. `isDestination`

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFiles.java
```diff
@@ -0,0 +1,41 @@
...
+{
+    /**
+     * Lists files belong to a Cassandra instance after excluding default set files to exclude
+     * {@link LiveMigrationConfiguration#filesToExclude()} and default set of directories to exclude
+     * {@link LiveMigrationConfiguration#directoriesToExclude()}.
+     *
+     * @return list of files a Cassandra instance has after excluding default files and folders.
+     * @throws IOException when failed to read any file or folder.
+     */
+    List<InstanceFileInfo> getFiles() throws IOException;
```
yifan-c: Depending on the FS, the number of files could exceed what `List` can hold, which is at most `int.MAX` number of elements. This extreme case is not likely to happen, but it would be good to consider it when designing the interface. For example, document that if there are exceeding number of files, it would truncate to at most `int.MAX`. Or, the method can return `Iterator` instead. 
Taking another look at the "list files" handler, since it is likely to have a lot of files on disk, I am wondering how large the response do we expect? Would it cause OOM or memory pressure?

yifan-c (commented):
Submitting my comments so far. 
Besides the inline comments, please do not forget to update `CHANGES.txt`

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
nvharikrishna: done

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFileInfo.java
nvharikrishna: done

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFileInfo.java
nvharikrishna: done

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
nvharikrishna: Initializing with unmodifiable list in the constructor.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/InstanceFilesListResponse.java
nvharikrishna: done

conf/sidecar.yaml
nvharikrishna: Thanks! it is a miss from my side to update this config.

conf/sidecar.yaml
nvharikrishna: done, also comment this entry as it enables migration.

server/src/main/java/org/apache/cassandra/sidecar/config/LiveMigrationConfiguration.java
nvharikrishna: yes, updated it

server/src/main/java/org/apache/cassandra/sidecar/config/LiveMigrationConfiguration.java
nvharikrishna: This configuration cannot be per task. Per task - could mean that one source host can be specified in one request and another source in the next request. The map should be determined by the admin/operator but not the caller. Accepting remote host as part of the migration request becomes security risk.

Cassandra won't be running when the sidecar is started at the destination. So it cannot fetch sidecar internal keyspace for the config. Does sidecar running in one host talks to Cassandra running in other hosts?

I would say this configuration is per cluster as it tells which sidecars are going to allow these endpoints/migration and which sidecars are not going to allow. 

> If the live migration configuration is expressed in the form of yaml config, it would require rolling restart of sidecar to pick up the configurations.

Agree that it would be painful, but I couldn't see any other way at the moment. Probably, sidecar supporting hot reloading of configuration would help.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/LiveMigrationConfigurationImpl.java
nvharikrishna: Good catch, thanks! added unit test.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationMap.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationMap.java
nvharikrishna: LiveMigrationConfigurationImpl default constructor is using empty sets and maps. By the way I see sidecar configuration is constructed, I don't think it will be null.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFiles.java
nvharikrishna: Tested with 70K files few months back, don't recollect the numbers, will test it again. At least I recollect that I did not find abnormal gc pauses with 8gb heap (could be subjective to heap environment/size).

As far as I know JDK11 doesn't support iterative file walking. List documentation says that `If this list contains more than Integer.MAX_VALUE elements, returns Integer.MAX_VALUE.`, I guess it itself acts as a limiting factor.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
nvharikrishna: Ran into null keys while unit testing (as some of the directories are optional). Changed it from HashMap to HashTable to catch null keys early. Null checks are already added to optional values. Changed it to HashMap and wrapped it with unmodifiable before returning.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
nvharikrishna: Updated it.

server/src/main/java/org/apache/cassandra/sidecar/config/LiveMigrationConfiguration.java
yifan-c: Thanks for the explanation! However, I am not convinced. 

> The map should be determined by the admin/operator but not the caller. Accepting remote host as part of the migration request becomes security risk.

Who is the target audience of this feature? I think they are the admin/operators. The security risk should not be resolved by how task is configured, but via authn & authz. In fact, this patch contains the access control already. Cluster admins can configure that only admins have the permission to invoke the APIs. It is a requirement already. 

> Cassandra won't be running when the sidecar is started at the destination. So it cannot fetch sidecar internal keyspace for the config. Does sidecar running in one host talks to Cassandra running in other hosts?

Sidecar has access to the pairing cassandra cluster only, but not other clusters. You are right, that the sidecar instances in the target cluster won't have access to Cassandra. 
Now, I am just thinking out. In CEP-40, the approach is a pull model, i.e. target sidecar instances pull from source cluster. What if the control is inverted? Briefly, in this sequence,
1. Operator submit task to source cluster (sidecar) via an API and the config is persisted
2. [Optional] an explicit step to start live migration
3. Each source sidecar instance behaves as a coordinator, who performs the steps of sync describe in CEP-40, i.e. contacts the assigned sidecar instance (according to the mapping in the config) to do the initial and last pull, validation, etc.

Does it sound right to you?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFiles.java
yifan-c: > iterative file walking

It should be achievable. I do not feel strong on it. Would like to avoid early optimization. 

Yep, thanks for adding the doc. That is good enough for now.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFilesImpl.java
```diff
@@ -0,0 +1,145 @@
...
+
+    public CassandraInstanceFilesImpl(InstanceMetadata instanceMetadata,
+                                      LiveMigrationConfiguration configuration)
+    {
+        this.instanceMetadata = instanceMetadata;
+        this.configuration = configuration;
+    }
+
+    @Override
+    public List<InstanceFileInfo> getFiles() throws IOException
```
yifan-c: nit: the code style prefer omitting `get` from `getters`. I will _not_ insist if you feel the current name is fine. But, if you do want to update, please take care of the other getters, e.g. `getDirVisitorList` in this class.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFilesImpl.java
```diff
@@ -0,0 +1,145 @@
...
+
+        Set<PathMatcher> fileExclusionMatchers = toPathMatchers(filesToExclude, placeholders, homeDir);
+        Set<PathMatcher> dirExclusionMatchers = toPathMatchers(dirsToExclude, placeholders, homeDir);
+
+        return Optional.of(new DirVisitor(homeDir, pathPrefix, fileExclusionMatchers, dirExclusionMatchers));
+    }
+
+    Set<PathMatcher> toPathMatchers(Set<String> exclusions, Set<String> homeDirPlaceholders, String homeDir)
+    {
+        if (null == exclusions)
```
yifan-c: nit: check empty case too. 

```suggestion
        if (null == exclusions || exclusions.isEmpty())
```

conf/sidecar.yaml
```diff
@@ -362,3 +362,10 @@ s3_client:
 #    uri:
 #    username:
 #    password:
+
+live_migration:
+  files_to_exclude: []
+  dirs_to_exclude:
+    - ${DATA_FILE_DIR}/*/*/snapshots # Excludes snapshot directories to copy to destination
```
yifan-c: can you please also add documents regarding the other available "placeholders"? Basically, we need to let end users know how to configure w/o having to look at the code.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFilesImpl.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        Path homeDirPath = Paths.get(homeDir);
+        if (!Files.exists(homeDirPath))
+        {
+            LOGGER.info("Directory {} does not exist.", homeDir);
+            return Optional.empty();
+        }
+
+        Objects.requireNonNull(placeholders, "placeholders are required");
+        Objects.requireNonNull(pathPrefix, "pathPrefix is required");
```
yifan-c: nit: move those 2 checks before the files existence check; can save a bit of I/O if any of the condition fails.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationPlaceholderUtil.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+public class LiveMigrationPlaceholderUtil
+{
+
+    public static final String DATA_FILE_DIR_PLACEHOLDER = "DATA_FILE_DIR";
+    public static final String COMMITLOG_DIR_PLACEHOLDER = "COMMITLOG_DIR";
+    public static final String HINTS_DIR_PLACEHOLDER = "HINTS_DIR";
+    public static final String SAVED_CACHES_DIR_PLACEHOLDER = "SAVED_CACHES_DIR";
+    public static final String CDC_RAW_DIR_PLACEHOLDER = "CDC_RAW_DIR";
+    public static final String LOCAL_SYSTEM_DATA_FILE_DIR_PLACEHOLDER = "LOCAL_SYSTEM_DATA_FILE_DIR";
```
yifan-c, line 40: nit: they could be enum. It would make contains check easier to reason about.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/MigrationFileVisitor.java
```diff
@@ -0,0 +1,123 @@
...
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(MigrationFileVisitor.class);
+    private final String homeDir;
+    private final List<Path> validFiles;
+    private final Set<PathMatcher> filesToExclude;
+    private final Set<PathMatcher> directoriesToExclude;
+
+    public MigrationFileVisitor(@NotNull final String homeDir,
+                                @NotNull final Set<PathMatcher> filesToExcludeMatchers,
+                                @NotNull final Set<PathMatcher> directoriesToExcludeMatchers)
```
yifan-c: 1. remove `final`s
2. I think there is no guarantee that implementation of `PathMatcher` implements hashcode and equals. I am not sure of the benefit of using a set to hold the `PathMatcher`s. I think a list would just do fine.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFilesImpl.java
```diff
@@ -0,0 +1,145 @@
...
+
+        Set<PathMatcher> fileExclusionMatchers = toPathMatchers(filesToExclude, placeholders, homeDir);
+        Set<PathMatcher> dirExclusionMatchers = toPathMatchers(dirsToExclude, placeholders, homeDir);
+
+        return Optional.of(new DirVisitor(homeDir, pathPrefix, fileExclusionMatchers, dirExclusionMatchers));
+    }
+
+    Set<PathMatcher> toPathMatchers(Set<String> exclusions, Set<String> homeDirPlaceholders, String homeDir)
```
yifan-c: I should probably make the comment here on the usage of `set`.

yifan-c (commented):
final batch of comments

server/src/main/java/org/apache/cassandra/sidecar/livemigration/MigrationFileVisitor.java
nvharikrishna: make sense, done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFilesImpl.java
nvharikrishna: updated it

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFilesImpl.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFilesImpl.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationPlaceholderUtil.java
nvharikrishna: Some of the placeholder are constructed dynamically based on no.of dirs available (DATA_FILE_DIR_0, DATA_FILE_DIR_1). I feel using String constants is simple and satisfies the requirements. Keeping them as string constants.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/CassandraInstanceFilesImpl.java
nvharikrishna: done

conf/sidecar.yaml
nvharikrishna: done

nvharikrishna (commented):
Addressed review comments.

server/src/main/java/org/apache/cassandra/sidecar/config/LiveMigrationConfiguration.java
nvharikrishna: > Sidecar has access to the pairing cassandra cluster only

This migration talks about migrating Cassandra instance with in a cluster rather than moving nodes from one cluster to another.

> Each source sidecar instance behaves as a coordinator, who performs the steps of sync describe in CEP-40,

During the data copy, destination must delete files no long present at source. In pull model scenarios, Source instances should handle deletions. I’m not entirely comfortable with this approach because ideally, Sidecars should manage tasks within their host. Allowing other Sidecars to have knowledge and involved without overstepping control could lead to unintended results.

In future, supporting migration from different sources like S3 may be required. In this case, a destination can just come up, pull the data, and start C* (discussed in the CEP-40
discussion thread).

yifan-c: > This migration talks about migrating Cassandra instance with in a cluster rather than moving nodes from one cluster to another.

Thanks for the correction. I was under the wrong impression of the migration. 

It is simpler when migrating within a cluster coordinated by Sidecar. The sidecar instances (both source and target) have access to Cassandra. The motivation of my idea is to avoid the operational pain from the configuration approach. The equivalent configuration can be persisted in the database. 

> In pull model scenarios, Source instances should handle deletions.

I guess you meant the "coordinated" scenario that I described? The target Sidecar instances can still run diff and delete in the coordinated mode. For example, a coordinator in Source issues a sync to the sidecar instance in Target and ask it to check diff and sync (delete files no longer present in source, and download files not yet downloaded) according to the persisted configuration. Vast majority of the flow is the same as described in CEP-40. The advantage is having all the control via the API (e.g. operators do not need to update configuration in the source instances). 

Or, we can also stick with the idea of having the "tooling" to coordinate, as depicted in CEP-40. In that case, the only difference is storing configuration in Cassandra. It should achieve the idea of not bouncing sidecar instances. 

That being said, it is not necessary to block on the merge of this patch (as you are adding the patches of the feature). Please re-think about the suggestion. We can talk more offline. 

> supporting migration from different sources like S3

I am not sure how S3 fits in the live-migration scenario. If you are thinking about restoring from a snapshot on S3, it is not live. If you are thinking about downloading from analytics bulk written data, it is not migration. The goal of live-migration is to moving data from one Cassandra node to another with zero downtime.

nvharikrishna: > The sidecar instances (both source and target) have access to Cassandra.

Assuming that Sidecar interacts only within local instances, the destination will not gain access to Cassandra until the data is copied, validated, and Cassandra starts running. The Cassandra instance at the source needs to be shut down before starting the final data copy. After this point, the source sidecar will no longer have access to Cassandra.

If Sidecar communicates with multiple Cassandra instances, the `driver_parameters.contact_points` list for the CQL session provider must be updated to include new nodes so that it can talk to new instances while the old nodes go down. This process needs config change and bounce before starting migrations to recognize the new contact points which is important if I migrate all nodes in the cluster or dc. Correct me if I made wrong assumptions.

Having this configuration independent of Cassandra (or any other component) allows to do the migration even when source/partial cluster/full cluster is down. I feel this offers good flexibility.

The concern about rolling out Sidecar to recognize migration config is valid (concern for me too), but I suggest approaching this by hot reloading the configuration instead of storing it in Cassandra as this feature involves changing the nodes of the cluster itself.

Let me know what you think.

PR #213, opened 2025-04-09
CASSSIDECAR-236 Fixing incorrect creation of MinuteBoundConfiguration & MillisecondBoundConfiguration in CdcConfigImpl

jyothsnakonisa:


server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcConfigImpl.java
```diff
@@ -179,7 +179,7 @@ public MinuteBoundConfiguration watermarkWindow()
     {
         // this prop sets the maximum duration age accepted by CDC, any mutations with write timestamps older than
         // the watermark window will be dropped with log message "Exclude the update due to out of the allowed time window."
-        return new MinuteBoundConfiguration(getInt(ConfigKeys.WATERMARK_SECONDS.lowcaseName, DEFAULT_WATERMARK_WINDOW), TimeUnit.SECONDS);
+        return new MinuteBoundConfiguration(getInt(ConfigKeys.WATERMARK_SECONDS.lowcaseName, DEFAULT_WATERMARK_WINDOW) / 60, TimeUnit.MINUTES);
```
bbotella: Should we change the resolution to Seconds instead? What if someone configures 30 seconds for the watermark? That integer would be truncated to zero right?

jyothsnakonisa: We can change that but should that be minute granularity or seconds granularity is the question. Also, note that it chages the API if we change it to SecondBound

bbotella: Then we should change the ConfigKey to WATERMARK_MINUTES.

jyothsnakonisa: We can, but that might break any existing configs that are being used

jyothsnakonisa: I think changing the MinuteBound to SecondBound is better compared to changing configkey as it will break at compile time.

server/src/test/java/org/apache/cassandra/sidecar/cdc/CdcConfigImplTest.java
```diff
@@ -118,6 +120,8 @@ void testReturnDefaultValuesWhenConfigsAreEmpty()
         assertThat(cdcConfig.env()).isEqualTo("");
         assertThat(cdcConfig.kafkaTopic()).isNull();
         assertThat(cdcConfig.logOnly()).isFalse();
+        assertThat(cdcConfig.watermarkWindow()).isEqualTo(new MinuteBoundConfiguration(72, TimeUnit.HOURS));
```
yifan-c: should this be `SecondBoundConfiguration`?

server/src/test/java/org/apache/cassandra/sidecar/cdc/CdcConfigImplTest.java
```diff
@@ -137,6 +146,8 @@ void testConfigsWhenConfigsAreNotEmpty()
         assertThat(cdcConfig.env()).isEqualTo("if");
         assertThat(cdcConfig.kafkaTopic()).isEqualTo("topic1");
         assertThat(cdcConfig.logOnly()).isFalse();
+        assertThat(cdcConfig.watermarkWindow()).isEqualTo(new MinuteBoundConfiguration(2, TimeUnit.MINUTES));
```
yifan-c: and this. should it be `SecondBoundConfiguration`?

yifan-c (commented):
LGTM overall. Left a question on the test.

PR #215, opened 2025-04-19
CASSSIDECAR-223: Create API that allows downloading of files listed by Live Migration

nvharikrishna:
Added endpoint for transferring files during live migration. The source C* instance's sidecar will serve this API and the destination sidecar uses it download files.

CASSSIDECAR-223

### DIFF
--- a/client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
+++ b/client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
@@ -150,6 +150,10 @@ public final class ApiEndpointsV1
 
     public static final String DIR_TYPE_PARAM = "dirType";
     public static final String DIR_INDEX_PARAM = "dirIndex";
+    // API endpoint allows files transfer of specific directories handled by Cassandra during live migration operation.
+    // dirType path parameter    : The type of directory (data, commitlog, etc.)
+    // dirIndex path parameter   : The index of the directory
+    // The remaining path ('/*') : Represents the relative path within the specified directory
     public static final String LIVE_MIGRATION_FILE_TRANSFER_API = LIVE_MIGRATION_FILES_API + "/:" + DIR_TYPE_PARAM
                                                                   + "/:" + DIR_INDEX_PARAM + "/*";
 
--- a/client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
+++ b/client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
@@ -18,6 +18,13 @@
 
 package org.apache.cassandra.sidecar.common;
 
+import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.LiveMigrationDirType.CDC_RAW_DIR;
+import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.LiveMigrationDirType.COMMIT_LOG_DIR;
+import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.LiveMigrationDirType.DATA_FIlE_DIR;
+import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.LiveMigrationDirType.HINTS_DIR;
+import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.LiveMigrationDirType.LOCAL_SYSTEM_DATA_FILE_DIR;
+import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.LiveMigrationDirType.SAVED_CACHES_DIR;
+
 /**
  * A constants container class for API endpoints of version 1.
  */
@@ -148,13 +155,50 @@ public final class ApiEndpointsV1
 
     public static final String LIVE_MIGRATION_FILES_API = LIVE_MIGRATION_API_PREFIX + "/files";
 
-    public static final String LIVE_MIGRATION_CDC_RAW_DIR_PATH = LIVE_MIGRATION_FILES_API + "/cdc_raw";
-    public static final String LIVE_MIGRATION_COMMITLOG_DIR_PATH = LIVE_MIGRATION_FILES_API + "/commitlog";
-    public static final String LIVE_MIGRATION_DATA_FILE_DIR_PATH = LIVE_MIGRATION_FILES_API + "/data";
-    public static final String LIVE_MIGRATION_HINTS_DIR_PATH = LIVE_MIGRATION_FILES_API + "/hints";
+    public static final String DIR_TYPE_PARAM = "dirType";
+    public static final String DIR_INDEX_PARAM = "dirIndex";
+    public static final String LIVE_MIGRATION_FILE_TRANSFER_API = LIVE_MIGRATION_FILES_API + "/:" + DIR_TYPE_PARAM
+                                                                  + "/:" + DIR_INDEX_PARAM + "/*";
+
+    /**
+     * Enum for holding different type of directories handled by Live Migration.
+     */
+    public enum LiveMigrationDirType
+    {
+        CDC_RAW_DIR("cdc_raw"),
+        COMMIT_LOG_DIR("commitlog"),
+        DATA_FIlE_DIR("data"),
+        HINTS_DIR("hints"),
+        LOCAL_SYSTEM_DATA_FILE_DIR("local_system_data"),
+        SAVED_CACHES_DIR("saved_caches");
+
+        private final String dirType;
+
+        LiveMigrationDirType(String dirType)
+        {
+            this.dirType = dirType;
+        }
+
+        public static LiveMigrationDirType find(String dirType)
+        {
+            for (LiveMigrationDirType type : values())
+            {
+                if (type.dirType.equals(dirType))
+                {
+                    return type;
+                }
+            }
+            return null;
+        }
+    }
+
+    public static final String LIVE_MIGRATION_CDC_RAW_DIR_PATH = LIVE_MIGRATION_FILES_API + "/" + CDC_RAW_DIR.dirType;
+    public static final String LIVE_MIGRATION_COMMITLOG_DIR_PATH = LIVE_MIGRATION_FILES_API + "/" + COMMIT_LOG_DIR.dirType;
+    public static final String LIVE_MIGRATION_DATA_FILE_DIR_PATH = LIVE_MIGRATION_FILES_API + "/" + DATA_FIlE_DIR.dirType;
+    public static final String LIVE_MIGRATION_HINTS_DIR_PATH = LIVE_MIGRATION_FILES_API + "/" + HINTS_DIR.dirType;
     public static final String LIVE_MIGRATION_LOCAL_SYSTEM_DATA_FILE_DIR_PATH = LIVE_MIGRATION_FILES_API
-                                                                                + "/local_system_data";
-    public static final String LIVE_MIGRATION_SAVED_CACHES_DIR_PATH = LIVE_MIGRATION_FILES_API + "/saved_caches";
+                                                                                + "/" + LOCAL_SYSTEM_DATA_FILE_DIR.dirType;
+    public static final String LIVE_MIGRATION_SAVED_CACHES_DIR_PATH = LIVE_MIGRATION_FILES_API + "/" + SAVED_CACHES_DIR.dirType;
 
 
     private ApiEndpointsV1()
--- a/client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
+++ b/client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
@@ -18,13 +18,6 @@
 
 package org.apache.cassandra.sidecar.common;
 
-import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.LiveMigrationDirType.CDC_RAW_DIR;
-import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.LiveMigrationDirType.COMMIT_LOG_DIR;
-import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.LiveMigrationDirType.DATA_FIlE_DIR;
-import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.LiveMigrationDirType.HINTS_DIR;
-import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.LiveMigrationDirType.LOCAL_SYSTEM_DATA_FILE_DIR;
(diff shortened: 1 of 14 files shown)


client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -147,13 +154,50 @@ public final class ApiEndpointsV1
...
-    public static final String LIVE_MIGRATION_HINTS_DIR_PATH = LIVE_MIGRATION_FILES_API + "/hints";
+    public static final String DIR_TYPE_PARAM = "dirType";
+    public static final String DIR_INDEX_PARAM = "dirIndex";
+    public static final String LIVE_MIGRATION_FILE_TRANSFER_API = LIVE_MIGRATION_FILES_API + "/:" + DIR_TYPE_PARAM
+                                                                  + "/:" + DIR_INDEX_PARAM + "/*";
+
+    /**
+     * Enum for holding different type of directories handled by Live Migration.
+     */
+    public enum LiveMigrationDirType
```
frankgh: this enum feels misplaced and also the paths are not really being used to build the route. Ideally, we should only have the routes that we consume in the `ApiEndpointsV1` class. I would keep the `LiveMigrationDirType` enum in the `livemigration` package, and then leverage the enum in `org.apache.cassandra.sidecar.livemigration.LiveMigrationInstanceMetadataUtil`

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
```diff
@@ -79,4 +79,5 @@ public class BasicPermissions
 
     // Live Migration permissions
     public static final Permission LIST_FILES = new DomainAwarePermission("LIVE_MIGRATION:LIST_FILES", CLUSTER_SCOPE);
+    public static final Permission STREAM_FILES = new DomainAwarePermission("LIVE_MIGRATION:STREAM", CLUSTER_SCOPE);
```
frankgh, line 86: I would keep a single permission for the feature , similar to what we have for CDC

frankgh (commented):
Added some comments , but I still need to review a few more files

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
nvharikrishna: Created a separate permission so that STREAM_FILES can be given only to Sidecars and LIST_FILES can be given to users. Doesn't it make sense?

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
nvharikrishna: Thanks!, sounds better.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
yifan-c: In what circumstances, you want to revoke the stream permission? 
According to the diagram in [CEP-40](https://cwiki.apache.org/confluence/display/CASSANDRA/CEP-40%3A+Data+Transfer+Using+Cassandra+Sidecar+for+Live+Migrating+Instances), list files API is invoked by sidecars, not users, right?

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -147,14 +147,10 @@ public final class ApiEndpointsV1
...
-    public static final String LIVE_MIGRATION_DATA_FILE_DIR_PATH = LIVE_MIGRATION_FILES_API + "/data";
-    public static final String LIVE_MIGRATION_HINTS_DIR_PATH = LIVE_MIGRATION_FILES_API + "/hints";
-    public static final String LIVE_MIGRATION_LOCAL_SYSTEM_DATA_FILE_DIR_PATH = LIVE_MIGRATION_FILES_API
-                                                                                + "/local_system_data";
-    public static final String LIVE_MIGRATION_SAVED_CACHES_DIR_PATH = LIVE_MIGRATION_FILES_API + "/saved_caches";
-
+    public static final String DIR_TYPE_PARAM = "dirType";
+    public static final String DIR_INDEX_PARAM = "dirIndex";
+    public static final String LIVE_MIGRATION_FILE_TRANSFER_API = LIVE_MIGRATION_FILES_API + "/:" + DIR_TYPE_PARAM
+                                                                  + "/:" + DIR_INDEX_PARAM + "/*";
```
yifan-c, line 158: nit: Please add a comment in the code to explain why match `/*`.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationDirType.java
```diff
@@ -0,0 +1,51 @@
...
+    {
+        for (LiveMigrationDirType type : values())
+        {
+            if (type.dirType.equals(dirType))
+            {
+                return type;
+            }
+        }
+        return null;
+    }
```
yifan-c: You could look up much faster from a pre-defined hashmap, instead of looping through all values.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileStreamHandler.java
```diff
@@ -0,0 +1,243 @@
...
+                                  @Nullable Void request)
+    {
+        String reqPath = URLDecoder.decode(rc.request().path(), StandardCharsets.UTF_8);
+
+        if (reqPath.contains("/../") || reqPath.endsWith("/.."))
+        {
+            LOGGER.warn("Tried to access file using relative path({}). Rejecting the request.", reqPath);
+            rc.response().setStatusCode(HttpResponseStatus.BAD_REQUEST.code()).end();
+            return;
+        }
```
yifan-c, line 130: In terms of validation, how is this section different from the section below? `reqPath` is same as `normalizedPath`, right?

```
        String normalizedPath = rc.normalizedPath();
        String localFile;

        try
        {
            localFile = LiveMigrationInstanceMetadataUtil.localPath(normalizedPath, instanceMeta);
        }
        catch (IllegalArgumentException e)
        {
            LOGGER.warn("Invalid path", e);
            rc.response().setStatusCode(HttpResponseStatus.NOT_FOUND.code()).end();
            return;
        }
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileStreamHandler.java
```diff
@@ -0,0 +1,243 @@
...
+                                          ExecutorPools executorPools,
+                                          CassandraInputValidator validator,
+                                          SidecarConfiguration sidecarConfiguration)
+    {
+        super(metadataFetcher, executorPools, validator);
+        this.liveMigrationConfiguration = sidecarConfiguration.liveMigrationConfiguration();
+    }
+
+    @Override
+    protected Void extractParamsOrThrow(RoutingContext context)
```
yifan-c, line 89: Looks like this method is only used for some validation. Not really value extraction.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileStreamHandler.java
```diff
@@ -0,0 +1,243 @@
...
+            rc.response().setStatusCode(HttpResponseStatus.NOT_FOUND.code()).end();
+            return;
+        }
+
+        Path path = Paths.get(localFile);
+
+        if (!Files.exists(path))
+        {
+            LOGGER.info("File {} not found.", localFile);
+            rc.response().setStatusCode(HttpResponseStatus.NOT_FOUND.code()).end();
```
yifan-c, line 165: return early here?

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileStreamHandler.java
```diff
@@ -0,0 +1,243 @@
...
+            LOGGER.warn("Invalid path", e);
+            rc.response().setStatusCode(HttpResponseStatus.NOT_FOUND.code()).end();
+            return;
+        }
+
+        Path path = Paths.get(localFile);
+
+        if (!Files.exists(path))
+        {
+            LOGGER.info("File {} not found.", localFile);
```
yifan-c: we use the `kv` log format consistently in the project. The log message could be this. Please also updated the other log messages in this file. 

```suggestion 
            LOGGER.info("Requested file is not found. file={}", localFile);
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileStreamHandler.java
```diff
@@ -0,0 +1,243 @@
...
+        }
+        if (isExcluded(path, instanceMeta))
+        {
+            LOGGER.debug("Requested file {} or one of its parent directories is excluded from Live Migration.", reqPath);
+            rc.response().setStatusCode(HttpResponseStatus.NOT_FOUND.code()).end();
+            return;
+        }
+
+        rc.put(FileStreamHandler.FILE_PATH_CONTEXT_KEY, localFile);
+        rc.next();
```
yifan-c: nit: dispatch the checks to `executorPools.service()`? I am not super concerned about the I/O access, but the `isExcluded` check seems heavy. 
On success, call `rc.next()`. On failure, do nothing, as `rc` should be ended already.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
```diff
@@ -54,6 +55,13 @@
 public class LiveMigrationInstanceMetadataUtil
 {
 
+    public static final String LIVE_MIGRATION_CDC_RAW_DIR_PATH = ApiEndpointsV1.LIVE_MIGRATION_FILES_API + "/" + CDC_RAW_DIR.dirType;
+    public static final String LIVE_MIGRATION_COMMITLOG_DIR_PATH = ApiEndpointsV1.LIVE_MIGRATION_FILES_API + "/" + COMMIT_LOG_DIR.dirType;
+    public static final String LIVE_MIGRATION_DATA_FILE_DIR_PATH = ApiEndpointsV1.LIVE_MIGRATION_FILES_API + "/" + DATA_FIlE_DIR.dirType;
+    public static final String LIVE_MIGRATION_HINTS_DIR_PATH = ApiEndpointsV1.LIVE_MIGRATION_FILES_API + "/" + HINTS_DIR.dirType;
+    public static final String LIVE_MIGRATION_LOCAL_SYSTEM_DATA_FILE_DIR_PATH = ApiEndpointsV1.LIVE_MIGRATION_FILES_API
+                                                                                + "/" + LOCAL_SYSTEM_DATA_FILE_DIR.dirType;
+    public static final String LIVE_MIGRATION_SAVED_CACHES_DIR_PATH = ApiEndpointsV1.LIVE_MIGRATION_FILES_API + "/" + SAVED_CACHES_DIR.dirType;
```
yifan-c, line 64: Why they are moved out of `ApiEndpointsV1.java`?

frankgh: it's based on this feedback: https://github.com/apache/cassandra-sidecar/pull/215#discussion_r2155736361

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
nvharikrishna: Proposing to keep the stream permission to Sidecar only by default. From CEP-40, 

image
list APIS can be exposed to admin/operator (for debugging purposes).  admin/operator can check the list, but may not be required to download the files.

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileStreamHandler.java
nvharikrishna: right, similar to any other handler extending `AbstractHandler<Void>` it uses the arguments provided as part of 'handleInternal' method.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileStreamHandler.java
nvharikrishna: request path can look like this `/api/v1/live-migration/files/data/0/ks/tb-1234/../../secrets`, but normalized path will not have `..`. [[more](https://vertx.io/docs/4.0.3/apidocs/io/vertx/ext/web/RoutingContext.html#normalizedPath--)]

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileStreamHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileStreamHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileStreamHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationDirType.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
yifan-c: Make sense. Consider adding comment in the live migration permissions to explain the expected scenario.

PR #217, opened 2025-05-16
CASSSIDECAR-247 add APP_HOME prefix for log config and log dir in cassandra-sidecar startup script

smiklosovic:
…tartup script


frankgh (approved):
+1

PR #221, opened 2025-06-04
CASSSIDECAR-255: Fix restore ranges of failed sidecar-managed restore jobs do not persist

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-255

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobConsistencyChecker.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobConsistencyChecker.java
@@ -100,6 +100,7 @@ public class RestoreJobConsistencyChecker
         RestoreJobProgressCollector collector = RestoreJobProgressCollectors.create(restoreJob, fetchPolicy);
         RestoreRangeStatus successCriteria = restoreJob.expectedNextRangeStatus();
         ConsistencyVerifier verifier = ConsistencyVerifiers.forConsistencyLevel(restoreJob.consistencyLevel, restoreJob.localDatacenter);
+        LOGGER.info("Checking restore job progress. jobId={} fetchPolicy={} successCriteria={}", restoreJob.jobId, fetchPolicy, successCriteria);
         Future<RestoreJobProgress> future = ringTopologyRefresher
                                             .replicaByTokenRangeAsync(restoreJob)
                                             .compose(topology -> findRangesAndConclude(restoreJob, successCriteria, topology, verifier, collector));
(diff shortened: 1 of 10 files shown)


server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobConsistencyChecker.java
```diff
@@ -100,6 +100,7 @@ public Future<RestoreJobProgress> check(RestoreJob restoreJob, RestoreJobProgres
         RestoreJobProgressCollector collector = RestoreJobProgressCollectors.create(restoreJob, fetchPolicy);
         RestoreRangeStatus successCriteria = restoreJob.expectedNextRangeStatus();
         ConsistencyVerifier verifier = ConsistencyVerifiers.forConsistencyLevel(restoreJob.consistencyLevel, restoreJob.localDatacenter);
+        LOGGER.info("Checking restore job progress. jobId={} fetchPolicy={} successCriteria={}", restoreJob.jobId, fetchPolicy, successCriteria);
```
sarankk, line 103: How frequently does the `/progress` endpoint get called? should we log this at debug?

server/src/test/java/org/apache/cassandra/sidecar/utils/SidecarClientProviderTest.java
```diff
@@ -84,29 +84,32 @@ class SidecarClientProviderTest
...
     {
-        CertificateBundle certificateAuthority = new CertificateBuilder().subject("CN=Apache Cassandra Root CA, OU=Certification Authority, O=Unknown, C=Unknown")
-                                                                         .alias("fakerootca")
-                                                                         .isCertificateAuthority(true)
-                                                                         .buildSelfSigned();
+        CertificateBundle certificateAuthority = new CertificateBuilder()
+                                                 .subject("CN=Apache Cassandra Root CA, OU=Certification Authority, O=Unknown, C=Unknown")
+                                                 .alias("fakerootca")
+                                                 .isCertificateAuthority(true)
+                                                 .buildSelfSigned();
```
yifan-c, line 91: just reformatting the code. Code style complain the line being too long

frankgh (approved):
+1 thanks for the fix

sarankk (approved):
+1 lgtm, 1 minor nit

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobConsistencyChecker.java
yifan-c: It depends on the client. 
In the context of Anayltics, the rate is at most once per second. I am adding this log message, because it is critical to confirm that client is requesting the consistency check. 
We over-estimate log verbosity in some places but often under-estimate in many other places. I think this one does not fall into the under-estimate category.

PR #224, opened 2025-06-09
CASSSIDECAR-258: Stage restore slices when in import phase already

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-258


frankgh (approved):
+1

PR #226, opened 2025-06-17
CASSSIDECAR-262: Make table stats permission table resource scoped

sarankk:


server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/FeaturePermission.java
```diff
@@ -52,6 +53,7 @@ public enum FeaturePermission
                           READ_SNAPSHOT,
                           DELETE_SNAPSHOT,
                           STREAM_SNAPSHOT,
+                          STATS_TABLE_SCOPED,
```
frankgh, line 56: This is required for CASSANALYTICS-36

frankgh (approved):
+1 Looks good to me

PR #227, opened 2025-06-19
CASSSIDECAR-263 Support a flag & implementation for only forward DNS resolution

anderoo:


server-common/src/main/java/org/apache/cassandra/sidecar/common/server/dns/DnsResolver.java
```diff
@@ -26,20 +28,38 @@
...
+        final String name;
+
+        DnsResolverEnum(String name)
+        {
+            this.name = name;
         }
 
-        @Override
-        public String reverseResolve(String address) throws UnknownHostException
+        public String toString()
```
frankgh: can we add the override annotation?
```suggestion
        @Override
        public String toString()
```

frankgh (commented):
Looks good. I think we should also update the sidecar.yaml file and add some user facing documentation. We have some tests in `SidecarConfigurationTest` which cover configuration tests. Can we add a test there for this change as well? Maybe refer to `org.apache.cassandra.sidecar.config.SidecarConfigurationTest#testCoordinationDefaultElectorateMembershipStrategy` as an example.

server/src/main/java/org/apache/cassandra/sidecar/config/ServiceConfiguration.java
```diff
@@ -179,4 +180,11 @@ default WorkerPoolConfiguration serverInternalWorkerPoolConfiguration()
      * @return the configuration relevant to the coordination functionality of Sidecar
      */
     CoordinationConfiguration coordinationConfiguration();
+
+    /**
+     * @return Sidecar's DNS Resolver to use. Default will be to provide forward and reverse DNS resolution which
+     * provides both the hostname & address. The alternative, resolveToIp, only provides forward resolution meaning
+     * the IP address will be returned for both hostname & address.
+     */
+    DnsResolver.DnsResolverEnum dnsResolver();
```
frankgh: Here we should use the interface instead.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/ServiceConfigurationImpl.java
```diff
@@ -449,6 +455,12 @@ public CoordinationConfiguration coordinationConfiguration()
         return coordinationConfiguration;
     }
 
+    @Override
+    @JsonProperty(value = DNS_RESOLVER_PROPERTY)
+    public DnsResolver.DnsResolverEnum dnsResolver() {
```
frankgh: here we should use the interface

server/src/main/java/org/apache/cassandra/sidecar/modules/UtilitiesModule.java
```diff
@@ -53,7 +53,7 @@ TimeProvider timeProvider()
     @Singleton
     DnsResolver dnsResolver()
     {
-        return DnsResolver.DEFAULT;
+        return DnsResolver.DnsResolverEnum.DEFAULT;
```
yifan-c: Nit: how about moving the enum declaration into its own Java file? So that we can have a simpler statement, without the 'DnsResolver.' portion

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/dns/DnsResolver.java
anderoo: Updated

server/src/main/java/org/apache/cassandra/sidecar/config/ServiceConfiguration.java
anderoo: Updated

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/ServiceConfigurationImpl.java
anderoo: Updated

conf/sidecar.yaml
```diff
@@ -158,6 +158,11 @@ sidecar:
       # The minimum value allowed for the cluster lease claim process task implementation is 30 seconds
       execute_interval: 100s
 
+    # DnsResolver to use (org.apache.cassandra.sidecar.common.server.dns.DnsResolver). Currently supported
+    # values are 'default' and 'resolveToIp'. Default will resolve hostnames to addresses and addresses to
+    # hostnames whereas resolveToIp will only resolve hostnames to addresses.
+    dnsResolver: default
```
frankgh: can we use snake case here to be consistent with the rest of the yaml?

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/ServiceConfigurationImpl.java
```diff
@@ -149,6 +151,9 @@ public class ServiceConfigurationImpl implements ServiceConfiguration
     @JsonProperty(value = COORDINATION)
     protected final CoordinationConfiguration coordinationConfiguration;
 
+    @JsonProperty(value = DNS_RESOLVER_PROPERTY)
+    protected final DnsResolver.DnsResolverEnum dnsResolver;
```
frankgh: The type here should also be interface `DnsResolver`

integration-framework/src/test/java/org/apache/cassandra/sidecar/testing/ResolveToIpDnsResolverTest.java
```diff
@@ -0,0 +1,44 @@
...
+package org.apache.cassandra.sidecar.testing;
+
+import org.apache.cassandra.sidecar.common.server.dns.DnsResolver;
+import org.junit.jupiter.api.Test;
+
+import java.net.UnknownHostException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ResolveToIpDnsResolverTest {
```
frankgh: This class might need reformatting
```suggestion
public class ResolveToIpDnsResolverTest
{
```

frankgh (commented):
Thanks for addressing the issues so promptly. I added a few additional comments.

conf/sidecar.yaml
anderoo: Updated

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/ServiceConfigurationImpl.java
anderoo: Updated

integration-framework/src/test/java/org/apache/cassandra/sidecar/testing/ResolveToIpDnsResolverTest.java
anderoo: Updated

server/src/main/java/org/apache/cassandra/sidecar/modules/UtilitiesModule.java
anderoo: Updated

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/ServiceConfigurationImpl.java
```diff
@@ -149,6 +152,9 @@ public class ServiceConfigurationImpl implements ServiceConfiguration
     @JsonProperty(value = COORDINATION)
     protected final CoordinationConfiguration coordinationConfiguration;
 
+    @JsonProperty(value = DNS_RESOLVER_PROPERTY)
+    protected final DnsResolverEnum dnsResolver;
```
frankgh: The type here should be `DnsResolver`
```suggestion
    protected final DnsResolver dnsResolver;
```

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/dns/DnsResolver.java
```diff
@@ -26,29 +26,17 @@
...
      *
      * @param hostname the hostname to resolve
      * @return IP address
      * @throws UnknownHostException when the host is not known
      */
-    String resolve(String hostname) throws UnknownHostException;
+    default String resolve(String hostname) throws UnknownHostException
+    {
+        return InetAddress.getByName(hostname).getHostAddress();
+    }
```
frankgh: I would preserve this interface and not add a default, but instead implement `resolve` in the `DnsResolverEnum`
```suggestion
    String resolve(String hostname) throws UnknownHostException;
```

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/SidecarConfigurationImpl.java
```diff
@@ -384,6 +384,10 @@ private static SimpleModule resolveYamlTypeMappings() throws IOException
                      module.addAbstractTypeMapping(configurationInterface, clazz);
                  });
 
+        // Manually add type mapping for DnsResolver
+        module.addAbstractTypeMapping(org.apache.cassandra.sidecar.common.server.dns.DnsResolver.class,
```
smiklosovic: why so long? we can just import each, no?

smiklosovic (approved):

frankgh (approved):
+1

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/dns/DnsResolvers.java
```diff
@@ -0,0 +1,85 @@
...
+ * address to hostname whereas the latter will only resolve hostname to address.
+ */
+public enum DnsResolvers implements DnsResolver
+{
+    /**
+     * Default implementation of the {@link DnsResolver} that uses the JDK's
+     * underlying DNS resolution mechanism
+     */
+    @JsonProperty("default")
+    DEFAULT("default")
```
smiklosovic: do we still need this name? is it still necessary? Is not this covered by JsonProperty already?

PR #228, opened 2025-06-24
CASSSIDECAR-264: Settings endpoint should return Service Unavailable …

frankgh:
…when Cassandra instance is down

Patch by Francisco Guerrero; reviewed by TBD for CASSSIDECAR-264


bbotella:
+1

PR #230, opened 2025-06-26
CASSSIDECAR-267: Fix CdcRawDirectorySpaceCleaner

bbotella:
Fix how cdc_raw folder is being accessed.

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
@@ -184,7 +184,17 @@ public class CdcRawDirectorySpaceCleaner implements PeriodicTask
         {
             if (instanceMetadata.cdcDir() != null)
             {
-                cleanUpCdcRawDirectory(new File(Optional.ofNullable(instanceMetadata.cdcDir()).get()));
+                try
+                {
+                    cleanUpCdcRawDirectory(new File(Optional.ofNullable(instanceMetadata.cdcDir()).get()));
+                }
+                catch (Exception e)
+                {
+                    LOGGER.warn("Couldn't clean up CDC dir", e);
+                }
+            }
+            else {
+                LOGGER.warn("Couldn't find CDC dir to clean up");
             }
         }
     }
--- a/server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
@@ -191,7 +191,7 @@ public class CdcRawDirectorySpaceCleaner implements PeriodicTask
                 }
                 catch (Exception e)
                 {
-                    LOGGER.warn("Unable to clean up CDC directory {}", cdcDir, e);
+                    LOGGER.warn("Unable to clean up CDC directory {} for instance {}", cdcDir, instanceMetadata, e);
                 }
             }
             else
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -182,10 +182,10 @@ protected void routineCleanUp()
     {
         for (InstanceMetadata instanceMetadata : instancesMetadata.instances())
         {
-            List<String> dataDirectories = instanceMetadata.dataDirs();
-            dataDirectories.stream()
-                           .map(dir -> new File(dir, CDC_DIR_NAME))
-                           .forEach(this::cleanUpCdcRawDirectory);
+            if (instanceMetadata.cdcDir() != null)
```
yifan-c: how about add a warning log message if cdcDir is null?
`CdcRawDirectorySpaceCleaner` is enabled explicitly via a property; if it is enabled, but cdc dir does not exist, it likely indicates some issue.

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -182,10 +182,10 @@ protected void routineCleanUp()
...
+            {
+                cleanUpCdcRawDirectory(new File(Optional.ofNullable(instanceMetadata.cdcDir()).get()));
```
yifan-c: How about catch all the exception from the method, log a warning, and continue with the next instance?

yifan-c (commented):
Please update changes.txt

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
frankgh: can we avoid using Optional here? I have the following suggestion

```
            String cdcDir = instanceMetadata.cdcDir();
            if (cdcDir != null)
            {
                try
                {
                    cleanUpCdcRawDirectory(new File(cdcDir));
                }
                catch (Exception e)
                {
                    LOGGER.warn("Unable to clean up CDC directory {}", cdcDir, e);
                }
            }
            else
            {
                LOGGER.warn("CDC directory is not configured. Skipping clean up");
            }
        }
```

frankgh: I also added some small changes to the wording of the log messages . Let me know if  this change makes sense to you

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -182,10 +182,22 @@ protected void routineCleanUp()
...
+            String cdcDir = instanceMetadata.cdcDir();
+            if (cdcDir != null)
+            {
+                try
+                {
+                    cleanUpCdcRawDirectory(new File(cdcDir));
+                }
+                catch (Exception e)
+                {
+                    LOGGER.warn("Unable to clean up CDC directory {}", cdcDir, e);
```
frankgh: apologies, I think we need to log the instance in the message too:
```suggestion
                    LOGGER.warn("Unable to clean up CDC directory {} for instance {}", cdcDir, instanceMetadata, e);
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -182,10 +182,22 @@ protected void routineCleanUp()
...
+                }
+            }
+            else
+            {
+                LOGGER.warn("CDC directory is not configured. Skipping clean up");
```
frankgh: Same here: 
```suggestion
                LOGGER.warn("CDC directory is not configured for instance {}. Skipping clean up", instanceMetadata);
```

yifan-c (approved):

frankgh (approved):
+1 thanks for addressing the comments so quickly!

PR #232, opened 2025-06-27
CASSSIDECAR-234: Add Getting Started and Configuration guide to sidecar docs

isaacreath:
This patch adds a quickstart guide in the user.adoc file and a configuration guide under configuring.adoc similar to the same guides available at cassandra.apache.org/doc/trunk for Cassandra.

To build the asciidocs website, you can run `./gradlew asciidoctor` which will put the docs HTML files in the `docs/build/` folder.


docs/build.gradle
```diff
@@ -22,4 +22,7 @@ apply plugin: 'org.asciidoctor.jvm.convert'
 asciidoctor {
     sourceDir = file("src")
     outputDir = file("build")
+    attributes(
+            'project-version': project.version
```
isaacreath, line 26: This creates the docs for the current version of Cassandra Sidecar (e.g. 0.2-SNAPSHOT for trunk), which does impact the quickstart guide. For testing, I've been overriding this to `0.1.0` since that's our current release, but when a new release is cut I would expect this to work properly when these docs are built off of the release tag.

docs/src/configuring.adoc
```diff
@@ -0,0 +1,292 @@
+# Configuring Cassandra Sidecar
```
isaacreath: As I was writing this, I was questioning the value of listing out all YAML options the way we do for Cassandra (https://cassandra.apache.org/doc/5.0/cassandra/managing/configuration/cass_yaml_file.html) or if the source of truth should just be the YAML with comments explaining how to use each key. I could see this going out of date pretty quickly.

yifan-c: yaml + comment is a better fit than a separate doc. As yaml is examined by code and guaranteed to be update date.

docs/src/user.adoc
```diff
@@ -1,15 +1,446 @@
...
+[[setup]]
 ## Setup
+In this section, we will cover the basic configurations required to connect Cassandra Sidecar to your Cassandra cluster. For a more in-depth guide on Cassandra Sidecar configuration, please see the <<configuring,Configuring>> section.
+
+Cassandra Sidecar's root directory will contain the following directories:
+
+* bin: This contains the script to start Cassandra Sidecar
+* conf: This is the location of the Cassandra Sidecar configuration
+* lib: This directory contains the JAR files which comprise the Cassandra Sidecar application
+* agents: This directory contains JAR files for JVM agents which attach to Cassandra Sidecar (such as Jolokia)
```
frankgh: I guess this is gone since [CASSSIDECAR-330](https://issues.apache.org/jira/browse/CASSSIDECAR-330): Remove Jolokia agent

frankgh (approved):
+1 looks good to me with one minor comment about the agents directory which no longer applies

docs/src/user.adoc
```diff
@@ -1,15 +1,446 @@
...
+*** `execute_interval`: This defines how often the cluster lease claim process will execute after the previous task has completed execution. The minimum allowed value is 30 seconds (`30s`). By default, 100 seconds (`100s`).
+
+[[vert.x]]
+### vert.x
+
+This section of the `sidecar.yaml` file is used configure Vert.x, which powers Cassandra Sidecar. The following properties are defined:
+
+* `filesystem_options`: This object defines the configuration used by Vert.x it defines the following fields:
+** `classpath_resolving_enabled`: When vert.x cannot find the file on the filesystem it tries to resolve the file from the classpath when this is set to `true`. Otherwise, vert.x will not attempt to resolve the file on the classpath. Default value is `false`.
```
frankgh: this just changed to true since [CASSSIDECAR-176](https://issues.apache.org/jira/browse/CASSSIDECAR-176): Add comprehensive OpenAPI documentation annotations

docs/src/user.adoc
```diff
@@ -1,15 +1,446 @@
...
+** `file_caching_enabled`: Defines whether or not to enable caching files on the real file system when the filesystem performs class path resolving. Set to `false` to disable caching. Default value is `false`.
```
frankgh: same for this one. It is true by default

yifan-c (approved):
IMO, the documents on the configurations fields are better located in the configuration file, `sidecar.yaml`, in order to not diverge in the future. It is easy to miss updating the `user.adoc`. 
I am not going to block on it for this patch. We can relocate the docs whenever it is needed.

PR #233, opened 2025-06-30
CASSSIDECAR-224 Enhancing Sidecar Client to download files for Live Migration

nvharikrishna:

CASSSIDECAR-224 Enhancing Sidecar Client to download files for Live Migration.

### DIFF
--- a/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
+++ b/client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
@@ -917,13 +917,12 @@ public class SidecarClient implements AutoCloseable, SidecarClientBlobRestoreExt
      * @param targetFilePath destination file path to save file coming from remote
      * @return a CompletableFuture representing the completion of the operation
      */
-    public CompletableFuture<Boolean> liveMigrationStreamFileAsync(SidecarInstance instance, String reqPath, String targetFilePath)
+    public CompletableFuture<Void> liveMigrationStreamFileAsync(SidecarInstance instance, String reqPath, String targetFilePath)
     {
         return executor.executeRequestAsync(requestBuilder().singleInstanceSelectionPolicy(instance)
                                                             .request(new StreamFileRequest(reqPath, targetFilePath))
                                                             .retryPolicy(new LiveMigrationDownloadRetryPolicy(defaultRetryPolicy, targetFilePath))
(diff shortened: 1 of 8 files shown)


client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -890,6 +894,38 @@ public <T> CompletableFuture<T> executeRequestAsync(RequestContext context)
...
+    /**
+     * Downloads requested file {@code reqPath} to the destination file provided with param {@code targetFile}
+     * during Live Migration.
+     *
+     * @param instance       the instance where the request will be executed
+     * @param reqPath        instance file path to copy
+     * @param targetFilePath destination file path to save file coming from remote
+     * @return a CompletableFuture representing the completion of the operation
+     */
+    public CompletableFuture<Boolean> liveMigrationStreamFileAsync(SidecarInstance instance, String reqPath, String targetFilePath)
```
yifan-c: just curious, the boolean here can only be `true`, making it effectively the same as `CompletableFuture<Void>`, right? Then, you can remove the transformation at line926

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
```diff
@@ -1811,6 +1816,90 @@ void testNodeUpdateNative() throws Exception
         });
     }
 
+    @Test
+    public void testLiveMigrationListInstanceFiles() throws ExecutionException, InterruptedException, JsonProcessingException
```
yifan-c: nit: please omit `public`, since it is unnecessary for the test.

yifan-c (commented):
lgtm overall.

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
nvharikrishna: done

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
nvharikrishna: done

frankgh (approved):
+1

PR #234, opened 2025-07-01
CASSSIDECAR-269: Allow restore jobs to restore to the local datacenter only

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-269

### DIFF
--- a/client-common/src/main/java/org/apache/cassandra/sidecar/common/data/RestoreJobConstants.java
+++ b/client-common/src/main/java/org/apache/cassandra/sidecar/common/data/RestoreJobConstants.java
@@ -42,9 +42,9 @@ public class RestoreJobConstants
      * of the belonging keyspace.
      * The option should be used with caution. To help build an educated decision, when only restoring to the local
      * datacenter, but the table replicates to multiple datacenters, it could cause a large amount of repair streaming
-     * traffic. Argurably, you might want to leverage the intra-node repair feature, then you can enable this option.
-     * Another use case that could justify the option is runing distinct restore jobs, one per datacenter, concurrently.
-     * In this case, there is an external coordinator that manges the resotre job in each datacenter.
+     * traffic. Arguably, you might want to leverage the intra-node repair feature, then you can enable this option.
+     * Another use case that could justify the option is running distinct restore jobs, one per datacenter, concurrently.
+     * In this case, there is an external coordinator that manges the restore job in each datacenter.
      */
     public static final String JOB_RESTORE_TO_LOCAL_DATA_CENTER_ONLY = "restoreToLocalDatacenterOnly";
     public static final String SLICE_ID = "sliceId";
--- a/client-common/src/main/java/org/apache/cassandra/sidecar/common/data/RestoreJobConstants.java
+++ b/client-common/src/main/java/org/apache/cassandra/sidecar/common/data/RestoreJobConstants.java
@@ -36,6 +36,17 @@ public class RestoreJobConstants
     public static final String JOB_CONSISTENCY_LEVEL = "consistencyLevel";
     public static final String JOB_OPERATION_REASON = "reason";
     public static final String JOB_LOCAL_DATA_CENTER = "localDatacenter";
+    /**
+     * A boolean field; when restoreToLocalDatacenterOnly is set to true, localDatacenter must also be specified, and
+     * the restore job will only restore to the specified localDatacenter, regardless of the replication configuration
+     * of the belonging keyspace.
+     * The option should be used with caution. To help build an educated decision, when only restoring to the local
+     * datacenter, but the table replicates to multiple datacenters, it could cause a large amount of repair streaming
+     * traffic. Argurably, you might want to leverage the intra-node repair feature, then you can enable this option.
+     * Another use case that could justify the option is runing distinct restore jobs, one per datacenter, concurrently.
+     * In this case, there is an external coordinator that manges the resotre job in each datacenter.
+     */
+    public static final String JOB_RESTORE_TO_LOCAL_DATA_CENTER_ONLY = "restoreToLocalDatacenterOnly";
     public static final String SLICE_ID = "sliceId";
     public static final String BUCKET_ID = "bucketId";
     public static final String SLICE_START_TOKEN = "startToken";
--- a/client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CreateRestoreJobRequestPayload.java
+++ b/client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CreateRestoreJobRequestPayload.java
@@ -33,6 +33,7 @@ import org.apache.cassandra.sidecar.common.data.RestoreJobSecrets;
 import org.apache.cassandra.sidecar.common.data.RestoreJobStatus;
 import org.apache.cassandra.sidecar.common.data.SSTableImportOptions;
 import org.apache.cassandra.sidecar.common.utils.Preconditions;
+import org.apache.cassandra.sidecar.common.utils.StringUtils;
 import org.jetbrains.annotations.Nullable;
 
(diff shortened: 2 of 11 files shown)


client-common/src/main/java/org/apache/cassandra/sidecar/common/data/RestoreJobConstants.java
```diff
@@ -36,6 +36,17 @@ public class RestoreJobConstants
...
     public static final String JOB_LOCAL_DATA_CENTER = "localDatacenter";
+    /**
+     * A boolean field; when restoreToLocalDatacenterOnly is set to true, localDatacenter must also be specified, and
+     * the restore job will only restore to the specified localDatacenter, regardless of the replication configuration
+     * of the belonging keyspace.
+     * The option should be used with caution. To help build an educated decision, when only restoring to the local
+     * datacenter, but the table replicates to multiple datacenters, it could cause a large amount of repair streaming
+     * traffic. Argurably, you might want to leverage the intra-node repair feature, then you can enable this option.
+     * Another use case that could justify the option is runing distinct restore jobs, one per datacenter, concurrently.
+     * In this case, there is an external coordinator that manges the resotre job in each datacenter.
```
frankgh: minor typos
```suggestion
     * traffic. Arguably, you might want to leverage the intra-node repair feature, then you can enable this option.
     * Another use case that could justify the option is running distinct restore jobs, one per datacenter, concurrently.
     * In this case, there is an external coordinator that manges the restore job in each datacenter.
```

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CreateRestoreJobRequestPayload.java
```diff
@@ -86,7 +89,8 @@ public CreateRestoreJobRequestPayload(@JsonProperty(JOB_ID) UUID jobId,
                                           @JsonProperty(JOB_IMPORT_OPTIONS) SSTableImportOptions importOptions,
                                           @JsonProperty(JOB_EXPIRE_AT) long expireAtInMillis,
                                           @JsonProperty(JOB_CONSISTENCY_LEVEL) String consistencyLevel,
-                                          @JsonProperty(JOB_LOCAL_DATA_CENTER) String localDatacenter)
+                                          @JsonProperty(JOB_LOCAL_DATA_CENTER) String localDatacenter,
```
frankgh, line 94: let's add the missing parameters to javadocs

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CreateRestoreJobRequestPayload.java
```diff
@@ -101,6 +105,9 @@ public CreateRestoreJobRequestPayload(@JsonProperty(JOB_ID) UUID jobId,
                              : importOptions;
         this.expireAtInMillis = expireAtInMillis;
         this.consistencyConfig = ConsistencyConfig.parseString(consistencyLevel, localDatacenter);
+        Preconditions.checkArgument(!localDatacenterOnly || StringUtils.isNotEmpty(localDatacenter),
+                                    "Must specify a non-empty localDatacenter when restoreToLocalDatacenterOnly is configured to true");
```
frankgh: I think the `non-empty` part is implied
```suggestion
                                    "Must specify a localDatacenter when restoreToLocalDatacenterOnly is true");
```

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -385,6 +399,41 @@ private void processSidecarManagedJobMaybe(RestoreJob job)
...
+    private boolean isLocalDatacenterExcluded(RestoreJob job)
+    {
+        if (!job.shouldRestoreToLocalDatacenterOnly)
+        {
+            return false;
+        }
+
+        if (localDatacenter == null)
+        {
+            LOGGER.debug("The restore job should restore only to the local datacenter, but the local datacetner is undetermined yet; skip this run");
```
frankgh:
```suggestion
            LOGGER.debug("The restore job should restore only to the local datacenter, but the local datacenter is undetermined yet; skip this run");
```

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -385,6 +399,41 @@ private void processSidecarManagedJobMaybe(RestoreJob job)
...
+            return true;
+        }
+
+        // when job should restore to local datacenter only, but the target datacenter is not the local one
+        return !localDatacenter.equalsIgnoreCase(job.localDatacenter);
```
frankgh: datacenter is case sensitive, so I don't think we should equal here with ignore case, even though it would be bad if someone decided to have `DATACENTER1` and `datacenter1`, but it would be possible

server/src/test/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscovererTest.java
```diff
@@ -379,6 +385,59 @@ void testAdjustRecencyDays()
...
+                                                        .consistencyLevel(ConsistencyLevel.LOCAL_QUORUM)
+                                                        .localDatacenter("dc2")
+                                                        .shouldRestoreToLocalDatacenterOnly(true)
+                                                        .build()));
+        executeBlocking();
+        verify(mockRangeAccessor, never()).create(any());
+    }
+
+    @Test
+    void testRestoreToLocalDatacenterOnlyJobIsOnHoldWhenLocalDatacenterInUndetermined()
```
frankgh: Is this what you meant?
```suggestion
    void testRestoreToLocalDatacenterOnlyJobIsOnHoldWhenLocalDatacenterIsUndetermined()
```

server/src/test/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscovererTest.java
```diff
@@ -379,6 +385,59 @@ void testAdjustRecencyDays()
...
+                                                        .expireAt(new Date(System.currentTimeMillis() + 10000L))
+                                                        .consistencyLevel(ConsistencyLevel.LOCAL_QUORUM)
+                                                        .localDatacenter("dc1")
+                                                        .shouldRestoreToLocalDatacenterOnly(true)
+                                                        .build()));
+
+        executeBlocking();
+        verify(ringTopologyRefresher, never()).register(any(), any());
+
+        // in the new run, the local datacenter is discovered. The discoverer proceeds further and register the job
```
frankgh, line 430:
```suggestion
        // in the new run, the local datacenter is discovered. The discoverer proceeds further and registers the job
```

frankgh (commented):
Looks good, I left a few minor comments

client-common/src/main/java/org/apache/cassandra/sidecar/common/data/RestoreJobConstants.java
yifan-c: ah. good catch! Somehow, typo check is disabled in my local. :|

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CreateRestoreJobRequestPayload.java
yifan-c: ok. done

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
yifan-c: Good catch!

yifan-c (commented):
Not sure how replying a comment becomes a review 🤷

frankgh (approved):
+1

PR #237, opened 2025-07-23
CASSSIDECAR-325: CdcRawDirectorySpaceCleaner always uses fallback aft…

frankgh:
…er the first time it is queried from system_views.settings

Patch by Francisco Guerrero; reviewed by TBD for CASSSIDECAR-325

### DIFF
--- a/server-common/src/main/java/org/apache/cassandra/sidecar/db/schema/AbstractSchema.java
+++ b/server-common/src/main/java/org/apache/cassandra/sidecar/db/schema/AbstractSchema.java
@@ -88,7 +88,7 @@ public abstract class AbstractSchema
         }
 
         prepareStatements(session);
-        logger.debug("{} is initialized!", this.getClass().getSimpleName());
+        logger.info("{} table schema is initialized", this.getClass().getSimpleName());
         return true;
     }
 
--- a/server/src/main/java/org/apache/cassandra/sidecar/modules/SidecarSchemaModule.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/modules/SidecarSchemaModule.java
@@ -62,9 +62,7 @@ public class SidecarSchemaModule extends AbstractModule
             }
             catch (Throwable cause)
             {
-                String message = "Failed to register table schema: " + tableSchema;
-                LOGGER.error(message);
-                throw new RuntimeException(message, cause);
+                throw new RuntimeException("Failed to register table schema: " + tableSchema, cause);
             }
         });
         return new SidecarSchema(vertx, configuration, sidecarInternalKeyspace);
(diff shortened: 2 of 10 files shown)


server-common/src/main/java/org/apache/cassandra/sidecar/db/schema/AbstractSchema.java
```diff
@@ -88,7 +88,7 @@ protected boolean initializeInternal(@NotNull Session session,
         }
 
         prepareStatements(session);
-        logger.debug("{} is initialized!", this.getClass().getSimpleName());
+        logger.info("{} table schema is initialized", this.getClass().getSimpleName());
```
frankgh, line 91: useful for debugging purposes. This gets logged a single time during startup and it produces useful information

server/src/main/java/org/apache/cassandra/sidecar/modules/SidecarSchemaModule.java
```diff
@@ -50,13 +55,16 @@ public SidecarSchema sidecarSchema(Vertx vertx,
...
             try
             {
                 sidecarInternalKeyspace.registerTableSchema(tableSchema);
             }
             catch (Throwable cause)
             {
-                throw new RuntimeException("Failed to register table schema: " + tableSchema, cause);
+                String message = "Failed to register table schema: " + tableSchema;
+                LOGGER.error(message);
+                throw new RuntimeException(message, cause);
```
yifan-c: The same error message should be printed by guice when bootstraping. But not against having another error log message here.

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -226,19 +226,21 @@ protected void cleanUpCdcRawDirectory(File cdcRawDirectory)
             return;
         }
 
-        long directorySize = FileUtils.directorySize(cdcRawDirectory);
-        long upperLimitBytes =
-        (long) (maxUsage() * cdcConfiguration.cdcRawDirectoryMaxPercentUsage());
+        long directorySizeBytes = FileUtils.directorySizeBytes(cdcRawDirectory);
+        long maxedUsageBytes = maxUsageBytes();
```
yifan-c: typo?
```suggestion
        long maxUsageBytes = maxUsageBytes();
```

yifan-c (approved):
Looks good. Some nits.

server/src/main/java/org/apache/cassandra/sidecar/modules/SidecarSchemaModule.java
frankgh: yeah good point

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
frankgh: yeah, typo, thanks !

bbotella (approved):

PR #240, opened 2025-07-27
CASSSIDECAR-1: Sidecar gradle run should not run integration tests

frankgh:
Patch by Francisco Guerrero; reviewed by TBD for CASSSIDECAR-1

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -336,14 +336,14 @@ gradle.taskGraph.whenReady { taskGraph ->
         def exclusionTasks = tasks.findAll {
             // integration tests
             it.name == "integrationTest" ||
-                    // distribution plugin tasks
-                    it.name.startsWith('dist') || it.name.startsWith('assembleDist') ||
-                    // tar, deb, zip, rpm tasks
-                    it.name.endsWith("Tar") || it.name.endsWith("Deb") || it.name.endsWith("Rpm") || it.name.endsWith("Zip") ||
-                    // docs
-                    it.name.contains("asciidoctor") || it.name == "copyDocs" ||
-                    // other tasks
-                    it.name.startsWith("spotbugs") || it.name == "rat"
+            // distribution plugin tasks
+            it.name.startsWith('dist') || it.name.startsWith('assembleDist') ||
+            // tar, deb, zip, rpm tasks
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -323,6 +325,33 @@ tasks.register('copyDocs', Copy) {
...
+            // integration tests
+            it.name == "integrationTest" ||
+                    // distribution plugin tasks
+                    it.name.startsWith('dist') || it.name.startsWith('assembleDist') ||
+                    // tar, deb, zip, rpm tasks
+                    it.name.endsWith("Tar") || it.name.endsWith("Deb") || it.name.endsWith("Rpm") || it.name.endsWith("Zip") ||
+                    // docs
+                    it.name.contains("asciidoctor") || it.name == "copyDocs" ||
+                    // other tasks
+                    it.name.startsWith("spotbugs") || it.name == "rat"
```
yifan-c: Can you use the continuation indent of 0, instead of 8? To be the same as Java.

build.gradle
```diff
@@ -323,6 +325,33 @@ tasks.register('copyDocs', Copy) {
     exclude "tmp"
 }
 
+// Exclude some tasks to speed up running ./gradlew run
+if ("run" in gradle.startParameter.taskNames) {
+    project.gradle.startParameter.excludedTaskNames.add("codeCheckTasks")
+    project.gradle.startParameter.excludedTaskNames.add("asciidoctor")
+}
+
+gradle.taskGraph.whenReady { taskGraph ->
```
yifan-c, line 334: good idea

yifan-c (commented):
Just a minor one on indentation. 

The change in the circleci config file makes sense. It took me a while to realize that the change is valid.

build.gradle
frankgh: this is automatically indented, but I can manually update it

yifan-c: yep. The project does not have the style setting for groovy. Maybe update the `ide/idea/codeStyleSettings.xml` to include the following too. 

```xml
  <codeStyleSettings language="Groovy">
    <indentOptions>
      <option name="CONTINUATION_INDENT_SIZE" value="0" />
    </indentOptions>
  </codeStyleSettings>
```

frankgh: good idea, I've pushed the change

PR #241, opened 2025-07-31
Fix rpm broken symlink

serban21:
CASSSIDECAR-327

CASSSIDECAR-218 renamed the rpm/deb package to `apache-cassandra-sidecar`, but kept the symlink to the same place that now doesn't exist (`/opt/cassandra-sidecar/bin/cassandra-sidecar`)


frankgh (approved):
+1 thanks for the fix

PR #246, opened 2025-08-05
CASSSIDECAR-333: Fix a bug that reads member_of field as a Set instead of List

sarankk:


server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -150,27 +150,29 @@ protected void warmUp(int availableRetries)
...
-            LOGGER.warn("Retries exhausted, unexpected error pre-warming cache={}", name);
+            logger.warn("Retries exhausted, unexpected error pre-warming cache={}", name);
             return;
         }
 
         try
         {
+            long startTimeNanos = System.nanoTime();
             cache.putAll(bulkLoadFunction.get());
+            logger.info("Cache warm completed successfully in {} nanoseconds", System.nanoTime() - startTimeNanos);
```
frankgh: let's use the cache name here like in the other log entries

frankgh (approved):
+1

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -46,7 +46,7 @@
  */
 public abstract class AuthCache<K, V>
 {
-    private static final Logger LOGGER = LoggerFactory.getLogger(AuthCache.class);
+    protected final Logger logger = LoggerFactory.getLogger(this.getClass());
```
yifan-c: I would prefer a dedicated LOGGER declared per class, from the perspective of clarity. I found it very hard to see the usage of logger that is defined in a superclass.

server/src/test/java/org/apache/cassandra/sidecar/db/SystemAuthDatabaseAccessorTest.java
```diff
@@ -48,4 +61,36 @@ void testAccessorThrowsWhenTableNotFound()
...
+        when(mockResult.all()).thenReturn(rows);
+        when(mockSession.execute((Statement) any())).thenReturn(mockResult);
+        SystemAuthDatabaseAccessor systemAuthDatabaseAccessor = new SystemAuthDatabaseAccessor(mockSystemAuthSchema,
+                                                                                               mockCqlSessionProvider,
+                                                                                               new PermissionFactoryImpl());
+        Map<String, Boolean> superUserCache = systemAuthDatabaseAccessor.findAllRolesToSuperuserStatus();
+        assertThat(superUserCache.size()).isEqualTo(2);
+        assertThat(superUserCache.get("super_role")).isTrue();
+        assertThat(superUserCache.get("non_super_role")).isTrue();
+    }
```
yifan-c: I am not sure this test is meaningful.. It just mocks everything and mock is based on the implementation being tested. 
I think it should test with the accessor connecting to the cassandra node in integration test. Please take `org.apache.cassandra.sidecar.db.RestoreSliceDatabaseAccessorIntTest` as an example.

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
frankgh: How can you leverage the logger of the implementing class though?

frankgh: right now the limitation we have is that all logs are logged by the AuthCache logger. But we have multiple implementations and there's no way to differentiate between the loggers.

sarankk: Yifan suggested adding another logger in child class if there is need. Currently we don't log anything from child classes

server/src/test/java/org/apache/cassandra/sidecar/db/SystemAuthDatabaseAccessorTest.java
sarankk: Added a different test

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
frankgh: Right, but your actual logger is the same always, so if you need to find something based on the implementing class name, you can't

sarankk: We still have `getClass()` in actual logger `private final Logger logger = LoggerFactory.getLogger(this.getClass());` just changed access modifier

PR #250, opened 2025-08-15
CASSSIDECAR-331: NullPointerException When Authentication Is Enabled but sidecar_internal Schema Is Disabled

sarankk:


server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/AuthenticationHandlerFactory.java
```diff
@@ -45,4 +46,14 @@ public interface AuthenticationHandlerFactory
     AuthenticationHandlerInternal create(Vertx vertx,
                                          AccessControlConfiguration accessControlConfiguration,
                                          Map<String, String> parameters) throws ConfigurationException;
+    
+    /**
+     * Validates that this authentication handler factory can be used with the given configuration.
```
frankgh, line 51:
```suggestion
     * Validates that this authentication handler factory can be used with the given configuration.
     * This gives the implementation flexibility to establish pre-requisites in order to properly function
     * correctly.
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/JwtAuthenticationHandlerFactory.java
```diff
@@ -65,4 +66,16 @@ protected JwtParameters parameterParser(Map<String, String> parameters)
...
+    
+    @Override
+    public void validatePrerequisites(SidecarConfiguration sidecarConfiguration) throws ConfigurationException
+    {
+        boolean isSidecarSchemaEnabled = sidecarConfiguration.serviceConfiguration()
+                                                             .schemaKeyspaceConfiguration()
+                                                             .isEnabled();
+        if (!isSidecarSchemaEnabled)
+        {
+            throw new ConfigurationException("JWT auth requires sidecar schema to be enabled for role processing");
```
frankgh:
```suggestion
            throw new ConfigurationException("JwtAuthenticationHandlerFactory requires Sidecar schema to be enabled for role processing");
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/MutualTlsAuthenticationHandlerFactory.java
```diff
@@ -106,4 +107,16 @@ private MutualTlsAuthenticationHandler createInternal(Vertx vertx,
...
+
+    @Override
+    public void validatePrerequisites(SidecarConfiguration sidecarConfiguration) throws ConfigurationException
+    {
+        boolean isSidecarSchemaEnabled = sidecarConfiguration.serviceConfiguration()
+                                                             .schemaKeyspaceConfiguration()
+                                                             .isEnabled();
+        if (!isSidecarSchemaEnabled)
+        {
+            throw new ConfigurationException("mTLS auth requires sidecar schema to be enabled for role processing");
```
frankgh: The user facing configuration that we document in sidecar.yaml is MutualTlsAuthenticationHandlerFactory, so I think we should say that instead.
```suggestion
            throw new ConfigurationException("MutualTlsAuthenticationHandlerFactory requires Sidecar schema to be enabled for role processing");
```

server/src/main/java/org/apache/cassandra/sidecar/modules/AuthModule.java
```diff
@@ -180,6 +181,10 @@ AuthorizationProvider authorizationProvider(SidecarConfiguration sidecarConfigur
         }
         if (config.className().equalsIgnoreCase(RoleBasedAuthorizationProvider.class.getName()))
         {
+            if (!sidecarConfiguration.serviceConfiguration().schemaKeyspaceConfiguration().isEnabled())
+            {
+                throw new ConfigurationException(config.className() + " requires sidecar schema to be enabled for role permissions storage");
```
frankgh:
```suggestion
                throw new ConfigurationException(config.className() + " requires Sidecar schema to be enabled for role permissions used by Sidecar");
```

frankgh (commented):
Looks good. Thanks for the fix. I have small comments

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -119,6 +119,18 @@ public Map<K, V> getAll()
         return Collections.unmodifiableMap(cache.asMap());
     }
 
+    /**
+     * Invalidate a key.
+     * @param k key to invalidate
+     */
+    public void invalidate(K k)
```
bbotella, line 126: Where is this method being used?

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/JwtAuthenticationHandlerFactory.java
```diff
@@ -65,4 +66,16 @@ protected JwtParameters parameterParser(Map<String, String> parameters)
...
+    @Override
+    public void validatePrerequisites(SidecarConfiguration sidecarConfiguration) throws ConfigurationException
+    {
+        boolean isSidecarSchemaEnabled = sidecarConfiguration.serviceConfiguration()
+                                                             .schemaKeyspaceConfiguration()
+                                                             .isEnabled();
+        if (!isSidecarSchemaEnabled)
+        {
+            throw new ConfigurationException("JwtAuthenticationHandlerFactory requires Sidecar schema to be enabled for role processing");
+        }
```
bbotella, line 79: Why not having this implementation in the default implementation of the interface?

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/MutualTlsAuthenticationHandlerFactory.java
```diff
@@ -106,4 +107,16 @@ private MutualTlsAuthenticationHandler createInternal(Vertx vertx,
...
+
+    @Override
+    public void validatePrerequisites(SidecarConfiguration sidecarConfiguration) throws ConfigurationException
+    {
+        boolean isSidecarSchemaEnabled = sidecarConfiguration.serviceConfiguration()
+                                                             .schemaKeyspaceConfiguration()
+                                                             .isEnabled();
+        if (!isSidecarSchemaEnabled)
+        {
+            throw new ConfigurationException("MutualTlsAuthenticationHandlerFactory requires Sidecar schema to be enabled for role processing");
```
bbotella, line 119: What about mentioning explicitly the flag that needs to be enabled on the config?

server/src/main/java/org/apache/cassandra/sidecar/modules/AuthModule.java
```diff
@@ -180,6 +181,10 @@ AuthorizationProvider authorizationProvider(SidecarConfiguration sidecarConfigur
         }
         if (config.className().equalsIgnoreCase(RoleBasedAuthorizationProvider.class.getName()))
         {
+            if (!sidecarConfiguration.serviceConfiguration().schemaKeyspaceConfiguration().isEnabled())
+            {
+                throw new ConfigurationException(config.className() + " requires Sidecar schema to be enabled for role permissions used by Sidecar");
```
bbotella, line 186: What about mentioning explicitly the flag that needs to be enabled on the config?

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/JwtAuthenticationHandlerFactory.java
frankgh: I don't think it makes sense to have this in the interface, because you don't know if the implementation requires sidecar schema. It only makes sense to have this check if you require sidecar schema for the implementation

PR #251, opened 2025-08-18
CASSSIDECAR-245: Implementation of CassandraClusterSchema

bbotella:
CASSSIDECAR-245: Implementation of CassandraClusterSchema


server/build.gradle
```diff
@@ -148,6 +148,10 @@ dependencies {
     implementation("io.swagger.core.v3:swagger-models:${project.swaggerVersion}")
     implementation("io.swagger.core.v3:swagger-jaxrs2:${project.swaggerVersion}")
 
+    implementation(group: "org.apache.cassandra", name: "cassandra-analytics-common", version: "${[project.analyticsVersion]}")
+    implementation(group: "org.apache.cassandra", name: "cassandra-analytics-cdc_spark3_2.12", version: "${[project.analyticsVersion]}")
+    implementation "com.esotericsoftware:kryo-shaded:${kryoVersion}"
```
frankgh, line 148: Let's  omit the kryo version. My fear is that the version number might get out of sync from what `cassandra-analytics-cdc_spark3_2.12` is bringing transitively.  I am also uneasy about `cassandra-analytics-cdc_spark3_2.12` pulling guava. I'd love to get rid of guava as a dependency from the upstream, but it is widely used.
```suggestion
```

server/src/main/java/org/apache/cassandra/bridge/CassandraBridgeFactory.java
```diff
@@ -0,0 +1,145 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.Arrays;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+import com.google.common.base.Preconditions;
```
frankgh: we should use Sidecar's preconditions in general, but in this case it's not needed. We should use `Objects.requireNonNull`

server/src/main/java/org/apache/cassandra/bridge/CassandraBridgeFactory.java
```diff
@@ -0,0 +1,145 @@
...
+    public CassandraBridge get(@NotNull CassandraVersionFeatures features)
+    {
+        return get(getCassandraVersion(features));
+    }
+
+    @NotNull
+    public CassandraBridge get(@NotNull CassandraVersion version)
+    {
+        String jarBaseName = version.jarBaseName();
+        Preconditions.checkNotNull(jarBaseName, "Cassandra version " + version + " is not supported");
```
frankgh:
```suggestion
        Objects.requireNonNull(jarBaseName, "Cassandra version " + version + " is not supported");
```

server/src/main/java/org/apache/cassandra/bridge/CassandraBridgeFactory.java
```diff
@@ -0,0 +1,145 @@
...
+                           }).toArray(URL[]::new);
+
+        return AccessController.doPrivileged(new PrivilegedAction<ClassLoader>()
+        {
+            public ClassLoader run()
+            {
+                return new PostDelegationClassLoader(urls, Thread.currentThread().getContextClassLoader());
+            }
+        });
+    }
```
frankgh: NIT:
```suggestion
                return AccessController.doPrivileged((PrivilegedAction<ClassLoader>) () ->
                                                                             new PostDelegationClassLoader(urls, Thread.currentThread().getContextClassLoader()));
```

server/src/main/java/org/apache/cassandra/sidecar/cdc/SidecarCdcStats.java
```diff
@@ -0,0 +1,303 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.cassandra.sidecar.cdc;
+
+/**
```
frankgh, line 22: I find the AI generated javadocs to be less useful than a concise description of what the method does. In general I'm not liking these

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CdcConfigurationImpl.java
```diff
@@ -35,10 +35,13 @@ public class CdcConfigurationImpl implements CdcConfiguration
     private static final Logger LOGGER = LoggerFactory.getLogger(CdcConfigurationImpl.class);
     public static final String IS_ENABLED_PROPERTY = "enabled";
     public static final String CONFIGURATION_REFRESH_TIME_PROPERTY = "config_refresh_time";
+    public static final String TABLE_SCHEMA_REFRESH_TIME_PROPERTY = "table_schema_refresh_time";
```
frankgh, line 38: can we add this to sidecar.yaml and document it there as well?

server/src/main/java/org/apache/cassandra/sidecar/db/CdcDatabaseAccessor.java
```diff
@@ -0,0 +1,298 @@
...
+import org.apache.cassandra.sidecar.utils.ByteBufUtils;
+import org.apache.cassandra.sidecar.utils.InstanceMetadataFetcher;
+import org.apache.cassandra.sidecar.utils.TokenSplitUtil;
+import org.apache.cassandra.spark.data.partitioner.Partitioner;
+import org.apache.cassandra.spark.utils.TableIdentifier;
+import org.apache.cassandra.spark.utils.ThrowableUtils;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Database accessor for CDC (Change Data Capture) state management operations.
```
frankgh, line 54: again this javadoc is not very useful and generally distracting

server/src/main/java/org/apache/cassandra/sidecar/db/CdcDatabaseAccessor.java
```diff
@@ -0,0 +1,298 @@
...
+ */
+@SuppressWarnings("resource")
+@Singleton
+public class CdcDatabaseAccessor extends DatabaseAccessor<CdcStatesSchema>
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(CdcDatabaseAccessor.class);
+    private final TableHistorySchema tableHistorySchema;
+    private final Provider<TokenSplitUtil> tokenSplitUtilProvider;
+    private volatile TokenSplitUtil tokenSplitUtil = null;
+    private volatile InstanceMetadataFetcher instanceMetadataFetcher;
```
frankgh: this should not be volatile, instead we should make it final
```suggestion
    private final InstanceMetadataFetcher instanceMetadataFetcher;
```

server/src/main/java/org/apache/cassandra/sidecar/db/CdcDatabaseAccessor.java
```diff
@@ -0,0 +1,298 @@
...
+    {
+        super(cdcStatesSchema, sessionProvider);
+        this.tableHistorySchema = tableHistorySchema;
+        this.tokenSplitUtilProvider = tokenSplitUtilProvider;
+        this.instanceMetadataFetcher = instanceMetadataFetcher;
+        this.cassandraBridgeFactory = cassandraBridgeFactory;
+    }
+
+    @VisibleForTesting
+    public CdcDatabaseAccessor(InstanceMetadataFetcher instanceMetadataFetcher,
```
frankgh: while this ctor is annotated with `@VisibleForTesting` I don't see it being used anywhere. Maybe you are missing tests for this class?

server/src/main/java/org/apache/cassandra/sidecar/db/CdcDatabaseAccessor.java
```diff
@@ -0,0 +1,298 @@
...
+        this.tableHistorySchema = tableHistorySchema;
+        this.tokenSplitUtilProvider = () -> tokenSplitUtil;
+        this.instanceMetadataFetcher = instanceMetadataFetcher;
+        this.cassandraBridgeFactory = cassandraBridgeFactory;
+    }
+
+    protected TokenSplitUtil tokenSplitUtil()
+    {
+        // cql connection must be initialized before we can initialize TokenSplitUtil class
+        if (this.tokenSplitUtil == null)
```
frankgh, line 82: do we need to worry about concurrent access to this method here?

server/src/main/java/org/apache/cassandra/sidecar/tasks/CassandraClusterSchema.java
```diff
@@ -0,0 +1,273 @@
...
+            promise.fail(t);
+        }
+    }
+
+    @Override
+    public ScheduleDecision scheduleDecision()
+    {
+        boolean shouldSkip = !sidecarConfiguration.serviceConfiguration().schemaKeyspaceConfiguration().isEnabled()
+                             || !sidecarConfiguration.serviceConfiguration().cdcConfiguration().isEnabled();
+        return shouldSkip ? ScheduleDecision.SKIP : ScheduleDecision.EXECUTE;
```
frankgh: NIT: I find this way more readable
```suggestion
                if (sidecarConfiguration.serviceConfiguration().schemaKeyspaceConfiguration().isEnabled() &&
            sidecarConfiguration.serviceConfiguration().cdcConfiguration().isEnabled())
        {
            return ScheduleDecision.EXECUTE;
        }
        return ScheduleDecision.SKIP;
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/CassandraClusterSchema.java
```diff
@@ -0,0 +1,273 @@
...
+ * <p>
+ * This class is thread-safe and designed as a singleton for dependency injection into CDC
+ * and other schema-dependent components.
+ *
+ * @see org.apache.cassandra.sidecar.db.CdcDatabaseAccessor
+ * @see org.apache.cassandra.bridge.CdcBridge
+ * @see org.apache.cassandra.spark.data.CqlTable
+ */
+@Singleton
+public class CassandraClusterSchema implements PeriodicTask
```
frankgh: I think the name of the class should reflect the fact that this will be some background task the periodically refreshes the schema

server/src/main/java/org/apache/cassandra/sidecar/tasks/CassandraClusterSchema.java
```diff
@@ -0,0 +1,273 @@
...
+
+    public void refresh()
+    {
+        NodeSettings nodeSettings = instanceFetcher.callOnFirstAvailableInstance(instance-> instance.delegate().nodeSettings());
+        CassandraBridge cassandraBridge = cassandraBridgeFactory.get(nodeSettings.releaseVersion());
+        CdcBridge cdcBridge = CdcBridgeFactory.getCdcBridge(cassandraBridge);
+
+        try
+        {
+            LOGGER.info("Checking for schema changes...");
```
frankgh: this at debug level?
```suggestion
            LOGGER.debug("Checking for schema changes...");
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/CassandraClusterSchema.java
```diff
@@ -0,0 +1,273 @@
...
+            {
+                LOGGER.info("Schema change detected, refreshing CDC tables");
+                currSchemaText.set(fullSchemaText);
+                Set<CqlTable> updatedCdcTables = buildCdcTables(fullSchemaText, databaseAccessor, tableIdCache, instanceFetcher, cassandraBridgeFactory);
+                LOGGER.info("Cdc enabled tables tables='{}'", 
+                            updatedCdcTables.stream()
+                                            .map(m -> String.format("%s.%s", m.keyspace(), m.table()))
+                                            .collect(Collectors.joining(",")));
+                cdcTables.set(updatedCdcTables);
+                cdcBridge.updateCdcSchema(updatedCdcTables, databaseAccessor.partitioner(),
```
frankgh: I don't think this `databaseAccessor.partitioner()` is required at all. We already have the partitioner information as a string in `nodeSettings.partitioner()`, with the caveat that we will need to check if the string has `.` and if it does we will need to do the logic we have inside `databaseAccessor.partitioner()`
```suggestion
                cdcBridge.updateCdcSchema(updatedCdcTables, Partitioner.valueOf(nodeSettings.partitioner()),
```

frankgh: and of course Partitioner is in spark, so we wouldn't be able to change it here. So we'll need to do the check before calling valueOF

server/src/main/java/org/apache/cassandra/sidecar/tasks/CassandraClusterSchema.java
```diff
@@ -0,0 +1,273 @@
...
+        catch (Throwable t)
+        {
+            LOGGER.error("Unexpected error while refreshing the schema", t);
+            throw t;
+        }
+    }
+
+    public DurationSpec delay()
+    {
+        return tableSchemaRefreshTime;
```
frankgh: We should keep it like this for when we support dynamic configuration reloading in the future.
```suggestion
        return sidecarConfiguration.serviceConfiguration().cdcConfiguration().tableSchemaRefreshTime()
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/CassandraClusterSchema.java
```diff
@@ -0,0 +1,273 @@
...
+            throw exception;
+        }
+        catch (Throwable t)
+        {
+            LOGGER.error("Unexpected error while refreshing the schema", t);
+            throw t;
+        }
+    }
+
+    public DurationSpec delay()
```
frankgh:
```suggestion
    @Override
    public DurationSpec delay()
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/CassandraClusterSchema.java
```diff
@@ -0,0 +1,273 @@
...
+                                                              .distinct() // remove duplicated keyspace strings
+                                                              .collect(Collectors.toMap(Function.identity(),
+                                                                                        keyspace -> CqlUtils.extractUdts(fullSchema, keyspace)));
+
+        Map<TableIdentifier, UUID> tableIds = createStmts.keySet()
+                                                         .stream()
+                                                         .collect(Collectors.toMap(Function.identity(),
+                                                                                   id -> tableIdCache.computeIfAbsent(id, tableIdLoaderFunction)));
+
+        NodeSettings nodeSettings = instanceFetcher.callOnFirstAvailableInstance(instance-> instance.delegate().nodeSettings());
```
frankgh: We already have the nodeSettings in the callsite for this method, let's just use that

frankgh: as a matter of fact, we have the bridge, we should just pass the bridge

server/src/main/java/org/apache/cassandra/sidecar/utils/ByteBufUtils.java
```diff
@@ -0,0 +1,313 @@
...
+ * operations throughout the sidecar system.
+ * <p>
+ * All methods in this class are static and thread-safe unless otherwise noted.
+ * The class maintains thread-local resources (such as charset decoders) to ensure
+ * optimal performance in multi-threaded environments.
+ *
+ * @see java.nio.ByteBuffer
+ * @see java.nio.charset.CharsetDecoder
+ */
+public class ByteBufUtils
```
frankgh, line 38: NIT, this class has a lot of final modifiers that are not necessary and do not conform with the Cassandra style guide

server/src/main/java/org/apache/cassandra/sidecar/utils/ByteBufUtils.java
frankgh, line 1: This class needs unit tests.

server/src/main/java/org/apache/cassandra/sidecar/utils/CdcUtil.java
```diff
@@ -21,12 +21,67 @@
...
+import java.util.stream.Collectors;
+
+import com.google.common.annotations.VisibleForTesting;
+
+import org.apache.cassandra.spark.utils.TableIdentifier;
+import org.jetbrains.annotations.NotNull;
 
 /**
- * Class with utility methods for CDC.
+ * Utility class providing CDC (Change Data Capture) specific operations and file handling.
```
frankgh, line 38: these AI generated javadocs are too much

frankgh (commented):
Took a first pass at this PR, and have a few comments.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/TableHistorySchema.java
```diff
@@ -0,0 +1,168 @@
...
+ * evolution in long-running CDC pipelines.
+ * <p>
+ * This class is thread-safe and designed as a singleton for dependency injection into
+ * components that require table schema history access.
+ *
+ * @see TableSchema
+ * @see org.apache.cassandra.sidecar.db.CdcDatabaseAccessor
+ */
+@Singleton
+public class TableHistorySchema extends TableSchema
```
frankgh: This schema should only be initialized on the lease holder
```suggestion
public class TableHistorySchema extends TableSchema implements ExecuteOnClusterLeaseholderOnly
```

server/src/main/java/org/apache/cassandra/sidecar/db/schema/TableHistorySchema.java
```diff
@@ -0,0 +1,168 @@
...
+
+        return ksMetadata.getTable(TABLE_SCHEMA_HISTORY) != null;
+    }
+
+    @Override
+    protected String createSchemaStatement()
+    {
+        return String.format("CREATE TABLE IF NOT EXISTS %s.%s (" +
+                             "  ks text," +
+                             "  tb text," +
```
frankgh: NIT: to maintain consistency with Cassandra column naming
```suggestion
                             "  keyspace_name text," +
                             "  table_name text," +
```

server/src/main/java/org/apache/cassandra/sidecar/db/schema/CdcStatesSchema.java
```diff
@@ -0,0 +1,169 @@
...
+ * }</pre>
+ * <p>
+ * This class is thread-safe and designed as a singleton for dependency injection into
+ * components that require CDC state table access.
+ *
+ * @see TableSchema
+ * @see org.apache.cassandra.sidecar.db.CdcDatabaseAccessor
+ */
+@Singleton
+public class CdcStatesSchema extends TableSchema
```
frankgh: should only be initialized in the cluster lease holder
```suggestion
public class CdcStatesSchema extends TableSchema implements ExecuteOnClusterLeaseholderOnly
```

server/src/main/java/org/apache/cassandra/sidecar/modules/CdcModule.java
```diff
@@ -85,6 +97,42 @@ PeriodicTask cdcRawDirectorySpaceCleanercPeriodicTask(CdcRawDirectorySpaceCleane
         return cleanerTask;
     }
 
+    @ProvidesIntoMap
+    @KeyClassMapKey(PeriodicTaskMapKeys.CassandraClusterSchemaTaskKey.class)
+    PeriodicTask cassandraClusterSchema(InstanceMetadataFetcher instanceMetadataFetcher,
```
frankgh:
```suggestion
    PeriodicTask cassandraClusterSchemaMonitor(InstanceMetadataFetcher instanceMetadataFetcher,
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/CassandraClusterSchemaMonitor.java
```diff
@@ -0,0 +1,233 @@
...
+                              instanceFetcher,
+                              cassandraBridge);
+    }
+
+    private static Set<CqlTable> buildCdcTables(@NotNull final String fullSchema,
+                                                @NotNull final Partitioner partitioner,
+                                                @NotNull final ConcurrentHashMap<TableIdentifier, UUID> tableIdCache,
+                                                @NotNull final Function<TableIdentifier, UUID> tableIdLoaderFunction,
+                                                @NotNull final InstanceMetadataFetcher instanceFetcher,
+                                                @NotNull final CassandraBridge cassandraBridge)
```
frankgh: we can omit final, and instanceFetcher is no longer needed
```suggestion
    private static Set<CqlTable> buildCdcTables(@NotNull String fullSchema,
                                                @NotNull Partitioner partitioner,
                                                @NotNull ConcurrentHashMap<TableIdentifier, UUID> tableIdCache,
                                                @NotNull Function<TableIdentifier, UUID> tableIdLoaderFunction,
                                                @NotNull CassandraBridge cassandraBridge)
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/CassandraClusterSchemaMonitor.java
```diff
@@ -0,0 +1,233 @@
...
+                              cdcDatabaseAccessor::getTableId,
+                              instanceFetcher,
+                              cassandraBridge);
+    }
+
+    private static Set<CqlTable> buildCdcTables(@NotNull String fullSchema,
+                                                @NotNull CdcDatabaseAccessor cdcDatabaseAccessor,
+                                                @NotNull ConcurrentHashMap<TableIdentifier, UUID> tableIdCache,
+                                                @NotNull InstanceMetadataFetcher instanceFetcher,
+                                                @NotNull final CassandraBridge cassandraBridge)
```
frankgh: instanceFetcher is no longer needed
```suggestion
    private static Set<CqlTable> buildCdcTables(@NotNull String fullSchema,
                                                @NotNull CdcDatabaseAccessor cdcDatabaseAccessor,
                                                @NotNull ConcurrentHashMap<TableIdentifier, UUID> tableIdCache,
                                                @NotNull CassandraBridge cassandraBridge)
```

server/src/main/java/org/apache/cassandra/sidecar/tasks/CassandraClusterSchemaMonitor.java
```diff
@@ -0,0 +1,233 @@
...
+            return ScheduleDecision.EXECUTE;
+        }
+        return ScheduleDecision.SKIP;
+    }
+
+    @VisibleForTesting
+    static Set<CqlTable> buildCdcTables(CdcDatabaseAccessor cdcDatabaseAccessor,
+                                        ConcurrentHashMap<TableIdentifier, UUID> tableIdCache,
+                                        @NotNull InstanceMetadataFetcher instanceFetcher,
+                                        @NotNull final CassandraBridge cassandraBridge)
```
frankgh: instanceFetcher is no longer needed
```suggestion
    static Set<CqlTable> buildCdcTables(CdcDatabaseAccessor cdcDatabaseAccessor,
                                        ConcurrentHashMap<TableIdentifier, UUID> tableIdCache,
                                        @NotNull CassandraBridge cassandraBridge)
```

frankgh (commented):
I added some additional comments

server/src/main/java/org/apache/cassandra/sidecar/db/CdcDatabaseAccessor.java
frankgh: this is unused. Let's remove it?

server/src/test/java/org/apache/cassandra/sidecar/db/CdcDatabaseAccessorTests.java
```diff
@@ -0,0 +1,522 @@
...
+        MockCdcStateV2 datastore = new MockCdcStateV2();
+        String jobId = UUID.randomUUID().toString();
+        CdcStatesSchema mockCdcStatesSchema = mock(CdcStatesSchema.class, RETURNS_DEEP_STUBS);
+        TableHistorySchema mockTableHistorySchema = mock(TableHistorySchema.class, RETURNS_DEEP_STUBS);
+        List<BigInteger> tokensBeforeShrink = TokenSplitUtil.splitTokens(numNodes, partitioner);
+        List<BigInteger> tokensAfterShrink = TokenSplitUtil.splitTokens(numNodes / 2, partitioner);
+        TokenSplitUtil tokenSplitUtil = new TokenSplitUtil(numNodes);
+
+        Provider<TokenSplitUtil> mockTokenSplitUtilProvider = mock(Provider.class);
+        when(mockTokenSplitUtilProvider.get()).thenReturn(tokenSplitUtil);
```
frankgh: NIT: 
```suggestion
        Provider<TokenSplitUtil> tokenSplitUtilProvider = () -> tokenSplitUtil;
```

server/src/test/java/org/apache/cassandra/sidecar/db/CdcDatabaseAccessorTests.java
```diff
@@ -0,0 +1,522 @@
...
+        Partitioner partitioner = Partitioner.Murmur3Partitioner;
+        MockCdcStateV2 datastore = new MockCdcStateV2();
+        String jobId = UUID.randomUUID().toString();
+        CdcStatesSchema mockCdcStatesSchema = mock(CdcStatesSchema.class, RETURNS_DEEP_STUBS);
+        TableHistorySchema mockTableHistorySchema = mock(TableHistorySchema.class, RETURNS_DEEP_STUBS);
+        List<BigInteger> tokensBeforeExpansion = TokenSplitUtil.splitTokens(numNodes, partitioner);
+        List<BigInteger> tokensAfterExpansion = TokenSplitUtil.splitTokens(numNodes * 2, partitioner);
+        TokenSplitUtil tokenSplitUtil = new TokenSplitUtil(numNodes);
+
+        Provider<TokenSplitUtil> mockTokenSplitUtilProvider = mock(Provider.class);
```
frankgh: NIT:
```suggestion
        Provider<TokenSplitUtil> tokenSplitUtilProvider = () -> tokenSplitUtil;
```

server/src/test/java/org/apache/cassandra/sidecar/db/CdcDatabaseAccessorTests.java
```diff
@@ -0,0 +1,522 @@
...
+import org.apache.cassandra.sidecar.db.schema.CdcStatesSchema;
+import org.apache.cassandra.sidecar.db.schema.TableHistorySchema;
+import org.apache.cassandra.sidecar.utils.InstanceMetadataFetcher;
+import org.apache.cassandra.sidecar.utils.TokenSplitUtil;
+import org.apache.cassandra.spark.data.partitioner.Partitioner;
+import org.jetbrains.annotations.NotNull;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
```
frankgh: let's migrate to `org.assertj.core.api.Assertions.assertThat` to be consistent with the repo

server/src/test/java/org/apache/cassandra/sidecar/utils/ByteBufUtilsTest.java
```diff
@@ -0,0 +1,101 @@
...
+import com.google.common.io.ByteStreams;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+/**
+ * Test for {@link ByteBufUtils}
+ */
+@SuppressWarnings("UnstableApiUsage")
```
frankgh: this looks like it might not be needed?

server/src/test/java/org/apache/cassandra/sidecar/utils/ByteBufUtilsTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    private static void testReadRemainingBytes(final String str) throws IOException
+    {
+        assertEquals(str, new String(ByteBufUtils.readRemainingBytes(new ByteArrayInputStream(str.getBytes()), str.length()), StandardCharsets.UTF_8));
+    }
+
+    private static void testSkipBytesFully(final byte[] ar) throws IOException
+    {
+        final int len = ar.length;
+        final ByteArrayDataInput in = ByteStreams.newDataInput(ar, 0);
```
frankgh: NIT:
```suggestion
       int len = ar.length;
        ByteArrayDataInput in = ByteStreams.newDataInput(ar, 0);
```

server/src/main/java/org/apache/cassandra/sidecar/utils/TokenSplitUtil.java
```diff
@@ -0,0 +1,200 @@
...
+    protected static int getOrInsert(TokenSplitConfigAccessor tokenSplitConfigAccessor, CdcConfig cdcConfig, InstanceMetadataFetcher fetcher)
+    {
+        Map<String, String> config = tokenSplitConfigAccessor.getConfig().getConfigs();
+        if (config != null && config.containsKey(NUM_SPLITS_KEY))
+        {
+            return Integer.parseInt(config.get(NUM_SPLITS_KEY));
+        }
+
+
+        int maxDcSize = fetcher.callOnFirstAvailableInstance(instanceMetadata -> instanceMetadata.delegate().metadata())
```
frankgh, line 91: if sidecar cannot connect to CQL this will prevent the startup of the service. We should not fail here. Also we should have a more descriptive error message when we throw. But in general we should never throw when building the DAG for guice.

server/src/main/java/org/apache/cassandra/sidecar/utils/CdcUtil.java
```diff
@@ -175,4 +192,119 @@ public static boolean matchIndexExtension(String fileName)
     {
         return fileName.endsWith(IDX_FILE_EXTENSION);
     }
+
+    /**
+     * @param schemaStr full cluster schema text.
+     * @return map of keyspace/table identifier to table create statements.
+     */
+    public static Map<TableIdentifier, String> extractCdcTables(@NotNull final String schemaStr)
```
frankgh: maybe add a unit test for this function?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/TableHistorySchema.java
```diff
@@ -0,0 +1,128 @@
...
+    public PreparedStatement selectVersionTableSchema()
+    {
+        return selectVersionTableSchema;
+    }
+
+    private static class CqlLiterals
+    {
+        static String insertTableSchema(SchemaKeyspaceConfiguration config)
+        {
+            return withTable("INSERT INTO %s.%s (ks, tb, version, created_at, table_schema) " +
```
frankgh: after the change above, we should use the new names for the keyspace, table columns
```suggestion
            return withTable("INSERT INTO %s.%s (keyspace_name, table_name, version, created_at, table_schema) " +
```

server/src/main/java/org/apache/cassandra/sidecar/db/schema/TableHistorySchema.java
```diff
@@ -0,0 +1,128 @@
...
+                             "VALUES (?, ?, ?, NOW(), ?)", config);
+        }
+
+        static String selectVersionTableSchema(SchemaKeyspaceConfiguration config)
+        {
+            return withTable("SELECT table_schema FROM %s.%s WHERE ks = ? AND tb = ? AND version = ?", config);
```
frankgh:
```suggestion
            return withTable("SELECT table_schema FROM %s.%s WHERE keyspace_name = ? AND table_name = ? AND version = ?", config);
```

frankgh (commented):
I think we still have some pending things to resolve.

frankgh (approved):
+1 looks good to me

PR #252, opened 2025-08-19
225 Live Migration data copy api

nvharikrishna:

[CASSSIDECAR-225](https://issues.apache.org/jira/browse/CASSSIDECAR-225) Created API for Creating Live Migration Tasks, data copy & get tasks.


client-common/src/test/java/org/apache/cassandra/sidecar/common/request/LiveMigrationDataCopyRequestTest.java
frankgh, line 1: Indentation in this file is a bit off. I think there's an [effort to standardize](https://github.com/apache/cassandra-sidecar/pull/235) on a style and make it automatic, but I think for now we should try to stay close to the conventions we have in the project

server/src/main/java/org/apache/cassandra/sidecar/concurrent/AsyncConcurrentTaskExecutor.java
```diff
@@ -0,0 +1,285 @@
...
+                                       int maxConcurrency)
+    {
+        Objects.requireNonNull(vertx, "vertx must not be null.");
+        Objects.requireNonNull(tasks, "tasks must not be null.");
+        if (maxConcurrency < 1)
+        {
+            throw (new IllegalArgumentException("maxConcurrency must be > 0"));
+        }
+
+        this.tasks = Collections.unmodifiableList(tasks);
```
frankgh: NIT: prefer newer JDK APIs
```suggestion
        this.tasks = List.copyOf(tasks);
```

server/src/main/java/org/apache/cassandra/sidecar/concurrent/AsyncConcurrentTaskExecutor.java
```diff
@@ -0,0 +1,285 @@
...
+                         {
+                             return task.call();
+                         }
+                         catch (Exception e)
+                         {
+                             return Future.failedFuture(e);
+                         }
+                     })
+                     .onComplete(ar -> {
+                         if (ar.succeeded() || !(ar.cause() instanceof CancellationException))
```
frankgh, line 195: does it make sense to log here if the task did not succeed?

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationDirType.java
```diff
@@ -28,7 +28,7 @@ public enum LiveMigrationDirType
 {
     CDC_RAW_DIR("cdc_raw"),
     COMMIT_LOG_DIR("commitlog"),
-    DATA_FIlE_DIR("data"),
+    DATA_FILE_DIR("data"),
```
frankgh, line 31: wow, it took me 10 seconds to realize the diff

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationTaskFactoryImpl.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public LiveMigrationTask create(final LiveMigrationDataCopyRequest request,
+                                    final String source,
+                                    final int port,
+                                    final InstanceMetadata instanceMetadata)
```
frankgh: NIT
```suggestion
    public LiveMigrationTask create(LiveMigrationDataCopyRequest request,
                                    String source,
                                    int port,
                                    InstanceMetadata instanceMetadata)
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
```diff
@@ -0,0 +1,430 @@
...
+    /**
+     * Transitions to the CLEANING state with updated file metadata.
+     *
+     * @param totalSize  total size of files at the source
+     * @param totalFiles total number of files at the source
+     * @return new OperationStatus instance in CLEANING state
+     * @throws IllegalStateTransitionException if current state cannot transition to CLEANING
+     */
+    @VisibleForTesting
+    public OperationStatus getCleaningState(long totalSize, int totalFiles)
```
frankgh: should this be renamed to indicate that a transition is occurring? for example `toCleaningState` ?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    /**
+     * Transitions to the PREPARING state.
+     *
+     * @return new OperationStatus instance in PREPARING state
+     * @throws IllegalStateTransitionException if current state cannot transition to PREPARING
+     */
+    @VisibleForTesting
+    public OperationStatus getPreparingState()
```
frankgh:
```suggestion
    public OperationStatus toPreparingState()
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
```diff
@@ -0,0 +1,430 @@
...
+    /**
+     * Transitions to the DOWNLOADING state with updated download metadata.
+     * Resets progress counters for the new download phase.
+     *
+     * @param bytesToDownload total size of data to be downloaded in bytes
+     * @param filesToDownload number of files to be downloaded
+     * @return new OperationStatus instance in DOWNLOADING state
+     * @throws IllegalStateTransitionException if current state cannot transition to DOWNLOADING
+     */
+    OperationStatus getDownloadingState(final long bytesToDownload, final int filesToDownload)
```
frankgh:
```suggestion
    OperationStatus toDownloadingState(long bytesToDownload, int filesToDownload)
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
```diff
@@ -0,0 +1,430 @@
...
+                                   new AtomicLong());
+    }
+
+    /**
+     * Transitions to the DOWNLOAD_COMPLETE state.
+     *
+     * @return new OperationStatus instance in DOWNLOAD_COMPLETE state
+     * @throws IllegalStateTransitionException if current state cannot transition to DOWNLOAD_COMPLETE
+     */
+    OperationStatus getDownloadCompleteState()
```
frankgh:
```suggestion
    OperationStatus toDownloadCompleteState()
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    /**
+     * Transitions to the SUCCESS state, indicating successful completion.
+     *
+     * @return new OperationStatus instance in SUCCESS state
+     * @throws IllegalStateTransitionException if current state cannot transition to SUCCESS
+     */
+    @VisibleForTesting
+    public OperationStatus getSuccessState()
```
frankgh:
```suggestion
    public OperationStatus toSuccessState()
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
```diff
@@ -0,0 +1,430 @@
...
+        return filesToDownload;
+    }
+
+    /**
+     * Returns the thread-safe atomic counter for files downloaded.
+     * This counter can be safely read and updated from multiple threads.
+     *
+     * @return atomic counter for files downloaded
+     */
+    public AtomicInteger filesDownloaded()
```
frankgh: I don't think we should expose the internal atomic integer values here, but instead expose methods that will internally update the counts. i.e. `markFileDownloaded()`

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
```diff
@@ -0,0 +1,430 @@
...
+        return filesDownloaded;
+    }
+
+    /**
+     * Returns the thread-safe atomic counter for bytes downloaded.
+     * This counter can be safely read and updated from multiple threads.
+     *
+     * @return atomic counter for bytes downloaded
+     */
+    public AtomicLong bytesDownloaded()
```
frankgh: In general -1 on exposing these values directly and instead providing business oriented methods that allow you to internally modify these values

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
```diff
@@ -0,0 +1,430 @@
...
+     *   <li><b>DOWNLOADING</b> - Actively downloading files from the source</li>
+     *   <li><b>DOWNLOAD_COMPLETE</b> - All files have been downloaded (terminal state)</li>
+     *   <li><b>SUCCESS</b> - Operation completed successfully (terminal state)</li>
+     *   <li><b>FAILED</b> - Operation failed due to an error (terminal state)</li>
+     *   <li><b>CANCELLED</b> - Operation was cancelled by user request (terminal state)</li>
+     * </ul>
+     * <p>
+     * Invalid transitions throw {@link IllegalStateTransitionException}.
+     */
+    public enum State
```
frankgh, line 422: I think it would be valuable to have unit tests for this specific enum that focus on testing the transitions

server/src/main/java/org/apache/cassandra/sidecar/modules/LiveMigrationModule.java
```diff
@@ -51,11 +58,113 @@ public class LiveMigrationModule extends AbstractModule
...
+                      .build();
+    }
+
+    @PATCH
+    @Operation(summary = "Cancel data copy task",
+    description = "Cancels an existing data copy task for live migration")
+    @APIResponse(description = "Data copy task cancelled successfully",
+                 responseCode = "200",
+                 content = @Content(mediaType = "application/json",
+                 schema = @Schema(type = SchemaType.OBJECT)))
```
frankgh:
```suggestion
                 schema = @Schema(implementation = LiveMigrationTaskResponse.class)))
```

server/src/main/java/org/apache/cassandra/sidecar/modules/LiveMigrationModule.java
```diff
@@ -51,11 +58,113 @@ public class LiveMigrationModule extends AbstractModule
...
+                      .build();
+    }
+
+    @GET
+    @Operation(summary = "Get data copy task",
+               description = "Retrieves the status and details of a specific data copy task by task ID")
+    @APIResponse(description = "Data copy task retrieved successfully",
+                 responseCode = "200",
+                 content = @Content(mediaType = "application/json",
+                 schema = @Schema(type = SchemaType.OBJECT)))
```
frankgh:
```suggestion
                 schema = @Schema(implementation = LiveMigrationTaskResponse.class)))
```

server/src/main/java/org/apache/cassandra/sidecar/modules/LiveMigrationModule.java
```diff
@@ -68,10 +177,10 @@ protected void configure()
                  schema = @Schema(type = SchemaType.OBJECT)))
     @ProvidesIntoMap
     @KeyClassMapKey(VertxRouteMapKeys.LiveMigrationFileStreamHandlerRouteKey.class)
-    VertxRoute downloadFileRoute(RouteBuilder.Factory factory,
-                                 LiveMigrationApiEnableDisableHandler liveMigrationApiEnableDisableHandler,
-                                 LiveMigrationFileStreamHandler liveMigrationFileStreamHandler,
-                                 FileStreamHandler fileStreamHandler)
+    public VertxRoute downloadFileRoute(RouteBuilder.Factory factory,
```
frankgh, line 181: I think these shouldn't be public. Any reason we are changing the visibility here?

server/src/test/java/org/apache/cassandra/sidecar/concurrent/AsyncConcurrentTaskExecutorTest.java
```diff
@@ -0,0 +1,487 @@
...
+
+    @Test
+    public void testSingleTaskSuccess(VertxTestContext context)
+    {
+        List<Callable<Future<String>>> tasks = Collections.singletonList(getDummySuccessTask("Task executed successfully"));
+        AsyncConcurrentTaskExecutor<String> concurrentTaskExecutor = new AsyncConcurrentTaskExecutor<>(vertx, tasks, 1);
+        List<Future<String>> taskFutures = concurrentTaskExecutor.start();
+        Future.join(taskFutures).onComplete(ar -> context.verify(() -> {
+            assertThat(ar).isNotNull();
+            assertThat(ar.result().list().get(0)).isEqualTo("Task executed successfully");
```
frankgh, line 102: I think we need to wrap the future in context.succeeding, otherwise errors will be swallowed 
```suggestion
        Future.join(taskFutures).onComplete(context.succeeding(result -> context.verify(() -> {
            assertThat(result).isNotNull();
            assertThat(result.list().get(0)).isEqualTo("Task executed successfully");
```

server/src/test/java/org/apache/cassandra/sidecar/concurrent/AsyncConcurrentTaskExecutorTest.java
```diff
@@ -0,0 +1,487 @@
...
+    }
+
+    @Test
+    public void testMultipleTasksSucceeds(VertxTestContext context)
+    {
+        List<Callable<Future<Boolean>>> tasks = getDummySuccessTasks(10, Boolean.TRUE);
+        AsyncConcurrentTaskExecutor<Boolean> concurrentTaskExecutor = new AsyncConcurrentTaskExecutor<>(vertx, tasks, 5);
+        List<Future<Boolean>> taskFutures = concurrentTaskExecutor.start();
+        Future.join(taskFutures).onComplete(ar -> context.verify(() -> {
+            assertThat(ar).isNotNull();
```
frankgh, line 114:
```suggestion
        Future.join(taskFutures).onComplete(context.succeeding(result -> context.verify(() -> {
            assertThat(result).isNotNull();
```

server/src/test/java/org/apache/cassandra/sidecar/concurrent/AsyncConcurrentTaskExecutorTest.java
```diff
@@ -0,0 +1,487 @@
...
+    }
+
+    @Test
+    public void testLotsOfTasks(VertxTestContext context)
+    {
+        List<Callable<Future<Integer>>> tasks = getDummySuccessTasks(100_000, 1);
+        AsyncConcurrentTaskExecutor<Integer> concurrentTaskExecutor = new AsyncConcurrentTaskExecutor<>(vertx, tasks, 50);
+        List<Future<Integer>> taskFutures = concurrentTaskExecutor.start();
+        Future.join(taskFutures).onComplete(ar -> context.verify(() -> {
+            assertThat(ar).isNotNull();
```
frankgh, line 131:
```suggestion
        Future.join(taskFutures).onComplete(context.succeeding(result -> context.verify(() -> {
            assertThat(result).isNotNull();
```

server/src/test/java/org/apache/cassandra/sidecar/concurrent/AsyncConcurrentTaskExecutorTest.java
```diff
@@ -0,0 +1,487 @@
...
+        for (int i = 0; i < 1000; i++)
+        {
+            Callable<Future<Integer>> task = getDummySuccessCounterTask(counter);
+            tasks.add(task);
+        }
+        TestAsyncTaskExecutor<Integer> concurrentTaskExecutor =
+        new TestAsyncTaskExecutor<>(vertx, tasks, maxConcurrency, counter);
+        List<Future<Integer>> taskFutures = concurrentTaskExecutor.start();
+
+        Future.join(taskFutures).onComplete(ar -> context.verify(() -> {
```
frankgh, line 270:
```suggestion
                Future.join(taskFutures).onComplete(ar -> context.verify(() -> {
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateDataCopyTaskHandler.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    protected void handleInternal(RoutingContext context,
+                                  HttpServerRequest httpRequest,
+                                  @NotNull String host,
+                                  SocketAddress remoteAddress,
+                                  LiveMigrationDataCopyRequest liveMigrationTaskRequest)
+    {
+        dataCopyTaskManager
```
frankgh, line 97: I see create task has some sort of concurrency limiter control. There is something similar implemented in `org.apache.cassandra.sidecar.handlers.sstableuploads.SSTableUploadHandler` leveraging the `org.apache.cassandra.sidecar.concurrent.ConcurrencyLimiter`

server/src/main/java/org/apache/cassandra/sidecar/modules/LiveMigrationModule.java
```diff
@@ -51,11 +58,113 @@ public class LiveMigrationModule extends AbstractModule
...
         bind(LiveMigrationMap.class).to(LiveMigrationMapSidecarConfigImpl.class);
+        bind(LiveMigrationTaskFactory.class).to(LiveMigrationTaskFactoryImpl.class);
     }
 
+    @GET
+    @Path(ApiEndpointsV1.LIVE_MIGRATION_DATA_COPY_TASKS_ROUTE)
+    @Operation(summary = "Create data copy task",
+               description = "Creates a new data copy task for live migration")
+    @APIResponse(description = "Data copy task created successfully",
+                 responseCode = "200",
```
frankgh: I think it is returning an ACCEPTED status code, which I believe is 202, no?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
frankgh, line 1: NIT : usage of `final` in this class.

frankgh (commented):
looks good so far, I have a few comments

client-common/src/test/java/org/apache/cassandra/sidecar/common/request/LiveMigrationDataCopyRequestTest.java
nvharikrishna: yeah, somehow missed it. Updated the code.

server/src/main/java/org/apache/cassandra/sidecar/concurrent/AsyncConcurrentTaskExecutor.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/concurrent/AsyncConcurrentTaskExecutor.java
nvharikrishna: Expecting the caller to log if required.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationDirType.java
nvharikrishna: yeah, I missed it last time.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationTaskFactoryImpl.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
nvharikrishna: right, already named the methods in State in the recommended way, but missed to change here. Updated similar methods too.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
nvharikrishna: Makes sense. Added methods to increment counters.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/OperationStatus.java
nvharikrishna: okay, done

server/src/main/java/org/apache/cassandra/sidecar/modules/LiveMigrationModule.java
nvharikrishna: right

server/src/main/java/org/apache/cassandra/sidecar/modules/LiveMigrationModule.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/modules/LiveMigrationModule.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/modules/LiveMigrationModule.java
nvharikrishna: Injellij showing `@GET method must be public` error. So, modified them to public.

server/src/test/java/org/apache/cassandra/sidecar/concurrent/AsyncConcurrentTaskExecutorTest.java
nvharikrishna: It failed when there are assertion failures or exceptions. Tried the following and the test did fail
```java
        Future.join(taskFutures).onComplete(ar -> context.verify(() -> {
            assertThat(1).isEqualTo(2);  // Making sure that it fails
            assertThat(ar).isNotNull();
            assertThat(ar.result().list().get(0)).isEqualTo("Task executed successfully");
            context.completeNow();
        }));
```

context.verify is doing this:

```java
  public VertxTestContext verify(ExecutionBlock block) {
    Objects.requireNonNull(block, "The block cannot be null");
    try {
      block.apply();
    } catch (Throwable t) {
      failNow(t);
    }
    return this;
  }
```

As per vertx documentation "_[verify](https://vertx.io/docs/apidocs/io/vertx/junit5/VertxTestContext.html#verify-io.vertx.junit5.VertxTestContext.ExecutionBlock-) to perform assertions, any exception thrown from the code block is considered as a test failure_", I feel context.verify is good enough.

server/src/test/java/org/apache/cassandra/sidecar/concurrent/AsyncConcurrentTaskExecutorTest.java
nvharikrishna: Don't need this extra indentation right?

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateDataCopyTaskHandler.java
nvharikrishna: Thanks for pointing to it. Saw usage of `ConcurrencyLimiter`. In this case, only one task is allowed to be in progress at a time per instance which is handled by concurrency map. If I have to use `ConcurrencyLimiter`, then need to maintain one for each instance and started looking bit complex. I feel current implementation is simple and adequate to for this scenario. Is that fine?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -0,0 +1,688 @@
...
+    final Consumer<OperationStatus> statusUpdater;
+    final InstanceMetadata instanceMetadata;
+    final LiveMigrationConfiguration liveMigrationConfiguration;
+    final SidecarClient sidecarClient;
+    final String source;
+    final int port;
+    final String logPrefix;
+    final ExecutorPools executorPools;
+    OperationStatus operationStatus;
+    AsyncConcurrentTaskExecutor<Void> concurrentTaskExecutor;
```
yifan-c: why are they package-private? Seems that all members are only accessed within the same class. Maybe just private?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -0,0 +1,688 @@
...
+        }
+
+        return performDownload(instanceFiles);
+    }
+
+    private Future<OperationStatus> performDownload(List<InstanceFileInfo> instanceFiles)
+    {
+        long downloadSize = calculateDownloadSize(instanceFiles);
+        LOGGER.info("{} Downloading {} files from {}:{}", logPrefix, instanceFiles.size(), source, port);
+        LOGGER.info("{} download size: {}", logPrefix, downloadSize);
```
yifan-c: let's combine both? There is no guarantee that the log messages are logged next to each other, unless there is one single logging.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -0,0 +1,688 @@
...
+            {
+                LOGGER.debug("{} Deleting unwanted file: {}", logPrefix, localPath.toAbsolutePath());
+                vertx.fileSystem().deleteBlocking(localPath.toString());
+                filesDeleted.incrementAndGet();
+            }
+        }
+
+        LOGGER.info("{} Deleted {} unwanted files from local.", logPrefix, filesDeleted.get());
+        LOGGER.info("{} Deleted {} unwanted directories from local.", logPrefix, directoriesDeleted.get());
+    }
```
yifan-c: combine both?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -0,0 +1,688 @@
...
+            // only empty directories and/or empty files left to create.
+            return Future.succeededFuture(filesToDownload);
+        }
+
+        long totalSize = instanceFilesListResponse.getTotalSize();
+
+        // Calculate percentage of data required to download
+        double toDownloadPercentage = ((totalSize - downloadSize) * 1.0) / totalSize;
+        LOGGER.info("{} Remaining size to download={} bytes", logPrefix, downloadSize);
+        LOGGER.info("{} Download percentage={}, success threshold={}", logPrefix, toDownloadPercentage, successThreshold);
```
yifan-c: 1. combine both?
2. I suggest changing the field names in the log message. See below. The reason is to have the fieldName a single word (camel case to be consistent with other log message) in order to be extracted by log analysis tooling in a meaningful way. Please check the other log messages in the patch too. Also let's make sure to use the consistent filed name, so aggregation in log analysis makes sense. 

```suggestion
        LOGGER.info("{} Remaining size to download={} bytes. downloadedPercentage={} successThreshold={}", 
                    logPrefix, downloadSize, toDownloadPercentage, successThreshold);
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -0,0 +1,688 @@
...
+
+        if (toDownloadPercentage <= successThreshold)
+        {
+            return Future.succeededFuture(filesToDownload);
+        }
+
+        LOGGER.info("{} Download size met the successThreshold({}).{}",
+                    logPrefix,
+                    successThreshold,
+                    " Skipping download of " + filesToDownload.size() + " files of size " + downloadSize);
```
yifan-c: can you integrate the string into the log message template?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -0,0 +1,688 @@
...
+                                  {
+                                      return Future.succeededFuture(this.updateState(OperationStatus::toDownloadCompleteState));
+                                  }
+                                  LOGGER.warn("{} Failed to download few files. Updating state to {}", logPrefix, State.FAILED);
+                                  return Future.succeededFuture(this.updateState(OperationStatus::tryFailureState));
+                              },
+                              ar -> {
+                                  LOGGER.error("{} Failed to download files", logPrefix, ar);
+                                  return Future.succeededFuture(this.updateState(OperationStatus::tryFailureState));
+                              });
```
yifan-c: Handle failures in one place and simplify a bit

```suggestion
        return Future.join(futureList)
                     .compose(res -> res.succeeded() ? Future.succeededFuture() : Future.failedFuture(res.cause()))
                     .compose(ar -> Future.succeededFuture(this.updateState(OperationStatus::toDownloadCompleteState)),
                              cause -> {
                                  LOGGER.error("{} Failed to download files.  Updating state to {}", logPrefix, State.FAILED, cause);
                                  return Future.succeededFuture(this.updateState(OperationStatus::tryFailureState));
                              });
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -0,0 +1,688 @@
...
+     *
+     * @param file File info of remote.
+     * @return Returns {@link Optional#empty()} if file gets created successfully,
+     * otherwise returns an Optional of exception
+     */
+    @VisibleForTesting
+    Future<Void> createEmptyFile(InstanceFileInfo file)
+    {
+        return localPathAsync(file.fileUrl, instanceMetadata)
+               .expecting(path -> path.getParent() != null)
```
yifan-c: I do not feel like `expecting` unless you supplied the description to enrich the exception. Otherwise, the exception thrown reads like `VertxException("Unexpected result: " + path, true)`. The exception does not have the context regarding `path.getParent()`, which makes it really confusing for operators. 
I would like to suggest validating in a `compose` block and throwing a live migration specific exception.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -0,0 +1,688 @@
...
+               .onSuccess(path -> vertx.fileSystem().mkdirs(path.getParent().toAbsolutePath().toString()))
+               .compose(path -> vertx.fileSystem().createFile(path.toString()).compose(v -> Future.succeededFuture(path)))
```
yifan-c: There is no guarantee that `onSuccess` must be called before `compose`. However, the implementation requires the sequence. 
Can you just combine the `compose` blocks into a single one?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -0,0 +1,688 @@
...
+
+                directoriesDeleted.incrementAndGet();
+            }
+            else
+            {
+                LOGGER.debug("{} Deleting unwanted file: {}", logPrefix, localPath.toAbsolutePath());
+                vertx.fileSystem().deleteBlocking(localPath.toString());
+                filesDeleted.incrementAndGet();
+            }
+        }
```
yifan-c, line 336: When looping through a list of files to delete, can you catch any exception and continue with the next file? Assuming that you do not want to stop after a failure.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -0,0 +1,688 @@
...
+        // directories at the end (because size is explicitly set to -1 for directories).
+        instanceFileInfoList.sort(Collections.reverseOrder(Comparator.comparingLong(instanceFileInfo -> instanceFileInfo.size)));
+
+        List<Callable<Future<Void>>> downloadTasks = new ArrayList<>();
+        SidecarInstanceImpl instance = new SidecarInstanceImpl(source, port);
+
+
+        for (InstanceFileInfo file : instanceFileInfoList)
+        {
+            if (file.fileType.equals(FileType.DIRECTORY))
```
yifan-c:
```suggestion
            if (file.fileType == FileType.DIRECTORY)
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -0,0 +1,688 @@
...
+            {
+                futureList.add(createDirectory(file));
+            }
+            else if (file.fileType.equals(FileType.FILE) && file.size == 0)
```
yifan-c:
```suggestion
            else if (file.fileType == FileType.FILE && file.size == 0)
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -0,0 +1,688 @@
...
+
+        LOGGER.info("{} Download size met the successThreshold({}).{}",
+                    logPrefix,
+                    successThreshold,
+                    " Skipping download of " + filesToDownload.size() + " files of size " + downloadSize);
+        return Future.succeededFuture(Collections.emptyList());
+    }
+
+    @VisibleForTesting
+    Future<OperationStatus> sortBySizeAndDownload(List<InstanceFileInfo> instanceFileInfoList)
```
yifan-c, line 396: Can you comment the reason of downloading the large files first? 
And.. my wondering.. if dictionaries are sorted the last, will the directories already be created on the downloading host when downloading the files?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationTaskImpl.java
```diff
@@ -0,0 +1,204 @@
...
+    public void start()
+    {
+        promise = Promise.promise();
+        int iteration = 0;
+        handleRetry(startInternal(iteration), iteration, request.maxIterations, promise);
+    }
+
+    public Future<OperationStatus> startInternal(int iteration)
+    {
+        downloader = LiveMigrationFileDownloader.builder()
```
yifan-c, line 118: What if the task is cancelled already? Do you want to check the state and avoid creating the downloader? Otherwise, it is possible to create downloader and start downloading when the task is already cancelled.

vertx-client/src/main/java/org/apache/cassandra/sidecar/client/VertxRequestExecutor.java
```diff
@@ -71,4 +75,18 @@ public void close() throws Exception
...
+        if (context.request() instanceof DownloadableRequest && throwable instanceof FileSystemException)
+        {
+            // If the request is a downloadable request and the exception is a FileSystemException,
+            // then it is an issue with local file system. In this case, we should not retry the request.
+            future.completeExceptionally(throwable);
+            return;
+        }
+
+        super.applyRetryPolicy(future, iterator, sidecarInstance, context, attempt, response, throwable);
+    }
```
yifan-c, line 91: Is it specific to live migration? If so, would you like to create a live-migration-specific retry policy instead?

yifan-c (commented):
It is a large patch, so I probably have missed a lot of things. Submitting my comments so far.

vertx-client/src/main/java/org/apache/cassandra/sidecar/client/VertxRequestExecutor.java
nvharikrishna: it is not specific to live migration, if the disk went bad or has issue with permissions, then no need to attempt to download again from server.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationTaskImpl.java
nvharikrishna: Task is started even before responding to task creation request. But makes sense to add the check. Done.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: oh yeah, missed it, updated it.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: If failed to delete one file, then operation is going to fail anyway. So, failing it early.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: Already have a comment at line: 395. Converted it to doc comment.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: updated the code.

nvharikrishna (commented):
Addressed review comments.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateDataCopyTaskHandler.java
frankgh: yeah, that's reasonable. I was pointing out what's available.

server/src/test/java/org/apache/cassandra/sidecar/concurrent/AsyncConcurrentTaskExecutorTest.java
frankgh: no, not sure why I suggested it :(

frankgh (approved):
+1 looks good to me. I have no further comments

PR #255, opened 2025-08-24
CASSSIDECAR-339: Add developer-facing docs on testing Sidecar

frankgh:
Patch by Francisco Guerrero; reviewed by TBD for CASSSIDECAR-339

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -37,7 +37,7 @@ This document provides a comprehensive guide to running tests for the Apache Cas
 ### Required Software
 
 1. **Java 11 or higher** (OpenJDK or Oracle)
-2. **Docker** for running integration tests
+2. **Docker** for running integration tests that leverage the S3MockContainer to test the S3 client. Unless your test requires the S3 Mock Container, please create a unit or integration test. 
 3. **Git** for cloning repositories
 
 ### Build Prerequisites
--- a/TESTING.md
+++ b/TESTING.md
@@ -139,7 +139,7 @@ Container-based tests run in Docker environments for additional isolation. For e
 tests for the Restore from S3 feature leverage the testcontainer-based S3 service
 (`com.adobe.testing.s3mock.testcontainers.S3MockContainer`). Only tests that require
 testcontainer-like functionality should live under `src/containerTest/`; otherwise,
-a unit or in JVM dtest integration test should suffice.
+a unit or in JVM dtest integration test is preferred.
 
 **Location**: `src/containerTest/` in relevant modules
 
@@ -228,7 +228,7 @@ Tests running on Java 11+ automatically receive optimized JVM arguments defined 
 For faster development cycles:
 
 ```bash
-# Run with code checks disabled
+# Run the Sidecar server with code checks disabled. This is useful for local manual testing of endpoints.
 ./gradlew run
 
 # Run specific test class
@@ -318,10 +318,12 @@ INTEGRATION_MAX_PARALLEL_FORKS=1 ./gradlew integrationTest
 Integration tests use `forkEvery = 1` to ensure test isolation. If you encounter test pollution:
 
 1. Verify each test properly cleans up resources
-2. Check for static state leakage
-3. Review test execution order
+2. Check for static state leakage within the failing test class.
+3. Review test execution order.
 
-### Debugging Integration Tests
+If the test state pollution is unavoidable within the test class, it is reasonable to split the tests into multiple, independent test classes.
+
+### Enabling debug logging for Integration Tests
 
 Enable debug logging by modifying `server/src/test/resources/logback-in-jvm-dtest.xml`:
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -172,7 +172,7 @@ Integration tests support several environment variables:
 
 | Variable | Default | Description |
 |----------|---------|-------------|
-| `INTEGRATION_MAX_HEAP_SIZE` | `3000M` | Maximum heap size for integration tests |
+| `INTEGRATION_MAX_HEAP_SIZE` | `3000M` | Maximum heap size for integration tests, per parallel fork |
 | `INTEGRATION_MAX_PARALLEL_FORKS` | `4` | Number of parallel test forks |
 | `INTEGRATION_MTLS_ENABLED` | `true` | Enable mTLS for integration tests |
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,0 +1,358 @@
+<!--
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+# Testing Guide for Apache Cassandra Sidecar
+
+This document provides a comprehensive guide to running tests for the Apache Cassandra Sidecar project.
+
+## Table of Contents
+
+- [Prerequisites](#prerequisites)
+- [Test Types](#test-types)
+- [Running Tests](#running-tests)
+- [Test Configuration](#test-configuration)
+- [Development Testing](#development-testing)
+- [Continuous Integration](#continuous-integration)
+- [Troubleshooting](#troubleshooting)
+
+## Prerequisites
+
+### Required Software
+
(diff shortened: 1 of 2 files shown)


TESTING.md
```diff
@@ -0,0 +1,358 @@
...
+# Run integration tests for specific module
+./gradlew :integration-tests:integrationTest
+
+# Run integration tests for specific module
+./gradlew :server:integrationTest
+```
+
+### Container Tests
+
+Container-based tests run in Docker environments for additional isolation.
```
JeetKunDoug: We should probably describe the cases where we use container tests (just the SBW-on-S3 code where we need the S3 mock, right?) so folks don't unintentionally add tests where they aren't needed.

TESTING.md
```diff
@@ -0,0 +1,358 @@
...
+- [Development Testing](#development-testing)
+- [Continuous Integration](#continuous-integration)
+- [Troubleshooting](#troubleshooting)
+
+## Prerequisites
+
+### Required Software
+
+1. **Java 11 or higher** (OpenJDK or Oracle)
+2. **Docker** for running integration tests
```
JeetKunDoug:
```suggestion
2. **Docker** for running integration tests that leverage the S3MockContainer to test the S3 client. Unless your test requires the S3 Mock Container, please create a unit or integration test. 
```

TESTING.md
```diff
@@ -0,0 +1,358 @@
...
+
+# Run with specific Cassandra versions
+./gradlew test -Dcassandra.sidecar.versions_to_test=4.1,5.1
+```
+
+### Test Execution Configuration
+
+Integration tests support several environment variables:
+
+| Variable | Default | Description |
```
JeetKunDoug: NIT: Reformat table so it's readable in text and not just rendered.

TESTING.md
```diff
@@ -0,0 +1,358 @@
...
+|----------|---------|-------------|
+| `INTEGRATION_MAX_HEAP_SIZE` | `3000M` | Maximum heap size for integration tests |
```
JeetKunDoug:
```suggestion
| `INTEGRATION_MAX_HEAP_SIZE` | `3000M` | Maximum heap size for integration tests, per parallel fork |
```

TESTING.md
```diff
@@ -0,0 +1,362 @@
...
+- [Test Configuration](#test-configuration)
+- [Development Testing](#development-testing)
+- [Continuous Integration](#continuous-integration)
+- [Troubleshooting](#troubleshooting)
+
+## Prerequisites
+
+### Required Software
+
+1. **Java 11 or higher** (OpenJDK or Oracle)
```
JeetKunDoug: Do we need to call out OpenJDK or Oracle? Is this just to try to limit our "support" to those, and avoid other JDKs? Because I don't think _we_ test with Oracle JDK anywhere, right?

TESTING.md
```diff
@@ -0,0 +1,362 @@
...
+
+- **REPO**: Cassandra git repository (default: `https://github.com/apache/cassandra.git`)
+- **BRANCHES**: Space-delimited list of branches (default: `"cassandra-4.0 cassandra-4.1 cassandra-5.0 trunk"`)
+
+Example with custom branches:
+```bash
+BRANCHES="cassandra-4.1 trunk" ./scripts/build-dtest-jars.sh
+```
+
+### Network Setup for Multi-Node Tests
```
JeetKunDoug, line 62: Do we also need to add entries to `/etc/hosts` for these to work correctly? I know I _did_ ages ago, and since it's persistent I never touched it again, but I think they are still necessary.

TESTING.md
```diff
@@ -0,0 +1,362 @@
...
+./gradlew :server:integrationTest
+```
+
+### Container Tests
+
+Container-based tests run in Docker environments for additional isolation. For example,
+tests for the Restore from S3 feature leverage the testcontainer-based S3 service
+(`com.adobe.testing.s3mock.testcontainers.S3MockContainer`). Only tests that require
+testcontainer-like functionality should live under `src/containerTest/`; otherwise,
+a unit or in JVM dtest integration test should suffice.
```
JeetKunDoug:
```suggestion
a unit or in JVM dtest integration test is preferred.
```

TESTING.md
```diff
@@ -0,0 +1,362 @@
...
+# Run all tests (unit + integration)
+./gradlew check
+
+# Run only unit tests
+./gradlew test
+
+# Run only integration tests
+./gradlew integrationTest
+
+# Skip integration tests
```
JeetKunDoug, line 199: Can you clarify the difference between this and `./gradlew test` - does it also run the container tests? Also, can you skip those with something like `-x containerTest` ?

TESTING.md
```diff
@@ -0,0 +1,362 @@
...
+Tests running on Java 11+ automatically receive optimized JVM arguments defined in `gradle/common/java11Options.gradle`.
+
+## Development Testing
+
+### Running Tests During Development
+
+For faster development cycles:
+
+```bash
+# Run with code checks disabled
```
JeetKunDoug: Run _what_ with checks disabled?
```suggestion
# Run the Sidecar server with code checks disabled. This is useful for local manual testing of endpoints.
```

TESTING.md
```diff
@@ -0,0 +1,362 @@
...
+
+### Common Issues
+
+#### "DTest jar not found"
+**Solution**: Run `./scripts/build-dtest-jars.sh` to build required dependencies.
+
+#### "Network interface not available"
+**Solution**: Set up network aliases as described in [Network Setup](#network-setup-for-multi-node-tests).
+
+#### "OutOfMemoryError during integration tests"
```
JeetKunDoug, line 340: We really shouldn't need this any more given we fork every test - should the guidance really be to figure out if there's a leak somewhere?

TESTING.md
```diff
@@ -0,0 +1,362 @@
...
+INTEGRATION_MAX_PARALLEL_FORKS=1 ./gradlew integrationTest
+```
+
+### Test Isolation Issues
+
+Integration tests use `forkEvery = 1` to ensure test isolation. If you encounter test pollution:
+
+1. Verify each test properly cleans up resources
+2. Check for static state leakage
+3. Review test execution order
```
JeetKunDoug: with `forkEvery = 1` there should be no possibility of static state leakage _between_ test classes. Maybe we need a bit more of a description here so people better understand when this might happen?
```suggestion
Integration tests use `forkEvery = 1` to ensure test isolation. If you encounter test pollution:

1. Verify each test properly cleans up resources
2. Check for static state leakage within the failing test class.
3. Review test execution order.

If the test state pollution is unavoidable within the test class, it is reasonable to split the tests into multiple, independent test classes.
```

TESTING.md
```diff
@@ -0,0 +1,362 @@
...
+
+### Debugging Integration Tests
```
JeetKunDoug:
```suggestion
### Enabling debug logging for Integration Tests
```

TESTING.md
```diff
@@ -0,0 +1,362 @@
...
+
+Enable debug logging by modifying `server/src/test/resources/logback-in-jvm-dtest.xml`:
+
+```xml
+<logger name="org.apache.cassandra.sidecar" level="DEBUG"/>
+```
+
+### Performance Testing
```
JeetKunDoug: This isn't really "performance testing" in the traditional sense of the word... I'm not sure what it is, other than "Troubleshooting long-running tests,"

TESTING.md
```diff
@@ -0,0 +1,362 @@
...
+
+#### Docker Issues
+- Ensure Docker daemon is running
+- Verify Docker has sufficient resources allocated
+- Check for port conflicts
+
+#### Java Version Issues
+- Verify Java 11+ compatibility
+- Check JVM arguments in CI environments
+- Review module path configurations
```
JeetKunDoug: Not sure what this is supposed to mean - can we be more descriptive of what "Review module path configurations" means in the context of "environment-specific issues" - are you talking about Classpath issues?

TESTING.md
```diff
@@ -0,0 +1,362 @@
...
+### Environment-Specific Issues
+
+#### Docker Issues
+- Ensure Docker daemon is running
+- Verify Docker has sufficient resources allocated
+- Check for port conflicts
+
+#### Java Version Issues
+- Verify Java 11+ compatibility
+- Check JVM arguments in CI environments
```
JeetKunDoug, line 387: Should we include a link to the `java11options.gradle` file to make sure they know _what_ they should be looking at?

JeetKunDoug (changes requested):
Thanks for putting this together - a bunch of small questions/suggestions.

TESTING.md
frankgh: Good point, I will remove that reference

TESTING.md
frankgh: I guess it depends on how many instances of cassandra are spun up for the test. But I agree that defaults should be sufficient for all our tests

TESTING.md
frankgh: not sure actually, I will remove it

TESTING.md
frankgh: This is documented below under `Quick Test Commands`. But I will mention how to skip container tests

JeetKunDoug:
:shipit: (also commented on the Jira)

PR #258, opened 2025-09-10
CASSSIDECAR-342: Sidecar endpoint for draining a node

sklaha:
### Sidecar endpoint for draining a node

* API for draining a node
* Unit tests
* Integration tests

CircleCI: https://app.circleci.com/pipelines/github/sklaha/cassandra-sidecar/37/workflows/52a84253-9715-4794-9dea-3f00a41c16ac


server/src/main/java/org/apache/cassandra/sidecar/job/NodeDrainJob.java
```diff
@@ -0,0 +1,140 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean isRunningOnCassandra()
+    {
+        String operationMode = storageOperations.operationMode();
+        NodeDrainStateEnum nodeDrainStateEnum = NodeDrainStateEnum.fromOperationMode(operationMode);
+        return nodeDrainStateEnum != null && nodeDrainStateEnum.jobStatus == OperationalJobStatus.RUNNING;
```
frankgh: I find this logic very confusing to follow, would it make more sense to just check whether the node is draining here instead?
```suggestion
        return nodeDrainStateEnum == NodeDrainStateEnum.DRAINING;
```

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDrainJob.java
```diff
@@ -0,0 +1,140 @@
...
+         */
+        public static NodeDrainStateEnum fromOperationMode(String operationMode)
+        {
+            try
+            {
+                return valueOf(operationMode);
+            }
+            catch (IllegalArgumentException | NullPointerException e)
+            {
+                return null;
```
frankgh, line 72: should we log at debug when this happens at the very least?

server/src/main/java/org/apache/cassandra/sidecar/modules/CassandraOperationsModule.java
```diff
@@ -156,6 +157,22 @@ VertxRoute cassandraNodeDecommissionRoute(RouteBuilder.Factory factory,
         return factory.buildRouteWithHandler(nodeDecommissionHandler);
     }
 
+    @PUT
+    @Path(ApiEndpointsV1.NODE_DRAIN_ROUTE)
+    @Operation(summary = "Drain node",
+               description = "Drains the Cassandra node by flushing memtables and stopping writes")
+    @APIResponse(description = "Node drain operation initiated successfully",
+                 responseCode = "200",
```
frankgh, line 165: I think we have two response codes 202 and 200, correct?

server/src/test/java/org/apache/cassandra/sidecar/handlers/NodeDrainHandlerTest.java
```diff
@@ -0,0 +1,243 @@
...
+     */
+    class NodeDrainTestModule extends AbstractModule
+    {
+        @Provides
+        @Singleton
+        public InstancesMetadata instanceMetadata()
+        {
+            final int instanceId = 100;
+            final String host = "127.0.0.1";
+            final InstanceMetadata instanceMetadata = mock(InstanceMetadata.class);
```
frankgh: NIT: remove final keyword from local variables
```suggestion
            int instanceId = 100;
            String host = "127.0.0.1";
            InstanceMetadata instanceMetadata = mock(InstanceMetadata.class);
```

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraNodeOperationsIntegrationTest.java
```diff
@@ -0,0 +1,120 @@
...
+        // Verify node status is DRAINED by checking the operationMode via stream stats endpoint
+        HttpResponse<Buffer> streamStatsResponse = getBlocking(
+        trustedClient().get(serverWrapper.serverPort, "localhost", ApiEndpointsV1.STREAM_STATS_ROUTE)
+                       .send());
+
+        assertThat(streamStatsResponse.statusCode()).isEqualTo(OK.code());
+
+        JsonObject streamStats = streamStatsResponse.bodyAsJsonObject();
+        assertThat(streamStats).isNotNull();
+        assertThat(streamStats.getString("operationMode")).isEqualTo("DRAINED");
```
frankgh: don't use thread.sleep, use loopassert instead
```suggestion
         // Wait up to 30 seconds for drain operation to complete
                loopAssert(30, 500, () -> {
            // Verify node status is DRAINED by checking the operationMode via stream stats endpoint
            HttpResponse<Buffer> streamStatsResponse = getBlocking(
            trustedClient().get(serverWrapper.serverPort, "localhost", ApiEndpointsV1.STREAM_STATS_ROUTE)
                           .send());

            assertThat(streamStatsResponse.statusCode()).isEqualTo(OK.code());

            JsonObject streamStats = streamStatsResponse.bodyAsJsonObject();
            assertThat(streamStats).isNotNull();
            assertThat(streamStats.getString("operationMode")).isEqualTo("DRAINED");
        });
```

frankgh (commented):
Looks good overall, I left a few minor comments

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDrainJob.java
sklaha: Updated.

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDrainJob.java
sklaha: Yes, I added an error log for this.

server/src/main/java/org/apache/cassandra/sidecar/modules/CassandraOperationsModule.java
sklaha: Yes, I added both of the statuses.

server/src/test/java/org/apache/cassandra/sidecar/handlers/NodeDrainHandlerTest.java
sklaha: updated.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraNodeOperationsIntegrationTest.java
sklaha: Done

frankgh (approved):
+1 thanks for the patch!

server/src/main/java/org/apache/cassandra/sidecar/handlers/NodeDecommissionHandler.java
```diff
@@ -89,22 +84,7 @@ public void handleInternal(RoutingContext context,
...
-            logger.error("Conflicting job encountered. reason={}", reason);
-            context.response().setStatusCode(HttpResponseStatus.CONFLICT.code());
-            context.json(new OperationalJobResponse(job.jobId(), OperationalJobStatus.FAILED, job.name(), reason));
-            return;
-        }
-
-        // Get the result, waiting for the specified wait time for result
-        job.asyncResult(executorPools.service(), config.operationalJobExecutionMaxWaitTime())
-           .onComplete(v -> OperationalJobUtils.sendStatusBasedResponse(context, job));
+        handleOperationalJob(this.jobManager, this.config, context, job);
```
arjunashok: nice

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDrainJob.java
```diff
@@ -0,0 +1,142 @@
...
+public class NodeDrainJob extends OperationalJob
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(NodeDrainJob.class);
+    private static final String OPERATION = "drain";
+    protected StorageOperations storageOperations;
+
+    /**
+     * Enum representing the various drain states of a Cassandra node.
+     */
+    public enum NodeDrainStateEnum
```
arjunashok, line 44: Should this be a generic `OperationMode` enum (minus the Enum suffix in the name)? This could then be leveraged by other jobs that interpret or map them into job states (eg. decommission). Unless you see a need for a drain specific enum.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraNodeOperationsIntegrationTest.java
```diff
@@ -0,0 +1,113 @@
...
+        assertThat(responseBody.getString("jobId")).isNotNull();
+        assertThat(responseBody.getString("jobStatus")).isIn(
+        OperationalJobStatus.CREATED.name(),
+        OperationalJobStatus.RUNNING.name(),
+        OperationalJobStatus.SUCCEEDED.name()
+        );
+
+        loopAssert(30, 500, () -> {
+            // Verify node status is DRAINED by checking the operationMode via stream stats endpoint
+            HttpResponse<Buffer> streamStatsResponse = getBlocking(
```
arjunashok, line 78: We should also validate that the operational job `status` endpoint behaves as expected.

arjunashok (commented):
Looks good, left some minor comments

server/src/main/java/org/apache/cassandra/sidecar/handlers/AbstractHandler.java
```diff
@@ -317,4 +324,32 @@ public static String extractHostAddressWithoutPort(HttpServerRequest request) th
...
+
+    /**
+     * Handles the submission and execution of an operational job.
+     *
+     * @param jobManager the manager responsible for submitting and tracking operational jobs
+     * @param config     the service configuration containing execution parameters
+     * @param context    the routing context for the HTTP request/response
+     * @param job        the operational job to be executed
+     */
+    protected void handleOperationalJob(OperationalJobManager jobManager, ServiceConfiguration config, RoutingContext context, OperationalJob job)
```
sarankk: Nice abstracting this out. Instead of adding this to `AbstractHandler` it might be better to add a new `AbstractOperationHandler` inject the `OperationalJobManager` and `SidecarConfiguration` to `OperationHandler` instead of passing them through `handleOperationalJob` method.

sarankk (commented):
Thanks Sudipta, lgtm. Left one comment

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraNodeOperationsIntegrationTest.java
sklaha: Yes, I added that test. Thanks for the suggestion.

server/src/main/java/org/apache/cassandra/sidecar/handlers/AbstractHandler.java
sklaha: I updated the code based on our discussion. Please check if it's good now.

server/src/main/java/org/apache/cassandra/sidecar/job/NodeDrainJob.java
sklaha: If we combine the enum for all operations, each job will need to store the mapping of running vs not running statuses. Please check the implementation of the status() for the context: https://github.com/apache/cassandra-sidecar/pull/258/files#diff-7dc78ce3affd3c55b0f2fd978064df7d3bbeb4e22b67f14cc02e64d4885a1b52R98

The other option is to keep the status enums separate for for job, as I implemented. Please let me know your thoughts.

server/src/main/java/org/apache/cassandra/sidecar/handlers/NodeDecommissionHandler.java
```diff
@@ -84,7 +85,11 @@ public void handleInternal(RoutingContext context,
     {
         StorageOperations operations = metadataFetcher.delegate(host).storageOperations();
         NodeDecommissionJob job = new NodeDecommissionJob(UUIDs.timeBased(), operations, isForce);
-        handleOperationalJob(this.jobManager, this.config, context, job);
+        this.jobManager.trySubmitJob(job,
+                                     (completedJob, exception) ->
+                                     OperationalJobUtils.sendStatusBasedResponse(context, completedJob, exception),
+                                     executorPools.service(),
```
sarankk, line 91: Nit: We already inject `ExecutorPools` in `OperationalJobManager`. We can use service pool from there instead of passing here.

sarankk (approved):
Thanks Sudipta, looks good to me. One minor comment.

arjunashok (approved):
LGTM, +1

PR #259, opened 2025-09-19
CASSSIDECAR-345 live migration status endpoint

nvharikrishna:

[CASSSIDECAR-345](https://issues.apache.org/jira/browse/CASSSIDECAR-345) Added Live Migration Status endpoint to persist live migration status

Summary of changes:

* Added an endpoint for persisting the live migration status (completed or not) into a file in staging folder of an instance. 
* When live migration status is updated as completed, then that instance will not accept any new live migration tasks or serve files. These safety check prevents re-migrating instances that were already migrated but config is not yet cleared.
* Also added an end point to delete the status which works only after cleaning up live migration configuration.


client-common/src/main/java/org/apache/cassandra/sidecar/common/request/LiveMigrationDataCopyRequest.java
```diff
@@ -58,18 +52,6 @@ public LiveMigrationDataCopyRequest(@JsonProperty("maxIterations") int maxIterat
...
     {
-        this(UUID.randomUUID().toString(), maxIterations, successThreshold, maxConcurrency);
-    }
-
-    public LiveMigrationDataCopyRequest(String id,
-                                        int maxIterations,
-                                        double successThreshold,
-                                        int maxConcurrency)
-    {
-        if (id == null || id.isEmpty())
```
nvharikrishna, line 69: Having request ID at in reqeust is bit confusing when this class is used at client. Handling requestId separately and returning it as part of response.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
```diff
@@ -96,21 +93,8 @@ public Future<LiveMigrationTask> createTask(@NotNull LiveMigrationDataCopyReques
...
         InstanceMetadata localInstanceMetadata = instancesMetadata.instanceFromHost(currentHost);
 
-        return vertx.executeBlocking(() -> liveMigrationMap.getSource(currentHost))
-                    .compose(source -> validateDataDirs(source, localInstanceMetadata))
-                    .compose(source -> createDataCopyTask(request, source, localInstanceMetadata));
-    }
-
-    Future<String> validateDataDirs(String source, InstanceMetadata localInstanceMetadata)
-    {
-        InstanceMetadata sourceMeta = instancesMetadata.instanceFromHost(source);
```
nvharikrishna, line 106: `instancesMetadata.instanceFromHost(source);` doesn't work if the source is not local. Hence removing it. Data copy will fail while downloading files.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/LiveMigrationDataCopyRequest.java
yifan-c: I actually think it makes sense to have ID supplied by the client, so that the request can be idempotent when client needs to retry.
It is common for client to provide the ID for server.
If client retry is not a consideration and the idea is ensure server-generated IDs. Please resolve the comment.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/LiveMigrationTaskResponse.java
```diff
@@ -66,13 +88,19 @@ public int port()
...
     public int maxIterations()
     {
-        return request.maxIterations;
+        return maxIterations;
+    }
+
+    @JsonProperty("maxConcurrency")
+    public int maxConcurrency()
+    {
+        return maxConcurrency;
```
yifan-c, line 97: just curious: this new field does not seem to be used, except in the test.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/LiveMigrationStatus.java
```diff
@@ -0,0 +1,99 @@
...
+        return state;
+    }
+
+
+    @Nullable
+    @JsonProperty("endTime")
+    public Long endTime()
+    {
+        return endTime;
+    }
```
yifan-c, line 68: When `state == COMPLETED`, should the `endTime` be non-null? If so, probably add a validation in the constructor to ensure `state != COMPLETED || endTime != null`

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/LiveMigrationStatus.java
```diff
@@ -0,0 +1,99 @@
...
+        if (object == null || getClass() != object.getClass()) return false;
+        LiveMigrationStatus status = (LiveMigrationStatus) object;
+        return state == status.state && Objects.equals(endTime, status.endTime);
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(state, endTime);
+    }
```
yifan-c, line 102: Are they implemented just in case? I do not think the methods are exercised in the code base.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationStatusTrackerImpl.java
```diff
@@ -0,0 +1,248 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(LiveMigrationStatusTrackerImpl.class);
+    private final Vertx vertx;
+    private final ObjectMapper objectMapper;
+
+    // Per-instance locks to ensure atomicity of file operations
+    private final ConcurrentHashMap<String, Object> instanceLocks = new ConcurrentHashMap<>();
+
+    @Inject
+    public LiveMigrationStatusTrackerImpl(Vertx vertx)
```
yifan-c: Inject `ExecutorPools` instead?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationStatusTrackerImpl.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Future<Void> setMigrationStatus(InstanceMetadata instanceMetadata,
+                                           LiveMigrationStatus status)
+    {
+        Promise<Void> promise = Promise.promise();
```
yifan-c: nit: the promise is unnecessary, `executeBlocking` returns a future already.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationStatusTrackerImpl.java
```diff
@@ -0,0 +1,248 @@
...
+            synchronized (getLockForInstance(instanceMetadata.host()))
+            {
+                try
+                {
+                    Path stagingDir = ensureStagingDir(instanceMetadata);
+                    Path statusFile = stagingDir.resolve(STATUS_FILE_NAME);
+
+                    if (!Files.exists(statusFile))
+                    {
+                        promise.fail(new IllegalArgumentException(
```
yifan-c: `IllegalStateException`?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationStatusTrackerImpl.java
```diff
@@ -0,0 +1,248 @@
...
+                        "Live Migration status file does not exist for instance " + instanceMetadata.host()));
```
yifan-c: Maybe the file is already deleted? Should it be handled gracefully?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationStatusTrackerImpl.java
```diff
@@ -0,0 +1,248 @@
...
+
+                    // Delete the file
+                    Files.delete(statusFile);
+                    LOGGER.info("Live migration status file deleted for instance {}", instanceMetadata.host());
+                    promise.complete();
+                    return null;
+                }
+                catch (IOException e)
+                {
+                    LOGGER.error("Failed to clear LiveMigrationStatus for instance {}", instanceMetadata.host(), e);
```
yifan-c: what should operator do in the failure mode? Maybe suggest manual file deletion?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationStatusTrackerImpl.java
nvharikrishna: sounds good, done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationStatusTrackerImpl.java
nvharikrishna: The deletion could be attempted before completion, or it could be re-attempted, or a client could simply retry the operation. The first two scenarios may require operator attention because the instance is in an unexpected state. Therefore, throwing an exception is necessary.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationStatusTrackerImpl.java
nvharikrishna: Yes, manual file deletion might be required. Updated the message.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationStatusTrackerImpl.java
nvharikrishna: right, updated

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationStatusTrackerImpl.java
nvharikrishna: yeah, missed it, done

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/LiveMigrationStatus.java
nvharikrishna: done, using `state == MigrationState.COMPLETED && endTime == null` to be clean.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/LiveMigrationStatus.java
nvharikrishna: `state` is an enum and `endTime` is a `Long`, should be fine right? I’m not sure if I understood your comment.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/LiveMigrationTaskResponse.java
nvharikrishna: It serves as an intimation to the client.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/LiveMigrationDataCopyRequest.java
nvharikrishna: > If client retry is not a consideration and the idea is ensure server-generated IDs.
yes, this is the expectation

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationMap.java
```diff
@@ -34,43 +34,49 @@ public interface LiveMigrationMap
      * Tells whether given instance is configured as source or not.
      *
      * @param instanceMeta Cassandra instance metadata
-     * @return true if given instance is configured as source for live migration.
+     * @return Future that completes with true if given instance is configured as source for live migration.
      */
-    default boolean isSource(@NotNull InstanceMetadata instanceMeta)
+    default Future<Boolean> isSource(@NotNull InstanceMetadata instanceMeta)
```
frankgh, line 39: 👍

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationStatusGetHandler.java
```diff
@@ -0,0 +1,87 @@
...
+        return null;
+    }
+
+    @Override
+    protected void handleInternal(RoutingContext routingContext, HttpServerRequest httpServerRequest,
+                                  @NotNull String host, SocketAddress socketAddress, Void unused)
+    {
+        InstanceMetadata instance = metadataFetcher.instance(host);
+        statusTracker.getMigrationStatus(instance)
+                     .compose(status -> routingContext.response().send(Json.encode(status)))
```
frankgh: NIT:
```suggestion
                     .compose(routingContext::json)
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
```diff
@@ -157,7 +141,7 @@ LiveMigrationTask createTask(LiveMigrationDataCopyRequest request,
                                  int port,
                                  InstanceMetadata localInstanceMetadata)
     {
-        return liveMigrationTaskFactory.create(request, source, port, localInstanceMetadata);
+        return liveMigrationTaskFactory.create(UUID.randomUUID().toString(), request, source, port, localInstanceMetadata);
```
frankgh: did you consider using a time-based UUID here instead?

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationStatusClearHandler.java
```diff
@@ -0,0 +1,111 @@
...
+                         {
+                             context.fail(wrapHttpException(HttpResponseStatus.SERVICE_UNAVAILABLE, e.getMessage(), e));
+                         }
+                     });
+    }
+
+    @Override
+    public Set<Authorization> requiredAuthorizations()
+    {
+        return Set.of(BasicPermissions.STREAM_FILES.toAuthorization());
```
frankgh: is this the authorization we want for this endpoint?

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationStatusClearHandler.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    protected void handleInternal(RoutingContext context, HttpServerRequest httpRequest, @NotNull String host, SocketAddress remoteAddress, Void request)
+    {
+        InstanceMetadata instanceMetadata = metadataFetcher.instance(host);
+        statusTracker.clearMigrationStatus(instanceMetadata)
+                     .onSuccess(v -> {
+                         LOGGER.info("Successfully cleared live migration status for instance {}", host);
+                         context.response().setStatusCode(HttpResponseStatus.OK.code()).end();
```
frankgh: generally we want to provide some feedback to the client, especially if the end user is a person. I recommend the following, which is what we do consistently across the codebase
```suggestion
                         context.json(OK_STATUS);
```

frankgh (commented):
Looks good in general, I have some minor comments

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationStatusClearHandler.java
nvharikrishna: done, thanks for sharing the tip 👍

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationStatusClearHandler.java
nvharikrishna: Good question! I am having second thoughts too. What if this permission is renamed as DATA_COPY? It is generic enough to cover the whole data-copy process of LiveMigration. What do you say?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
nvharikrishna: Thought that a random UUID is good enough instead of using a class from a non-JDK library to generate one. But I see that Sidecar is already using `com.datastax.driver.core.utils.UUIDs`.  Made the changes.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationStatusGetHandler.java
nvharikrishna: done

frankgh (approved):
+1

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/LiveMigrationStatus.java
yifan-c: Sorry. Missed it. My original comment was basically asking whether the `LiveMigrationStatus` is used in set or as map key. It seems not, which means the two methods will not be exercised.

yifan-c (approved):

PR #260, opened 2025-09-23
CASSSIDECAR-346 Sidecar side of CEP-55

smiklosovic:
This is not eligible for review yet as server side is not in yet either.


server/src/main/java/org/apache/cassandra/sidecar/handlers/role/RoleGenerationHandler.java
```diff
@@ -0,0 +1,164 @@
...
+                                  @NotNull String host,
+                                  SocketAddress remoteAddress,
+                                  GenerateRoleRequest request)
+    {
+        executorPools.service()
+                     .executeBlocking(() ->
+                                      {
+                                          CassandraAdapterDelegate delegate = metadataFetcher.delegate(host);
+
+                                          if (delegate.version().major < 5 && !request.requestBody().sidecarGeneration)
```
smiklosovic, line 82: for now just this to be able to test it somehow, will be bumped when Cassandra 6 is out.

PR #262, opened 2025-09-23
CASSSIDECAR-348: Add observability around authN and authZ caches

sarankk:


server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -61,14 +63,16 @@ protected AuthCache(String name,
...
+                        CacheConfiguration cacheConfiguration,
+                        CacheStatsCounter cacheMetrics)
     {
         this.name = name;
         this.vertx = vertx;
         this.internalPool = executorPools.internal();
         this.loadFunction = loadFunction;
         this.bulkLoadFunction = bulkLoadFunction;
         this.config = cacheConfiguration;
+        this.cacheMetrics = cacheMetrics;
```
frankgh: we should validate this is not null
```suggestion
        this.cacheMetrics = Objects.requireNonNull(cacheMetrics, "cacheMetrics is required");
```

frankgh (commented):
Looks good, I added a minor comment

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -119,13 +123,25 @@ public Map<K, V> getAll()
...
     private LoadingCache<K, V> initCache()
     {
         return Caffeine.newBuilder()
-                       // setting refreshAfterWrite and expireAfterWrite to same value makes sure no stale
-                       // data is fetched after expire time
-                       .refreshAfterWrite(config.expireAfterAccess().quantity(), config.expireAfterAccess().unit())
-                       .expireAfterWrite(config.expireAfterAccess().quantity(), config.expireAfterAccess().unit())
+                       // The cache keeps the entry until it's last access has expired. This avoids repeated calls to
+                       // db for the same key during high volume requests.
+                       .expireAfterAccess(config.expireAfterAccess().quantity(), config.expireAfterAccess().unit())
```
yifan-c: `refreshAfterWrite` ensures the load to the database is capped, i.e _at most_ once every D duration per entry, which "avoids repeated calls during high volume requests" already. When the entry is stable, an **async** call is made to load the data and the stale data is returned while fetch. Current thread is not blocked.
Meanwhile, `expireAfterAccess` could keep the potentially stale entry in the cache for longer duration, and even fewer DB access. However, when data is expired, it is removed from cache immediately. And the subsequent calls to `get()` will **block** the threads until the new value is loaded. It can potentially cause thundering herd problem, if multiple threads are getting the same expired entry.

I would suggest keeping `refreshAfterWrite` only. It gives a good balance between capping the DB access (load is predictable) and freshness of data.

server/src/test/java/org/apache/cassandra/sidecar/acl/RoleAuthorizationsCacheTest.java
```diff
@@ -106,7 +118,11 @@ void testCacheSizeAlwaysOne() throws InterruptedException
 
         // New entries fetched during refreshes
         assertThat(cache.getAuthorizations("test_role2").size()).isOne();
+        assertThat(sidecarMetrics.server().cache().rolePermissionsCacheMetrics.snapshot().hitCount()).isZero();
         assertThat(cache.getAll().size()).isOne();
+        assertThat(cache.getAuthorizations("test_role2").size()).isOne();
+        assertThat(sidecarMetrics.server().cache().rolePermissionsCacheMetrics.snapshot().hitCount()).isOne();
+
```
yifan-c: Can you also assert on the load success count and potentially introduce a scenario that test the cache miss count?

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -119,13 +123,25 @@ public Map<K, V> getAll()
...
 
+    /**
+     * Invalidate a key.
+     * @param k key to invalidate
+     */
+    public void invalidate(K k)
+    {
+        if (cache != null)
+        {
+            cache.invalidate(k);
```
yifan-c, line 136: Please add a log message when invalidating a key. It should be useful to confirm the invalidation for debugging purpose. Assuming the frequency should be low, the log level can be `INFO`. 
The method is not currently used.

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
sarankk: It could be beneficial to allow configuring both with different config values, for removing stale entries. earlier we were using same config value for both and behavior of that would be equivalent to having only refreshAfterWrite. Instead shall I add one more config value to allow configuring expireAfterAccess and refreshAfterWrite separately?  @yifan-c

yifan-c: By configuring both and with the same value, the behavior is the same as `expireAfterAccess`. Not equivalent to `refreshAfterWrite`. Given that the configuration already declares `expire_after_access`, to not break the compatibility, I would suggest to add a new configuration field `refresh_after_write` and encourage to only configure `refresh_after_write` in the document.

frankgh (approved):
+1

PR #263, opened 2025-09-29
CASSSIDECAR-349: Thread race issue when adding excluded metric in Fil…

frankgh:
…teringMetricRegistry

Patch by Francisco Guerrero; reviewed by TBD for CASSSIDECAR-349

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/metrics/FilteringMetricRegistry.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/metrics/FilteringMetricRegistry.java
@@ -22,6 +22,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/cassandra/sidecar/metrics/FilteringMetricRegistry.java
```diff
@@ -283,15 +284,22 @@ private <T extends Metric> T typeChecked(Metric metric, Class<T> type)
...
+        });
+
+        if (registeredExcludedMetric.get())
+        {
             // allMetrics needs to be recomputed when an excluded metric is registered
             synchronized (this)
             {
                 allMetrics = null;
             }
+        }
```
yifan-c, line 300: Now the update sequence looks correct. W/o the patch, it is possible that `allMetrics` is updated _before_ `excludedMetric`

yifan-c (approved):

PR #242, opened 2025-07-31
CASSSIDECAR-272: Add /api/v2/cassandra/settings which will return Cassandra configurations stored in system_views.settings

isaacreath:
This patch adds a new endpoint (`/api/v2/cassandra/settings`) which differs from `/api/v1/cassandra/settings` in the following ways:
1) Adds authorization onto endpoint
2) Gives the caller the ability to read the configurations from `system_views.settings`.
3) Reformats the object returned from the API to provide additional structure, like so:

```
{
    "cassandra": { } // Key-value pairs containing the data from system_view.settings.
    "jmx": {} // The settings available via JMX by querying storage service / gossip mbeans. This data was at the top level in the original api/v1/settings object.
    "sidecar": {} // This contains sidecar configuration information. 
}
```

PR #264, opened 2025-09-29
CASSSIDECAR-272: Add /api/v2/cassandra/settings which will return Cassandra configurations stored in system_views.settings

isaacreath:
Fresh PR originally from https://github.com/apache/cassandra-sidecar/pull/242.

This PR adds a new API, `/api/v2/cassandra/settings` which resolves the settings of a node from the `system_views.settings` table from the Cassandra instance being requested. Notably, this differs from the `/api/v1/cassandra/settings` table in the following ways:

1) `/api/v2/cassandra/settings` is an authorized endpoint, whereas `/api/v1/cassandra/settings` is not. All callers of `/api/v2/cassandra/settings` will either need to be an admin identity, or the identity provided must have the ability to read the `system_views.settings` table.

2) This API no longer contains the sidecar version. This will be moved into a separate endpoint for sidecar V2 API.

3) This API no longer contains settings information only available over JMX (e.g. tokens, release version). These will also need to be moved into a separate API.

The response body of this new endpoint is as follows:

```
{
  "nodeSettings": {
       "columns_per_table_warn_threshold": "-1",
       "cdc_total_space_in_mb": "0",
       ... // Repeat for all options available. 
  } 
}
```

### DIFF
--- a/adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
+++ b/adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
@@ -107,9 +107,9 @@ public class CassandraAdapter implements ICassandraAdapter
 
     @Override
     @NotNull
-    public Map<String, String> cqlNodeSettings()
+    public Map<String, String> v2NodeSettings()
     {
-        throw new UnsupportedOperationException("CQL node settings are not provided by this adapter");
+        throw new UnsupportedOperationException("V2 node settings are not provided by this adapter");
     }
 
     @Override
--- a/adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
+++ b/adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
@@ -19,6 +19,7 @@
 package org.apache.cassandra.sidecar.adapters.base;
 
 import java.net.InetSocketAddress;
+import java.util.Map;
 import java.util.Objects;
 
 import com.datastax.driver.core.ConsistencyLevel;
@@ -102,6 +103,13 @@ public class CassandraAdapter implements ICassandraAdapter
     public NodeSettings nodeSettings()
     {
         throw new UnsupportedOperationException("Node settings are not provided by this adapter");
+    }
+
+    @Override
+    @NotNull
+    public Map<String, String> cqlNodeSettings()
+    {
+        throw new UnsupportedOperationException("CQL node settings are not provided by this adapter");
     }
 
     @Override
--- a/client-common/src/main/java/org/apache/cassandra/sidecar/common/response/v2/V2NodeSettings.java
+++ b/client-common/src/main/java/org/apache/cassandra/sidecar/common/response/v2/V2NodeSettings.java
@@ -1,0 +1,113 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.cassandra.sidecar.common.response.v2;
+
+import java.util.Map;
+import java.util.Objects;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.cassandra.sidecar.common.DataObjectBuilder;
+import org.apache.cassandra.sidecar.common.response.NodeSettings;
+
+/**
+ * V2NodeSettings stores settings information for Cassandra.
+ */
+public class V2NodeSettings
+{
+    // Contains Cassandra node settings from system_views.setting table.
+    @JsonProperty("nodeSettings")
+    private final Map<String, String> nodeSettings;
+
+    /**
+     * Constructs a new {@link NodeSettings}.
+     */
+    public V2NodeSettings()
+    {
+        this(builder());
+    }
+
+    public V2NodeSettings(V2NodeSettings.Builder builder)
+    {
+        this.nodeSettings = builder.nodeSettings;
+
+    }
+
+    @JsonProperty("nodeSettings")
+    public Map<String, String> nodeSettings()
+    {
+        return nodeSettings;
+    }
+
+    public static Builder builder()
+    {
+        return new Builder();
+    }
+
+    @Override
+    public boolean equals(Object o)
+    {
+        if (o == null || getClass() != o.getClass()) return false;
+        V2NodeSettings that = (V2NodeSettings) o;
+        return Objects.equals(nodeSettings, that.nodeSettings);
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(nodeSettings);
+    }
+
+    /**
+     * A builder class to enable construction of V2NodeSettings objects.
+     */
+    public static final class Builder implements DataObjectBuilder<Builder, V2NodeSettings>
+    {
+        private Map<String, String> nodeSettings;
+
+        private Builder()
+        {
+        }
+
+        @Override
+        public V2NodeSettings.Builder self()
+        {
+            return this;
+        }
+
+        /**
+         * Sets a value for the Cassandra settings map and returns a reference to this Builder enabling method chaining.
+         * @param nodeSettings Map of Cassandra setting name to setting value.
+         * @return a reference to this Builder
+         */
+        public V2NodeSettings.Builder nodeSettings(Map<String, String> nodeSettings)
+        {
+            return update(b -> b.nodeSettings = nodeSettings);
+        }
+
+        /**
+         * Returns a {@code NodeSettings} built from the parameters previously set.
+         *
+         * @return a {@code NodeSettings} built with parameters of this {@code NodeSettings.Builder}
+         */
+        @Override
+        public V2NodeSettings build()
+        {
+            return new V2NodeSettings(this);
+        }
+    }
+}
--- a/client-common/src/main/java/org/apache/cassandra/sidecar/common/response/v2/V2NodeSettings.java
+++ b/client-common/src/main/java/org/apache/cassandra/sidecar/common/response/v2/V2NodeSettings.java
@@ -20,6 +20,7 @@ package org.apache.cassandra.sidecar.common.response.v2;
 import java.util.Map;
 import java.util.Objects;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.cassandra.sidecar.common.response.NodeSettings;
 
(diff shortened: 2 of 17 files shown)


isaacreath:
https://app.circleci.com/pipelines/circleci/2pKxPuEmc4jySZHiGcJrJE/LTM5bKmo57v3FcPSxgh5mk?branch=CASSSIDECAR-272-config-api CI link

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
```diff
@@ -92,6 +93,13 @@ public NodeSettings nodeSettings()
         throw new UnsupportedOperationException("Node settings are not provided by this adapter");
     }
 
+    @Override
+    @NotNull
+    public Map<String, String> cqlNodeSettings()
```
frankgh: This is a smell, we are leaking implementation details to the interface. Maybe for the legacy node settings method we can rename it internally as `legacyNodeSettings` or `v1NodeSettings`? Alternatively, we can name this method `v2NodeSettings`? What do you think?

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/v2/V2NodeSettings.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(nodeSettings);
+    }
+
+    /**
+     * A builder class to enable construction of V2NodeSettings objects.
+     */
+    public static final class Builder implements DataObjectBuilder<Builder, V2NodeSettings>
```
frankgh: even though I love the builder pattern, it seems like for this use case it is not really necessary. I would just keep a constructor

```
    @JsonCreator
    public V2NodeSettings(@JsonProperty("nodeSettings") Map<String, String> nodeSettings)
    {
        this.nodeSettings = nodeSettings;
    }
```

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/v2/V2NodeSettingsIntegrationTest.java
```diff
@@ -0,0 +1,110 @@
...
+        cluster.getFirstRunningInstance().nodetool("disablebinary");
+        ensureSettingsBecomeUnavailable("CQL NodeSettings unavailable");
+
+        // Re-enable NTR, settings should become available again.
+        cluster.getFirstRunningInstance().nodetool("enablebinary");
+        ensureSettingsAvailable();
+
+        // Changing a configuration should eventually reflect in settings API.
+        cluster.getFirstRunningInstance().nodetool("setconcurrency", "READ", "10");
+        ensureSettingsAvailable();
```
frankgh: I think during this verification we need to make sure that the value for concurrency is changed. The current validation is not really making sure this value is 10, so maybe we can add validation for that?

server/src/main/java/org/apache/cassandra/sidecar/CassandraSidecarDaemon.java
```diff
@@ -73,12 +73,12 @@ static boolean close(Server app)
             app.close()
                .toCompletionStage()
                .toCompletableFuture()
-               .get(1, TimeUnit.MINUTES);
+               .get(2, TimeUnit.MINUTES);
```
frankgh, line 76: any reason we want to change this value?

server/src/main/java/org/apache/cassandra/sidecar/handlers/v2/cassandra/V2NodeSettingsHandler.java
```diff
@@ -0,0 +1,76 @@
...
+                                                    .nodeSettings(cqlSettings)
+                                                    .build();
+        context.json(v2nodeSettings);
+    }
+
+    @Override
+    public Set<Authorization> requiredAuthorizations()
+    {
+        Authorization authorization = BasicPermissions.READ_SETTINGS.toAuthorization("data/system_views/settings");
+        return Set.of(authorization);
```
frankgh: I think this is not correct, I checked with @sarankk and she suggested we do the following instead:
```suggestion
                Set<String> eligibleResources = Set.of(DATA_SCOPE.variableAwareResource(),
                                               // Keyspace access to system_views
                                               "data/system_views",
                                               // Access to all tables in keyspace system_views
                                               "data/system_views/*",
                                               // Access to the settings table in the system_views keyspace
                                               "data/system_views/settings");
        return Set.of(CassandraPermissions.SELECT.toAuthorization(eligibleResources));
```

server/src/main/java/org/apache/cassandra/sidecar/modules/multibindings/VertxRouteMapKeys.java
```diff
@@ -88,6 +89,11 @@ interface CassandraNodeSettingsRouteKey extends RouteClassKey
         HttpMethod HTTP_METHOD = HttpMethod.GET;
         String ROUTE_URI = ApiEndpointsV1.NODE_SETTINGS_ROUTE;
     }
+    interface V2CassandraNodeSettingsRouteKey extends RouteClassKey
```
frankgh: let's preserve the alphabetical order here as mentioned in https://github.com/apache/cassandra-sidecar/blob/trunk/server/src/main/java/org/apache/cassandra/sidecar/modules/multibindings/VertxRouteMapKeys.java#L35

server/src/main/java/org/apache/cassandra/sidecar/db/SystemViewsDatabaseAccessor.java
```diff
@@ -94,12 +95,24 @@ public Long cdcTotalSpaceBytesSetting() throws SchemaUnavailableException
     public Map<String, String> getSettings(String... names) throws SchemaUnavailableException
     {
         BoundStatement statement = tableSchema.selectSettings().bind(Arrays.asList(names));
+        return querySettings(statement);
+    }
+
+    @NotNull
+    public Map<String, String> getSettings() throws SchemaUnavailableException
```
frankgh: NIT, maybe we can rename it to something more descriptive like `allSettings`
```suggestion
    public Map<String, String> allSettings() throws SchemaUnavailableException
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
```diff
@@ -88,4 +88,8 @@ public class BasicPermissions
     // Lifecycle permissions
     public static final Permission READ_LIFECYCLE = new DomainAwarePermission("LIFECYCLE:READ", CLUSTER_SCOPE);
     public static final Permission MODIFY_LIFECYCLE = new DomainAwarePermission("LIFECYCLE:MODIFY", CLUSTER_SCOPE);
+
+    // Cassandra Settings Permissions
+    // Caller must have been granted SELECT on `system_view.settings` to hit the /api/v2/cassandra/settings API.
+    public static final Permission READ_SETTINGS = new StandardPermission("SELECT", TABLE_SCOPE);
```
frankgh: I think we will not need this change

frankgh (commented):
Looks good in general. I've left a few comments

server/src/main/java/org/apache/cassandra/sidecar/CassandraSidecarDaemon.java
isaacreath: I noticed in CI that sometimes it would take sidecar more than 1 minute to exit which would cause test failures (even outside of this branch). I suppose we could look into this further so that we do not need to increase this value.

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraAdapter.java
isaacreath: I think v2NodeSettings is a good approach.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/BasicPermissions.java
isaacreath: Addressed in https://github.com/apache/cassandra-sidecar/pull/264/commits/bbb94f7972d10813a3f7ebb1bde3a10dd1fed8ad

server/src/main/java/org/apache/cassandra/sidecar/modules/multibindings/VertxRouteMapKeys.java
isaacreath: Addressed as part of https://github.com/apache/cassandra-sidecar/pull/264/commits/3b5a9b8148abb6749b2272c67cb655fea764e08a

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/v2/V2NodeSettingsIntegrationTest.java
isaacreath: Addressed in https://github.com/apache/cassandra-sidecar/pull/264/commits/3b5a9b8148abb6749b2272c67cb655fea764e08a

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/v2/V2NodeSettings.java
isaacreath: Cleaned up in https://github.com/apache/cassandra-sidecar/pull/264/commits/0874f71eb2bc8502a65098c9dee6c58103f6a980

isaacreath:
Latest CI link: https://app.circleci.com/pipelines/circleci/2pKxPuEmc4jySZHiGcJrJE/LTM5bKmo57v3FcPSxgh5mk/32/workflows/d86357fb-299c-4a75-85c0-99e3a40a7689

server/src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -380,6 +411,23 @@ public NodeSettings nodeSettings() throws CassandraUnavailableException
...
+     */
+    @Override
+    @NotNull
+    public Map<String, String> v2NodeSettings() throws CassandraUnavailableException
+    {
+        if (nodeSettingsFromCql == null)
+        {
+            throw new CassandraUnavailableException(CQL, "CQL NodeSettings unavailable");
+        }
+        return nodeSettingsFromCql;
```
frankgh: There's a chance for a race here, where one thread is retrieving the settings and another is updating settings to null in line 322 for example. I would recommend changing this code to avoid any potential race issues
```suggestion
        Map<String, String> settings = nodeSettingsFromCql;
        if (settings == null)
        {
            throw new CassandraUnavailableException(CQL, "CQL NodeSettings unavailable");
        }
        return settings;
```

server/src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -242,7 +251,7 @@ protected synchronized void jmxHealthCheck()
     /**
      * Performs health checks by utilizing the native protocol
      */
-    protected void nativeProtocolHealthCheck()
+    protected synchronized void nativeProtocolHealthCheck()
```
frankgh: why is this synchronized? We guarantee a single thread here. Any reason to change it ?

server/src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -281,6 +291,26 @@ protected void nativeProtocolHealthCheck()
             {
                 notifyNativeConnection();
             }
+
+            if (isNativeUp.get())
```
frankgh: this will always be true, I believe. Any case where this might not be true?

server/src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
isaacreath: The goal was to ensure that we're explicitly guarding the modification to `nodeSettingsFromCql` in case we have other callers of this function in the future, but I think it's fine that we are guarding this through the `isHealthCheckActive` compare and set in `healthCheck()`

frankgh (approved):
+1

sarankk (approved):
+1 Thanks Isaac. Looks good to me

isaacreath:
Thank you both for the reviews!

PR #265, opened 2025-09-30
CASSSIDECAR-344: Sidecar endpoint for moving a node to a new token

sklaha:
### Sidecar endpoint for moving a node to a new token

* API for moving a node to a new token
* Unit tests
* Integration tests

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 0.3.0
 -----
+ * Upgrade caffeine dependency (CASSSIDECAR-332)
  * Add Live Migration Status endpoint to persist Live Migration status and for safety (CASSSIDECAR-345)
  * Sidecar endpoint for draining a node (CASSSIDECAR-342)
  * Avoid resuming stream early during SSTable upload (CASSSIDECAR-359)
--- a/gradle.properties
+++ b/gradle.properties
@@ -26,6 +26,7 @@ guiceVersion=7.0.0
 slf4jVersion=2.0.17
 logbackVersion=1.5.15
 jacksonVersion=2.14.3
+caffeineVersion=3.2.2
 dtestApiVersion=0.0.17
 assertjCoreVersion=3.24.2
 # trunk is currently 5.1 - update when trunk moves
--- a/server/src/main/java/org/apache/cassandra/sidecar/metrics/CacheStatsCounter.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/metrics/CacheStatsCounter.java
@@ -25,6 +25,7 @@ import java.util.concurrent.atomic.LongAdder;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.github.benmanes.caffeine.cache.stats.CacheStats;
 import com.github.benmanes.caffeine.cache.stats.StatsCounter;
 import org.checkerframework.checker.index.qual.NonNegative;
@@ -102,16 +103,7 @@ public class CacheStatsCounter implements StatsCounter
     }
 
     @Override
-    public void recordEviction()
-    {
-        recordEviction(1);
-    }
-
-    /**
-     * @deprecated
-     */
-    @Deprecated
-    public void recordEviction(int weight)
+    public void recordEviction(int weight, RemovalCause cause)
     {
         evictions.metric.update(weight);
     }


client-common/src/main/java/org/apache/cassandra/sidecar/common/request/NodeMoveRequest.java
```diff
@@ -0,0 +1,48 @@
...
+public class NodeMoveRequest extends JsonRequest<OperationalJobResponse>
+{
+    /**
+     * Constructs a request to execute a node move operation
+     *
+     * @param newToken the new token for the node to move to
+     */
+    public NodeMoveRequest(String newToken)
+    {
+        super(ApiEndpointsV1.NODE_MOVE_ROUTE + "?newToken=" + newToken);
```
frankgh: should we make the `newToken` be part of the body payload instead of having it as a query string?

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/StorageOperations.java
```diff
@@ -174,4 +174,11 @@ default void outOfRangeDataCleanup(@NotNull String keyspace, @NotNull String tab
      * @return the current compaction throughput in megabytes per second, or 0 if throughput cannot be determined
      */
     int getCompactionThroughputMbPerSec();
+
+    /**
+     * Triggers the node move operation to move the node to a new token.
+     *
+     * @param newToken the new token for the node to move to
+     */
+    void move(String newToken);
```
frankgh: I think we need to handle the case where the token does not belong to the node. In Cassandra, it looks like we throw an `IOException` when token is not valid. Also we should have a test around this.

frankgh (commented):
Looks good in general. I added a few comments

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/NodeMoveRequest.java
arjunashok: +1, should be in the body since it is part of the actual resource being updated rather than an optional parameter/filter.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraNodeOperationsIntegrationTest.java
```diff
@@ -0,0 +1,161 @@
...
+            HttpResponse<Buffer> streamStatsResponse = getBlocking(
+            trustedClient().get(serverWrapper.serverPort, "localhost", ApiEndpointsV1.STREAM_STATS_ROUTE)
+                           .send());
+
+            assertThat(streamStatsResponse.statusCode()).isEqualTo(OK.code());
+
+            JsonObject streamStats = streamStatsResponse.bodyAsJsonObject();
+            assertThat(streamStats).isNotNull();
+            // The operationMode should be either NORMAL (completed) or MOVING (in progress)
+            assertThat(streamStats.getString("operationMode")).isIn("NORMAL", "MOVING");
```
arjunashok: The concern with this is that we get a false negative when no move happened i.e. node remains in `NORMAL` state. We would need a more concrete way to validate that a move happened - eg. add data to ensure the move takes longer and evaluate validating the end state via ring or token-ranges endpoint.

server/src/main/java/org/apache/cassandra/sidecar/job/NodeMoveJob.java
```diff
@@ -0,0 +1,71 @@
...
+        super(jobId);
+        this.newToken = newToken;
+        this.storageOperations = storageOps;
+    }
+
+    @Override
+    public boolean isRunningOnCassandra()
+    {
+        String operationMode = storageOperations.operationMode();
+        return OPERATION_MODE_MOVING.equals(operationMode);
```
arjunashok: Question is, do we want to fail-fast when we discover the node is not in a `NORMAL` state here. It would seem reasonable to do so, but worth checking Cassandra's behavior across versions.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/NodeMoveRequest.java
sklaha: I accidentally deleted the commits in the PR from my fork, which automatically closed the PR. I’m unable to reopen it. I’ve created a new PR that addresses this comment: https://github.com/apache/cassandra-sidecar/pull/275

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/StorageOperations.java
sklaha: I accidentally deleted the commits in the PR from my fork, which automatically closed the PR. I’m unable to reopen it. I’ve created a new PR that addresses this comment: https://github.com/apache/cassandra-sidecar/pull/275

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraNodeOperationsIntegrationTest.java
sklaha: Done here: https://github.com/apache/cassandra-sidecar/pull/265

PR #266, opened 2025-10-01
CASSSIDECAR-343: Sidecar endpoint for flushing a node

sklaha:
### Sidecar endpoint for flushing a node

* API for flushing a node
* Unit tests
* Integration tests


CASSSIDECAR-343.patch
frankgh, line 1: can we remove this?

CASSSIDECAR-344.patch
frankgh, line 1: can we remove this?

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/NodeFlushRequest.class
JeetKunDoug, line 1: Please remove this class file (not sure how it landed here)

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/NodeFlushIntegrationTest.java
```diff
@@ -0,0 +1,293 @@
...
+import static org.apache.cassandra.sidecar.common.data.OperationalJobStatus.FAILED;
+import static org.apache.cassandra.sidecar.common.data.OperationalJobStatus.SUCCEEDED;
+import static org.apache.cassandra.testing.utils.AssertionUtils.getBlocking;
+import static org.apache.cassandra.testing.utils.AssertionUtils.loopAssert;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration tests for node flush operation
+ */
+public class NodeFlushIntegrationTest extends SharedClusterSidecarIntegrationTestBase
```
arjunashok, line 48: Can we validate that an actual flush operation was triggered/performed on Cassandra? One option worth evaluating would be to validate the memtable size/count around the flush operation.

arjunashok (commented):
Please also add `Changes.txt` entry

PR #267, opened 2025-10-01
CASSSIDECAR-350: Fix restore job metrics and log adjustments

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-350

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/concurrent/ConcurrencyLimiter.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/concurrent/ConcurrencyLimiter.java
@@ -21,6 +21,9 @@ package org.apache.cassandra.sidecar.concurrent;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.IntSupplier;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /**
  * A class that provides functionality for a concurrency limiter where implementing consumers can try to acquire a
  * permit before executing an operation, and later releasing the permit when the operation is completed. This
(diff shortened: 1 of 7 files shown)


server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -245,10 +244,11 @@ private void executeInternal()
         jobDiscoveryRecencyDays = Math.max(context.earliestInDays, restoreJobConfig.jobDiscoveryMinimumRecencyDays());
         LOGGER.info("Exit job discovery. " +
                     "inflightJobsCount={} " +
+                    "delay={} " +
```
frankgh, line 247: you want to keep this one at info level?

server/src/main/java/org/apache/cassandra/sidecar/concurrent/ConcurrencyLimiter.java
```diff
@@ -95,7 +99,14 @@ public void releasePermit()
     {
         permits.updateAndGet(current -> {
             int next = current - 1;
-            return Math.max(next, 0);
+            if (next < 0)
+            {
+                // log a warning with stacktrace and reset the acquired permits to 0
+                LOGGER.warn("Over-release detected!", new IllegalStateException("ConcurrencyLimiter permits over-released"));
```
frankgh, line 105: I assume we want the stacktrace here to have an idea where the issue is coming from.

frankgh (approved):
+1 looks good to me

server/src/main/java/org/apache/cassandra/sidecar/concurrent/ConcurrencyLimiter.java
yifan-c: yes

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
yifan-c: yes. Would be useful to still see the summary on exiting.

PR #268, opened 2025-10-01
CASSSIDECAR-351: Migrate RoleBasedAuthorizationIntegrationTest to int…

frankgh:
…egration-framework

Patch by Francisco Guerrero; reviewed by Saranya Krishnakumar for CASSSIDECAR-351


server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CacheConfigurationImpl.java
```diff
@@ -24,20 +24,24 @@
...
  */
 public class CacheConfigurationImpl implements CacheConfiguration
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(CacheConfigurationImpl.class);
+    private static final long DEFAULT_MAXIMUM_SIZE = 100;
+    private static final boolean DEFAULT_ENABLED = true;
+    private static final int DEFAULT_WARMUP_RETRIES = 5;
+    private static final MillisecondBoundConfiguration DEFAULT_WARMUP_RETRY_INTERVAL = MillisecondBoundConfiguration.parse("1s");
 
     protected MillisecondBoundConfiguration expireAfterAccess;
```
sarankk, line 42: Since we have a builder, let's remove the set methods for other parameters

sarankk (approved):
Thanks Francisco for this refractor. Lgtm. Left one comment, also we are missing update in CHANGES.txt

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/CacheConfigurationImpl.java
frankgh: The setters are used when we have the legacy names in the yaml , for example `expire_after_access_millis` is the legacy property. New properties should not have a setter because they should not have a legacy yaml property

PR #269, opened 2025-10-10
CASSSIDECAR-355: Avoid creating objects in the CassandraAdapter imple…

frankgh:
…mentation

Patch by Francisco Guerrero; reviewed by TBD for CASSSIDECAR-355


adapters/adapters-cassandra41/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra41/Cassandra41Adapter.java
```diff
@@ -49,8 +48,9 @@ public Cassandra41Adapter(DnsResolver dnsResolver,
      * {@inheritDoc}
      */
     @Override
-    @NotNull
```
sarankk, line 52: Nit: why do we remove @NotNull?

sarankk (approved):
+1 Thanks Francisco, Lgtm

adapters/adapters-cassandra41/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra41/Cassandra41Adapter.java
frankgh: good catch, probably missed during the refactor, I assume it's the method in the base class does not have the annotation so IDE is not complaining

PR #270, opened 2025-10-16
CASSSIDECAR-357: Add cache for Authorization layer

sarankk:


integration-tests/src/integrationTest/org/apache/cassandra/sidecar/acl/RoleBasedAuthorizationIntegrationTest.java
```diff
@@ -463,23 +470,24 @@ void testEndpointRequiringMultipleActions()
...
-                    snapshotFiles.snapshotFilesInfo()
-                                 .stream()
-                                 .filter(info -> info.fileName.endsWith("-Data.db"))
-                                 .sorted(Comparator.comparing(o -> o.fileName))
-                                 .collect(Collectors.toList());
-                    assertThat(filesToStream).isNotNull().isNotEmpty();
-                    componentDownloadUrl[0] = filesToStream.get(0).componentDownloadUrl();
-                });
+            // Wait for cache to expire without accessing it
+            Uninterruptibles.sleepUninterruptibly(3, TimeUnit.SECONDS);
```
yifan-c: nit: 3 seconds is long for unit test. How about explicitly invalidate the cache? I do not think cache eviction is the behavior you want to test in this test case. Invalidating the cache explicitly is deterministic and avoids potential flakiness.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/acl/RoleBasedAuthorizationIntegrationTest.java
```diff
@@ -488,19 +496,22 @@ void testEndpointRequiringMultipleActions()
 
             // STREAM SSTable request requires both Sidecar SNAPSHOT:STREAM permission and Cassandra's SELECT
             // permission on a table it accesses data.
-            loopAssert(2, 100, () -> {
-                // request denied without SELECT permission
-                verifyAccess(HttpMethod.GET, componentDownloadUrl[0], nonAdminClientKeystorePath, assertStatus(HttpResponseStatus.FORBIDDEN));
-            });
+
+            // Wait for cache to expire without accessing it
+            Uninterruptibles.sleepUninterruptibly(3, TimeUnit.SECONDS);
```
yifan-c: nit: avoid 3 seconds sleep if possible.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/acl/RoleBasedAuthorizationIntegrationTest.java
```diff
@@ -488,19 +496,22 @@ void testEndpointRequiringMultipleActions()
...
             // grant SELECT permission to non_admin_test_role
             grantTablePermission(session, "multiple_permissions_required_test_keyspace", "test_table", "non_admin_test_role");
         }, sslOptions);
 
-        loopAssert(2, 100, () -> {
-            // request denied without SELECT permission
-            verifyAccess(HttpMethod.GET, componentDownloadUrl[0], nonAdminClientKeystorePath, assertStatus(HttpResponseStatus.OK));
-        });
+        // Wait for cache to expire without accessing it
+        Uninterruptibles.sleepUninterruptibly(3, TimeUnit.SECONDS);
```
yifan-c: nit: avoid 3 seconds sleep if possible.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/acl/RoleBasedAuthorizationIntegrationTest.java
```diff
@@ -656,6 +667,110 @@ void testGrantingCdcFeaturePermission() throws Exception
...
+                                              mtlsTestHelper.trustStorePath(),
+                                              mtlsTestHelper.trustStorePassword());
+        withAuthenticatedSession(cluster.get(1), "cassandra", "cassandra", session -> {
+            session.execute(String.format("DELETE FROM sidecar_internal.role_permissions_v1 " +
+                                          "WHERE role = '%s' AND resource = 'data/%s'", "non_admin_test_role",
+                                          "non_admin_cache_revocation_test_keyspace"));
+        }, sslOptions);
+
+        // Wait for cache to expire without accessing it
+        Uninterruptibles.sleepUninterruptibly(3, TimeUnit.SECONDS);
```
yifan-c: nit: avoid 3 seconds sleep if possible.

server/src/main/java/org/apache/cassandra/bridge/CassandraBridgeFactory.java
```diff
@@ -133,8 +133,8 @@ public ClassLoader buildClassLoader(String... resourceNames)
                                }
                            }).toArray(URL[]::new);
 
-        return AccessController.doPrivileged((PrivilegedAction<ClassLoader>) () ->
-                                                                             new PostDelegationClassLoader(urls, Thread.currentThread().getContextClassLoader()));
+        return AccessController.doPrivileged(
+        (PrivilegedAction<ClassLoader>) () -> new PostDelegationClassLoader(urls, Thread.currentThread().getContextClassLoader()));
```
yifan-c: Is it just reformatting? If so, please revert to preserve the git history.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/AuthorizationCacheKeyImpl.java
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+        return true;
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(roles, variablesHashCode(variables));
+    }
```
yifan-c: The hashcode calculation is expensive. Can you pre-calculate the hashcode? 
Also, a big question, is the `variables` immutable? If it is mutable, it cannot be used as key.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/AuthorizationCacheKeyImpl.java
```diff
@@ -0,0 +1,113 @@
...
+    private int variablesHashCode(MultiMap variables)
+    {
+        if (variables == null)
+        {
+            return 0;
+        }
+        int hash = 0;
+        for (String name : variables.names())
+        {
+            hash += Objects.hash(name, variables.getAll(name));
```
yifan-c: Using addition for hashcode seems to be alerting. It will allow collisions more easily. Can you multiply by a prime? for example

```java
hash = hash * 31 + Objects.hash(name, variables.getAll(name))
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/AuthorizationCacheKeyImpl.java
```diff
@@ -0,0 +1,113 @@
...
+ */
+public class AuthorizationCacheKeyImpl implements AuthorizationCacheKey
+{
+    private final List<String> roles;
+    private final MultiMap variables;
+
+    public AuthorizationCacheKeyImpl(User user, MultiMap variables)
+    {
+        this.roles = extractCassandraRoles(user);
+        // Store a copy, otherwise Cache is not able to identity 2 keys with same values in MultiMap as same.
```
yifan-c: guess `identify` is what you want.
```suggestion
        // Store a copy, otherwise Cache is not able to identify 2 keys with same values in MultiMap as same.
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/CachedAuthorizationHandler.java
```diff
@@ -0,0 +1,157 @@
...
+    {
+        CacheConfiguration permissionCacheConfig = accessControlConfiguration.permissionCacheConfiguration();
+        if (permissionCacheConfig.expireAfterAccess() == null)
+        {
+            throw new ConfigurationException("Authorization handler cache must be configured with expireAfterAccess");
+        }
+        Caffeine<Object, Object> cacheBuilder = Caffeine.newBuilder()
+                                                        .maximumSize(permissionCacheConfig.maximumSize())
+                                                        .recordStats(() -> cacheMetrics);
+        if (permissionCacheConfig.expireAfterAccess() != null)
```
yifan-c: null is already checked for `expireAfterAccess` at line 135. Do you mean `refreshAfterWrite()`? or is it just redundant?

yifan-c (commented):
Looks good overall. Please check out the comments.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/CachedAuthorizationHandler.java
sarankk: It was just redundant, removing it.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/CachedAuthorizationHandler.java
```diff
@@ -0,0 +1,157 @@
...
+
+            super.handle(ctx);
+            if (!ctx.failed())
+            {
+                ctxNextCalled.set(true);
+                long durationNanos = System.nanoTime() - startTimeNanos;
+                authMetrics.authorizationTime.metric.update(durationNanos, TimeUnit.NANOSECONDS);
+                return true;
+            }
+            return false;
```
yifan-c: I am concerned about the complexity in this mapping function. It is supposed to be simple and short.

sarankk: Pulled the cache compute into a separate method.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/AuthorizationCacheKeyImpl.java
```diff
@@ -0,0 +1,69 @@
...
+public class AuthorizationCacheKeyImpl implements AuthorizationCacheKey
+{
+    private final List<String> roles;
+    private final Map<String, String> variables;
+
+    public AuthorizationCacheKeyImpl(User user, MultiMap variables)
+    {
+        this.roles = extractCassandraRoles(user);
+        // store an immutable copy of variables in a Map
+        this.variables = variables.entries()
```
frankgh: can we precompute the hash here?

server/src/main/java/org/apache/cassandra/sidecar/routes/RouteBuilder.java
```diff
@@ -206,25 +214,29 @@ public static class Factory
...
+            this.sidecarMetrics = sidecarMetrics;
         }
 
         public RouteBuilder builderForRoute()
         {
             return new RouteBuilder(accessControlConfiguration,
                                     authorizationProvider,
                                     adminIdentityResolver,
-                                    authZParameterValidateHandler);
+                                    authZParameterValidateHandler,
```
frankgh: NIT, we should instead modify the ctor to take the Factory. i.e new RouteBuilder(this);

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/acl/authorization/RoleBasedAuthorizationIntegrationTest.java
```diff
@@ -257,7 +268,7 @@ protected Function<SidecarConfigurationImpl.Builder, SidecarConfigurationImpl.Bu
                                                       Map.of());
 
             CacheConfiguration permissionCacheConfiguration = CacheConfigurationImpl.builder()
-                                                                                    .expireAfterAccess(MillisecondBoundConfiguration.parse("100ms"))
+                                                                                    .expireAfterAccess(MillisecondBoundConfiguration.parse("300ms"))
```
frankgh: can we keep it 100ms?

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/acl/authorization/RoleBasedAuthorizationIntegrationTest.java
```diff
@@ -463,8 +475,10 @@ void testEndpointRequiringMultipleActions()
                                     "data/multiple_permissions_required_test_keyspace/test_table",
                                     "SNAPSHOT:READ");
 
+            invalidateAuthorizationHandlerCaches();
+
             // wait for cache refresh
-            loopAssert(2, 100, () -> {
+            loopAssert(4, 100, () -> {
```
frankgh: can we keep this at 2?

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/acl/authorization/RoleBasedAuthorizationIntegrationTest.java
```diff
@@ -486,9 +500,11 @@ void testEndpointRequiringMultipleActions()
                                     "data/multiple_permissions_required_test_keyspace/test_table",
                                     "SNAPSHOT:STREAM");
 
+            invalidateAuthorizationHandlerCaches();
+
             // STREAM SSTable request requires both Sidecar SNAPSHOT:STREAM permission and Cassandra's SELECT
             // permission on a table it accesses data.
-            loopAssert(2, 100, () -> {
+            loopAssert(4, 100, () -> {
```
frankgh: can we keep this at 2?

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/acl/authorization/RoleBasedAuthorizationIntegrationTest.java
```diff
@@ -497,7 +513,9 @@ void testEndpointRequiringMultipleActions()
             grantTablePermission(session, "multiple_permissions_required_test_keyspace", "test_table", "non_admin_test_role");
         }, sslOptions);
 
-        loopAssert(2, 100, () -> {
+        invalidateAuthorizationHandlerCaches();
+
+        loopAssert(4, 100, () -> {
```
frankgh: can we keep this at 2?

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/AuthMetrics.java
```diff
@@ -32,17 +33,23 @@ public class AuthMetrics
     private static final String DOMAIN = SERVER_PREFIX + ".Auth";
     public final NamedMetric<Counter> jwtPemRefreshFailures;
     public final NamedMetric<Counter> jwtPemRefreshSuccesses;
+    public final NamedMetric<Timer> authorizationTime;
```
frankgh, line 40: I think we should document that this authorization time is only taking into account authorizations that are not cached

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/AuthorizationCacheKeyImpl.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Implementation of {@link AuthorizationCacheKey}, uniquely represents an authorization request with user and resource
+ * context.
+ */
+public class AuthorizationCacheKeyImpl implements AuthorizationCacheKey
+{
+    private final List<String> roles;
+    private final Map<String, String> variables;
+
+    public AuthorizationCacheKeyImpl(User user, MultiMap variables)
```
frankgh: NIT : can we use the underlying interface here?
```suggestion
    public AuthorizationCacheKeyImpl(User user, Iterable<Map.Entry<String, String>> variables)
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/CachedAuthorizationHandler.java
```diff
@@ -0,0 +1,171 @@
...
+import static org.apache.cassandra.sidecar.utils.AuthUtils.extractIdentities;
+
+/**
+ * {@link CachedAuthorizationHandler} caches all authorization requests using {@link AuthorizationCacheKey}.
+ */
+public class CachedAuthorizationHandler extends AuthorizationHandlerImpl
+{
+    static final String POPULATE_AUTHORIZATION_CACHES_ENV_NAME = "populate_authorization_caches";
+    static final Set<Cache<AuthorizationCacheKey, Boolean>> AUTHORIZATION_CACHES = ConcurrentHashMap.newKeySet();
+    private static final HttpException FORBIDDEN_EXCEPTION = new HttpException(403);
```
frankgh, line 60: will we be losing the stacktrace by precreating the exception? is the stack trace something we care about?

sarankk: No, even from `super.handle` we only get FORBIDDEN exception.

frankgh: yeah, I saw what the super class was doing, I am questioning whether that's correct or not

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/acl/authorization/RoleBasedAuthorizationIntegrationTest.java
```diff
@@ -769,6 +944,28 @@ private HttpResponse<Buffer> createRequest(WebClient client, HttpMethod method,
...
+
+        // Now block for response
+        for (int i = 0; i < times; i++)
+        {
+            HttpResponse<Buffer> response = getBlocking(futures.get(i));
+            assertThat(response.statusCode()).isEqualTo(expectedResponseCode);
+        }
+    }
+
+    private Future<HttpResponse<Buffer>> createUnblockingRequest(WebClient client, HttpMethod method, String route)
```
frankgh: NIT 
```suggestion
    private Future<HttpResponse<Buffer>> createNonBlockingRequest(WebClient client, HttpMethod method, String route)
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/CachedAuthorizationHandler.java
sarankk: What context are you referring you that we are missing, we don't get stack trace during these calls. We only see 403 exception.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/CachedAuthorizationHandler.java
sarankk: Switched to AsyncCache as suggested

yifan-c (approved):
lgtm

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/acl/authorization/RoleBasedAuthorizationIntegrationTest.java
```diff
@@ -656,6 +667,157 @@ void testGrantingCdcFeaturePermission() throws Exception
                                                HttpResponseStatus.INTERNAL_SERVER_ERROR.code());
     }
 
+    @Test
+    void testAuthorizationCaching()
+    {
+        SidecarMetrics metrics = serverWrapper.injector.getInstance(SidecarMetrics.class);
+
+        CacheStats baseline = metrics.server().cache().authorizationCacheMetrics.snapshot();
```
frankgh, line 675: is this needed?

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/CachedAuthorizationHandler.java
```diff
@@ -0,0 +1,201 @@
...
+        User user = ctx.user();
+        AuthorizationContext authorizationContext = AuthorizationContext.create(user);
+        if (this.variableHandler != null)
+        {
+            this.variableHandler.accept(ctx, authorizationContext);
+        }
+
+        AtomicBoolean ctxNextCalled = new AtomicBoolean(false);
+        AuthorizationCacheKey key = AuthorizationCacheKey.create(handlerId, authorizationContext);
+        CompletableFuture<Boolean> authorizationFuture = checkAuthorization(key, ctx, ctxNextCalled, startTimeNanos);
```
frankgh: `checkAuthorization` should not return a CompletableFuture. The code base is using vertx and we should consistently use vertx Futures instead here. Vertx provides ways to convert a CompletableFuture to Future, `Future.fromCompletionStage`.

If we start having an inconsistent code base it will become harder for new contributors to understand when to use one or the other. Having a single way of doing things will simplify this.

```suggestion
        Future<Boolean> authorizationFuture = checkAuthorization(key, ctx, ctxNextCalled, startTimeNanos);
```

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/CachedAuthorizationHandler.java
```diff
@@ -0,0 +1,201 @@
...
+
+        User user = ctx.user();
+        AuthorizationContext authorizationContext = AuthorizationContext.create(user);
+        if (this.variableHandler != null)
+        {
+            this.variableHandler.accept(ctx, authorizationContext);
+        }
+
+        AtomicBoolean ctxNextCalled = new AtomicBoolean(false);
+        AuthorizationCacheKey key = AuthorizationCacheKey.create(handlerId, authorizationContext);
```
frankgh: Let's defer key creation for when we need it. The key won't be used in the case where the cache is not enabled.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/AuthorizationCacheKeyImpl.java
```diff
@@ -0,0 +1,93 @@
...
+        {
+            // Vert.x HeadersMultimap and HeadersMultimap.MapEntry does not implement equals or hashCode,
+            // hence we store flattened variables in a Set
+            Set<String> flattenedVariables = new HashSet<>();
+            for (Map.Entry<String, String> entry : variables)
+            {
+                // We convert to lower case, since Vert.x Multimap representation is case insensitive for variables stored
+                flattenedVariables.add(entry.getKey().toLowerCase() + ":" + entry.getValue());
+            }
+            this.variables = Set.copyOf(flattenedVariables);
```
frankgh: we can probably avoid creating a copy here. The flattenedVariables set does not escape this class. If you prefer to keep an immutable set, I think we should use `Collections.unmodifiableSet` which avoids copying the values again

server/src/main/java/org/apache/cassandra/sidecar/metrics/server/AuthMetrics.java
frankgh, line 40: maybe we can add a comment like this?
```suggestion
    /**
     * Captures the time to successfully authorize non-cached requests. Cached authorization requests will
     * not be recorded in this metric
     */
    public final NamedMetric<Timer> authorizationTime;
```

frankgh (changes requested):

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/acl/authorization/RoleBasedAuthorizationIntegrationTest.java
sarankk: Yes it is since cache stats are shared between test cases

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/AuthorizationCacheKeyImpl.java
sarankk: Discussed we use the created Set as is.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/CachedAuthorizationHandler.java
sarankk: We have lot of usage of `CompletableFuture` in Sidecar project in client subproject.

frankgh: the reason for the client to use CompletableFuture is because the core project client project does not have any vertx dependencies

frankgh (approved):
+1 looks good to me!

PR #272, opened 2025-10-30
CASSSIDECAR-360: Sidecar API Endpoint for Nodetool Compaction Stop

shalnisundram:
Add bespoke API endpoint for nodetool compaction stop.

This implementation carries out functionality currently performed when execution command `nodetool stop`, which is a short-running operation that requests ongoing compactions of a specified` type` or `id` to be stopped.
 
See [CASSANDRA-20021 JIRA ticket](https://issues.apache.org/jira/browse/CASSSIDECAR-360) - more details can be found on ongoing effort to build APIs for nodetool C* cluster management operations.

patch by Shalni Sundram


adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
```diff
@@ -53,4 +53,27 @@ public List<Map<String, String>> getCompactions()
...
+
+        // compactionId takes precedence over type if both are provided
+        if (compactionId != null && !compactionId.trim().isEmpty())
+        {
+            proxy.stopCompactionById(compactionId);
+        }
+        else
+        {
+            // Handler guarantees compactionType is non-null/non-empty
+            proxy.stopCompaction(compactionType);
```
bbotella: Maybe we should add a @NotNull annotation to compactionType?

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -143,7 +143,7 @@ public final class ApiEndpointsV1
     public static final String NODE_DRAIN_ROUTE = API_V1 + CASSANDRA + "/operations/drain";
     public static final String STREAM_STATS_ROUTE = API_V1 + CASSANDRA + "/stats/streams";
     public static final String TABLE_STATS_ROUTE = API_V1 + CASSANDRA + PER_KEYSPACE + PER_TABLE + "/stats";
-
+    public static final String COMPACTION_STOP_ROUTE = API_V1 + CASSANDRA + "/operations/compaction/stop";
```
bbotella: nit: I see that `/operations/` is repeated across some routes. What about extracting it to a static final?

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/CompactionStopRequest.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package org.apache.cassandra.sidecar.common.request;
+
+import io.netty.handler.codec.http.HttpMethod;
+import org.apache.cassandra.sidecar.common.ApiEndpointsV1;
+import org.apache.cassandra.sidecar.common.response.OperationalJobResponse;
+
+/**
+ * Represents a request to execute node decommission operation
```
bbotella: Is node decommission the same as compaction stop?

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CompactionStopRequestPayload.java
```diff
@@ -0,0 +1,82 @@
...
+ * <p>Valid JSON:</p>
+ * <pre>
+ *   { "compaction_type": "COMPACTION", "compaction_id": "abc-123" }
+ *   { "compaction_type": "VALIDATION" }
+ * </pre>
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class CompactionStopRequestPayload
+{
+    private final String compactionType;
```
bbotella, line 42: Should this be an enum instead of a free string?

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/CompactionStopResponse.java
```diff
@@ -0,0 +1,225 @@
...
+ * Response class for the Compaction Stop API
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class CompactionStopResponse
+{
+    public static final String COMPACTION_TYPE = "compaction_type";
+    public static final String COMPACTION_ID = "compaction_id";
+    public static final String STATUS = "status";
+    public static final String ERROR_CODE = "error_code";
+    public static final String REASON = "reason";
```
bbotella: nit: What about appending `_KEY` to the constant names?

server/src/main/java/org/apache/cassandra/sidecar/modules/CassandraOperationsModule.java
```diff
@@ -21,37 +21,14 @@
...
-import org.apache.cassandra.sidecar.handlers.NativeUpdateHandler;
-import org.apache.cassandra.sidecar.handlers.NodeDecommissionHandler;
-import org.apache.cassandra.sidecar.handlers.NodeDrainHandler;
-import org.apache.cassandra.sidecar.handlers.OperationalJobHandler;
-import org.apache.cassandra.sidecar.handlers.RingHandler;
-import org.apache.cassandra.sidecar.handlers.SchemaHandler;
-import org.apache.cassandra.sidecar.handlers.StreamStatsHandler;
-import org.apache.cassandra.sidecar.handlers.TableStatsHandler;
-import org.apache.cassandra.sidecar.handlers.TokenRangeReplicaMapHandler;
+import org.apache.cassandra.sidecar.handlers.*;
```
bbotella: Please revert this. We try to avoid the * imports

server/src/main/java/org/apache/cassandra/sidecar/modules/CassandraOperationsModule.java
```diff
@@ -109,6 +86,25 @@ VertxRoute cassandraCompactionStatsRoute(RouteBuilder.Factory factory,
...
     }
 
+    @POST  // Note: POST, not PUT, since we're sending data in the body
+    @Path(ApiEndpointsV1.COMPACTION_STOP_ROUTE)
+    @Operation(summary = "Stop compaction operation",
+            description = "Stops a compaction operation on the Cassandra node")
+    @APIResponse(description = "Compaction stop operation completed",
+            responseCode = "200",
+            content = @Content(mediaType = "application/json",
+                    schema = @Schema(implementation = CompactionStopResponse.class)))
```
bbotella: Please add the @APIResponse for the error codes as well

shalnisundram: See [lines 149-189 of CompactionStopHandler.java](https://github.com/shalnisundram/cassandra-sidecar/blob/6d42f45dbdbbc30da2149990272d2494ca437227/server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java#L149) - I do some custom error handling here, as particular api response codes will have different messages thrown depending on the input. Let me know if I should still edit this section in CassandraOperationsModule.

server/src/main/java/org/apache/cassandra/sidecar/modules/CassandraOperationsModule.java
shalnisundram: Thanks for letting me know - fixed.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/CompactionStopRequest.java
shalnisundram: they're not the same that's a typo - thanks for catching. Fixed.

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
shalnisundram: See change on [CassandraCompactionManagerOperations line 77 ](https://github.com/shalnisundram/cassandra-sidecar/blob/6d42f45dbdbbc30da2149990272d2494ca437227/adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java#L77). Not exactly an annotation but a requireNonNull check. Let me know if that's acceptable.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CompactionStopRequestPayload.java
shalnisundram: Chose deliberately to use `String`, as I figured this would encourage forward compatibility with future C* versions if more compactionTypes are added later.  i.e., would want to avoid sidecar forcing all C* versions into a single enum definition, and would rather leave this responsibility to cassandra itself. Would appreciate @arjunashok's input for this though.

sklaha (commented):
Thank you for the PR. Looks good at a high level. Please run `./gradlew clean; ./gradlew build` which will show you check style issues. I am leaving my other comments inline.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CompactionStopRequestPayload.java
sklaha: You are already defining the compaction types as a set in AbstractHandler.SUPPORTED_COMPACTION_TYPES. Better to make it an enum instead of a set.

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
```diff
@@ -0,0 +1,190 @@
...
+        operations.stopCompaction(compactionId, compactionType);
+
+        // Return success response
+        return CompactionStopResponse.builder()
+                .compactionType(compactionType)
+                .compactionId(compactionId)
+                .status("SUCCESS")
+                .errorCode("200 OK")
+                .reason("Operation Succeeded")
+                .build();
```
sklaha: Do we need these fields? Doesn't HTTP 200 response provide the same information? Also, these fields being set for failure scenarios.

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Override extractParamsOrThrow to support compactionStop param constraints
+     * Method extracts and validates compaction stop request from routing context
+     *
+     * @param context the routing context
+     * @return validated CompactionStopRequestPayload
+     */
+    @Override
+    protected CompactionStopRequestPayload extractParamsOrThrow(RoutingContext context)
```
sklaha, line 134: What happens when the request has both compaction id and compaction type?

sklaha: Looks like compaction id takes precedence. It could be better to just fail the request with appropriate error message.

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
```diff
@@ -53,4 +55,27 @@ public List<Map<String, String>> getCompactions()
         return jmxClient.proxy(CompactionManagerJmxOperations.class, COMPACTION_MANAGER_OBJ_NAME)
                         .getCompactions();
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void stopCompaction(String compactionId, String compactionType)
```
sklaha: I would prefer to have two separate methods for compaction operations based on ID and type, rather than having a single method that accepts two parameters and actually uses only one.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CompactionStopIntegrationTest.java
```diff
@@ -0,0 +1,218 @@
...
+        CompactionStopResponse stopResponse = response.bodyAsJson(CompactionStopResponse.class);
+        assertThat(stopResponse).isNotNull();
+        assertThat(stopResponse.status()).isEqualTo("SUCCESS");
+        assertThat(stopResponse.compactionType()).isEqualTo("COMPACTION");
+        assertThat(stopResponse.errorCode()).isEqualTo("200 OK");
+        assertThat(stopResponse.reason()).isEqualTo("Operation Succeeded");
+    }
+
+    @Test
+    void testStopCompactionByIdSuccess()
```
sklaha: In the current implementation, a request with a non-existent compaction ID silently succeeds. The JMX call simply succeeds without any action being taken. It would be more informative to update the Cassandra JMX implementation to return a boolean value or some other indicator. I will seek input from others on this.

https://github.pie.apple.com/slaha/aci-cassandra/blob/38ab7d02e0979074f3de1c7f90ce70668d2492b9/src/java/org/apache/cassandra/db/compaction/CompactionManager.java#L2306

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CompactionStopIntegrationTest.java
```diff
@@ -0,0 +1,218 @@
...
+import static org.apache.cassandra.testing.TestUtils.DC1_RF1;
+import static org.apache.cassandra.testing.TestUtils.TEST_KEYSPACE;
+import static org.apache.cassandra.testing.TestUtils.TEST_TABLE_PREFIX;
+import static org.apache.cassandra.testing.utils.AssertionUtils.getBlocking;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration tests for the Compaction Stop API endpoint
+ */
+class CompactionStopIntegrationTest extends SharedClusterSidecarIntegrationTestBase
```
sklaha, line 58: Can we add a test that triggers compaction and effectively stops ongoing compactions using the API? You can get an idea of how to trigger a compaction from the `testCompactionStatsRetrieval`.

https://github.com/apache/cassandra-sidecar/blob/c987f2956b22b457cf3ac68408657e706d0b9a1b/integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraStatsIntegrationTest.java#L285

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CompactionStopRequestPayload.java
shalnisundram: Done

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
shalnisundram: Some small changes in the handler - the 200 response shouldn't appear upon a failure - these are caught in `extractParamsOrThrow()` on line 133 of this file.
Other than that, we'd decided in the original design to include these fields - they won't be exposed automatically, but let me know if there's a better way of including them in the response.

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
shalnisundram: Done - split into two functions

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
shalnisundram: See comment to previous message above - functions split with `stopById()` taking precedence.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CompactionStopIntegrationTest.java
shalnisundram: Good idea and thanks for the reference. Added two integration tests in this file - one for stopping compactions by type and the other by ID. Tests check this by comparing list of active compactions before and after new API call.

Made SSTables large + only considered compactions under 50% complete to catch and stop them before they finish on their own.

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -139,16 +140,18 @@ public final class ApiEndpointsV1
...
     public static final String OPERATIONAL_JOB_ROUTE = API_V1 + CASSANDRA + PER_OPERATIONAL_JOB;
-    public static final String NODE_DECOMMISSION_ROUTE = API_V1 + CASSANDRA + "/operations/decommission";
-    public static final String NODE_DRAIN_ROUTE = API_V1 + CASSANDRA + "/operations/drain";
+    public static final String NODE_DECOMMISSION_ROUTE = API_V1 + CASSANDRA + OPERATIONS + "/decommission";
     public static final String STREAM_STATS_ROUTE = API_V1 + CASSANDRA + "/stats/streams";
     public static final String TABLE_STATS_ROUTE = API_V1 + CASSANDRA + PER_KEYSPACE + PER_TABLE + "/stats";
-
+    public static final String COMPACTION_STOP_ROUTE = API_V1 + CASSANDRA + OPERATIONS + "/compaction/stop";
+    public static final String NODE_DRAIN_ROUTE = API_V1 + CASSANDRA + OPERATIONS + "/drain";
+    public static final String LIFECYCLE_ROUTE = API_V1 + CASSANDRA + "/lifecycle";
```
arjunashok: These changes are unrelated to your PR, possibly as a result of the merge from the main branch.  You might want to rebase your changes on top of the main/develop branch contents, so the PR only has your changes.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/CompactionStopRequest.java
```diff
@@ -0,0 +1,46 @@
...
+package org.apache.cassandra.sidecar.common.request;
+
+import io.netty.handler.codec.http.HttpMethod;
+import org.apache.cassandra.sidecar.common.ApiEndpointsV1;
+import org.apache.cassandra.sidecar.common.response.OperationalJobResponse;
+
+/**
+ * Represents a request to execute node stop operation
+ */
+public class CompactionStopRequest extends JsonRequest<OperationalJobResponse>
```
arjunashok: Should this be `extends JsonRequest<CompactionStopResponse>`

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/CompactionStopResponse.java
```diff
@@ -0,0 +1,225 @@
...
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class CompactionStopResponse
+{
+    public static final String COMPACTION_TYPE_KEY = "compaction_type";
+    public static final String COMPACTION_ID_KEY = "compaction_id";
+    public static final String STATUS_KEY = "status";
+    public static final String ERROR_CODE_KEY = "error_code";
+    public static final String REASON_KEY = "reason";
+
+    private final String compactionType;
```
arjunashok, line 36: This should use the enum for consistency.

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/CompactionStopResponse.java
```diff
@@ -0,0 +1,225 @@
...
+/**
+ * Response class for the Compaction Stop API
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class CompactionStopResponse
+{
+    public static final String COMPACTION_TYPE_KEY = "compaction_type";
+    public static final String COMPACTION_ID_KEY = "compaction_id";
+    public static final String STATUS_KEY = "status";
+    public static final String ERROR_CODE_KEY = "error_code";
```
arjunashok: Let's evaluate the intended usage of the `error_code` field in determining the values. Based on the current name, it seems to suggest this is a  status code (currently returning `200 OK`) 

The 200 status code would be redundant in this case, as it is implied with the HTTP response code. 

The other option is to do away with this field since we have a `reason` field to denote an arbitrary string reason when there is an error (or a non-200 response).

arjunashok: Also, can `status` be an enum here? Seems like it can be one of 3 or 4 values - `SUBMITTED`, `PENDING` or `FAILED`

server/src/main/java/org/apache/cassandra/sidecar/modules/CassandraOperationsModule.java
```diff
@@ -112,6 +98,27 @@ VertxRoute cassandraCompactionStatsRoute(RouteBuilder.Factory factory,
...
+    @APIResponse(responseCode = "500", description = "Internal server error")
+    @APIResponse(responseCode = "400", description = "Internal server error on getting the ring view")
```
arjunashok: Is this the appropriate description? `400` implies bad request, so it should not state "Internal server error"

shalnisundram: Yes thanks for catching that. Changed to "Invalid request - malformed JSON body or invalid compaction parameters"

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/CompactionStopResponse.java
shalnisundram: Original status options for the command were just two - "PENDING" and "FAILED". For this API it's "SUBMITTED" (more clear I think) and "FAILED".
Status changed to enum - see CompactionStopStatus.java for new enum class

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/CompactionStopResponse.java
shalnisundram: Done

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/CompactionStopRequest.java
shalnisundram: Yes - changed

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
shalnisundram: Rebased - changed file now only includes compaction stop additions related to new compaction stop imports and routes

client/src/main/java/org/apache/cassandra/sidecar/client/SidecarClient.java
```diff
@@ -752,6 +753,20 @@ public CompletableFuture<CompactionStatsResponse> compactionStats(SidecarInstanc
                                             .build());
     }
 
+    /**
+     * Executes the compaction stats request using the default retry policy and provided {@code instance}.
```
arjunashok: Nit: update to `compaction stop`

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CompactionStopIntegrationTest.java
```diff
@@ -0,0 +1,461 @@
...
+                        .sendBuffer(buffer(payload))
+                        .expecting(HttpResponseExpectation.SC_OK)
+        );
+
+        assertThat(response.statusCode()).isEqualTo(HttpResponseStatus.OK.code());
+        CompactionStopResponse stopResponse = response.bodyAsJson(CompactionStopResponse.class);
+        assertThat(stopResponse).isNotNull();
+        assertThat(stopResponse.status()).isEqualTo(CompactionStopStatus.SUBMITTED);
+        assertThat(stopResponse.compactionType()).isEqualTo("COMPACTION");
+        assertThat(stopResponse.reason()).isEqualTo("Operation Succeeded");
```
arjunashok: No longer applicable, correct?

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CompactionStopIntegrationTest.java
```diff
@@ -0,0 +1,461 @@
...
+        // Generate many SSTables with large data to ensure longer compactions
+        generateSSTables(TEST_TABLE, 200);  // More SSTables with 10KB rows = much longer compaction
+
+        // Trigger compaction in background
+        Thread compactionThread = new Thread(() -> {
+            cluster.stream().forEach(instance ->
+                    instance.nodetool("compact", TEST_KEYSPACE, TEST_TABLE.table())
+            );
+        });
+        compactionThread.start();
```
arjunashok: `nodetool compact` is non-blocking by default and will return immediately. We like;y do not need to run this on a separate thread.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CompactionStopIntegrationTest.java
```diff
@@ -0,0 +1,461 @@
...
+                    assertThat(stopResponse.statusCode()).isEqualTo(HttpResponseStatus.OK.code());
+                    logger.info("Compaction stop called successfully for ID: {} at {}% progress", compactionId, progressAtStart);
+
+                    // Verify the compaction with this ID is no longer in active compactions
+                    // Since it was at < 50% when we called stop, it's extremely unlikely to have
+                    // naturally completed in the verification window (would need to compact remaining
+                    // ~1GB+ in a few seconds)
+                    boolean compactionStopped = false;
+                    for (int verifyAttempt = 0; verifyAttempt < 10; verifyAttempt++) {
+                        Thread.sleep(300);
```
arjunashok: This makes the test brittle and can introduce flakiness. Lets use `AssertionUtils.loopAssert()` instead for both the polling loop and for the verification.

arjunashok (commented):
Looking good for the most part. Added some minor comments.

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
```diff
@@ -53,4 +55,32 @@ public List<Map<String, String>> getCompactions()
...
                         .getCompactions();
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public void stopCompactionById(String compactionId)
+    {
+        CompactionManagerJmxOperations proxy = jmxClient.proxy(
```
bbotella: Do you need to declare the proxy if compactionId == null? Can't we do this inside the if?

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
bbotella: nit: I see the `operations` part of the path in several static variables here. Let's extract it to its own private static

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CompactionStopRequestPayload.java
```diff
@@ -0,0 +1,82 @@
...
+        return this.compactionId;
+    }
+
+    @Override
+    public String toString()
+    {
+        return "CompactionStopRequestPayload{" +
+               "compactionType='" + compactionType + '\'' +
+               ", compactionId='" + compactionId + '\'' +
+               '}';
```
bbotella:
```suggestion
        return "CompactionStopRequestPayload{" +
               "compactionType='" + compactionType + "'"  +
               ", compactionId='" + compactionId + "'"  +
               "}";
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
```diff
@@ -0,0 +1,169 @@
...
+        CompactionType compactionType = request.compactionType();
+        String compactionId = request.compactionId();
+
+        // Convert enum to string for the operation (if not null)
+        // Use .name() to get uppercase, which Cassandra's OperationType expects
+        String compactionTypeStr = compactionType != null ? compactionType.name() : null;
+
+        // Attempt to stop the compaction
+        // If compactionId is provided, use it (takes precedence over type)
+        if (compactionId != null && !compactionId.trim().isEmpty()) {
```
bbotella: I see this pattern of "if compactionID then do otherwise do by type" in several different places. What do you think about abstracting that inside operations, with a generic method that does `stopCompaction(@Nullable String compactionId, @Nullable String compactionType)`, and performs that if internally?

Sorry if this has already been discussed.

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+        catch (IllegalArgumentException e)
+        {
+            logger.warn("Bad request. {}", e.getMessage());
+            throw wrapHttpException(HttpResponseStatus.BAD_REQUEST, e.getMessage());
+        }
+
+        // Validate that at least one field is provided
+        if (payload.compactionType() == null &&
+            (payload.compactionId() == null || payload.compactionId().trim().isEmpty()))
```
bbotella: I also see this pattern of `== null` or `== null or empty` across different places. What about adding boolean public methods on payload that return if they are valid? Something like `isCompactionTypeValid()` and `isCompactionIdValid`?

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CompactionStopIntegrationTest.java
shalnisundram: Omitted the reason() field - the status , type, and id fields still stand in the response so left those in

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CompactionStopIntegrationTest.java
shalnisundram: gotcha - omitted separate thread.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CompactionStopIntegrationTest.java
shalnisundram: Tried with `loopAssert`, but the tests will typically fail since a condition must be met to pass. Since trying to catch an active compaction will usually not succeed, I kept the manual loop structure in. 
To improve the test, added an `assumeTrue()` at the end of the test so it doesn't pass by default when not finding an active compaction. Now the test is ignored after all attempts: https://github.com/apache/cassandra-sidecar/blob/def6e6fb36cb7fa05bf9e3a0237d8a407cd3e8cc/integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CompactionStopIntegrationTest.java#L451

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
shalnisundram: Yeah in the latter case we would have a single method that accepts two params but only uses one, which others expressed was not preferable. Thus, preserved the `stopByID()` and `stopCompaction()` structure already used by `node stop`

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
shalnisundram: Done. Added the private static, but only used it for the `node stop` route rather than `node drain` and `node decommission` routes which also uses this path, as those changes would not be directly related to this PR.

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
shalnisundram: Good catch yes - changed.

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
shalnisundram: Done - added those functions in `CompactionStopRequestPayload.java`, although didn't use those functions for the pattern present in `CassandraCompactionManagerOperations.java`, as those methods don't work with the `payload` object.
https://github.com/apache/cassandra-sidecar/blob/0a9b3f458d0788a1dd8d517d35bb4e4e4c1754a1/client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CompactionStopRequestPayload.java#L77

adapters/adapters-base/src/test/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperationsTest.java
```diff
@@ -0,0 +1,127 @@
...
+    @Test
+    void testStopCompactionJmxProxyCalledOnce()
+    {
+        // Test JMX proxy obtained exactly once per call
+        compactionManagerOperations.stopCompactionById("test-id");
+
+        verify(mockJmxClient, times(1))
+            .proxy(CompactionManagerJmxOperations.class, COMPACTION_MANAGER_OBJ_NAME);
+    }
+}
```
jyothsnakonisa: Please add a new line in the end of the file

client-common/src/test/java/org/apache/cassandra/sidecar/common/request/data/CompactionStopRequestPayloadTest.java
```diff
@@ -0,0 +1,179 @@
...
+        CompactionStopRequestPayload lowerCase = new CompactionStopRequestPayload(CompactionType.COMPACTION, "Test-ID-123");
+        String json = MAPPER.writeValueAsString(lowerCase);
+        assertThat(json).contains("compaction");
+        assertThat(json).contains("Test-ID-123");
+
+        CompactionStopRequestPayload deser = MAPPER.readValue(json, CompactionStopRequestPayload.class);
+        assertThat(deser.compactionType()).isEqualTo(CompactionType.COMPACTION);
+        assertThat(deser.compactionId()).isEqualTo("Test-ID-123");
+    }
+}
```
jyothsnakonisa: Please add a new line at the end of the file

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
```diff
@@ -0,0 +1,168 @@
...
+        if (!payload.atLeastOneParamProvided())
+        {
+            logger.warn("Bad request. Both compaction_type and compaction_id are missing.");
+            throw wrapHttpException(HttpResponseStatus.BAD_REQUEST,
+                    "At least one of 'compaction_type' or 'compaction_id' must be provided");
+        }
+
+        return payload;
+    }
+}
```
jyothsnakonisa: Please add a new line here

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
```diff
@@ -0,0 +1,168 @@
...
+     *
+     * @param operations the compaction manager operations
+     * @param request    the request payload containing compaction_type and/or compaction_id
+     * @return CompactionStopResponse with the operation result
+     */
+    private CompactionStopResponse stopCompaction(CompactionManagerOperations operations,
+                                                  CompactionStopRequestPayload request)
+    {
+        CompactionType compactionType = request.compactionType();
+        String compactionId = request.compactionId();
```
jyothsnakonisa, line 105: Do you want to trim the compactionID? I see that you are doing it in the other parts of the code

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
```diff
@@ -53,4 +55,31 @@ public List<Map<String, String>> getCompactions()
...
+    public void stopCompactionById(String compactionId)
+    {
+        // compactionId takes precedence over type if both are provided
+        if (compactionId != null && !compactionId.trim().isEmpty()) {
+            CompactionManagerJmxOperations proxy = jmxClient.proxy(
+                    CompactionManagerJmxOperations.class,
+                    COMPACTION_MANAGER_OBJ_NAME
+            );
+            proxy.stopCompactionById(compactionId);
+        }
```
jyothsnakonisa, line 81: Do you want to throw an exception when the compactionId is invalid here and in the `stopCompaction` method

jyothsnakonisa (approved):

jyothsnakonisa:
Looks good overall, few minor comments

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
shalnisundram: Won't trim here since in many cases the compactionId will be null, so want to avoid NPEs from attempting to trim in these scenarios. Have a null check and trim already in `hasValidCompactionId()` so will keep that structure
Though realized from this I wasn't returning the trimmed ID in `stopCompactionById()` so made that change there + respective tests

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
shalnisundram: Done - thrown IllegalArgumentException. Thanks for catching that

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
```diff
@@ -53,4 +55,37 @@ public List<Map<String, String>> getCompactions()
...
+    @Override
+    public void stopCompactionById(String compactionId)
+    {
+        // compactionId takes precedence over type if both are provided
+        if (compactionId != null && !compactionId.trim().isEmpty()) {
+            CompactionManagerJmxOperations proxy = jmxClient.proxy(
+                    CompactionManagerJmxOperations.class,
+                    COMPACTION_MANAGER_OBJ_NAME
+            );
+            proxy.stopCompactionById(compactionId);
```
sarankk: Should we use the trimmed `compactionId` here?

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
```diff
@@ -20,10 +20,12 @@
 
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import org.apache.cassandra.sidecar.adapters.base.jmx.CompactionManagerJmxOperations;
 import org.apache.cassandra.sidecar.common.server.CompactionManagerOperations;
 import org.apache.cassandra.sidecar.common.server.JmxClient;
+import org.jetbrains.annotations.NotNull;
```
sarankk: Nit: Remove unused import

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
```diff
@@ -53,4 +55,37 @@ public List<Map<String, String>> getCompactions()
...
+    {
+        // compactionId takes precedence over type if both are provided
+        if (compactionId != null && !compactionId.trim().isEmpty()) {
+            CompactionManagerJmxOperations proxy = jmxClient.proxy(
+                    CompactionManagerJmxOperations.class,
+                    COMPACTION_MANAGER_OBJ_NAME
+            );
+            proxy.stopCompactionById(compactionId);
+        } else {
+            throw new IllegalArgumentException("compaction id is null or empty");
```
sarankk: Nit: can you also add the compaction id in error message. Easy to debug with that.

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
```diff
@@ -53,4 +55,37 @@ public List<Map<String, String>> getCompactions()
...
+        if (compactionType != null && !compactionType.trim().isEmpty()) {
+            CompactionManagerJmxOperations proxy = jmxClient.proxy(
+                    CompactionManagerJmxOperations.class,
+                    COMPACTION_MANAGER_OBJ_NAME
+            );
+
+            proxy.stopCompaction(Objects.requireNonNull(compactionType,
+                    "compactionType must not be null when compactionId is not provided"));
+        } else {
+            throw new IllegalArgumentException("compaction type is null or empty");
```
sarankk: Nit: can you also add the compaction type in error message. Easy to debug with that.

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/jmx/CompactionManagerJmxOperations.java
```diff
@@ -34,4 +34,14 @@ public interface CompactionManagerJmxOperations
      * @return list of compaction info maps
      */
     List<Map<String, String>> getCompactions();
+
+    /**
+     * Stop compaction by type
```
sarankk, line 39: Can you also that in Javadoc, that it throws exception for invalid compaction type, similarly javadoc for compaction by Id.

client-common/src/main/java/org/apache/cassandra/sidecar/common/ApiEndpointsV1.java
```diff
@@ -18,6 +18,8 @@
 
 package org.apache.cassandra.sidecar.common;
 
+import java.lang.ref.PhantomReference;
```
sarankk: Nit: Remove unused import

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CompactionStopRequestPayload.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Creates a new CompactionStopRequestPayload
+     *
+     * @param compactionType the type of compaction to stop (e.g., COMPACTION, VALIDATION, etc.)
+     * @param compactionId   optional ID of a specific compaction to stop
+     */
+    @JsonCreator
+    public CompactionStopRequestPayload(
+    @JsonProperty(value = "compaction_type") CompactionType compactionType,
```
sarankk: Nit: for parameter names shall we use camel case, just for consistency?. Other request payloads use camel case

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/CompactionStopResponse.java
```diff
@@ -0,0 +1,170 @@
...
+import org.apache.cassandra.sidecar.common.data.CompactionStopStatus;
+import org.apache.cassandra.sidecar.common.data.CompactionType;
+
+/**
+ * Response class for the Compaction Stop API
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class CompactionStopResponse
+{
+    public static final String COMPACTION_TYPE_KEY = "compaction_type";
```
sarankk: Nit: here too shall we use camel case for parameter names for consistency?

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/CompactionStopResponse.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Constructs a new {@link CompactionStopResponse}.
+     *
+     * @param compactionType the type of compaction that was requested to stop
+     * @param compactionId   the ID of the compaction that was requested to stop
+     * @param status         the status of the stop operation (e.g., "PENDING", "FAILED")
+     */
+    @JsonCreator
+    public CompactionStopResponse(@JsonProperty(COMPACTION_TYPE_KEY) CompactionType compactionType,
```
sarankk: Nit: can you reuse these in request payload too. Feel free to ignore this comment.

client-common/src/test/java/org/apache/cassandra/sidecar/common/request/data/CompactionStopRequestPayloadTest.java
```diff
@@ -0,0 +1,179 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.cassandra.sidecar.common.data.CompactionType;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
```
sarankk: Nit: Unused import.

client/src/testFixtures/java/org/apache/cassandra/sidecar/client/SidecarClientTest.java
```diff
@@ -2003,7 +2068,7 @@ void testNodeLifecycleInfo() throws Exception
         assertThat(result).isNotNull();
         assertThat(result.currentState()).isEqualTo(CassandraState.RUNNING);
         assertThat(result.desiredState()).isEqualTo(CassandraState.RUNNING);
-        assertThat(result.status()).isEqualTo(OperationStatus.CONVERGED);
+        assertThat(result.status()).isEqualTo(Lifecycle.OperationStatus.CONVERGED);
```
sarankk, line 2144: Nit: These changes seem unrelated to PR.

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/CompactionManagerOperations.java
```diff
@@ -32,4 +32,22 @@ public interface CompactionManagerOperations
...
+     * Stops compaction based on compaction ID or type.
+     * If compactionId is provided, it takes precedence over compactionType.
+     *
+     * @param compactionId   the compaction ID to stop (nullable)
+     * @throws IllegalArgumentException if both parameters are null or empty
+     */
+    void stopCompactionById(String compactionId);
+
+    /**
+     * Stops compaction based on compaction ID or type.
```
sarankk: This comment is confusing. This method is only for stopping with compaction type right? similarly comment for `stopCompactionById` method.

server/src/main/java/org/apache/cassandra/sidecar/modules/CassandraOperationsModule.java
```diff
@@ -112,6 +115,28 @@ VertxRoute cassandraCompactionStatsRoute(RouteBuilder.Factory factory,
...
 
+    @POST  // Note: POST, not PUT, since we're sending data in the body
+    @Path(ApiEndpointsV1.COMPACTION_STOP_ROUTE)
+    @Operation(summary = "Stop compaction operation",
+            description = "Stops a compaction operation on the Cassandra node")
+    @APIResponse(description = "Compaction stop operation completed",
+            responseCode = "200",
+            content = @Content(mediaType = "application/json",
+                    schema = @Schema(implementation = CompactionStopResponse.class)))
+    @APIResponse(responseCode = "500", description = "Internal server error")
```
sarankk: Ideally we should not see `500` response from handler. 500 means we did not foresee some error and handle it properly. I would suggest removing it from list of expected responses.

server/src/main/java/org/apache/cassandra/sidecar/modules/CassandraOperationsModule.java
```diff
@@ -112,6 +115,28 @@ VertxRoute cassandraCompactionStatsRoute(RouteBuilder.Factory factory,
         return factory.buildRouteWithHandler(compactionStatsHandler);
     }
 
+    @POST  // Note: POST, not PUT, since we're sending data in the body
```
sarankk: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/PUT Http docs say we can send request content with `PUT` requests too. Difference between them is, `PUT` requests are idempotent. Gossip update route is an example https://github.com/apache/cassandra-sidecar/blob/6cf4b4c9c818efac825af999fb47ac545f75d430/server/src/main/java/org/apache/cassandra/sidecar/modules/CassandraOperationsModule.java#L395. Lets update this to `PUT` and remove note?

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CompactionType.java
```diff
@@ -0,0 +1,74 @@
...
+package org.apache.cassandra.sidecar.common.data;
+
+import java.util.Locale;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+/**
+ * Supported compaction types based on Cassandra's OperationType enum
+ */
+public enum CompactionType
```
sarankk, line 30: I do not see `MAJOR_COMPACTION` type here from `OperationType` class, is that intentional?

client-common/src/main/java/org/apache/cassandra/sidecar/common/data/CompactionType.java
```diff
@@ -0,0 +1,74 @@
...
+    ROW_CACHE_SAVE,
+    COUNTER_CACHE_SAVE,
+    CLEANUP,
+    SCRUB,
+    UPGRADE_SSTABLES,
+    INDEX_BUILD,
+    TOMBSTONE_COMPACTION,
+    ANTICOMPACTION,
+    VERIFY,
+    WRITE,
```
sarankk: `WRITE` operation seems to be not a compaction type, can you double check?

sarankk (commented):
+1 Thanks Shalni, overall looks great to me. Left some comments, I am +1 once the comments are addressed.

server/src/main/java/org/apache/cassandra/sidecar/modules/CassandraOperationsModule.java
shalnisundram: Yes would prefer an idempotent operation - thanks. Fix made

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CompactionType.java
shalnisundram: Thanks for the catch - I may have been looking at another C* version. Just changed it to reflect this list: https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/db/compaction/OperationType.java

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/CompactionManagerOperations.java
shalnisundram: Correct - clarification made in both Javadocs in most recent push

client-common/src/main/java/org/apache/cassandra/sidecar/common/data/CompactionStopStatus.java
```diff
@@ -0,0 +1,17 @@
+package org.apache.cassandra.sidecar.common.data;
```
sarankk, line 18: Missing license for file

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
```diff
@@ -53,4 +55,56 @@ public List<Map<String, String>> getCompactions()
...
+            throw new IllegalArgumentException("compaction type " + compactionType + " is null or empty");
+        }
+    }
+
+    @Override
+    public List<String> supportedCompactionTypes()
+    {
+        return Arrays.stream(CompactionType.values())
+                .map(CompactionType::name)
+                .collect(Collectors.toList());
```
arjunashok: Optimize to return a `Set` for lookups, or optionally, a static set in the enum. 

eg.
```
    private static final List<String> NAMES = Arrays.stream(values())
            .map(CompactionType::name)
            .collect(Collectors.toUnmodifiableList());
```

adapters/adapters-cassandra41/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra41/CompactionType.java
```diff
@@ -0,0 +1,78 @@
...
+package org.apache.cassandra.sidecar.adapters.cassandra41;
+
+import java.util.Locale;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+
+/**
+ * Supported compaction types based on Cassandra's OperationType enum
+ */
+public enum CompactionType
```
arjunashok: Let's be explicit about the version this corresponds to in the class name. eg. `Cassandra41CompactionType`

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperations.java
```diff
@@ -53,4 +55,56 @@ public List<Map<String, String>> getCompactions()
...
+        else
+        {
+            throw new IllegalArgumentException("compaction type " + compactionType + " is null or empty");
+        }
+    }
+
+    @Override
+    public List<String> supportedCompactionTypes()
+    {
+        return Arrays.stream(CompactionType.values())
```
sarankk: Instead of creating a stream of values each time, can you compute it once and return it?

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraFactory.java
```diff
@@ -31,9 +31,9 @@
 import org.apache.cassandra.sidecar.db.schema.TableSchemaFetcher;
 
 /**
- * Factory to produce the 4.0 adapter
+ * Factory to produce the Cassandra 5.x adapter
  */
-@MinimumVersion("4.0.0")
+@MinimumVersion("5.0.0")
```
sarankk: `adapters-base` should still support 4.0.0 version, can you revert changes in this class? We will use the `Cassandra50Factory` you added for 5.0 version

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CompactionType.java
```diff
@@ -0,0 +1,78 @@
...
+        {
+            return null;
+        }
+        try
+        {
+            return valueOf(name.trim().toUpperCase(Locale.ROOT));
+        }
+        catch (IllegalArgumentException unknownEnum)
+        {
+            String validTypes = String.join(", ",
```
sarankk: Nit: create this list once

adapters/adapters-base/src/test/java/org/apache/cassandra/sidecar/adapters/base/CassandraCompactionManagerOperationsTest.java
```diff
@@ -0,0 +1,115 @@
...
+        // Test trim does not result in empty string
+        String compactionId = "  abc-123  ";
+        compactionManagerOperations.stopCompactionById(compactionId);
+
+        verify(mockJmxOperations, times(1)).stopCompactionById(compactionId);
+        verify(mockJmxOperations, times(0)).stopCompaction(org.mockito.ArgumentMatchers.anyString());
+    }
+
+    @Test
+    void testStopCompactionAllSupportedTypes()
```
sarankk, line 90: Can you add a test case for unsupported compaction type?

adapters/adapters-cassandra41/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra41/Cassandra41Adapter.java
```diff
@@ -31,7 +32,7 @@
 import org.jetbrains.annotations.NotNull;
 
 /**
- * A {@link ICassandraAdapter} implementation for Cassandra 4.1 and later
+ * A {@link ICassandraAdapter} implementation for Cassandra 4.x
```
sarankk: Supported compaction types list seem to be same between 4.0 and 4.1. We need not maintain `CompactionType` separately for 4.1, it can reuse 4.0's types. Can you revert changes in this class too?

adapters/adapters-cassandra41/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra41/Cassandra41CompactionManagerOperations.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.stream.Collectors;
+
+import org.apache.cassandra.sidecar.adapters.base.CassandraCompactionManagerOperations;
+import org.apache.cassandra.sidecar.common.server.JmxClient;
+
+/**
+ * Cassandra 4.1 implementation of CompactionManagerOperations that excludes MAJOR_COMPACTION
+ * which is not supported in Cassandra 4.x versions
+ */
+public class Cassandra41CompactionManagerOperations extends CassandraCompactionManagerOperations
```
sarankk: Related to earlier comment, this class is not required.

adapters/adapters-cassandra41/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra41/Cassandra41Factory.java
```diff
@@ -31,9 +31,9 @@
 import org.apache.cassandra.sidecar.db.schema.TableSchemaFetcher;
 
 /**
- * Factory to produce the 4.1 adapter
+ * Factory to produce the Cassandra 4.x adapter
  */
-@MinimumVersion("4.1.0")
+@MinimumVersion("4.0.0")
```
sarankk: We should keep the 4.1 factory as is. Can you revert changes here.

adapters/adapters-cassandra41/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra41/CompactionType.java
sarankk: Supported compaction types seem to be same for 4.1 and 4.0, we can reuse 4.0's types instead of maintaining a separate list. Can you remove changes in 4.1 adpator classes?

adapters/adapters-cassandra50/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra50/Cassandra50Adapter.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.cassandra.sidecar.common.server.ICassandraAdapter;
+import org.apache.cassandra.sidecar.common.server.JmxClient;
+import org.apache.cassandra.sidecar.common.server.StorageOperations;
+import org.apache.cassandra.sidecar.common.server.dns.DnsResolver;
+import org.apache.cassandra.sidecar.common.server.utils.DriverUtils;
+import org.apache.cassandra.sidecar.db.schema.TableSchemaFetcher;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A {@link ICassandraAdapter} implementation for Cassandra 4.x
```
sarankk: Can you update comment to say correct version?

adapters/adapters-cassandra50/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra50/Cassandra50CompactionManagerOperations.java
```diff
@@ -0,0 +1,56 @@
...
+        super(jmxClient);
+    }
+
+    /**
+     * {@inheritDoc}
+     *
+     * Cassandra 5.x supports MAJOR_COMPACTION operation type
+     */
+    @Override
+    public List<String> supportedCompactionTypes()
```
sarankk, line 54: 👍

adapters/adapters-cassandra50/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra50/Cassandra50StorageOperations.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.cassandra.sidecar.common.server.dns.DnsResolver;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import static org.apache.cassandra.sidecar.adapters.base.jmx.StorageJmxOperations.STORAGE_SERVICE_OBJ_NAME;
+
+/**
+ * An implementation of the {@link StorageOperations} that interfaces with Cassandra 5.0 and later
+ */
+public class Cassandra50StorageOperations extends CassandraStorageOperations
```
sarankk: can you extend from `Cassandra41StorageOperations`. You need not override the 2 methods if we extend from 4.1

adapters/adapters-cassandra50/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra50/CompactionType.java
```diff
@@ -0,0 +1,79 @@
...
+    VALIDATION,
+    INDEX_BUILD,
+    VIEW_BUILD,
+    COMPACTION,
+    TOMBSTONE_COMPACTION,
+    KEY_CACHE_SAVE,
+    ROW_CACHE_SAVE,
+    COUNTER_CACHE_SAVE,
+    INDEX_SUMMARY,
+    MAJOR_COMPACTION;
```
sarankk, line 48: 👍

adapters/adapters-cassandra50/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra50/CompactionType.java
```diff
@@ -0,0 +1,79 @@
...
+        {
+            return null;
+        }
+        try
+        {
+            return valueOf(name.trim().toUpperCase(Locale.ROOT));
+        }
+        catch (IllegalArgumentException unknownEnum)
+        {
+            String validTypes = String.join(", ",
```
sarankk: Nit: compute this once.

adapters/adapters-cassandra50/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra50/Cassandra50CompactionManagerOperations.java
```diff
@@ -0,0 +1,56 @@
...
+    {
+        return Arrays.stream(CompactionType.values())
```
sarankk: Nit: compute this once

server-common/src/main/java/org/apache/cassandra/sidecar/common/server/CompactionManagerOperations.java
```diff
@@ -32,4 +32,23 @@ public interface CompactionManagerOperations
...
+
+    /**
+     * Stops compaction based on type if no compaction ID is provided.
+     *
+     * @param compactionType   the compaction ID to stop (nullable)
+     * @throws IllegalArgumentException if both parameters are null or empty, or of the provided compactionType is unsupported
+     */
+    void stopCompaction(String compactionType);
+
+    List<String> supportedCompactionTypes();
```
sarankk, line 54: Can you add java doc for this method too to add a note on how types can differ between versions

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/CompactionStopResponse.java
```diff
@@ -0,0 +1,169 @@
...
+import org.apache.cassandra.sidecar.common.DataObjectBuilder;
+import org.apache.cassandra.sidecar.common.data.CompactionStopStatus;
+
+/**
+ * Response class for the Compaction Stop API
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class CompactionStopResponse
+{
+    public static final String COMPACTION_TYPE_KEY = "compactionType";
```
sarankk, line 33: Nit: can you use these String constants in request payload class too?

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
```diff
@@ -0,0 +1,175 @@
...
+    public void handleInternal(RoutingContext context,
+                               HttpServerRequest httpRequest,
+                               @NotNull String host,
+                               SocketAddress remoteAddress,
+                               CompactionStopRequestPayload request)
+    {
+        CompactionManagerOperations compactionManagerOps = metadataFetcher.delegate(host).compactionManagerOperations();
+
+        executorPools.service()
+                .executeBlocking(() -> stopCompaction(compactionManagerOps, request))
```
sarankk: Nit: incorrect alignment

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
```diff
@@ -0,0 +1,175 @@
...
+    {
+        String compactionType = request.compactionType();
+        String compactionId = request.compactionId();
+
+        // Validate compaction type supported by this Cassandra version
+        if (compactionType != null && !operations.supportedCompactionTypes().contains(compactionType))
+        {
+            throw new IllegalArgumentException(
+                String.format("Compaction type '%s' not supported in this Cassandra version. " +
+                             "Supported types: %s",
```
sarankk: Nit: incorrect alignment

server/src/main/java/org/apache/cassandra/sidecar/handlers/CompactionStopHandler.java
```diff
@@ -0,0 +1,175 @@
...
+     * @return CompactionStopResponse with the operation result
+     */
+    private CompactionStopResponse stopCompaction(CompactionManagerOperations operations,
+                                                  CompactionStopRequestPayload request)
+    {
+        String compactionType = request.compactionType();
+        String compactionId = request.compactionId();
+
+        // Validate compaction type supported by this Cassandra version
+        if (compactionType != null && !operations.supportedCompactionTypes().contains(compactionType))
```
sarankk: Since we are performing this check in stopCompaction call, we need not repeat it here again

sarankk (commented):
Changes look great to me, left final set of comments.

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraFactory.java
```diff
@@ -42,9 +41,9 @@ public class CassandraFactory implements ICassandraFactory
 
     public CassandraFactory(DnsResolver dnsResolver, DriverUtils driverUtils, TableSchemaFetcher tableSchemaFetcher)
     {
-        this.dnsResolver = Objects.requireNonNull(dnsResolver, "dnsResolver is required");
```
sarankk, line 45: Can you revert these changes?

adapters/adapters-cassandra41/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra41/Cassandra41Factory.java
```diff
@@ -42,9 +41,9 @@ public class Cassandra41Factory implements ICassandraFactory
 
     public Cassandra41Factory(DnsResolver dnsResolver, DriverUtils driverUtils, TableSchemaFetcher tableSchemaFetcher)
     {
-        this.dnsResolver = Objects.requireNonNull(dnsResolver, "dnsResolver is required");
```
sarankk, line 45: Can you revert these changes, they seem to be unrelated to PR

adapters/adapters-cassandra41/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra41/Cassandra41Adapter.java
```diff
@@ -50,7 +50,7 @@ public Cassandra41Adapter(DnsResolver dnsResolver,
      */
     @Override
     @NotNull
-    protected StorageOperations createStorageOperations(DnsResolver dnsResolver, JmxClient jmxClient)
+    public StorageOperations storageOperations()
```
sarankk: Can you revert this change too

sarankk (approved):
+1 Thanks Shalni for great work.

PR #273, opened 2025-11-03
CASSSIDECAR-361: Fast Cassandra Input Validator

frankgh:
Patch by Francisco Guerrero; reviewed by TBD for CASSSIDECAR-361

### DIFF
--- a/conf/sidecar.yaml
+++ b/conf/sidecar.yaml
@@ -375,6 +375,23 @@ metrics:
 #      value: "vertx.eventbus.*"              # exclude all metrics starts with vertx.eventbus
 
 cassandra_input_validation:
+  validator:
+    # Implementation to use for the validation of Casandra inputs. Out of the box, Cassandra Sidecar provides
+    # org.apache.cassandra.sidecar.utils.{RegexBasedCassandraInputValidator, FastCassandraInputValidator}.
+    #
+    # - RegexBasedCassandraInputValidator default implementation the uses regular expressions to perform validations
+    # - FastCassandraInputValidator optimized implementation that does not use regular expressions. This implementation
+    #                               ignores the regular expressions configured in the cassandra_input_validation
+    #                               configuration.
+  - class_name: org.apache.cassandra.sidecar.utils.RegexBasedCassandraInputValidator
+    # Configuration parameters that are only applicable to the FastCassandraInputValidator implementation
+    #parameters:
+    #  # Comma-separated list of terminations allowed for the component name. By default, allowed component
+    #  # names can only end in .db, .cql, .json, .crc32, or TOC.txt
+    #  valid_terminations: ".db,.cql,.json,.crc32,TOC.txt"
+    #  # Comma-separated list of terminations allowed for the restricted component name. By default, allowed
+    #  # restricted component names con only end in .db or TOC.txt
+    #  valid_restricted_terminations: ".db,TOC.txt"
   forbidden_keyspaces:
     - system_schema
     - system_traces
@@ -386,8 +403,8 @@ cassandra_input_validation:
     - sidecar_internal
   allowed_chars_for_directory: "[a-zA-Z][a-zA-Z0-9_]{0,47}"
   allowed_chars_for_quoted_name: "[a-zA-Z_0-9]{1,48}"
-  allowed_chars_for_component_name: "[a-zA-Z0-9_-]+(.db|.cql|.json|.crc32|TOC.txt)"
-  allowed_chars_for_restricted_component_name: "[a-zA-Z0-9_-]+(.db|TOC.txt)"
+  allowed_chars_for_component_name: "[a-zA-Z0-9_-]+(\\.db|\\.cql|\\.json|\\.crc32|TOC\\.txt)"
+  allowed_chars_for_restricted_component_name: "[a-zA-Z0-9_-]+(\\.db|TOC\\.txt)"
 
 blob_restore:
   job_discovery_active_loop_delay: 5m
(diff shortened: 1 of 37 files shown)


conf/sidecar.yaml
```diff
@@ -386,8 +403,8 @@ cassandra_input_validation:
     - sidecar_internal
   allowed_chars_for_directory: "[a-zA-Z][a-zA-Z0-9_]{0,47}"
   allowed_chars_for_quoted_name: "[a-zA-Z_0-9]{1,48}"
-  allowed_chars_for_component_name: "[a-zA-Z0-9_-]+(.db|.cql|.json|.crc32|TOC.txt)"
-  allowed_chars_for_restricted_component_name: "[a-zA-Z0-9_-]+(.db|TOC.txt)"
+  allowed_chars_for_component_name: "[a-zA-Z0-9_-]+(\\.db|\\.cql|\\.json|\\.crc32|TOC\\.txt)"
```
frankgh, line 406: the regex was incorrect in the original implementation and the error surfaced with a unit test

docs/src/user.adoc
```diff
@@ -402,11 +402,14 @@ The `metrics` configuration defines how Cassandra Sidecar exposes metrics over J
 
 This section defines input validations for Cassandra keyspace and directory names which are used for SSTable imports. The following properties are defined:
 
+* `validator`: The implementation to use for the validation of Casandra inputs.
+** `class_name`: The name of the class implementing the CassandraInputValidator interface. Out of the box Cassandra Sidecar provides org.apache.cassandra.sidecar.utils.{RegexBasedCassandraInputValidator, FastCassandraInputValidator}.
+** `parameters`: Configuration parameters that are only applicable to the FastCassandraInputValidator implementation.
```
sarankk, line 407: Can you also add the new `FastCassandraInputValidator` based parameters `valid_terminations`, `valid_restricted_terminations` to docs?

server/src/test/resources/config/sidecar_validation_configuration.yaml
```diff
@@ -25,11 +25,16 @@ sidecar:
...
+      valid_restricted_terminations: ".xml"
   forbidden_keyspaces:
     - a
     - b
     - c
   allowed_chars_for_directory: "[a-z]+"
   allowed_chars_for_quoted_name: "[A-Z]+"
-  allowed_chars_for_component_name: "(.db|.cql|.json|.crc32|TOC.txt)"
-  allowed_chars_for_restricted_component_name: "(.db|TOC.txt)"
+  allowed_chars_for_component_name: "(\\.db|\\.cql|\\.json|\\.crc32|TOC\\.txt)"
```
sarankk, line 39: Nit: Remove regex validator parameters since they are not relevant.

server/src/main/java/org/apache/cassandra/sidecar/utils/FastCassandraInputValidator.java
```diff
@@ -0,0 +1,314 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public Name validateKeyspaceName(@NotNull String keyspace) throws NullPointerException
+    {
+        Name name = new Name(keyspace);
+        validateNameLength(name, "keyspace", KEYSPACE_NAME_LENGTH);
+        validateNamePattern(name, "keyspace");
+
+        if (validationConfiguration.forbiddenKeyspaces().contains(name.name()))
```
sarankk: Nit: Can we move forbidden check before length and pattern validation?

server/src/main/java/org/apache/cassandra/sidecar/utils/RegexBasedCassandraInputValidator.java
```diff
@@ -0,0 +1,192 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Name validateKeyspaceName(@NotNull String keyspace)
+    {
+        Name name = new Name(keyspace);
+        validateNamePattern(name, "keyspace");
+
+        if (validationConfiguration.forbiddenKeyspaces().contains(name.name()))
```
sarankk, line 64: Nit: can we move forbidden check before validating pattern

server/src/main/java/org/apache/cassandra/sidecar/utils/FastCassandraInputValidator.java
```diff
@@ -0,0 +1,314 @@
...
+     *
+     * @param componentName     the name of the component to validate
+     * @param validTerminations a list of valid terminations for the component name
+     */
+    protected void validateComponentName(String componentName, List<String> validTerminations)
+    {
+        char c;
+        int lastIndexOfAllowedTermination = lastIndexOfAllowedTermination(componentName, validTerminations);
+
+        if (lastIndexOfAllowedTermination < 1)
```
sarankk, line 223: Could we have a custom component in Cassandra with 1 letter SSTable format?

sarankk (commented):
+1 Looks good to me. Some comments

server/src/test/resources/config/sidecar_validation_configuration.yaml
frankgh: We still use these values to validate in unit tests. See https://github.com/frankgh/cassandra-sidecar/blob/CASSSIDECAR-361/server/src/test/java/org/apache/cassandra/sidecar/config/SidecarConfigurationTest.java#L109

server/src/main/java/org/apache/cassandra/sidecar/utils/FastCassandraInputValidator.java
frankgh: No, that would not be possible today. I don't think it's reasonable to have something like that either , so we can assume it will not happen in the future

sarankk (approved):

PR #277, opened 2025-11-07
CASSSIDECAR-340: Add default process-based lifecycle provider to support the lifecycle API

pauloricardomg:
# Changes summary

## Functionality

* `ProcessRuntimeConfiguration`: represents the runtime configuration of a cassandra process, like environment variables, cassandra_home, cassandra_conf, etc. defined in `sidecar.yaml`
* `ProcessLifecycleProvider.start`: constructs Cassandra runtime configuration from instance and global parameters and uses java `ProcessBuilder` API to start a Cassandra process writing a pidfile to the lifecycle state directory.
* `ProcessLifecycleProvider.stop`: uses the pidfile to fetch process handle and send SIGKILL to stop the Cassandra process gracefully.
* `ProcessLifecycleProvider.isRunning`: a Cassandra process is considered running if a PID file exists in the lifecycle state directory and a process with that PID is currently running.
* `InstanceMetadata` and `InstanceConfiguration`: expose `storageDir` and instance-specific `lifecycleOptions`

## Documentation

* Add `examples/lifecycle/README.md` with an example on starting/stopping cassandra via lifecycle API
* Update default sidecar.yaml with example and comments on how to enable `ProcessLifecycleProvider`

## Infrastructure

* Update `build.gradle`, `integrationTestTask.gradle` and `gradle.properties` to support running  `ProcessLifecycleProvider` integration tests
* Add `scripts/build-cassandra-tarball.sh` and update scripts/build-dtest-jars.sh` to build tarballs (requirement for running integration tests) 
* Update `.circleci/config.yml` to download cassandra tarballs for `ProcessLifecycleProviderIntegrationTest` and run integration tests for each supported version (4.0/5.0/5.1)

## Testing

* Add unit tests for `ProcessLifecycleProvider` to check runtime configuration is being constructed correctly and basic workflow (start/stop/status) with `FakeLifecycleProvider`
* Add unit tests for `ProcessRuntimeConfiguration` to test start command construction and precondition validations
* `ProcessLifecycleProviderIntegrationTest`: configures a local Cassandra install from a tarball in a temporary directory and run tests from LifecycleProviderIntegrationTester (start/stop/status)
  * `TestFileUtils`: contains utilities to install the cassandra tarball for `ProcessLifecycleProviderIntegrationTest`


examples/lifecycle/setup.sh
```diff
@@ -0,0 +1,54 @@
...
+  wget -P ${TMP_DIR} ${TARBALL_URL}
+
+  echo "Extracting tarball"
+  tar -xvzf ${TMP_DIR}/${TARBALL_NAME} -C $(dirname $CASSANDRA_HOME)
+fi
+
+echo "Creating configuration"
+cp -r ${CASSANDRA_HOME}/conf/* ${CASSANDRA_CONF}
+sed "s#\$cassandraHome#${CASSANDRA_HOME}#g" ${SIDECAR_YAML_TEMPLATE} > ${SIDECAR_YAML}
+sed -i "s#\$baseDir#${NODE_DIR}#g" ${SIDECAR_YAML}
```
arjunashok: This was complaining on a Mac. The format that worked was - 

```
sed -i '' "s#\$baseDir#${NODE_DIR}#g" ${SIDECAR_YAML}
```

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,232 @@
...
+
+    private void startCassandra(InstanceMetadata instance)
+    {
+        ProcessRuntimeConfiguration runtimeConfig = getRuntimeConfiguration(instance);
+        try
+        {
+            String stdoutLocation = getStdoutLocation(runtimeConfig.instanceName());
+            String stderrLocation = getStderrLocation(runtimeConfig.instanceName());
+            String pidFileLocation = getPidFileLocation(runtimeConfig.instanceName());
+            ProcessBuilder processBuilder = runtimeConfig.buildStartCommand(pidFileLocation,
```
arjunashok, line 152: Seems like the additional files created here  - stdout/stderr - are appended-to without rotation/cleanup and could result in disk space issues. 

Also, there is a possibility of stale PID files if the process terminates abnormally. We would need some cleanup mechanism for such cases.

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,232 @@
...
+                                                                            stdoutLocation,
+                                                                            stderrLocation);
+            LOG.info("Starting Cassandra instance {} with command: {}", runtimeConfig.instanceName(), processBuilder.command());
+
+            Process process = processBuilder.start();
+            process.waitFor(CASSANDRA_PROCESS_TIMEOUT_MS, TimeUnit.MILLISECONDS); // blocking call, make async?
```
arjunashok: As you have called out, consider using some sort of a periodic check for process, Instead of blocking the thread. Likewise for the stop process below.

```suggestion
        vertx.setPeriodic(1000, timerId -> {
            if (isCassandraProcessRunning(instance)) {
                vertx.cancelTimer(timerId);
                LOG.info("Started Cassandra instance {} with PID {}", runtimeConfig.instanceName(), 
                         readPidFromFile(Path.of(pidFileLocation)));
            } else if (System.currentTimeMillis() > startTime + CASSANDRA_PROCESS_TIMEOUT_MS) {
                vertx.cancelTimer(timerId);
                LOG.error("Failed to start Cassandra instance {} within timeout", runtimeConfig.instanceName());
                // Handle timeout error
            }
        });
```

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,232 @@
...
+        try
+        {
+            String pidFileLocation = getPidFileLocation(casCfg.instanceName());
+            Long pid = readPidFromFile(Path.of(pidFileLocation));
+            Optional<ProcessHandle> processHandle = ProcessHandle.of(pid);
+            if (processHandle.isPresent())
+            {
+                LOG.info("Stopping process of Cassandra instance {} with PID {}.", casCfg.instanceName(), pid);
+                CompletableFuture<ProcessHandle> terminationFuture = processHandle.get().onExit();
+                processHandle.get().destroy();  // blocking call, make async?
```
arjunashok: Consider a force destroy `destroyForcibly()` on timeout with appropriate logging.

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,232 @@
...
+                terminationFuture.get(CASSANDRA_PROCESS_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            }
+            else
+            {
+                LOG.warn("No process running for Cassandra instance {} with PID {}.", casCfg.instanceName, pid);
```
arjunashok: Call the method for or consistency.
```suggestion
                LOG.warn("No process running for Cassandra instance {} with PID {}.", casCfg.instanceName(), pid);
```

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,232 @@
...
+            {
+                defaultJvmProperties.put(entry.getKey(), entry.getValue());
+            }
+            else if (entry.getKey().startsWith("env."))
+            {
+                defaultEnvVars.put(entry.getKey().replaceAll("env.", ""), entry.getValue());
+            }
+        }
+        this.lifecycleDir = params.get(OPT_STATE_DIR);
+        this.defaultCassandraHome = params.get(OPT_CASSANDRA_HOME);
```
arjunashok, line 78: Should we fail-fast if these directories do not exist or do not have appropriate permissions?

arjunashok (commented):
Added some comments. Also, please add an entry for these changes into `Changes.txt`.

examples/lifecycle/setup.sh
pauloricardomg: Good catch, fixed on 9d08a6ddfdb3622f338c052edf60b6cbf494b900

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
pauloricardomg: > Seems like the additional files created here - stdout/stderr - are appended-to without rotation/cleanup and could result in disk space issues.

These files are just startup logs, not the entire Cassandra logs which are logged to `$CASSANDRA_LOG_DIR` with proper rotation. In order to not grow the scope of this PR while keeping the file sizes bounded I updated this patch to overwrite the pervious startup logs instead of appending indefinitely on c1eb0117e767d4ceef63fea8b462c81678c882fa. If needed we can add append+rotation support later.

> Also, there is a possibility of stale PID files if the process terminates abnormally. We would need some cleanup mechanism for such cases.

Good call. I've enhanced `ProcessLifecycleProvider.isCassandraProcessRunning` to auto-remove the PID file when the PID file exists but no Cassandra process is running, as well ensure the PID file is removed on `ProcessLifecycleProvider.stop` (0b65e3e23780126ad759ecd2bfa852e4d1630666).

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
pauloricardomg: My understanding is that `vertx.setPeriodic` will return immediatelly right ? In which case, `LifecycleProvider.start` could return before the Cassandra process is initialized.

The issue is that [LifecycleManager.submitStartTask](https://github.com/pauloricardomg/cassandra-sidecar/blob/CASSSIDECAR-340/server/src/main/java/org/apache/cassandra/sidecar/lifecycle/LifecycleManager.java#L146-L147) assumes `LifecycleProvider.start` is synchronous and will only return after the process is guaranteed to be started. Are you suggesting making `LifecycleProvider.start` asynchronous and return a `Future` instead? Since this is a departure from the current synchronous design and may require many changes I would prefer to do this in a follow-up ticket if that works for you.

arjunashok: Yep, valid point about the architectural impact of going fully async. Follow-up ticket makes sense.

I was hoping to move away from blocking the worker thread  from waiting on a kernel system call, towards a Java synchronization primitive (like CountDownLatch), which is more efficient and interruptible.  Though I recognize this would be an intermediate step - if we're planning to eventually adopt an async interface. We'd need to weigh whether this intermediate improvement justifies the additional churn.

Additionally `process.waitFor(timeout)` returns `false` when it times out, but we don't check the return value. Wondering if it is worth detecting the status of the process start for better failure handling.

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,313 @@
...
+    */
+    protected static Optional<String> getCommandLinePlatformIndependent(ProcessHandle processHandle)
+    {
+        long pid = processHandle.pid();
+        try
+        {
+            ProcessBuilder pb = new ProcessBuilder("ps", "-p", String.valueOf(pid), "-o", "args=");
+            Process proc = pb.start();
+            try (BufferedReader reader = new BufferedReader(
+                    new InputStreamReader(proc.getInputStream())))
```
arjunashok: To address the spotbugs error
```suggestion
                    new InputStreamReader(proc.getInputStream(), StandardCharsets.UTF_8)))
```

arjunashok (commented):
Thanks for addressing the comments. Looks good. Added a comment to fix a spot bugs failure. 

Please add an entry to `Changes.txt`, and share a link to the CI build.

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
pauloricardomg: done on [56ca1f6](https://github.com/apache/cassandra-sidecar/pull/277/commits/56ca1f6c985f5e198d23f28ea9611da21c9d6167)

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
pauloricardomg: fixed on [369363a](https://github.com/apache/cassandra-sidecar/pull/277/commits/369363a4146d5c8d581b1cbd58f40f5e0c51c193)

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
pauloricardomg: updated on [b643899](https://github.com/apache/cassandra-sidecar/pull/277/commits/b643899ce6968dc05b0cb9fcb7ea51c1bd54021a)

pauloricardomg:
Thanks for the review @arjunashok - updated github actions to support new intergration tests for lifecycle provider. All the checks have passed.

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/LifecycleManager.java
```diff
@@ -148,7 +148,7 @@ private Future<Void> submitStartTask(String instanceId)
             }
             catch (Exception e)
             {
-                LOG.error("Failed to start instance {}: {}", instanceId, e.getMessage());
+                LOG.error("Failed to start instance {}: {}", instanceId, e.getMessage(), e);
```
frankgh: NIT:
```suggestion
                LOG.error("Failed to start instance {}", instanceId, e);
```

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/LifecycleManager.java
```diff
@@ -166,7 +166,7 @@ private Future<Void> submitStopTask(String instanceId)
             }
             catch (Exception e)
             {
-                LOG.error("Failed to stop instance {}: {}", instanceId, e.getMessage());
+                LOG.error("Failed to stop instance {}: {}", instanceId, e.getMessage(), e);
```
frankgh: NIT:
```suggestion
                LOG.error("Failed to stop instance {}", instanceId, e);
```

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessRuntimeConfiguration.java
```diff
@@ -0,0 +1,237 @@
...
+        // Redirect output to logs
+        processBuilder.redirectOutput(ProcessBuilder.Redirect.to(new File(stdoutFileLocation)));
+        processBuilder.redirectError(ProcessBuilder.Redirect.to(new File(stderrFileLocation)));
+
+        // Set working directory
+        processBuilder.directory(cassandraHome().toFile());
+        return processBuilder;
+    }
+
+    public String toString()
```
frankgh, line 168: NIT (style) : add override annotation
```suggestion
    @Override
    public String toString()
```

server/src/main/java/org/apache/cassandra/sidecar/config/InstanceConfiguration.java
```diff
@@ -104,4 +105,9 @@ public interface InstanceConfiguration
      * @return the password for the JMX role for the JMX service for the Cassandra instance
      */
     String jmxRolePassword();
+
+    /**
+     * @return The lifecycle options for this Cassandra instance
+     */
+    Map<String, String> lifecycleOptions();
```
frankgh, line 112: any reason for having this as a generic map, instead of using strongly typed convention?

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
```diff
@@ -136,6 +145,12 @@ public String stagingDir()
         return stagingDir;
     }
 
+    @Override
+    public String storageDir()
+    {
+        return  storageDir;
```
frankgh: NIT (formatting)
```suggestion
        return storageDir;
```

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
```diff
@@ -188,6 +203,16 @@ public InstanceMetrics metrics()
         return metrics;
     }
 
+    /**
+     * @return The lifecycle options for this Cassandra instance
+     */
+    @Override
+    @JsonProperty("lifecycle_options")
```
frankgh: This class should not have the json property annotation. This class gets built from the configuration classes in https://github.com/apache/cassandra-sidecar/pull/277/files#diff-4de249cbcebfa53deecf4ac89d82a9419b32aea0a418fa56b5ed3b56b104a688R253
```suggestion
    @NonNull
```

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
```diff
@@ -23,10 +23,12 @@
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
 
 import com.codahale.metrics.MetricRegistry;
+import com.fasterxml.jackson.annotation.JsonProperty;
```
frankgh:
```suggestion
```

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
```diff
@@ -89,6 +94,10 @@ protected InstanceMetadataImpl(Builder builder)
         hintsDir = builder.resolveHintsDir();
         savedCachesDir = builder.resolveSavedCachesDir();
         localSystemDataFileDir = FileUtils.maybeResolveHomeDirectory(builder.localSystemDataFileDir);
+        lifecycleOptions = builder.lifecycleOptions != null
+                                 ? Collections.unmodifiableMap(builder.lifecycleOptions)
+                                 : Collections.emptyMap();
```
frankgh: NIT (formatting)
```suggestion
        lifecycleOptions = builder.lifecycleOptions != null
                           ? Collections.unmodifiableMap(builder.lifecycleOptions)
                           : Collections.emptyMap();
```

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,354 @@
...
+ * Manage the lifecycle of Cassandra instances running on local processes
  */
 public class ProcessLifecycleProvider implements LifecycleProvider
 {
+    static final String OPT_CASSANDRA_HOME = "cassandra_home";
+    static final String OPT_CASSANDRA_CONF_DIR = "cassandra_conf_dir";
+    static final String OPT_CASSANDRA_LOG_DIR = "cassandra_log_dir";
+    static final String OPT_STATE_DIR = "state_dir";
+
+    protected static final Logger LOG = LoggerFactory.getLogger(ProcessLifecycleProvider.class);
```
frankgh: Let's keep the same naming across the project for consistency 
```suggestion
    protected static final Logger LOGGER = LoggerFactory.getLogger(ProcessLifecycleProvider.class);
```

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,354 @@
...
+    public static final long CASSANDRA_PROCESS_TIMEOUT_MS = Long.getLong("cassandra.sidecar.lifecycle.process.timeout.ms", 120_000L);
```
frankgh: Any reason why this is configured as a system property instead of having it configured in the sidecar.yaml file?

conf/sidecar.yaml
```diff
@@ -455,3 +458,15 @@ live_migration:
...
   max_concurrent_downloads: 20            # Maximum number of concurrent downloads allowed
+
+# Configuration to allow sidecar start and stop Cassandra instances via the lifecycle API (disabled by default)
+lifecycle:
+  enabled: false
+  provider:
+    class_name: org.apache.cassandra.sidecar.lifecycle.ProcessLifecycleProvider
+    parameters:
+      state_dir: /var/lib/cassandra-sidecar/lifecycle # The directory where the process state is stored
+      cassandra_home: /opt/cassandra # The default Cassandra installation directory
```
frankgh, line 473: similar to how we configure environment variables can we also use a prefix for java system properties. This would allow for configuring system properties that apply to the Cassandra process but are not Cassandra-core specific, for example a JMX property could be configured like this: `sys.com.sun.management.jmxremote.authenticate: true`.

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,354 @@
...
+
+    private final String lifecycleDir;
+    private final String defaultCassandraHome;
+    private final Map<String, String> defaultJvmProperties = new HashMap<>();
+    private final Map<String, String> defaultEnvVars = new HashMap<>();
+
     public ProcessLifecycleProvider(Map<String, String> params)
     {
-        // Params unused for now
+        // Extract any JVM properties or environment variables from the params
```
frankgh, line 65: can we annotate `params` in the constructor parameter with `@NotNull`?

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,354 @@
...
+        for (Map.Entry<String, String> entry : params.entrySet())
+        {
+            if (entry.getKey().startsWith("cassandra."))
+            {
+                defaultJvmProperties.put(entry.getKey(), entry.getValue());
+            }
+            else if (entry.getKey().startsWith("env."))
+            {
+                defaultEnvVars.put(entry.getKey().replaceAll("env.", ""), entry.getValue());
```
frankgh: can we use replace first instead here?
```suggestion
                defaultEnvVars.put(entry.getKey().replaceFirst("^env\\.", ""), entry.getValue());
```

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,354 @@
...
+        {
+            throw new RuntimeException("Failed to stop process for Cassandra instance " + casCfg.instanceName() + " due to " + t.getMessage(), t);
+        }
+    }
+
+    @VisibleForTesting
+    protected ProcessRuntimeConfiguration getRuntimeConfiguration(InstanceMetadata instance)
+    {
+        String cassandraHome = Optional.ofNullable(instance.lifecycleOptions().get(OPT_CASSANDRA_HOME))
+                                       .orElse(defaultCassandraHome);
```
frankgh: NIT, simplify:
```suggestion
        String cassandraHome = instance.lifecycleOptions().getOrDefault(OPT_CASSANDRA_HOME, defaultCassandraHome);
```

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,354 @@
...
+     */
+    private boolean isCassandraProcessRunning(InstanceMetadata instance)
+    {
+        Path pidFilePath = Path.of(getPidFileLocation(instance.host()));
+        try
+        {
+            // Case 1: PID file does not exist or is not readable
+            if (!Files.isRegularFile(pidFilePath) || !Files.isReadable(pidFilePath))
+            {
+                LOG.debug("PID file does not exist or is not readable for instance {} at path {}", instance.host(), pidFilePath);
```
frankgh: We should rely on the instance's toString() implementation for the log entries here:
```suggestion
                LOG.debug("PID file does not exist or is not readable for instance {} at path {}", instance, pidFilePath);
```

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
```diff
@@ -18,38 +18,354 @@
...
+     * 3. Verifies the process is a Cassandra instance by checking for
+     *    org.apache.cassandra.service.CassandraDaemon in the command line. Returns true if
+     *    the command line contains the Cassandra daemon class or cannot be determined.
+     * 4. If the process is running but is not a Cassandra process, returns false and deletes
+     *    the stale PID file.
+     *
+     * @param instance the instance metadata containing host information
+     * @return true if the instance is running as a Cassandra process, false otherwise
+     */
+    private boolean isCassandraProcessRunning(InstanceMetadata instance)
```
frankgh, line 263: can we debug at warn level here instead of debug? Any reason why we debug was used?

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessRuntimeConfiguration.java
```diff
@@ -0,0 +1,237 @@
...
+    {
+        return cassandraHome.resolve(Path.of("bin", "stop-server"));
+    }
+
+    private Path cassandraYaml()
+    {
+        return cassandraConfDir.resolve(Path.of("cassandra.yaml"));
+    }
+
+    public void validateStart() throws IllegalArgumentException
```
frankgh: should we throw a ConfigurationException here instead?

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessRuntimeConfiguration.java
```diff
@@ -0,0 +1,237 @@
...
+    {
+        return cassandraConfDir.toString();
+    }
+
+    public Path cassandraBin()
+    {
+        return cassandraHome.resolve(Path.of("bin", "cassandra"));
+    }
+
+    public Path stopServerBin()
```
frankgh: this is unused, do we need it? Looking at the implementation of the script, it looks like it only prints information and it doesn't actually do anything. So maybe remove this since it won't be used

arjunashok (approved):
LGTM, +1. Thanks for addressing the comments.

Couple of nits:
- Consider addressing the comment [here](https://github.com/apache/cassandra-sidecar/pull/277/changes#r2618206454) to handle timeout return value.
- Add entry to `Changes.txt`

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessRuntimeConfiguration.java
```diff
@@ -0,0 +1,237 @@
...
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Represents the configuration for a Cassandra process instance.
+ */
+public class ProcessRuntimeConfiguration
+{
+    static class Builder
```
frankgh: I think we should stick to the existing Builder  pattern that we have in the repository implementing `DataObjectBuilder`. Also, the `with` pattern in the builder is also diverging from the existing codebase.

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
frankgh: Looking at where this code is running , it seems like we always are calling this in a blocking thread. So I think it's fine to keep the code as is, under the assumption that the entry point for the call of this code will be run in a blocking thread.

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
frankgh: I agree with Paulo here, I think it's fine to keep a single stdout / stderr file for the startup of the process. I'd expect minimal log entries in these files.

.circleci/config.yml
```diff
@@ -150,8 +154,10 @@ jobs:
       - checkout
       - attach_workspace:
           at: dtest-jars
+      - attach_workspace:
+          at: cassandra-tarballs
       - run: ./scripts/install-shaded-dtest-jar-local.sh
-      - run: ./gradlew --no-daemon -PdtestVersion=4.0.16 -Dcassandra.sidecar.versions_to_test="4.0" integrationTestHeavyWeight --stacktrace
+      - run: ./gradlew --no-daemon -PdtestVersion=4.0.16 -PtarballVersion=4.0 -Dcassandra.sidecar.versions_to_test="4.0" jar integrationTestHeavyWeight --stacktrace
```
frankgh: should we reuse `cassandra.sidecar.versions_to_test` for the `tarballVersion`? will it ever differ?

frankgh, line 160: just out of curiosity why do we need to execute `jar` before running the integration tests?

examples/lifecycle/setup.sh
```diff
@@ -0,0 +1,54 @@
...
+
+echo "Creating directories"
+mkdir -p ${CASSANDRA_HOME} ${CASSANDRA_LOG_DIR} ${CASSANDRA_CONF} ${CASSANDRA_STORAGE_DIR} ${SIDECAR_LIFECYCLE_DIR} ${TMP_DIR}
+
+if [ -f ${CASSANDRA_HOME}/bin/cassandra ]; then
+  echo "Cassandra already installed at ${CASSANDRA_HOME}, skipping install"
+else
+  echo "Installing Cassandra at ${CASSANDRA_HOME}"
+  echo "Downloading ${TARBALL_URL}"
+  wget -P ${TMP_DIR} ${TARBALL_URL}
```
frankgh: can we use curl here instead? I think curl is more widely available than wget
```suggestion
  curl -L -o ${TMP_DIR}/$(basename ${TARBALL_URL}) ${TARBALL_URL}
```

examples/lifecycle/setup.sh
```diff
@@ -0,0 +1,54 @@
...
+
+  echo "Extracting tarball"
+  tar -xvzf ${TMP_DIR}/${TARBALL_NAME} -C $(dirname $CASSANDRA_HOME)
+fi
+
+echo "Creating configuration"
```
frankgh: NIT:
```suggestion
echo "Creating Sidecar configuration"
```

examples/lifecycle/setup.sh
```diff
@@ -0,0 +1,54 @@
...
+mkdir -p ${CASSANDRA_HOME} ${CASSANDRA_LOG_DIR} ${CASSANDRA_CONF} ${CASSANDRA_STORAGE_DIR} ${SIDECAR_LIFECYCLE_DIR} ${TMP_DIR}
+
+if [ -f ${CASSANDRA_HOME}/bin/cassandra ]; then
+  echo "Cassandra already installed at ${CASSANDRA_HOME}, skipping install"
+else
+  echo "Installing Cassandra at ${CASSANDRA_HOME}"
+  echo "Downloading ${TARBALL_URL}"
+  wget -P ${TMP_DIR} ${TARBALL_URL}
+
+  echo "Extracting tarball"
```
frankgh: NIT
```suggestion
  echo "Extracting Cassandra tarball"
```

examples/lifecycle/setup.sh
```diff
@@ -0,0 +1,54 @@
...
+  tar -xvzf ${TMP_DIR}/${TARBALL_NAME} -C $(dirname $CASSANDRA_HOME)
+fi
+
+echo "Creating configuration"
+cp -r ${CASSANDRA_HOME}/conf/* ${CASSANDRA_CONF}
+sed "s#\$cassandraHome#${CASSANDRA_HOME}#g" ${SIDECAR_YAML_TEMPLATE} > ${SIDECAR_YAML}
+sed -i '' "s#\$baseDir#${NODE_DIR}#g" ${SIDECAR_YAML}
```
frankgh, line 54: NIT add message when setup is complete:
```suggestion
sed -i '' "s#\$baseDir#${NODE_DIR}#g" ${SIDECAR_YAML}

echo "Setup complete!"
```

frankgh (commented):
Looks good! I have added some comments. I also have [this commit](https://github.com/frankgh/cassandra-sidecar/commit/5c45cba3ce0aa54d7f9cf7723a8e5d90fc304c23#r173996368) with some suggestions. Some of the suggestions in the commit address the comments I left in the PR, but there are other comments that are not addressed.

.circleci/config.yml
pauloricardomg: This is needed because the test `Server` requires the file `server-common/src/main/resources/sidecar.version` to be initialized and this is done by the jar task, otherwise the test [fails ](https://app.circleci.com/pipelines/circleci/KaZH6B8gKQFCZv6WTtM3yk/7N9zyX9UjY83jRQygcQ3Mf/32/workflows/45bbabf6-1bf8-4853-9225-ef0e1873f09d/jobs/433) with:

```java
ProcessLifecycleProviderIntegrationTest > initializationError FAILED
    com.google.inject.ProvisionException at ProcessLifecycleProviderIntegrationTest.java:144
        Caused by: java.lang.IllegalStateException at ProcessLifecycleProviderIntegrationTest.java:144
            Caused by: java.lang.NullPointerException at ProcessLifecycleProviderIntegrationTest.java:144
```

conf/sidecar.yaml
pauloricardomg: good call, I see this is already included in the reviews [commit](https://github.com/pauloricardomg/cassandra-sidecar/commit/75c614da8888ff37ebf0a8c2ace98bd12e3a50f7)

server/src/main/java/org/apache/cassandra/sidecar/config/InstanceConfiguration.java
pauloricardomg: This is to allow supporting custom instance options for downstream providers. For example, a HTTP-based lifecycle provider could have a different control endpoint per instance that would be specified as a lifecycle option, but this option would not make sense in a process lifecycle or docker provider. This would not be possible if we have a strongly typed object.

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessLifecycleProvider.java
pauloricardomg: no reason, extracted to configuration on 60c6d47d15581d36ecb785865a0041aecc71ba47

pauloricardomg:
handled timeout on 2d61c77b and extract process timeout to configuration 60c6d47

frankgh (approved):
+1 thanks for this contribution!

PR #294, opened 2025-12-01
CASSSIDECAR-243: Implementation of CDCPublisher

bbotella:
Patch by James Berragan, Jyothsna Konisa, Bernardo Botella; reviewed by   for CASSSIDECAR-243

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
@@ -76,12 +76,6 @@ public class CachingSchemaStore implements SchemaStore
         this.cdcConfig = cdcConfig;
         this.sidecarCdcStats = sidecarCdcStats;
 
-        ThrowingRunnable configChangeCallback = () -> {
-            LOGGER.info("Services configuration changed. Reloading publisher...");
-            loadPublisher();
-            publishSchemas();
-        };
-        this.cdcConfig.registerConfigChangeListener(configChangeCallback);
         configureSidecarServerEventListeners();
     }
 
--- a/server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
@@ -45,7 +45,6 @@ import org.apache.cassandra.cdc.kafka.KafkaOptions;
 import org.apache.cassandra.cdc.schemastore.SchemaStore;
 import org.apache.cassandra.cdc.schemastore.SchemaStorePublisherFactory;
 import org.apache.cassandra.cdc.schemastore.TableSchemaPublisher;
-import org.apache.cassandra.sidecar.common.server.ThrowingRunnable;
 import org.apache.cassandra.sidecar.db.TableHistoryDatabaseAccessor;
 import org.apache.cassandra.sidecar.db.schema.SidecarSchema;
 import org.apache.cassandra.sidecar.tasks.CassandraClusterSchemaMonitor;
--- a/server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
@@ -85,7 +85,6 @@ public class CachingSchemaStore implements SchemaStore
                        SidecarSchema sidecarSchema,
                        CqlToAvroSchemaConverter cqlToAvroSchemaConverter)
     {
-        super();
         this.cassandraClusterSchemaMonitor = cassandraClusterSchemaMonitor;
         this.tableHistoryDatabaseAccessor = tableHistoryDatabaseAccessor;
         this.sidecarSchema = sidecarSchema;
@@ -102,7 +101,11 @@ public class CachingSchemaStore implements SchemaStore
 
     private void loadPublisher()
     {
-        KafkaOptions kafkaOptions = () -> cdcConfig.kafkaConfigs();
+        KafkaOptions kafkaOptions = cdcConfig::kafkaConfigs;
+        if (this.publisher != null)
+        {
+            this.publisher.close();
+        }
         this.publisher = SchemaStorePublisherFactory.DEFAULT.buildPublisher(kafkaOptions);
     }
 
@@ -171,9 +174,9 @@ public class CachingSchemaStore implements SchemaStore
                 }
                 return v;
             });
-            loadPublisher();
-            publishSchemas();
         }
+        loadPublisher();
+        publishSchemas();
         // Remove any old schema entries for deleted tables, this operation can be done in the end as this is
         // only for removing stale entries and no one is going to use these entries once the table is removed.
         // This doesn't have to be an atomic operation.
--- a/server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
@@ -1,0 +1,260 @@
+package org.apache.cassandra.sidecar.cdc;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+
+import com.google.common.annotations.VisibleForTesting;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import io.vertx.core.Vertx;
+import io.vertx.core.eventbus.EventBus;
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericDatumReader;
+import org.apache.avro.generic.GenericDatumWriter;
+import org.apache.avro.generic.GenericRecord;
+
+import org.apache.cassandra.cdc.avro.AvroSchemas;
+import org.apache.cassandra.cdc.avro.CqlToAvroSchemaConverter;
+import org.apache.cassandra.cdc.kafka.KafkaOptions;
+import org.apache.cassandra.cdc.schemastore.SchemaStore;
+import org.apache.cassandra.cdc.schemastore.SchemaStorePublisherFactory;
+import org.apache.cassandra.cdc.schemastore.TableSchemaPublisher;
+import org.apache.cassandra.sidecar.common.server.ThrowingRunnable;
+import org.apache.cassandra.sidecar.db.TableHistoryDatabaseAccessor;
+import org.apache.cassandra.sidecar.db.schema.SidecarSchema;
+import org.apache.cassandra.sidecar.tasks.CassandraClusterSchemaMonitor;
+import org.apache.cassandra.spark.data.CqlTable;
+import org.apache.cassandra.spark.utils.TableIdentifier;
+
+import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_CDC_CACHE_WARMED_UP;
+import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_SERVER_START;
+import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_SIDECAR_SCHEMA_INITIALIZED;
+
+/**
+ * Schemas cache to be used by CDC event serialization.
+ */
+@Singleton
+public class CachingSchemaStore implements SchemaStore
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(CachingSchemaStore.class);
+    private final Map<TableIdentifier, SchemaCacheEntry> avroSchemasCache = new ConcurrentHashMap<>();
+    private final CassandraClusterSchemaMonitor cassandraClusterSchemaMonitor;
+    private final SidecarSchema sidecarSchema;
+    private final TableHistoryDatabaseAccessor tableHistoryDatabaseAccessor;
+    private final Vertx vertx;
+    private final CdcConfigImpl cdcConfig;
+    @Nullable volatile TableSchemaPublisher publisher;
+    private final CqlToAvroSchemaConverter cqlToAvroSchemaConverter;
+    private final SidecarCdcStats sidecarCdcStats;
+
+    @Inject
+    CachingSchemaStore(Vertx vertx,
+                       CassandraClusterSchemaMonitor cassandraClusterSchemaMonitor,
+                       TableHistoryDatabaseAccessor tableHistoryDatabaseAccessor,
+                       CdcConfigImpl cdcConfig,
+                       SidecarCdcStats sidecarCdcStats,
+                       SidecarSchema sidecarSchema,
+                       CqlToAvroSchemaConverter cqlToAvroSchemaConverter)
+    {
+        super();
+        this.cassandraClusterSchemaMonitor = cassandraClusterSchemaMonitor;
+        this.tableHistoryDatabaseAccessor = tableHistoryDatabaseAccessor;
+        this.sidecarSchema = sidecarSchema;
+        this.cqlToAvroSchemaConverter = cqlToAvroSchemaConverter;
+        this.avroSchemasCache.putAll(createSchemaCache(cassandraClusterSchemaMonitor.getCdcTables()));
+        AvroSchemas.registerLogicalTypes();
+        cassandraClusterSchemaMonitor.addSchemaChangeListener(this::onSchemaChanged);
+        this.vertx = vertx;
+        this.cdcConfig = cdcConfig;
+        this.sidecarCdcStats = sidecarCdcStats;
+
+        ThrowingRunnable configChangeCallback = () -> {
+            LOGGER.info("Services configuration changed. Reloading publisher...");
+            loadPublisher();
+            publishSchemas();
+        };
+        this.cdcConfig.registerConfigChangeListener(configChangeCallback);
+        configureSidecarServerEventListeners();
+    }
+
+    private void loadPublisher()
+    {
+        KafkaOptions kafkaOptions = new KafkaOptions()
+        {
+            public Map<String, Object> kafkaConfigs()
+            {
+                return cdcConfig.kafkaConfigs();
+            }
+        };
+        this.publisher = SchemaStorePublisherFactory.DEFAULT.buildPublisher(kafkaOptions);
+    }
+
+    private void configureSidecarServerEventListeners()
+    {
+        EventBus eventBus = vertx.eventBus();
+
+        eventBus.localConsumer(ON_SERVER_START.address(), startMessage -> {
+            eventBus.localConsumer(ON_SIDECAR_SCHEMA_INITIALIZED.address(), message -> {
+                LOGGER.debug("Sidecar Schema initialized message={}", message);
+                Set<CqlTable> refreshedCdcTables = cassandraClusterSchemaMonitor.getCdcTables();
+                for (CqlTable cqlTable : refreshedCdcTables)
+                {
+                    TableIdentifier tableIdentifier = TableIdentifier.of(cqlTable.keyspace(), cqlTable.table());
+                    avroSchemasCache.compute(tableIdentifier, (k, v) ->
+                    {
+                        tableHistoryDatabaseAccessor.insertTableSchemaHistory(cqlTable.keyspace(), cqlTable.table(), cqlTable.createStatement());
+                        return v;
+                    });
+                }
+                loadPublisher();
+                publishSchemas();
+            });
+        });
+    }
+
+    private void publishSchemas()
+    {
+        Set<CqlTable> refreshedCdcTables = cassandraClusterSchemaMonitor.getCdcTables();
+        for (CqlTable cqlTable : refreshedCdcTables)
+        {
+            TableIdentifier tableIdentifier = TableIdentifier.of(cqlTable.keyspace(), cqlTable.table());
+            avroSchemasCache.compute(tableIdentifier, (k, v) ->
+            {
+                if (null != publisher)
+                {
+                    Schema schema = cqlToAvroSchemaConverter.convert(cqlTable);
+                    TableSchemaPublisher.SchemaPublishMetadata metadata = new TableSchemaPublisher.SchemaPublishMetadata();
+                    metadata.put("name", cqlTable.table());
+                    metadata.put("namespace", cqlTable.keyspace());
+                    publisher.publishSchema(schema.toString(false), metadata);
+                    sidecarCdcStats.capturePublishedSchema();
+                }
+                return new SchemaCacheEntry(cqlTable, cqlToAvroSchemaConverter);
+            });
+        }
+    }
+
+    @VisibleForTesting
+    void onSchemaChanged()
+    {
+        Set<CqlTable> refreshedCdcTables = cassandraClusterSchemaMonitor.getCdcTables();
+        for (CqlTable cqlTable : refreshedCdcTables)
+        {
+            TableIdentifier tableIdentifier = TableIdentifier.of(cqlTable.keyspace(), cqlTable.table());
+            avroSchemasCache.compute(tableIdentifier, (k, v) ->
+            {
+                if (v == null || !v.tableSchema().equals(cqlTable.createStatement()))
+                {
+                    if (sidecarSchema.isInitialized())
+                    {
+                        tableHistoryDatabaseAccessor.insertTableSchemaHistory(cqlTable.keyspace(), cqlTable.table(), cqlTable.createStatement());
+                    }
+                    LOGGER.info("Re-generating Avro Schema after schema change keyspace={} table={}", tableIdentifier.keyspace(), tableIdentifier.table());
+                    return new SchemaCacheEntry(cqlTable, cqlToAvroSchemaConverter);
+                }
+                return v;
+            });
+            loadPublisher();
+            publishSchemas();
+        }
+        // Remove any old schema entries for deleted tables, this operation can be done in the end as this is
+        // only for removing stale entries and no one is going to use these entries once the table is removed.
+        // This doesn't have to be an atomic operation.
+        avroSchemasCache.keySet().retainAll(refreshedCdcTables.stream().map(cqlTable -> TableIdentifier.of(cqlTable.keyspace(), cqlTable.table())).collect(Collectors.toList()));
+        vertx.eventBus().publish(ON_CDC_CACHE_WARMED_UP.address(), "Cdc cache warmed up");
+    }
+
+    @Override
+    public Schema getSchema(String namespace, String name)
+    {
+        TableIdentifier tableIdentifier = getTableIdentifierFromNamespace(namespace);
+        return avroSchemasCache.computeIfAbsent(tableIdentifier, k ->
+        {
+            LOGGER.warn("Unknown table for getting schema keyspace={} table={}", tableIdentifier.keyspace(), tableIdentifier.table());
+            throw new RuntimeException("Unable to get schema for unknown table " + tableIdentifier);
+        }).schema;
+    }
+
+    @Override
+    public GenericDatumWriter<GenericRecord> getWriter(String namespace, String name)
+    {
+        TableIdentifier tableIdentifier = getTableIdentifierFromNamespace(namespace);
+        return avroSchemasCache.computeIfAbsent(tableIdentifier, k -> {
+            LOGGER.warn("Unknown table for getting writer keyspace={} table={}", tableIdentifier.keyspace(), tableIdentifier.table());
+            throw new RuntimeException("Unable to get writer for unknown table " + tableIdentifier);
+        }).writer;
+    }
+
+    @Override
+    public GenericDatumReader<GenericRecord> getReader(String namespace, String name)
+    {
+        TableIdentifier tableIdentifier = getTableIdentifierFromNamespace(namespace);
+        return avroSchemasCache.computeIfAbsent(tableIdentifier, k -> {
+            LOGGER.warn("Unknown table for getting reader keyspace={} table={}", tableIdentifier.keyspace(), tableIdentifier.table());
+            throw new RuntimeException("Unable to get reader for unknown table " + tableIdentifier);
+        }).reader;
+    }
+
+    @Override
+    public String getVersion(String namespace, String name)
+    {
+        TableIdentifier tableIdentifier = getTableIdentifierFromNamespace(namespace);
+        return avroSchemasCache.computeIfAbsent(tableIdentifier, k -> {
+            LOGGER.warn("Unknown table for getting reader keyspace={} table={}", tableIdentifier.keyspace(), tableIdentifier.table());
+            throw new RuntimeException("Unable to get reader for unknown table " + tableIdentifier);
+        }).schemaUuid;
+    }
+
+    public Map<String, Schema> getSchemas()
+    {
+        return avroSchemasCache.values().stream()
+                               .collect(Collectors.toMap(e -> e.schema.getNamespace(), e -> e.schema));
+    }
+
+    private Map<TableIdentifier, SchemaCacheEntry> createSchemaCache(Set<CqlTable> cdcTables)
+    {
+
+        return cdcTables.stream()
+                        .collect(Collectors.toMap(cqlTable -> TableIdentifier.of(cqlTable.keyspace(), cqlTable.table()),
+                                                  cqlTable -> new SchemaCacheEntry(cqlTable, cqlToAvroSchemaConverter))
+                        );
+    }
+
+    private TableIdentifier getTableIdentifierFromNamespace(String namespace)
+    {
+        String[] namespaceParts = namespace.split("\\.");
+        return TableIdentifier.of(namespaceParts[0], namespaceParts[1]);
+    }
+
+    private static class SchemaCacheEntry
+    {
+        private final CqlTable table;
+        private final Schema schema;
+        private final String schemaUuid;
+        private final GenericDatumWriter<GenericRecord> writer;
+        private final GenericDatumReader<GenericRecord> reader;
+
+        private SchemaCacheEntry(CqlTable table,
+                                 CqlToAvroSchemaConverter cqlToAvroSchemaConverter)
+        {
+            this.table = table;
+            this.schema = cqlToAvroSchemaConverter.convert(table);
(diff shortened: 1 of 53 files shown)


server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -0,0 +1,271 @@
...
+public class CachingSchemaStore implements SchemaStore
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(CachingSchemaStore.class);
+    private final Map<TableIdentifier, SchemaCacheEntry> avroSchemasCache = new ConcurrentHashMap<>();
+    private final CassandraClusterSchemaMonitor cassandraClusterSchemaMonitor;
+    private final SidecarSchema sidecarSchema;
+    private final TableHistoryDatabaseAccessor tableHistoryDatabaseAccessor;
+    private final Vertx vertx;
+    private final CdcConfigImpl cdcConfig;
+    @Nullable volatile TableSchemaPublisher publisher;
```
jyothsnakonisa: This could be private.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -0,0 +1,271 @@
...
+import org.apache.cassandra.sidecar.tasks.CassandraClusterSchemaMonitor;
+import org.apache.cassandra.spark.data.CqlTable;
+import org.apache.cassandra.spark.utils.TableIdentifier;
+
+import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_CDC_CACHE_WARMED_UP;
+import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_SERVER_START;
+import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_SIDECAR_SCHEMA_INITIALIZED;
+
+/**
+ * Schemas cache to be used by CDC event serialization.
```
jyothsnakonisa: Should we add detailed documentation?

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -0,0 +1,271 @@
...
+
+    private void loadPublisher()
+    {
+        KafkaOptions kafkaOptions = new KafkaOptions()
+        {
+            public Map<String, Object> kafkaConfigs()
+            {
+                return cdcConfig.kafkaConfigs();
+            }
+        };
```
jyothsnakonisa: Can this be lambda instead?

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -0,0 +1,271 @@
...
+        {
+            TableIdentifier tableIdentifier = TableIdentifier.of(cqlTable.keyspace(), cqlTable.table());
+            avroSchemasCache.compute(tableIdentifier, (k, v) ->
+            {
+                if (null != publisher)
+                {
+                    Schema schema = cqlToAvroSchemaConverter.convert(cqlTable);
+                    TableSchemaPublisher.SchemaPublishMetadata metadata = new TableSchemaPublisher.SchemaPublishMetadata();
+                    metadata.put("name", cqlTable.table());
+                    metadata.put("namespace", cqlTable.keyspace());
```
jyothsnakonisa: Can we have constants for the keys?

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -0,0 +1,271 @@
...
+            LOGGER.warn("Unknown table for getting reader keyspace={} table={}", tableIdentifier.keyspace(), tableIdentifier.table());
+            throw new RuntimeException("Unable to get reader for unknown table " + tableIdentifier);
+        }).schemaUuid;
+    }
+
+    public Map<String, Schema> getSchemas()
+    {
+        return avroSchemasCache.values().stream()
+                               .collect(Collectors.toMap(e -> e.schema.getNamespace(), e -> e.schema));
+    }
```
jyothsnakonisa: Unused method, do we need it in future patches?

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcAvroSerializer.java
```diff
@@ -0,0 +1,42 @@
...
+package org.apache.cassandra.sidecar.cdc;
+
+import org.apache.cassandra.bridge.CassandraBridgeFactory;
+import org.apache.cassandra.cdc.TypeCache;
+import org.apache.cassandra.cdc.kafka.AvroGenericRecordSerializer;
+import org.apache.cassandra.cdc.schemastore.SchemaStore;
+import org.apache.cassandra.sidecar.utils.InstanceMetadataFetcher;
+
+/**
+ * Taking a schema, this class serializes a CDC into AVRO format
```
jyothsnakonisa: May be better description for the class, how about the following?

> Serializer to convert Cassandra CDC events into Avro GenericRecord objects.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcManager.java
```diff
@@ -0,0 +1,263 @@
...
+import org.apache.cassandra.sidecar.concurrent.TaskExecutorPool;
+import org.apache.cassandra.sidecar.coordination.RangeManager;
+import org.apache.cassandra.sidecar.db.CdcDatabaseAccessor;
+import org.apache.cassandra.sidecar.utils.InstanceMetadataFetcher;
+import org.apache.cassandra.spark.utils.AsyncExecutor;
+import org.jetbrains.annotations.NotNull;
+
+
+/**
+ * Class handling CDC iterators
```
jyothsnakonisa: Please add better documentation.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
```diff
@@ -0,0 +1,401 @@
...
+import static org.apache.cassandra.sidecar.server.SidecarServerEvents.ON_SERVER_STOP;
+
+/**
+ * Class that handles CDC life cycle
+ */
+@Singleton
+public class CdcPublisher implements Handler<Message<Object>>, PeriodicTask
+{
+    static final Logger LOGGER = LoggerFactory.getLogger(CdcPublisher.class);
+    static final long INITIALIZATION_LOOP_DELAY_MILLIS = 1000;
```
jyothsnakonisa: Please make them private.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
```diff
@@ -0,0 +1,401 @@
...
+    private final TokenRingProvider tokenRingProvider;
+    private final SchemaSupplier schemaSupplier;
+    private final CdcSidecarInstancesProvider sidecarInstancesProvider;
+    private final InstanceMetadataFetcher instanceMetadataFetcher;
+    private final ClusterConfigProvider clusterConfigProvider;
+    private final SidecarCdcClient.ClientConfig clientConfig;
+    private final ICdcStats cdcStats;
+    private final SidecarConfiguration sidecarConfiguration;
+    private CdcManager cdcManager;
+    Serializer<CdcEvent> avroSerializer;
```
jyothsnakonisa: Please make this private

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
```diff
@@ -0,0 +1,401 @@
...
+        }
+
+        Map<String, String> sslConfigMap = new HashMap<>();
+
+        sslConfigMap.put("enabled", sslConfiguration.enabled() + "");
+        sslConfigMap.put("preferOpenSSL", sslConfiguration.preferOpenSSL() + "");
+        sslConfigMap.put("clientAuth", sslConfiguration.clientAuth());
+        sslConfigMap.put("cipherSuites", String.join(",", sslConfiguration.cipherSuites()));
+        sslConfigMap.put("secureTransportProtocols", String.join(",", sslConfiguration.secureTransportProtocols()));
+        sslConfigMap.put("handshakeTimeout", sslConfiguration.handshakeTimeout().toString());
```
jyothsnakonisa: Should we make these and other keys constants?

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
```diff
@@ -0,0 +1,401 @@
...
+        {
+            if (this.rangeManager == null)
+            {
+                this.rangeManager = new ContentionFreeRangeManager(vertx, tokenRingProvider);
+            }
+            String localDc = rangeManager.getLocalDcSafe();
+            if (conf.datacenter() != null && !conf.datacenter().isEmpty() && !conf.datacenter().equals(localDc))
+            {
+                LOGGER.info("Cdc not enabled in this DC localDc={} cdcDc={}", localDc, conf.datacenter());
+                return;
```
jyothsnakonisa: Return not needed here and other if else blocks

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcSchemaSupplier.java
```diff
@@ -0,0 +1,123 @@
...
+import org.apache.cassandra.spark.data.CqlTable;
+import org.apache.cassandra.spark.data.ReplicationFactor;
+import org.apache.cassandra.spark.data.partitioner.Partitioner;
+import org.apache.cassandra.spark.utils.CqlUtils;
+import org.apache.cassandra.spark.utils.TableIdentifier;
+import org.jetbrains.annotations.NotNull;
+
+
+/**
+ * Class providing the schema for CDC
```
jyothsnakonisa: May be more descriptive documentation about the class?

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
```diff
@@ -0,0 +1,401 @@
...
+        }
+    }
+
+    private void initialize()
+    {
+        try
+        {
+            if (this.rangeManager == null)
+            {
+                this.rangeManager = new ContentionFreeRangeManager(vertx, tokenRingProvider);
```
jyothsnakonisa: Why are we hardcoding range manager to be `ContentionFreeRangeManager` shouldn't that be injected?

`ContentionFreeRangeManager` always succeeds on propose and release ownership.

server/src/main/java/org/apache/cassandra/sidecar/coordination/RangeManager.java
```diff
@@ -0,0 +1,416 @@
...
+    protected synchronized void onSidecarUp(SidecarInstance instance)
+    {
+        // on Sidecar detected to be UP, initiate proposal to release ownership of its primary ranges
+        final String dc = getLocalDcSafe();
+        if (dc == null)
+        {
+            LOGGER.error("DC null, cannot respond to onSidecarUp event");
+            return;
+        }
+        Map<String, Set<TokenRange>> primaryRanges = toSet(tokenRingProvider.getPrimaryRanges(instance, dc));
```
jyothsnakonisa, line 241: Why is sidecar releasing ownership if the peer is up? Don't we run into situation where all sidecar instances release ownership and there will be no owner for the token range?

server/src/main/java/org/apache/cassandra/sidecar/coordination/TokenRingProvider.java
```diff
@@ -68,7 +67,7 @@ public TokenRingProvider(InstancesMetadata instancesMetadata, InstanceMetadataFe
      * @param dc          data center
      * @return map of token ranges per Cassandra instance IP
      */
-    protected abstract Map<String, List<Range<BigInteger>>> getAllTokenRanges(Partitioner partitioner, @Nullable String dc);
+    protected abstract Map<String, List<TokenRange>> getAllTokenRanges(Partitioner partitioner, @Nullable String dc);
```
jyothsnakonisa, line 70: Wondering why didn't we use `TokenRange` earlier, seems like we have that class available from before

server/src/main/java/org/apache/cassandra/sidecar/db/CdcDatabaseAccessor.java
```diff
@@ -181,31 +216,4 @@ ResultSetFuture selectCdcRange(String jobId, int split)
     {
         return session().executeAsync(tableSchema.select().bind(jobId, (short) split));
     }
-
-    public ResultSetFuture insertTableSchemaHistory(String keyspace, String tableName, String schema)
```
jyothsnakonisa, line 185: Are you sure that you might not need these in the future?

server/src/main/java/org/apache/cassandra/sidecar/db/CdcDatabaseAccessor.java
```diff
@@ -129,6 +138,32 @@ public List<ResultSetFuture> storeStateAsync(String jobId,
...
 
+    /**
+     * Load cdc state for a given jobId and token range and merge into canonical view
+     *
+     * @param stats SidecarCdcStats to publish metrics
+     * @param jobId Cdc job id
+     * @param range token range
+     * @return merged SidecarCdcState object that merges previous state objects that overlap with token range to given canonical view of Cdc state.
+     */
+    public Optional<CdcState> loadSidecarCdcState(SidecarCdcStats stats, String jobId, TokenRange range)
```
jyothsnakonisa: Looks like this is not used anywheere, I am assuming that it will be used later

server/src/main/java/org/apache/cassandra/sidecar/db/TableHistoryDatabaseAccessor.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.cassandra.sidecar.db.schema.TableHistorySchema;
+
+/**
+ * Database accessor for Table History operations.
+ */
+@SuppressWarnings("resource")
+@Singleton
+public class TableHistoryDatabaseAccessor extends DatabaseAccessor<TableHistorySchema>
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(TableHistoryDatabaseAccessor.class);
```
jyothsnakonisa: Unused logger.

server/src/main/java/org/apache/cassandra/sidecar/db/TableHistoryDatabaseAccessor.java
```diff
@@ -0,0 +1,78 @@
...
+
+    public ResultSetFuture insertTableSchemaHistory(String keyspace, String tableName, String schema)
+    {
+        UUID schemaUuid = UUID.nameUUIDFromBytes(schema.getBytes(StandardCharsets.UTF_8));
+        return session().executeAsync(tableSchema
+                                      .insertTableSchema()
+                                      .bind(keyspace, tableName, schemaUuid, schema));
+    }
+
+    public String tableSchemaFromVersion(String keyspace, String tableName, String version)
```
jyothsnakonisa: Unused method.

server/src/main/java/org/apache/cassandra/sidecar/modules/CdcModule.java
```diff
@@ -269,4 +326,156 @@ public SidecarCdcStats sidecarCdcStats()
...
+                            ).collect(Collectors.toSet());
+            }
+
+            public Partitioner partitioner()
+            {
+                NodeSettings nodeSettings = instanceMetadataFetcher.callOnFirstAvailableInstance(instance-> instance.delegate().nodeSettings());
+                String[] parts = nodeSettings.partitioner().split("\\.");
+                return Partitioner.valueOf(parts[parts.length - 1]);
+            }
+        };
```
jyothsnakonisa: How about moving this to a separate class for better readability?

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcConfigRefresherNotifierTask.java
```diff
@@ -0,0 +1,163 @@
...
+@Singleton
+public class CdcConfigRefresherNotifierTask implements PeriodicTask
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(CdcConfigRefresherNotifierTask.class);
+    private final SidecarConfiguration sidecarConfiguration;
+    private final CdcConfigAccessor cdcConfigAccessor;
+    private final KafkaConfigAccessor kafkaConfigAccessor;
+    private final Vertx vertx;
+    private final List<ThrowingRunnable> configChangeListeners = Collections.synchronizedList(new ArrayList<>());
+    public ConfigMappings configMappings = new ConfigMappings();
```
jyothsnakonisa: can be private

server/src/main/java/org/apache/cassandra/sidecar/tasks/ClusterTopologyMonitor.java
```diff
@@ -0,0 +1,310 @@
...
+        }
+    }
+
+    /**
+     * Returns the delay between topology refresh cycles.
+     */
+    @Override
+    public DurationSpec delay()
+    {
+//        return MillisecondBoundConfiguration.parse(sidecarConfiguration.serviceConfiguration().refreshClusterTopology());
```
jyothsnakonisa: Can you please remove the commented code.

server/src/main/java/org/apache/cassandra/sidecar/tasks/ClusterTopologyMonitor.java
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    /**
+     * Refreshes topology information for all datacenters and publishes change events.
+     */
+    protected void refresh()
+    {
+        for (String dc : tokenRingProvider.dcs())
+        {
+            @Nullable final Map<String, List<TokenRange>> prev = perDcRanges.get(dc);
```
jyothsnakonisa: We might not need `@nullable` for variable and  also noticed that other classes do not have final for local variables but this class seems to have final for local variable. Please remove them for consistency across this file.

server/src/test/java/org/apache/cassandra/sidecar/cdc/CdcConfigImplTest.java
```diff
@@ -150,103 +137,11 @@ void testConfigsWhenConfigsAreNotEmpty()
         assertThat(cdcConfig.persistDelay()).isEqualTo(new MillisecondBoundConfiguration(5, TimeUnit.SECONDS));
     }
 
-    @Test
-    void testConfigChanged() throws Exception
```
jyothsnakonisa, line 154: Why did you remove this test? I think this is valuable to have a test for checking callback and config updating when config is changed.

jyothsnakonisa (commented):
Looks good overall, leaving some comments.

server/src/main/java/org/apache/cassandra/sidecar/db/CdcDatabaseAccessor.java
bbotella: We can always add it back if we need it.

server/src/test/java/org/apache/cassandra/sidecar/cdc/CdcConfigImplTest.java
bbotella: This became a periodic task, and the callback was moved to a vertx signal instead.

jyothsnakonisa: But I think the test is sill valid right? may be we should have the test to check if the vertx signal is made when config is changed

jyothsnakonisa:
Thanks @bbotella for addressing my comments. There is this comment https://github.com/apache/cassandra-sidecar/pull/294#discussion_r2604697127 which is not resolved. We need to look into that codepath as it is critical for leader election and mutation publishing.

Also, can you please check if there is a way to add an end to end test that checks if the mutations are published(in-memory) might not need to have kafka for the test.

Other than those two everything else looks good to me.

jyothsnakonisa (approved):
Thanks Bernardo for the patch!

server/src/main/java/org/apache/cassandra/sidecar/codecs/BigIntegerCodec.java
```diff
@@ -0,0 +1,84 @@
...
+        pos.add(4 + ar.length);
+        return new BigInteger(ar);
+    }
+
+    /**
+     * Decodes a BigInteger from the wire buffer at the specified position.
+     */
+    public BigInteger decodeFromWire(int pos, Buffer buf)
+    {
+        return decodeFromWire(new MutableInt(pos), buf);
```
jberragan: I think the `MutableInt` is only required if we need to track the change in position in a wrapper serializer, this would be marginally faster as:

```suggestion
        return new BigInteger(CommonCodecs.BYTE_ARRAY.decodeFromWire(pos, buf));
```

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcConfigImpl.java
```diff
@@ -17,78 +17,64 @@
...
-    private final CdcConfiguration cdcConfiguration;
+    private final Vertx vertx;
     private final CdcConfigAccessor cdcConfigAccessor;
-    private final KafkaConfigAccessor kafkaConfigAccessor;
-    private final List<ThrowingRunnable> configChangeListeners = Collections.synchronizedList(new ArrayList<>());
-    private final ConfigRefreshNotifier configRefreshNotifier;
-    private volatile Map<String, String> kafkaConfigMappings = Map.of();
-    private volatile Map<String, String> cdcConfigMappings = Map.of();
+
+    private Map<String, String> kafkaConfigMappings = Map.of();
```
jberragan: These look like they should be `volatile`

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcManager.java
```diff
@@ -0,0 +1,282 @@
...
+                                                                      cdcStats,
+                                                                      taskExecutorPool);
+                                    }
+                                    catch (IOException e)
+                                    {
+                                        throw new RuntimeException(e);
+                                    }
+                                });
+                            }))
+                   .collect(Collectors.toList());
```
jberragan, line 167: Result of `.collect(Collectors.toList());` is ignored/not used.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcManager.java
```diff
@@ -0,0 +1,282 @@
...
+    List<SidecarCdc> buildCdcConsumers()
+    {
+        Map<String, Set<TokenRange>> ownedRanges = rangeManager.ownedTokenRanges();
+        if (ownedRanges == null || ownedRanges.isEmpty())
+        {
+            throw new IllegalStateException("No owned token ranges right now, cql session may still be initializing.");
+        }
+
+        // NEW: Deduplicate by (instanceId, tokenRange) to prevent duplicate consumers
+        Map<String, SidecarCdc> uniqueConsumers = new HashMap<>();
```
jberragan:
```suggestion
        Map<String, SidecarCdc> uniqueConsumers = new HashMap<>(ownedRanges.values().stream().mapToInt(Set::size).sum());
```

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcManager.java
```diff
@@ -0,0 +1,282 @@
...
+            if (resolveToSameAddress(instanceIp, configuredHost))
+            {
+                return instance.id();
+            }
+        }
+        LOGGER.warn("Requested IP {} does not match with any instances", instanceIp);
+        return -1;
+    }
+
+    private boolean resolveToSameAddress(String host1, String host2)
```
jberragan: It seems like this only expects ip addresses to be input, and so will not perform DNS resolution. Can we make that explicit? We can also make the method static and add some unit tests here.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcManager.java
```diff
@@ -0,0 +1,282 @@
...
+    public void stopConsumers()
+    {
+        consumers.forEach(SidecarCdc::stop);
+    }
+
+    private Integer getInstanceId(String instanceIp)
+    {
+        for (InstanceMetadata instance : instanceFetcher.allLocalInstances())
+        {
+            String configuredHost = instance.ipAddress();
```
jberragan: This is an ip address, no?

```suggestion
            String configuredIpAddress = instance.ipAddress();
```

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcConfigImpl.java
jberragan: There not accessed with `synchronized` so better to make them `volatile`

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
```diff
@@ -0,0 +1,404 @@
...
+    private final SidecarCdcStats sidecarCdcStats;
+    private final SchemaSupplier schemaSupplier;
+    private final CdcSidecarInstancesProvider sidecarInstancesProvider;
+    private final InstanceMetadataFetcher instanceMetadataFetcher;
+    private final ClusterConfigProvider clusterConfigProvider;
+    private final SidecarCdcClient.ClientConfig clientConfig;
+    private final ICdcStats cdcStats;
+    private final SidecarConfiguration sidecarConfiguration;
+    private CdcManager cdcManager;
+    private Serializer<CdcEvent> avroSerializer;
```
jberragan: Can be final:

```suggestion
    private final Serializer<CdcEvent> avroSerializer;
    private final Provider<RangeManager> rangeManagerProvider;
```

jberragan (approved):
Mostly nits, +1 (nb)

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -0,0 +1,263 @@
...
+                    {
+                        tableHistoryDatabaseAccessor.insertTableSchemaHistory(cqlTable.keyspace(), cqlTable.table(), cqlTable.createStatement());
+                    }
+                    LOGGER.info("Re-generating Avro Schema after schema change keyspace={} table={}", tableIdentifier.keyspace(), tableIdentifier.table());
+                    return new SchemaCacheEntry(cqlTable, cqlToAvroSchemaConverter);
+                }
+                return v;
+            });
+            loadPublisher();
+            publishSchemas();
```
yifan-c: Those 2 methods are called in _every_ iteration. Looks like they are included in the for-loop by mistake.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -0,0 +1,263 @@
...
+        this.cdcConfig = cdcConfig;
+        this.sidecarCdcStats = sidecarCdcStats;
+
+        configureSidecarServerEventListeners();
+    }
+
+    private void loadPublisher()
+    {
+        KafkaOptions kafkaOptions = () -> cdcConfig.kafkaConfigs();
+        this.publisher = SchemaStorePublisherFactory.DEFAULT.buildPublisher(kafkaOptions);
```
yifan-c, line 112: `publisher` is a `Closeable` object. We should close the previous object when assigning a new one.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcManager.java
```diff
@@ -0,0 +1,282 @@
...
+                                                                      cdcStats,
+                                                                      taskExecutorPool);
+                                    }
+                                    catch (IOException e)
+                                    {
+                                        throw new RuntimeException(e);
+                                    }
+                                });
+                            }))
+                   .collect(Collectors.toList());;
```
yifan-c: double `;`

server/src/main/java/org/apache/cassandra/sidecar/server/SidecarServerEvents.java
```diff
@@ -113,6 +113,10 @@ public enum SidecarServerEvents
      * has lost the cluster-wide lease.
      */
     ON_SIDECAR_GLOBAL_LEASE_LOST,
+
+    ON_CDC_CONFIGURATION_CHANGED,
+    ON_CDC_CACHE_WARMED_UP,
+    ON_CDC_CONFIG_MAPPINGS_CHANGED,
```
yifan-c, line 133: Add javadoc

server/src/main/java/org/apache/cassandra/sidecar/modules/CdcModule.java
```diff
@@ -269,4 +315,127 @@ public SidecarCdcStats sidecarCdcStats()
...
+                                () -> new ContentionFreeRangeManager(vertx, tokenRingProvider));
+    }
+
+    @Singleton
+    @ProvidesIntoMap
+    @KeyClassMapKey(PeriodicTaskMapKeys.CdcConfigRefresherNotifierKey.class)
+    PeriodicTask cdcConfigRefresherNotifier(Vertx vertx,
+                                            SidecarConfiguration sidecarConfiguration,
+                                    KafkaConfigAccessor kafkaConfigAccessor,
+                                    CdcConfigAccessor cdcConfigAccessor)
```
yifan-c: indentation

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -0,0 +1,263 @@
...
+    @Inject
+    CachingSchemaStore(Vertx vertx,
+                       CassandraClusterSchemaMonitor cassandraClusterSchemaMonitor,
+                       TableHistoryDatabaseAccessor tableHistoryDatabaseAccessor,
+                       CdcConfigImpl cdcConfig,
+                       SidecarCdcStats sidecarCdcStats,
+                       SidecarSchema sidecarSchema,
+                       CqlToAvroSchemaConverter cqlToAvroSchemaConverter)
+    {
+        super();
```
yifan-c: `super()` is redundant, because it is just implementing the interface `SchemaStore`

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -0,0 +1,263 @@
...
+        this.sidecarSchema = sidecarSchema;
+        this.cqlToAvroSchemaConverter = cqlToAvroSchemaConverter;
+        this.avroSchemasCache.putAll(createSchemaCache(cassandraClusterSchemaMonitor.getCdcTables()));
+        AvroSchemas.registerLogicalTypes();
+        cassandraClusterSchemaMonitor.addSchemaChangeListener(this::onSchemaChanged);
+        this.vertx = vertx;
+        this.cdcConfig = cdcConfig;
+        this.sidecarCdcStats = sidecarCdcStats;
+
+        configureSidecarServerEventListeners();
```
yifan-c: When CDC is disabled, we should skip those all together. It does not make sense to register listeners, etc. 

Ideally, we should not mount the `CdcModule` entirely, if the feature is not enabled. It is outside of the scope of this PR.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
```diff
@@ -0,0 +1,404 @@
...
+        this.cdcStats = cdcStats;
+        this.sidecarConfiguration = sidecarConfiguration;
+        this.avroSerializer = avroSerializer;
+        this.rangeManagerProvider = rangeManagerProvider;
+        vertx.eventBus().localConsumer(RangeManager.RangeManagerEvents.ON_TOKEN_RANGE_CHANGED.address(), this);
+        vertx.eventBus().localConsumer(RangeManager.LeadershipEvents.ON_TOKEN_RANGE_GAINED.address(), this);
+        vertx.eventBus().localConsumer(RangeManager.LeadershipEvents.ON_TOKEN_RANGE_LOST.address(), this);
+        vertx.eventBus().localConsumer(ON_SERVER_STOP.address(), this);
+        vertx.eventBus().localConsumer(ON_CDC_CACHE_WARMED_UP.address(), this);
+        vertx.eventBus().localConsumer(ON_CDC_CONFIGURATION_CHANGED.address(), new ConfigChangedHandler());
```
yifan-c: Skip those when CDC is disabled.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
```diff
@@ -0,0 +1,404 @@
...
+
+        SslConfig sslConfig = SslConfig.create(sslConfigMap);
+        return new SslConfigSecretsProvider(sslConfig);
+    }
+
+    public EventConsumer eventConsumer(CdcConfig conf,
+                                       Serializer<CdcEvent> avroSerializer)
+    {
+        KafkaProducer<String, byte[]> producer = new KafkaProducer<>(conf.kafkaConfigs());
+        KafkaPublisher kafkaPublisher = new KafkaPublisher(TopicSupplier.staticTopicSupplier(conf.kafkaTopic()),
```
yifan-c: Both `producer` and `kafkaPublisher` are `Closeable`. I do not see them get closed anywhere. We are create a new eventConsumer on every publisher restart, meaning leaks.

yifan-c (changes requested):
When CDC is disabled,

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
bbotella: I will create a Jira

yifan-c: Thanks. Just to clarify that conditionally register the listeners are still wanted in this patch.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
bbotella: Should this be part of the Jira mentioned in previous comment?

yifan-c: Those should be behind the condition whether cdc is enabled.

yifan-c (approved):

PR #308, opened 2026-01-06
CASSSIDECAR-369: Database access during auth flow is blocking event-loop thread

sarankk:


server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -92,16 +95,28 @@ protected LoadingCache<K, V> cache()
...
      * @param k key
-     * @return The current value of {@code K} if cached or loaded.
+     * @return A {@link Future} containing the current value of {@code K} if cached or loaded.
      * <p>
      * See {@link LoadingCache#get(Object)} for possible exceptions.
      */
-    public V get(K k)
+    public Future<V> get(K k)
+    {
+        return servicePool.executeBlocking(() -> getCached(k), false);
```
yifan-c, line 116: Unconditionally dispatching the execution to a service pool could hurt performance. In most cases, we expect cache hit and it should return immediately, instead of incurring context switch. 

You can do a non-blocking check by using `getIfPresent()` and only schedule `getCached` on cache missing. Note that it still permits concurrent calls for a small time window. It is a good balance between code simplicity and performance. 

I would not suggest to save the `Future`, as it could lead to wrong thread continuation, if not used carefully. 

Please update all the other similar use cases that unconditionally executes in servicePool.

server/src/main/java/org/apache/cassandra/sidecar/acl/AuthCache.java
```diff
@@ -92,16 +95,28 @@ protected LoadingCache<K, V> cache()
...
+    }
+
+    /**
+     * Helper method to retrieve value from cache synchronously. Used by {@link #get(K)} which offloads
+     * this to a worker thread.
+     *
+     * @param k key
+     * @return The current value of {@code K} if cached or loaded
```
yifan-c: nit: use lower case `the`. Because the sentence here is `@return the current ...`. It is common to use lower case letters following `@return`, `@param`, etc. in javadoc.

server/src/main/java/org/apache/cassandra/sidecar/acl/authorization/CachedAuthorizationHandler.java
```diff
@@ -264,15 +266,27 @@ protected void checkOrFetchAuthorizations(Promise<Boolean> promise, Authorizatio
...
+                     .map(cf -> {
+                         for (int i = 0; i < cf.size(); i++)
+                         {
+                             if (Boolean.TRUE.equals(cf.resultAt(i)))
+                             {
+                                 return true;
+                             }
+                         }
+                         return false;
+                     });
```
yifan-c: Leverage `Future.any` to get result sooner without waiting for all futures to complete.

```suggestion
        for (String identity : identities)
        {
            adminFutures.add(adminIdentityResolver.isAdmin(identity)
                                                  .compose(isAdmin -> isAdmin ? Future.succeededFuture() : Future.failedFuture("Not admin")));
        }
        return Future.any(adminFutures)
                     .compose(success -> Future.succeededFuture(true),
                              failure -> Future.succeededFuture(false));
```

sarankk: Good idea, updated it.

server/src/main/java/org/apache/cassandra/sidecar/acl/authentication/CassandraIdentityExtractor.java
```diff
@@ -44,22 +45,46 @@ public CassandraIdentityExtractor(AdminIdentityResolver adminIdentityResolver,
...
+                            // Sidecar recognizes identities in identity_to_role table as authenticated
+                            Future<Boolean> inCacheFuture
+                            = identityToRoleCache.containsKey(identity).compose(hasRole -> hasRole
+                                                                                           ? Future.succeededFuture()
+                                                                                           : Future.failedFuture("no role mapped"));;
+
+                            Future<Boolean> isValidFuture
+                            = Future.any(isAdminFuture, inCacheFuture)
+                                    .compose(success -> Future.succeededFuture(true),
+                                             failure -> Future.succeededFuture(false));
```
yifan-c: nit: consider moving to Future utilities, if this pattern is repeating. It is fine to ignore for now.

yifan-c: Both `isAdmin` and `containsKey` triggers the loading function in the `IdentityToRoleCache`. 
The two concurrent futures could lead to the function being evaluated twice, according to the `AuthCache.get` behavior. 
Should we eval `isAdmin` first, then `identityToRoleCache.containsKey`? The second does not need to trigger the loading function again, since `isAdmin` call already either populate the cache or not.

yifan-c (approved):
One minor issue in `CassandraIdentityExtractor`. OK with it not being addressed.

PR #309, opened 2026-01-25
CASSSIDECAR-226 Adding endpoint for verifying files post data copy during live migration

nvharikrishna:
CASSSIDECAR-226 Adding an endpoint for verifying files between source and destination post data copy.

This implementation uses a two-task approach (data copy + file verification) rather than inline digest verification during data copy (as originally proposed in CEP-40). This design choice is motivated by:

1. Performance Efficiency: The data copy task executes multiple iterations internally. Even with successThreshold=1.0, the task might require at least two internal iterations (iteration 0: download → DOWNLOAD_COMPLETE, iteration 1: verify threshold → SUCCESS). Inline digest verification would calculate digests twice per file (once in each iteration), doubling the I/O cost. With separate tasks, digests are calculated once after data stabilizes.
2. Code Simplicity: Separating digest verification from file copying provides clear separation of concerns, making each task easier to understand, test, and maintain.
3. Operational Flexibility: Users can run verification independently, repeat it if needed, or skip it for non-critical migrations. Inline verification would make this mandatory overhead.

Here are the endpoint details:

**Sample files verification task submission request:**
```bash
curl -X POST http://dest-host.example.com:9043/api/v1/live-migration/files-verification-tasks \
  -H "Content-Type: application/json" \
  -d '{
    "maxConcurrency": 10,
    "digestAlgorithm": "MD5"
  }'
```

It supports XXHash32 algorithm too and `seed` as additional input in the payload.

**Sample response:**

```bash
{
  "taskId": "b8e4f3d2-5c6b-5d9e-0f2g-3b4c5d6e7f8g",
  "statusUrl": "/api/v1/live-migration/files-verification-tasks/b8e4f3d2-5c6b-5d9e-0f2g-3b4c5d6e7f8g"
}
```

**Fetching files verification task status**
```bash
curl -X GET http://dest-host.example.com:9043/api/v1/live-migration/files-verification-tasks/b8e4f3d2-5c6b-5d9e-0f2g-3b4c5d6e7f8g
```

**Sample response:**
```bash
{
  "id": "b8e4f3d2-5c6b-5d9e-0f2g-3b4c5d6e7f8g ",
  "digestAlgorithm": "md5",
  "seed": null,
  "state": "COMPLETED",
  "source": "localhost1",
  "port": 9043,
  "filesNotFoundAtSource": 0,
  "filesNotFoundAtDestination": 0,
  "metadataMatched": 379,
  "metadataMismatches": 0,
  "digestMismatches": 0,
  "digestVerificationFailures": 0,
  "filesMatched": 323
}
```

Also made additional changes to ensure that either data copy task or file verification task can be executed at any point of time.


client-common/src/main/java/org/apache/cassandra/sidecar/common/request/LiveMigrationFileDigestRequest.java
```diff
@@ -0,0 +1,76 @@
...
+    {
+        Objects.requireNonNull(requestURI, "requestURI cannot be null");
+        Objects.requireNonNull(digestAlgorithm, "digestAlgorithm cannot be null");
+        if (digestAlgorithm.isEmpty())
+        {
+            throw new IllegalArgumentException("digestAlgorithm cannot be empty");
+        }
+
+        String fullURI = seed != null
+                         ? String.format("%s?%s=%s&%s=%d", requestURI, DIGEST_ALGORITHM_PARAM, digestAlgorithm, SEED_PARAM, seed)
```
yifan-c: One learning from the RestoreJob work is that the custom seed does not provide benefit for data integrity validation, but only adds code complexity. I would just drop the support of custom seed support to simplify the implementation, and use the fixed seed 0, which also makes the client-server communication simpler. 
Not strong on removing the seed support, but feel ideal to do so.

nvharikrishna: I agree with the code complexity and simplifying code suggestions. I can remove the support for seed for live migration.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateFilesVerificationTaskHandler.java
```diff
@@ -0,0 +1,172 @@
...
+        {
+            LiveMigrationFilesVerificationRequest request =
+            Json.decodeValue(context.body().buffer(), LiveMigrationFilesVerificationRequest.class);
+
+            if (request.maxConcurrency() > liveMigrationConfiguration.maxConcurrentFileRequests())
+            {
+                throw new IllegalArgumentException("Invalid maxConcurrency " + request.maxConcurrency() +
+                                                   ". It cannot be greater than " +
+                                                   liveMigrationConfiguration.maxConcurrentFileRequests());
+            }
```
yifan-c: In `FilesVerificationTaskManager`, it handles `maxConcurrency` differently. Can you address the inconsistency or the duplication? There seems to be sufficient to have one validation only. 

```
        if (request.maxConcurrency() > maxPossibleConcurrency)
        {
            return Future.failedFuture(
            new LiveMigrationInvalidRequestException("max concurrency can not be more than " + maxPossibleConcurrency));
        }
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateFilesVerificationTaskHandler.java
```diff
@@ -0,0 +1,172 @@
...
+                            String statusUrl = LIVE_MIGRATION_FILES_VERIFICATION_TASKS_ROUTE + "/" + task.id();
+                            context.json(new LiveMigrationTaskCreationResponse(task.id(), statusUrl));
+                        })
+
+                        .onFailure(throwable -> {
+                            if (throwable instanceof LiveMigrationTaskInProgressException)
+                            {
+                                LOGGER.error("Cannot start a new files verification task for host {} " +
+                                             "while another live migration task is in progress.", host);
+                                context.fail(wrapHttpException(FORBIDDEN, throwable.getMessage(), throwable));
```
yifan-c: Should the status code be 409 Conflict, instead of Forbidden? 
Forbidden typically means no permission to perform an action, not the cause here.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/FilesVerificationTaskManager.java
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * Manages the lifecycle of file digest verification tasks during live migration operations.
+ * This manager ensures that only one {@link LiveMigrationTask} can be active per instance at a time,
+ * preventing concurrent migration operations that could impact system resources.
+ * Tasks are created using the {@link LiveMigrationFilesVerificationTaskFactory} and
+ * executed asynchronously to validate file integrity between source and destination nodes.
+ */
+@Singleton
+public class FilesVerificationTaskManager
```
yifan-c, line 50: Should `FilesVerificationTaskManager` and `DataCopyTaskManager` have a common base class? There are several almost identical methods, e.g. `getAllTasks()`, `getTask()` and `cancelTask()`.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationTaskManager.java
```diff
@@ -0,0 +1,154 @@
...
+        InstanceMetadata localInstance = instancesMetadata.instanceFromHost(currentHost);
+        if (localInstance == null)
+        {
+            throw new IllegalStateException("No instance found for host: " + currentHost);
+        }
+        if (!currentTasks.containsKey(localInstance.id()))
+        {
+            return Collections.emptyList();
+        }
+        return Collections.singletonList(currentTasks.get(localInstance.id()));
```
yifan-c: `localInstance.id()` could potentially be removed at this step due to race condition. Instead, let's get the value at line#99 and return based on whether value is null or not.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFilesVerificationTask.java
```diff
@@ -0,0 +1,643 @@
...
+
+    private Future<InstanceFilesListResponse> fetchSourceFileList()
+    {
+        return Future.fromCompletionStage(
+                     sidecarClient.liveMigrationListInstanceFilesAsync(new SidecarInstanceImpl(source, port)))
+                     .onFailure(cause -> LOGGER.error("{} Failed to obtain list of files from source {}:{}",
+                                                      logPrefix, source, port, cause));
+    }
+
+    private @NotNull Future<List<InstanceFileInfo>> compareFilesMeta(List<InstanceFileInfo> localFiles,
```
yifan-c: nit: prefer `compareFilesMetadata`

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFilesVerificationTask.java
```diff
@@ -0,0 +1,643 @@
...
+import static org.apache.cassandra.sidecar.livemigration.LiveMigrationInstanceMetadataUtil.localPath;
+
+/**
+ * Verifies file integrity between source and destination instances during live migration
+ * by comparing file lists and validating file digests using configurable digest algorithms.
+ *
+ * <p>The verification process consists of three stages:
+ * <ol>
+ *   <li>Fetch file lists from both source and destination instances concurrently</li>
+ *   <li>Compare file metadata (size, type, modification time) - fails fast on mismatches</li>
```
yifan-c: I do not think `compareFilesMeta()` (again, prefer `compareFilesMetadata()`) fail fast. It loops through all files and collects all the errors. Please update the java doc to reflect the actual implementation.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFilesVerificationTask.java
```diff
@@ -0,0 +1,643 @@
...
+    {
+        String digestAlgorithm = digestResponse.digestAlgorithm;
+        if (digestAlgorithm.equalsIgnoreCase(MD5Digest.MD5_ALGORITHM))
+        {
+            return Future.succeededFuture(new MD5Digest(digestResponse.digest));
+        }
+        else if (digestAlgorithm.equalsIgnoreCase(XXHash32Digest.XXHASH_32_ALGORITHM))
+        {
+            return Future.succeededFuture(new XXHash32Digest(digestResponse.digest));
+        }
```
yifan-c: Should it be in `DigestAlgorithmFactory`?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/FilesVerificationTaskManager.java
```diff
@@ -0,0 +1,197 @@
...
+                                                                                             String source,
+                                                                                             InstanceMetadata localInstanceMetadata)
+    {
+        String timeUuid = UUIDs.timeBased().toString();
+        return Future.succeededFuture(taskFactory.create(timeUuid,
+                                                         source,
+                                                         sidecarConfiguration.serviceConfiguration().port(),
+                                                         request,
+                                                         localInstanceMetadata));
+    }
```
yifan-c: It is a private method and synchronous. The only reason is to chain `compose`. I think you can avoid wrapping `Future`, and just return `LiveMigrationTask`.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/FilesVerificationTaskManager.java
```diff
@@ -0,0 +1,197 @@
...
+                   {
+                       newTask.start();
+                       LOGGER.info("Accepted new files digest verification task for instance={} taskId={}",
+                                   localInstanceMetadata.id(), newTask.id());
+                       return Future.succeededFuture(newTask);
+                   }
+                   else
+                   {
+                       return Future.failedFuture(new LiveMigrationTaskInProgressException(
+                       "Another files digests verification is in progress for instance=" + localInstanceMetadata.id()));
```
yifan-c:
```suggestion
                       "Another files digest verification is in progress for instance=" + localInstanceMetadata.id()));
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFilesVerificationTask.java
```diff
@@ -0,0 +1,643 @@
...
+               .onFailure(cause -> LOGGER.error("{} Failed to verify file {}", logPrefix, fileInfo.fileUrl, cause));
+    }
+
+    private Future<Digest> getSourceFileDigest(InstanceFileInfo fileInfo)
+    {
+        return Future.fromCompletionStage(sidecarClient.liveMigrationFileDigestAsync(new SidecarInstanceImpl(source, port),
+                                                                                     fileInfo.fileUrl,
+                                                                                     request.digestAlgorithm()))
+                     .compose(this::toDigest);
+    }
```
yifan-c, line 324: It makes 1 http request to source per file to get the digest. According to `LiveMigrationConcurrencyLimitHandler`, `TOO_MANY_REQUESTS` can be thrown. There is no retry implemented to handle it, due to `SingleInstanceSelectionPolicy` + default retry policy. I think you want to add custom retry policy for the applicable requests. 

Beside no retry and fail silently, 1 request per file seems to ensure slowness already. Maybe we should revisit the design decision later.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateFilesVerificationTaskHandler.java
nvharikrishna: CONFLICT makes more sense. Updated it.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateFilesVerificationTaskHandler.java
nvharikrishna: De-duplicated the checks

server/src/main/java/org/apache/cassandra/sidecar/livemigration/FilesVerificationTaskManager.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/FilesVerificationTaskManager.java
nvharikrishna: make sense, updated it

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFilesVerificationTask.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFilesVerificationTask.java
nvharikrishna: Felt both Digest and DigestResponse are not very relative to DigestAlgorithm. So, did not placed it DigestAlgorithmFactory. Moved it to DigestResponse.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationTaskManager.java
nvharikrishna: Fetching task only once and returning the value.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFilesVerificationTask.java
nvharikrishna: I intended to say that the task won’t proceed to verifying file digests and will stop here if there are metadata mismatches. However, I understand your point. I’ve updated the text to avoid any confusion.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFilesVerificationTask.java
nvharikrishna: After giving some more thought, I feel the 429 status code (TOO_MANY_REQUESTS) is not appropriate and the 503 (SERVICE UNAVAILABLE) is more appropriate as maxConcurrentFileRequests (used by `LiveMigrationConcurrencyLimitHandler`) is a server-side concurrency cap shared across all clients, and an individual client hitting this limit hasn't done anything wrong - the server is simply busy with other requests. 503 describes this situation. 

SidecarclientProvider is using `ExponentialBackoffRetryPolicy` as the default retry policy, which does retries. If we suspect that the default retry policy can be overridden, then I can explicitly initiate an instance of ExponentialBackoffRetryPolicy and use it.

What do you think?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/FilesVerificationTaskManager.java
nvharikrishna: Started with inheriting LiveMigrationTaskManager, but had to switch to association. Since FilesVerificationTaskManager and DataCopyTaskManager are different types, Guice creates a separate instance for each, meaning each would get its own currentTasks map (instance to task map in LiveMigrationTaskManager). This breaks the invariant that only one task of any type can be active per instance at a time (line 44).  Don't want to use a static mutable map for inheritance. So, used association.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFilesVerificationTask.java
yifan-c: 5xx status code are considered server side error (unrelated to client activities). 
4xx status codes are errors triggered by client. 
IMO, 429 makes sense in this scenario. The error happens **only** when there are enough concurrent requests are issued by client. 
The default retry policy does not react on 429 status code and no retries are performed.

nvharikrishna: The concurrency limit used here (`sidecarConfiguration.liveMigrationConfiguration().maxConcurrentFileRequests()`) is global across all clients, not per-client. A client sending its very first request can get rejected because other clients have exhausted the pool.

429 means "the user has sent too many requests" — implying the specific client is at fault, which I feel misleading here. 503 could also mean "the server is unable to handle the request due to a temporary overload" - which I think describes the situation in this case.

If this were a per-client throttle, 429 would be correct. Since maxConcurrentFileRequests is a global limit protecting the source node, I feel 503 is more accurate.

Happy to change it to 429 if you feel differently though! Open to any other suggestions as well.

CHANGES.txt
```diff
@@ -1,5 +1,6 @@
 0.3.0
 -----
+ * Adding endpoint for verifying files post data copy during live migration (CASSSIDECAR-226)
```
frankgh, line 3: should move under 0.4.0

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/DigestResponse.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @JsonCreator
+    public DigestResponse(@JsonProperty("digest") String digest,
+                          @JsonProperty("digestAlgorithm") String digestAlgorithm)
+    {
+        Objects.requireNonNull(digest, "digest is required");
+        Objects.requireNonNull(digestAlgorithm, "digestAlgorithm is required");
+        this.digest = digest;
+        this.digestAlgorithm = digestAlgorithm;
+    }
```
frankgh, line 46: very small NIT, no need to change it
```suggestion
        this.digest = Objects.requireNonNull(digest, "digest is required");
        this.digestAlgorithm = Objects.requireNonNull(digestAlgorithm, "digestAlgorithm is required");
```

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/LiveMigrationFilesVerificationResponse.java
```diff
@@ -0,0 +1,186 @@
...
+                                                  @JsonProperty("metadataMismatches") int metadataMismatches,
+                                                  @JsonProperty("digestMismatches") int digestMismatches,
+                                                  @JsonProperty("digestVerificationFailures") int digestVerificationFailures,
+                                                  @JsonProperty("filesMatched") int filesMatched)
+    {
+        Objects.requireNonNull(id, "id of files verification task must be specified");
+        Objects.requireNonNull(state, "state of files verification task must be specified");
+        this.id = id;
+        this.digestAlgorithm = digestAlgorithm;
+        this.state = state;
```
frankgh: Small NIT, feel free to ignore:
```suggestion
        this.id = Objects.requireNonNull(id, "id of files verification task must be specified");
        this.digestAlgorithm = digestAlgorithm;
        this.state = Objects.requireNonNull(state, "state of files verification task must be specified");
```

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/LiveMigrationTaskCreationResponse.java
```diff
@@ -0,0 +1,66 @@
...
+
+    @JsonCreator
+    public LiveMigrationTaskCreationResponse(@JsonProperty("taskId") String taskId,
+                                             @JsonProperty("statusUrl") String statusUrl)
+    {
+        Objects.requireNonNull(taskId, "taskId cannot be null");
+        Objects.requireNonNull(statusUrl, "statusUrl cannot be null");
+
+        this.taskId = taskId;
+        this.statusUrl = statusUrl;
```
frankgh: Small NIT, feel free to ignore
```suggestion
        this.taskId = Objects.requireNonNull(taskId, "taskId cannot be null");
        this.statusUrl = Objects.requireNonNull(statusUrl, "statusUrl cannot be null");
```

client-common/src/test/java/org/apache/cassandra/sidecar/common/response/LiveMigrationFilesVerificationResponseTest.java
```diff
@@ -0,0 +1,322 @@
...
+        assertThat(deserialized.source()).isEqualTo(original.source());
+        assertThat(deserialized.port()).isEqualTo(original.port());
+        assertThat(deserialized.filesNotFoundAtSource()).isEqualTo(original.filesNotFoundAtSource());
+        assertThat(deserialized.filesNotFoundAtDestination()).isEqualTo(original.filesNotFoundAtDestination());
+        assertThat(deserialized.metadataMatched()).isEqualTo(original.metadataMatched());
+        assertThat(deserialized.metadataMismatches()).isEqualTo(original.metadataMismatches());
+        assertThat(deserialized.digestMismatches()).isEqualTo(original.digestMismatches());
+        assertThat(deserialized.digestVerificationFailures()).isEqualTo(original.digestVerificationFailures());
+        assertThat(deserialized.filesMatched()).isEqualTo(original.filesMatched());
+        assertThat(deserialized.isVerificationSuccessful()).isEqualTo(original.isVerificationSuccessful());
```
frankgh: Let's move to a helper method and reuse for other tests.

client-common/src/test/java/org/apache/cassandra/sidecar/common/response/LiveMigrationFilesVerificationResponseTest.java
```diff
@@ -0,0 +1,322 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+class LiveMigrationFilesVerificationResponseTest
+{
+    private final ObjectMapper mapper = new ObjectMapper();
+
+    @Test
+    void testSerializationRoundTrip() throws Exception
```
frankgh: we can simplify with a ParameterizedTest

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCancelFilesVerificationTaskHandler.java
```diff
@@ -0,0 +1,112 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import static org.apache.cassandra.sidecar.acl.authorization.BasicPermissions.DATA_COPY;
+import static org.apache.cassandra.sidecar.utils.HttpExceptions.wrapHttpException;
+
+/**
+ * Handler for canceling an active live migration files verification task.
+ * Accepts a task ID and cancels the corresponding verification task on the specified host.
+ */
+public class LiveMigrationCancelFilesVerificationTaskHandler extends AbstractHandler<String> implements AccessProtected
```
frankgh, line 52: Annotate with Singleton
```suggestion
@Singleton
public class LiveMigrationCancelFilesVerificationTaskHandler extends AbstractHandler<String> implements AccessProtected
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCancelFilesVerificationTaskHandler.java
```diff
@@ -0,0 +1,112 @@
...
+                                                           FilesVerificationTaskManager taskManager)
+    {
+        super(metadataFetcher, executorPools, validator);
+        this.taskManager = taskManager;
+    }
+
+    @Override
+    protected String extractParamsOrThrow(RoutingContext context)
+    {
+        return context.pathParam("taskId");
```
frankgh: can we do some validation of this parameter here? I don't think it can be null, but can it be empty? Is there some parameter validation that can be done ?

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationConcurrencyLimitHandler.java
```diff
@@ -0,0 +1,66 @@
...
+        new ConcurrencyLimiter(() -> sidecarConfiguration.liveMigrationConfiguration().maxConcurrentFileRequests());
+    }
+
+    @Override
+    public void handle(RoutingContext rc)
+    {
+        if (!concurrencyLimiter.tryAcquire())
+        {
+            LOGGER.warn("Too many concurrent live migration file requests. Path={}", rc.request().path());
+            rc.fail(wrapHttpException(HttpResponseStatus.SERVICE_UNAVAILABLE,
```
frankgh, line 58: Should the status code be a 429 instead? maybe we should follow the implementation of `org.apache.cassandra.sidecar.utils.FileStreamer#acquire` ?

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateFilesVerificationTaskHandler.java
```diff
@@ -0,0 +1,161 @@
...
+import static io.netty.handler.codec.http.HttpResponseStatus.SERVICE_UNAVAILABLE;
+import static org.apache.cassandra.sidecar.acl.authorization.BasicPermissions.DATA_COPY;
+import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.LIVE_MIGRATION_FILES_VERIFICATION_TASKS_ROUTE;
+import static org.apache.cassandra.sidecar.utils.HttpExceptions.wrapHttpException;
+
+/**
+ * HTTP handler for creating file digest verification tasks during live migration.
+ * Manages concurrent verification tasks per instance and orchestrates the verification process.
+ */
+public class LiveMigrationCreateFilesVerificationTaskHandler extends AbstractHandler<LiveMigrationFilesVerificationRequest> implements AccessProtected
```
frankgh, line 62:
```suggestion
@Singleton
public class LiveMigrationCreateFilesVerificationTaskHandler extends AbstractHandler<LiveMigrationFilesVerificationRequest> implements AccessProtected
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateFilesVerificationTaskHandler.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    @Override
+    protected void handleInternal(RoutingContext context,
+                                  HttpServerRequest httpRequest,
+                                  @NotNull String host,
+                                  SocketAddress remoteAddress,
+                                  LiveMigrationFilesVerificationRequest request)
+    {
+        LOGGER.debug("Received files verification request for host {} with maxConcurrency {}",
```
frankgh: this log entry might be redundant, as we already have it in `org.apache.cassandra.sidecar.handlers.AbstractHandler#handle` where we log the request. I guess we just need to make sure that the `request` has a `toString()` method

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateFilesVerificationTaskHandler.java
```diff
@@ -0,0 +1,161 @@
...
+        try
+        {
+            localInstanceMetadata = metadataFetcher.instance(host);
+        }
+        catch (NoSuchCassandraInstanceException e)
+        {
+            LOGGER.error("Failed to fetch instance metadata for host={}", host);
+            context.fail(wrapHttpException(SERVICE_UNAVAILABLE, e));
+            return;
+        }
```
frankgh, line 124: There's no need to handle these exceptions here. It is already explicitly handled in `org.apache.cassandra.sidecar.handlers.AbstractHandler#determineHttpException`
```suggestion
        InstanceMetadata localInstanceMetadata = metadataFetcher.instance(host);
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationDigestHandlerWrapper.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.cassandra.sidecar.common.request.LiveMigrationFileDigestRequest;
+
+import static org.apache.cassandra.sidecar.common.request.LiveMigrationFileDigestRequest.DIGEST_ALGORITHM_PARAM;
+
+/**
+ * Wrapper handler that conditionally delegates to {@link LiveMigrationFileDigestHandler} based on the presence
+ * of the {@link LiveMigrationFileDigestRequest#DIGEST_ALGORITHM_PARAM} query parameter, otherwise
+ * passes control to the next handler in the chain.
+ */
+public class LiveMigrationDigestHandlerWrapper implements Handler<RoutingContext>
```
frankgh, line 35:
```suggestion
@Singleton
public class LiveMigrationDigestHandlerWrapper implements Handler<RoutingContext>
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileDigestHandler.java
```diff
@@ -0,0 +1,131 @@
...
+        String digestAlgorithmParam = getDigestAlgorithmParam(context);
+        try
+        {
+            return digestAlgorithmFactory.getDigestAlgorithm(digestAlgorithmParam, 0);
+        }
+        catch (IllegalArgumentException e)
+        {
+            LOGGER.error("Unexpected error while getting digest algorithm for {}", digestAlgorithmParam, e);
+            throw wrapHttpException(HttpResponseStatus.BAD_REQUEST, e.getMessage());
+        }
```
frankgh: I think it's redundant to handle the `IllegalArgumentException` here. This will be correctly handled on an exception here: `org.apache.cassandra.sidecar.handlers.AbstractHandler#processFailure`
```suggestion
        String digestAlgorithmParam = getDigestAlgorithmParam(context);
        return digestAlgorithmFactory.getDigestAlgorithm(digestAlgorithmParam, 0);
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileDigestHandler.java
```diff
@@ -0,0 +1,131 @@
...
+        String file = context.get(FileStreamHandler.FILE_PATH_CONTEXT_KEY);
+
+        if (file == null)
+        {
+            LOGGER.error("File path not found in context");
+            context.fail(wrapHttpException(INTERNAL_SERVER_ERROR, "File path not available"));
+            return;
+        }
+        calculateDigest(vertx, file, digestAlgorithm)
+        .onComplete(ar -> {
```
frankgh, line 99: Why don't we re-use the existing machinery for digest verification? it feels like we are rewriting the same work instead of adjusting to the existing implementation.

```
        DigestVerifier digestVerifier = digestVerifierFactory.verifier(httpRequest.headers());
        digestVerifier.verify(file)
````

File: server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationGetAllFilesVerificationTasksHandler.java [hunk 27]
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.cassandra.sidecar.livemigration.LiveMigrationTask;
+
+import static org.apache.cassandra.sidecar.acl.authorization.BasicPermissions.DATA_COPY;
+import static org.apache.cassandra.sidecar.handlers.AbstractHandler.extractHostAddressWithoutPort;
+
+/**
+ * Handler for retrieving all active live migration files verification tasks.
+ * Returns a list of all verification tasks running on the local host.
+ */
+public class LiveMigrationGetAllFilesVerificationTasksHandler implements Handler<RoutingContext>, AccessProtected
```
frankgh, line 43:
```suggestion
@Singleton
public class LiveMigrationGetAllFilesVerificationTasksHandler implements Handler<RoutingContext>, AccessProtected
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationGetFilesVerificationTaskHandler.java
```diff
@@ -0,0 +1,110 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import static org.apache.cassandra.sidecar.acl.authorization.BasicPermissions.DATA_COPY;
+import static org.apache.cassandra.sidecar.utils.HttpExceptions.wrapHttpException;
+
+/**
+ * Handler for retrieving a specific live migration files verification task by task ID.
+ * Returns the task details if found, or a 404 error if the task does not exist on the specified host.
+ */
+public class LiveMigrationGetFilesVerificationTaskHandler extends AbstractHandler<String> implements AccessProtected
```
frankgh, line 52:
```suggestion
@Singleton
public class LiveMigrationGetFilesVerificationTaskHandler extends AbstractHandler<String> implements AccessProtected
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationGetFilesVerificationTaskHandler.java
```diff
@@ -0,0 +1,110 @@
...
+    protected void handleInternal(RoutingContext context,
+                                  HttpServerRequest httpRequest,
+                                  @NotNull String host,
+                                  SocketAddress remoteAddress,
+                                  String taskId)
+    {
+        try
+        {
+            LiveMigrationTask<LiveMigrationFilesVerificationResponse> task = taskManager.getTask(taskId, host);
+            LOGGER.info("Found live migration task with taskId={} on host={}", taskId, host);
```
frankgh: seems too much, can we lower to debug?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFilesVerificationTaskFactory.java
```diff
@@ -0,0 +1,78 @@
...
+                                                                            LiveMigrationFilesVerificationRequest request,
+                                                                            InstanceMetadata localInstanceMetadata)
+    {
+        return LiveMigrationFilesVerificationTask.builder()
+                                                 .id(id)
+                                                 .source(source)
+                                                 .port(port)
+                                                 .vertx(vertx)
+                                                 .executorPools(executorPools)
+                                                 .sidecarClient(sidecarClientProvider.get())
```
frankgh, line 71: What's the side-effect of calling `sidecarClientProvider.get()`? I don't remember off the top of my head. Can you ensure that calling `sidecarClientProvider.get()` does not create a new client? We should reuse a single client throughout the application.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationTaskManager.java
```diff
@@ -0,0 +1,152 @@
...
+     * Attempts to submit a new task for the specified instance.
+     * Only one task (of any type) can be active per instance at a time.
+     *
+     * @param instanceId the instance ID
+     * @param newTask    the task to submit
+     * @return true if the task was accepted, false if another task is already in progress
+     */
+    public boolean submitTask(int instanceId, LiveMigrationTask<?> newTask)
+    {
+        return currentTasks.compute(instanceId, (integer, taskInMap) -> {
```
frankgh: NIT
```suggestion
        return currentTasks.compute(instanceId, (ignored, taskInMap) -> {
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationTaskManager.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    @SuppressWarnings("ConstantValue")
+    public List<LiveMigrationTask<?>> getAllTasks(@NotNull String currentHost)
+    {
+        InstanceMetadata localInstance = instancesMetadata.instanceFromHost(currentHost);
+        if (localInstance == null)
+        {
+            throw new IllegalStateException("No instance found for host: " + currentHost);
+        }
+
```
frankgh: this check is not required. instanceFromHost returns non-null results

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationTaskManager.java
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    @SuppressWarnings("ConstantValue")
+    private LiveMigrationTask<?> getLiveMigrationTask(@NotNull String taskId, @NotNull String currentHost)
+    {
+        InstanceMetadata localInstance = instancesMetadata.instanceFromHost(currentHost);
+        if (localInstance == null)
+        {
+            throw new IllegalStateException("No instance found for host: " + currentHost);
+        }
```
frankgh: not necessary to do a null check here
```suggestion
```

frankgh (commented):
Looks good in general. I've added a few comments

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/DigestResponse.java
nvharikrishna: Done

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/LiveMigrationFilesVerificationResponse.java
nvharikrishna: done

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/LiveMigrationTaskCreationResponse.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCancelFilesVerificationTaskHandler.java
nvharikrishna: done

client-common/src/test/java/org/apache/cassandra/sidecar/common/response/LiveMigrationFilesVerificationResponseTest.java
nvharikrishna: done

client-common/src/test/java/org/apache/cassandra/sidecar/common/response/LiveMigrationFilesVerificationResponseTest.java
nvharikrishna: Now it appears only once after using parameterised tests.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCancelFilesVerificationTaskHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateFilesVerificationTaskHandler.java
nvharikrishna: removed it

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateFilesVerificationTaskHandler.java
nvharikrishna: Wanted to throw 503 instead of 500. So catching it and wrapping it into HttpException.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationDigestHandlerWrapper.java
nvharikrishna: Realised that I missed to mark multiple other handlers as Singleton. Marked other handlers too as Singleton.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileDigestHandler.java
nvharikrishna: Done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileDigestHandler.java
nvharikrishna: > digestVerifier.verify(file)

Is returning path and 'DigestVerifier' is validating path against known digest. So re-using `org.apache.cassandra.sidecar.utils.AsyncFileDigestCalculator#calculateDigest(io.vertx.core.Vertx, java.lang.String, org.apache.cassandra.sidecar.utils.DigestAlgorithm)` which is used by DigestVerifier too.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationGetAllFilesVerificationTasksHandler.java
nvharikrishna: Done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationGetFilesVerificationTaskHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationGetFilesVerificationTaskHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFilesVerificationTaskFactory.java
nvharikrishna: SidecarClientProvider is Singleton and `.get()` is returning its internal initialized client. Only one client instance is used.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationTaskManager.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationTaskManager.java
nvharikrishna: removed

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationCreateFilesVerificationTaskHandler.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationConcurrencyLimitHandler.java
nvharikrishna: It is discussed here: https://github.com/apache/cassandra-sidecar/pull/309#discussion_r2868621502

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationTaskManager.java
nvharikrishna: removed

frankgh: I still see the null check here, can you check?

nvharikrishna: Somehow missed to push this change, pushed it now.

frankgh (approved):
+1 thanks for the patch!

yifan-c (approved):

PR #312, opened 2026-02-03
CASSSIDECAR-399 Adding storage_port to instance configuration and metadata

nvharikrishna:
Added support for configuring the storage_port in Cassandra instance metadata and configuration. This enhancement exposes the storage port through the InstanceMetadata and InstanceConfiguration interfaces, allowing Sidecar to be aware of and utilize this configuration parameter.

For CASSSIDECAR-399


server/src/test/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImplTest.java
```diff
@@ -200,6 +200,56 @@ void testResolvesDefaultDirectories()
...
+                                                        .host(HOST)
+                                                        .port(PORT)
+                                                        .metricRegistry(METRIC_REGISTRY)
+                                                        .storageDir(rootDir)
+                                                        .build();
+        assertThat(metadata.storagePort()).isEqualTo(7000);
+    }
+
+    @Test
+    void testCustomStoragePort()
```
frankgh: NIT: prefer parameterized test
```suggestion
    @ParameterizedTest(name = "{index} => storagePort={0}")
    @ValueSource(ints = { 1, 65535, 7005 })
    void testCustomStoragePort(int customStoragePort)
```

server/src/test/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImplTest.java
```diff
@@ -200,6 +200,56 @@ void testResolvesDefaultDirectories()
...
+    {
+        String rootDir = tempDir.toString();
+        for (int customStoragePort : new int[]{ 1, 65535, 7005 })
+        {
```
frankgh: Not needed when using a parameterized test
```suggestion
```

server/src/test/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImplTest.java
```diff
@@ -200,6 +200,56 @@ void testResolvesDefaultDirectories()
...
+                                                            .storagePort(customStoragePort)
+                                                            .metricRegistry(METRIC_REGISTRY)
+                                                            .storageDir(rootDir)
+                                                            .build();
+            assertThat(metadata.storagePort()).isEqualTo(customStoragePort);
+        }
+    }
+
+    @Test
+    void testInvalidStoragePort()
```
frankgh: NIT: prefer parameterized test
```suggestion
    @ParameterizedTest(name = "{index} => invalid storagePort={0}")
    @ValueSource(ints = { 0, -1, 65536 })
```

frankgh (commented):
Looks good! I have a small suggestion

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
```diff
@@ -237,6 +254,7 @@ public static class Builder implements DataObjectBuilder<Builder, InstanceMetada
         protected String host;
         protected String ipAddress;
         protected int port;
+        protected int storagePort = DEFAULT_STORAGE_PORT;
```
skoppu22: I view the Sidecar as a service exposing data/config of Cassandra servers in the cluster, so developers/users can work with Cassandra accordingly. Is it OK to use Sidecar's local DEFAULT_STORAGE_PORT value here? Why don't we just give whatever we got from the Cassandra server? If we get null/nothing from the server, we can either set something like -1 or error message in the Sidecar's response, somehow indicating to the caller that Sidecar unable to get this value from the server? My perspective is, using Sidecar's own local values may mislead developers/users while debugging an issue with Cassandra.

server/src/test/java/org/apache/cassandra/sidecar/config/SidecarConfigurationTest.java
```diff
@@ -588,6 +588,41 @@ void testRepairConfiguration() throws IOException
...
+                                     "    jmx_host: 127.0.0.1\n" +
+                                     "    jmx_port: 7199\n" +
+                                     "    jmx_ssl_enabled: false";
+        SidecarConfigurationImpl configWithStoragePort = SidecarConfigurationImpl.fromYamlString(yamlWithStoragePort);
+        assertThat(configWithStoragePort.cassandraInstances()).isNotNull().hasSize(1);
+        InstanceConfiguration instance1 = configWithStoragePort.cassandraInstances().get(0);
+        assertThat(instance1.storagePort()).isEqualTo(7001);
+
+        // Test without storage_port (should return null)
+        String yamlWithoutStoragePort = "cassandra_instances:\n" +
```
skoppu22, line 616: Can we split this second part into a different test function, like testYamlWithoutStoragePort() etc, that will make clear to the reader that we are testing this scenario?

server/src/test/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImplTest.java
nvharikrishna: done

server/src/test/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImplTest.java
nvharikrishna: done

server/src/test/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImplTest.java
nvharikrishna: done

server/src/test/java/org/apache/cassandra/sidecar/config/SidecarConfigurationTest.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
nvharikrishna: Thanks for the thoughtful review! Let me share the reasoning behind using DEFAULT_STORAGE_PORT:

1. Independence from Cassandra process availability

Goal is to let sidecar function even when Cassandra process is not running or unreachable. This is particularly important in scenarios where:
- Sidecar is started before Cassandra
- Cassandra is temporarily unavailable
- During initialization or live migration (CEP-40) when Cassandra may not yet be part of the ring

2. 7000 port - following Cassandra's default value

The DEFAULT_STORAGE_PORT = 7000 is not Sidecar's own local value but rather the default value that Cassandra itself uses (defined in cassandra.yaml and Config.java). By using this default, it is mirroring Cassandra's own behavior when the value isn't explicitly configured. This should actually help developers since it matches Cassandra's defaults. Of course, if an operator has configured a non-standard storage_port in their Cassandra instance, they should configure the same value in Sidecar's configuration.

Cassandra uses IP + port for node identification, and Sidecar needs to correctly identify nodes even when Cassandra is temporarily unreachable. Using null, -1, or error messages when we can't reach Cassandra would actually hinder operations in the above scenarios. I feel the default value simply provides a sensible fallback.

Does this reasoning make sense?

frankgh (approved):
+1 thanks for the patch

conf/sidecar.yaml
```diff
@@ -24,6 +24,11 @@ cassandra_instances:
     host: localhost1
     port: 9042
 
+    # The storage port used for inter-node communication between Cassandra nodes in a cluster.
+    # This port is used for internal gossip protocol, data replication, and repair operations.
+    # Default: 7000
+    #storage_port: 7000
```
yifan-c, line 30: Appreciate the document

server/src/main/java/org/apache/cassandra/sidecar/cluster/instance/InstanceMetadataImpl.java
skoppu22: Thanks for details.

skoppu22 (approved):
Looks good

nvharikrishna:
FYI, just now pushed a changes to initialize default storage port in InstanceConfigurationImpl class (instead of InstanceMetadataImpl) to inline with other configuration implementations.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
```diff
@@ -33,9 +33,15 @@
...
 {
+    /**
+     * Default storage port for Cassandra inter-node communication.
+     */
+    public static final int DEFAULT_STORAGE_PORT = 7000;
+
     protected final int id;
     protected final String host;
     protected final int port;
+    protected final Integer storagePort;
```
frankgh: should this be int now that it's always provided?

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
```diff
@@ -119,6 +127,17 @@ public int port()
         return port;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @JsonProperty("storage_port")
+    @Nullable
+    public Integer storagePort()
```
frankgh: can we make this int as well? And the interface should also be a primitive

frankgh (commented):
I have a couple of comments after the changes

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
nvharikrishna: Makes sense, done

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/InstanceConfigurationImpl.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/modules/ConfigurationModule.java
```diff
@@ -240,21 +240,29 @@ private static InstanceMetadata buildInstanceMetadata(Vertx vertx,
...
+                                                                    .id(cassandraInstance.id())
+                                                                    .host(host, dnsResolver)
+                                                                    .port(port);
+
+        // Only set storagePort if explicitly configured, otherwise use builder default
+        Integer configuredStoragePort = cassandraInstance.storagePort();
+        if (configuredStoragePort != null)
+        {
+            builder.storagePort(configuredStoragePort);
+        }
```
frankgh: is this check still applicable? `configuredStoragePort` should never be null

yifan-c (commented):
Looks good. Waiting for green CI.

server/src/main/java/org/apache/cassandra/sidecar/modules/ConfigurationModule.java
nvharikrishna: Right, not applicable as storagePort is a primitive now. Pushed the changes.

PR #313, opened 2026-02-04
CASSSIDECAR-400- Fixing sidecar’s ProcessLifecycleProviderIntegrationTest() failure in root mode

shruti-p-s:
Patch by Shruti Sekaran; Reviewed by TBD for [CASSSIDECAR-400](https://issues.apache.org/jira/browse/CASSSIDECAR-400)

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessRuntimeConfiguration.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessRuntimeConfiguration.java
@@ -53,7 +53,7 @@ public class ProcessRuntimeConfiguration
     private final Map<String, String> extraJvmOptions;
     private final Map<String, String> extraEnvironmentVariables;
 
-    private ProcessRuntimeConfiguration(Builder builder)
+    protected ProcessRuntimeConfiguration(Builder builder)
     {
         instance = builder.instance;
         cassandraHome = Path.of(builder.cassandraHome);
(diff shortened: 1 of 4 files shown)


server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessRuntimeConfiguration.java
```diff
@@ -128,6 +128,7 @@ public ProcessBuilder buildStartCommand(String pidFileLocation, Path stdoutFileL
         startCassandraCmd.add(cassandraBin().toString());
         startCassandraCmd.add("-p");
         startCassandraCmd.add(pidFileLocation);
+        startCassandraCmd.add("-R"); // Allow running as root (required for Cassandra 5.0+ when running as root)
```
frankgh: I think adding this flag to production code is dangerous. Looking at [CASSANDRA-8142](https://issues.apache.org/jira/browse/CASSANDRA-8142) the idea is to prevent this behavior by default for good reasons. I would propose that we alternatively look into having the ability to provide optional extra args that we can configure for testing purposes.

frankgh (commented):
I have some concerns about adding the flag by default.

server/src/main/java/org/apache/cassandra/sidecar/lifecycle/ProcessRuntimeConfiguration.java
pauloricardomg: +1

shruti-p-s: Makes sense, I've updated the argument as a test override

sarankk (approved):
+1 Thanks Shruti, Lgtm

frankgh (approved):
+1 thanks for the patch

PR #315, opened 2026-02-11
CASSSIDECAR-401 Incorrect SSL Configuration Keys in CdcPublisher.secretsProvider()

jyothsnakonisa:


bbotella (approved):
+1

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
```diff
@@ -165,27 +153,20 @@ public SecretsProvider secretsProvider()
...
-        sslConfigMap.put(SSL_SECURE_TRANSPORT_PROTOCOLS_KEY, String.join(",", sslConfiguration.secureTransportProtocols()));
-        sslConfigMap.put(SSL_HANDSHAKE_TIMEOUT_KEY, sslConfiguration.handshakeTimeout().toString());
-
         if (sslConfiguration.isKeystoreConfigured())
         {
             KeyStoreConfiguration keystore = sslConfiguration.keystore();
-            sslConfigMap.put(SSL_KEYSTORE_PATH_KEY, keystore.path());
-            sslConfigMap.put(SSL_KEYSTORE_PASSWORD_KEY, keystore.password());
-            sslConfigMap.put(SSL_KEYSTORE_TYPE_KEY, keystore.type());
+            sslConfigMap.put(MapUtils.lowerCaseKey(SslConfig.KEYSTORE_PATH), keystore.path());
```
jmckenzie-dev: My own ignorance here - why do we `lowerCaseKey` them here? Asking claude indicates Spark's DataSource V2 API uses CaseInsensitiveStringMap which tracks from poking around at that code. Makes me wonder why we have all the `MapUtils.java` stuff in analytics instead of just using [CaseInsensitiveStringMap](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/map/CaseInsensitiveMap.html) ourselves... 

Certainly beyond the scope of this patch but might be worth some follow up thought if it'd clean up the codebase.

jmckenzie-dev: Might be worth adding a comment here as to why we're moving things lowercase like this since we have a split in locality between 2 codebases w/this formatting dependency. And/or just use the CaseInsensitiveStringMap on our `sslConfigMap` here if you wanted to. 🤷🏻 

Not a hard requirement at all - PR is good the way it is. Just trying to think through how to simplify things.

jmckenzie-dev (approved):
Nit / questions but nothing blocking. Looks good - solid fix and good test coverage. Good work!

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
jyothsnakonisa: `MapUtils.getOrDefault` method looks for lowercase keys and hence we have to convert keys to lowercase and pass it to SslConfig class in `cassandra-analytics`. I don't see a reason why MapUtils class should do that, may be we can clean these things up in the future.

jyothsnakonisa: Thanks for the suggestion to use a caseinsensitive map, that's much cleaner.

PR #316, opened 2026-02-11
CASSSIDECAR-406: Fix ClusterTopologyMonitor delay parsing error due to missing time unit suffix

jyothsnakonisa:


yifan-c (approved):

bbotella (approved):

conf/sidecar.yaml
```diff
@@ -344,6 +339,13 @@ healthcheck:
   initial_delay: 0ms
   execute_interval: 30s
 
+# Cluster Topology Monitor settings
+# Monitors Cassandra cluster topology changes and publishes events when detected
+cluster_topology_monitor:
+  enabled: true
+  initial_delay: 0s
+  execute_interval: 1000ms  # Interval between topology refresh cycles
```
jmckenzie-dev, line 352: What led to choosing once per second as our default? Is this potentially too high or too low? Coming into this naively it's not clear to me who the consumers of this information are and whether once per second is sufficient for their needs, overkill, etc.

Actionable feedback beyond my questions: adding a little more context on the comment above as to what this information is and how it's used would help provide context on why the values listed were selected. There's a lot of this in `cassandra.yaml` in the cassandra codebase if you want to see some examples. Kind of commenting on the "why" not just the "what".

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/SidecarConfigurationImpl.java
```diff
@@ -415,6 +429,10 @@ public static class Builder implements DataObjectBuilder<Builder, SidecarConfigu
         = new PeriodicTaskConfigurationImpl(true,
                                             MillisecondBoundConfiguration.ZERO,
                                             MillisecondBoundConfiguration.parse("30s"));
+        private PeriodicTaskConfiguration clusterTopologyMonitorConfiguration
+        = new PeriodicTaskConfigurationImpl(true,
+                                            MillisecondBoundConfiguration.ZERO,
+                                            MillisecondBoundConfiguration.parse("1000ms"));
```
jmckenzie-dev, line 435: If we have raw values in `sidecar.yaml` but then have a duplicate / DRY violation here where we repeat the value, this is going to be a potential source of future bugs. Not recommending you break w/tradition here, just that maybe we consider a follow up radar to have some of these values pull from a parsed .yaml file instead of duplicating the values hard-coded like this.

To be fair, as I think about this I realize we largely do the same thing in cassandra in `Config.java`. ☹️ So maybe we just keep things as they are. 🤷🏻

server/src/test/java/org/apache/cassandra/sidecar/tasks/ClusterTopologyMonitorTest.java
```diff
@@ -0,0 +1,495 @@
...
+        assertThat(customMonitor.initialDelay()).isEqualTo(MillisecondBoundConfiguration.parse("0s"));
+        assertThat(customMonitor.delay()).isEqualTo(MillisecondBoundConfiguration.parse("2000s"));
+    }
+
+    @Test
+    void testConfigurationFromYaml() throws IOException
+    {
+        // When running tests from the server module, working directory is the server directory
+        // So we need to go up one level to find conf/sidecar.yaml at project root
+        Path configPath = Paths.get("../conf/sidecar.yaml");
```
jmckenzie-dev: Will this relative pathing always hold true? Is there some kind of execution env var we could instead rely on that wouldn't be brittle?

jmckenzie-dev: Thinking something along the lines of "unit test wide usable accessor to get a hold of `sidecar.yaml`".

jmckenzie-dev (approved):
Looks good. Left some stylistic questions / nits, but nothing that I'd block review on.

frankgh:
Please update CHANGES.txt before merging

conf/sidecar.yaml
jyothsnakonisa: @jmckenzie-dev there was a default hardcoded value for the delay prior to this patch which was 1000ms. In this patch I am making the delay configurable and used the older value as default.

I agree, will add more comments around the props and what could be a reasonable value for the props.

server/src/test/java/org/apache/cassandra/sidecar/tasks/ClusterTopologyMonitorTest.java
jyothsnakonisa: @jmckenzie-dev May be not, we can create a new yaml file just for test in test resources.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/SidecarConfigurationImpl.java
jyothsnakonisa: You are right, I followed same convention `clusterTopologyMonitorConfiguration` considering other methods in `SidecarConfigurationImpl`. However these defaults will get replaced by the values in config file here https://github.com/apache/cassandra-sidecar/blob/bab2da8431296a3fe147397683b071d0dd7eafa0/server/src/main/java/org/apache/cassandra/sidecar/CassandraSidecarDaemon.java#L47

PR #317, opened 2026-02-13
CASSSIDECAR-308 CDC: Add end-to-end CDC integration tests

jyothsnakonisa:
Adding Integration test for CDC
- Needs analytics 0.3.0 version for the test to pass and the analytics release in progress
- Refactored code in MtlsTestHelper to reduce code redundancy

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -46,5 +46,5 @@ swaggerVersion=2.2.21
 kryoVersion=4.0.2
 # OSHI dependencies
 oshiVersion=6.9.0
-analyticsVersion=0.100
+analyticsVersion=0.3.0
 kafkaClientVersion=3.7.0
--- a/gradle.properties
+++ b/gradle.properties
@@ -46,5 +46,5 @@ swaggerVersion=2.2.21
 kryoVersion=4.0.2
 # OSHI dependencies
 oshiVersion=6.9.0
-analyticsVersion=0.2.0
+analyticsVersion=0.100
 kafkaClientVersion=3.7.0
--- a/integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
+++ b/integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
@@ -796,8 +796,8 @@ public abstract class SharedClusterIntegrationTestBase
         }
 
         public static SidecarConfigurationImpl.Builder defaultConfigurationBuilder(
-        MtlsTestHelper mtlsTestHelper,
-        Function<SidecarConfigurationImpl.Builder, SidecarConfigurationImpl.Builder> configurationOverrides)
+                MtlsTestHelper mtlsTestHelper,
+                Function<SidecarConfigurationImpl.Builder, SidecarConfigurationImpl.Builder> configurationOverrides)
         {
             ServiceConfiguration conf = TestServiceConfiguration.builder()
                                                                 .schemaKeyspaceConfiguration(SchemaKeyspaceConfigurationImpl.builder()
--- a/integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
+++ b/integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
@@ -36,6 +36,7 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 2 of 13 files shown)


integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
```diff
@@ -775,48 +796,23 @@ public LifecycleProvider lifecycleProvider()
         }
 
         public static SidecarConfigurationImpl.Builder defaultConfigurationBuilder(
-        MtlsTestHelper mtlsTestHelper, Function<SidecarConfigurationImpl.Builder, SidecarConfigurationImpl.Builder> configurationOverrides)
+        MtlsTestHelper mtlsTestHelper,
```
jmckenzie-dev: The indentation here makes me twitch. :) Do we normally do front-of-line like this in this repo? I see it in the prior line you're refactoring so see that you inherit it, but it's definitely not my favorite. Not a blocker, just a question / observation.

integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/TestCdcConfig.java
```diff
@@ -0,0 +1,154 @@
...
+    @Override
+    public String datacenter()
+    {
+        return "datacenter1";
+    }
+
+    @Override
+    public SecondBoundConfiguration watermarkWindow()
+    {
+        return new SecondBoundConfiguration(259200, TimeUnit.SECONDS);
```
jmckenzie-dev: Total nit, but why use a massive # of seconds here instead of `72, TimeUnit.HOURS` or `3, TimeUnit.DAYS`?

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/cdc/CdcIntegrationTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.apache.cassandra.testing.TestUtils.DC1_RF1;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration test for CDC functionality.
+ * Tests that mutations on CDC-enabled tables are captured and published to TestCdcEventConsumer.
+ */
+public class CdcIntegrationTest extends SharedClusterCdcSidecarIntegrationTestBase
```
jmckenzie-dev, line 42: This looks like solid "happy path does this actually work" testing. Do we have tests that randomly generate data, form a model of what we would expect the kafka messages to look like from that data, then run through the CDC pipeline and then confirm the end state to the modeled data?
If not - be a good follow up JIRA. :)

jmckenzie-dev: Also - when we go that route we'll probably want to refactor out some shared helpers from this test to use in both.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/testing/SharedClusterCdcSidecarIntegrationTestBase.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Override
+    protected void beforeClusterProvisioning()
+    {
+        // CDC integration test is only compatible with Cassandra 4.1 due to cassandra-analytics library version
+        SimpleCassandraVersion version = SimpleCassandraVersion.create(testVersion.version());
+        assumeThat(version.major)
+                .as("CDC test requires Cassandra 4.x")
+                .isEqualTo(4);
```
jmckenzie-dev, line 89: Greater than or equal to?

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/testing/SharedClusterCdcSidecarIntegrationTestBase.java
```diff
@@ -0,0 +1,197 @@
...
+        assumeThat(version.minor)
+                .as("CDC test requires Cassandra 4.1")
+                .isEqualTo(1);
```
jmckenzie-dev: Along w/the above comment, we should probably check that overall we're >= 4.1. Probably need some slightly more complex logic here but not much.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/testing/SharedClusterCdcSidecarIntegrationTestBase.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @Override
+    protected Function<SidecarConfigurationImpl.Builder, SidecarConfigurationImpl.Builder> configurationOverrides()
+    {
+        return builder -> {
+            // Override service configuration to use specific port for CDC tests
+            ServiceConfiguration existingConfig = builder.build().serviceConfiguration();
+            ServiceConfiguration cdcServiceConfig = ServiceConfigurationImpl.builder()
+                                                                            .host(existingConfig.host())
+                                                                            .port(9043)  // TODO: Make this port dynamically allocated
```
jmckenzie-dev, line 109: About this... should we go ahead and do this now? Or, if CI is very stable w/the current, have a follow up JIRA for a helper method that'll find an open port and then use it?

gradle.properties
```diff
@@ -46,5 +46,5 @@ swaggerVersion=2.2.21
 kryoVersion=4.0.2
 # OSHI dependencies
 oshiVersion=6.9.0
-analyticsVersion=0.2.0
+analyticsVersion=0.100
```
jmckenzie-dev: As discussed offline, this needs to be updated now that the new analytics version is out.

jmckenzie-dev (changes requested):
Handful of nits. Only non-trivial thing is the version comparison being hard-coded to 4.1 for tests as I think we need >= 4.1 on that comparison.

With that change + the analytics version update it's a 👍🏻 from me.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/testing/SharedClusterCdcSidecarIntegrationTestBase.java
jyothsnakonisa: In cassandra-analytics, we are using port number that is passed in `SidecarCdcClient.ClientConfig` instead of using the port number from `CdcDynamicSidecarInstancesProvider` and hence test is not able to use correct port number in sidecar client when dynamic ports are being used for sidecar. 

This needs a fix in cassandra-analytics, which I made a note of and this fix will go in next release. Until then, it is good to have an integration test(even with fixed port) to validate several cdc related changes that we are making. We can do a followup JIRA for the test to use dynamic port later.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/testing/SharedClusterCdcSidecarIntegrationTestBase.java
jyothsnakonisa: You are right it should be >= here.

integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/TestCdcConfig.java
jyothsnakonisa: Just a default that got copied from other classes, will update it :)

integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/SharedClusterIntegrationTestBase.java
jyothsnakonisa: No we don't its a mistake, thanks for catching it.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/cdc/CdcIntegrationTest.java
jyothsnakonisa: Sure, we can add randomized  integration tests going forward in follow up JIRAs.

jmckenzie-dev (approved):
👍🏻

bbotella (approved):

PR #318, opened 2026-02-13
CASSSIDECAR-407 Pluggable withCredentials provider for file access

burmanm:
Add new settings under driver_parameters to support pluggable withCredentials provider. Add two possible implementations, ConfigProvider (which is the current implementation) and a new one, FileProvider which reads username and password from the disk.

Also deprecate the old username/password parameters, but do not remove support for them (simply transform their input to the ConfigProvider).


server/src/main/java/org/apache/cassandra/sidecar/modules/ConfigurationModule.java
```diff
@@ -121,11 +129,41 @@ CQLSessionProvider cqlSessionProvider(Vertx vertx,
...
+
+        ParameterizedClassConfiguration config = driverConfiguration.authProvider();
+        if (config == null)
+        {
+            // Fallback to the old one
+            return new ConfigProvider(
+                Map.of("username", driverConfiguration.username(), "password", driverConfiguration.password()));
+        }
+
+        Map<String, String> namedParameters = config.namedParameters();
```
burmanm: This looks hacky because it is. Originally I had here:

```java
if (config.namedParameters() == null)
{
    throw new ConfigurationException(...)
}
```

But the findbugs wasn't accepting that, claiming the later access to namedParameters would potentially cause NPE in the next lines in this function. So to please the findbugs, I'm creating an empty namedParameters here and letting the providers throw the ConfigurationException.

frankgh:
Thanks for the patch, I will take a look at it today

server/src/main/java/org/apache/cassandra/sidecar/cluster/auth/CqlAuthProvider.java
```diff
@@ -0,0 +1,28 @@
...
+
+package org.apache.cassandra.sidecar.cluster.auth;
+
+/**
+ * Provides authentication details used by Sidecar's internal Cassandra CQL client.
+ */
+public interface CqlAuthProvider
+{
+    String username();
+    String password();
```
yifan-c, line 34: Please add javadoc, although the method name is self-explanatory.

server/src/main/java/org/apache/cassandra/sidecar/cluster/auth/FileProvider.java
```diff
@@ -0,0 +1,94 @@
...
+  public String username()
+  {
+      return readSecret(usernamePath, USERNAME_PATH_PARAM);
+  }
+
+  @Override
+  public String password()
+  {
+      return readSecret(passwordPath, PASSWORD_PATH_PARAM);
+  }
```
yifan-c, line 93: It reads from the file every single time per call. Do you expect the username/password to change? 

And the indentation is wrong.

server/src/main/java/org/apache/cassandra/sidecar/cluster/auth/ConfigProvider.java
```diff
@@ -0,0 +1,64 @@
...
+  public String username()
+  {
+      return username;
+  }
+
+  @Override
+  public String password()
+  {
+      return password;
+  }
```
yifan-c, line 63: indentation space is wrong - please use 4 spaces.

server/src/main/java/org/apache/cassandra/sidecar/modules/ConfigurationModule.java
```diff
@@ -121,11 +128,45 @@ CQLSessionProvider cqlSessionProvider(Vertx vertx,
...
+
+        if (config.className().equalsIgnoreCase(ConfigProvider.class.getName()))
+        {
+            return new ConfigProvider(namedParameters);
+        }
+        if (config.className().equalsIgnoreCase(FileProvider.class.getName()))
+        {
+            return new FileProvider(namedParameters);
+        }
+        throw new ConfigurationException("Unrecognized cql auth_provider " + config.className() + " set");
```
yifan-c, line 177: To truly make it pluggable, we need to use reflection to construct the instance from the supplied class name, instead of throwing.
For example, 

```suggestion
        try 
        {
            Class<?> cls = Class.forName(config.className());
            return (CqlAuthProvider) cls.getDeclaredConstructor(Map.class).newInstance(namedParameters);
        } 
        catch (ReflectiveOperationException e) 
        {
            throw new ConfigurationException("Failed to instantiate auth_provider " + config.className(), e);
        }
```

server/src/main/java/org/apache/cassandra/sidecar/modules/ConfigurationModule.java
```diff
@@ -121,11 +128,45 @@ CQLSessionProvider cqlSessionProvider(Vertx vertx,
...
+    CqlAuthProvider cqlAuthProvider(SidecarConfiguration sidecarConfiguration)
+    {
+        DriverConfiguration driverConfiguration = sidecarConfiguration.driverConfiguration();
+
+        ParameterizedClassConfiguration config = driverConfiguration.authProvider();
+        if (config == null)
+        {
+          // Fallback to the old one
+          String username = driverConfiguration.username() != null ? driverConfiguration.username() : "";
+          String password = driverConfiguration.password() != null ? driverConfiguration.password() : "";
```
yifan-c: I do not think defaulting to empty string is the old behavior, which allows `null` for username and password.
The reason of defaulting to the empty string seems to only to conform the `Map.of()` constructor. A workaround could be wrap HashMap (which permits `null` value) with unmodifiable.

yifan-c (commented):
Thanks for the patch! It is a good idea to have the auth provider pluggable.

server/src/main/java/org/apache/cassandra/sidecar/modules/ConfigurationModule.java
burmanm: I was aware of that (mentioned in the JIRA ticket), but this is not how the rest of the project has done it so I modified this one to be the same. 

https://github.com/apache/cassandra-sidecar/blob/trunk/server/src/main/java/org/apache/cassandra/sidecar/modules/LifecycleModule.java#L79

https://github.com/apache/cassandra-sidecar/blob/trunk/server/src/main/java/org/apache/cassandra/sidecar/modules/UtilitiesModule.java#L110

Should all these cases also use ClassLoader? It feels a bit out of context for this PR.

yifan-c: For the lifecycle and utilities modules, the pluggability is implemented via overriding the Guice provider. Imagining in the internal-sidecar fork, you have a internal module that provides a different `LifecycleProvider`. 

So, you are right to call it out. The alternative is to update `cqlAuthProvider` to be `@Provides`

server/src/main/java/org/apache/cassandra/sidecar/cluster/auth/FileProvider.java
burmanm: I'd expect the password/username to be modifiable, yes. Same as with TLS secrets. But if you want, I can modify this to be read-once and need to restart sidecar for changes.

Why isn't there automated test for the indentations and a formatter?

yifan-c: There is an "opinionated" guide on setting up the project in IntellJ (https://github.com/apache/cassandra-sidecar/blob/trunk/CONTRIBUTING.md#integration-with-intellij-idea). It should take care of the indentation. 
There is no formatter to help, if you are not using IntelliJ, ATM.

yifan-c: > I'd expect the password/username to be modifiable, yes. Same as with TLS secrets.

ok. I guess it is for the scenarios such as password rotation.

yifan-c:
Tests are failing. please take a look

integration-tests/src/integrationTest/resources/config/sidecar.yaml.template
```diff
@@ -267,8 +267,19 @@ access_control:
...
+  #   class_name: org.apache.cassandra.sidecar.cluster.auth.FileProvider
+  #   parameters:
+  #     username_path: /opt/sidecar/superuser/username
+  #     password_path: /opt/sidecar/superuser/password
+    class_name: org.apache.cassandra.sidecar.cluster.auth.ConfigProvider
+    parameters:
+      username: cassandra
+      password: cassandra
+#  username: cassandra
+#  password: cassandra
```
frankgh: NIT, not required to keep :

```suggestion
```

server/src/main/java/org/apache/cassandra/sidecar/cluster/CQLSessionProviderImpl.java
```diff
@@ -119,8 +122,10 @@ public CQLSessionProviderImpl(List<InetSocketAddress> contactPoints,
         this.localInstances = localInstances;
         this.localDc = localDc;
         this.numAdditionalConnections = numAdditionalConnections;
-        this.username = username;
-        this.password = password;
+        Map<String, String> namedParameters = new HashMap<>();
+        namedParameters.put("username", username);
+        namedParameters.put("password", password);
```
frankgh, line 127: tiny NIT:


```suggestion
        Map<String, String> namedParameters = Map.of("username", username,
                                                     "password", password);
```

server/src/main/java/org/apache/cassandra/sidecar/modules/ConfigurationModule.java
```diff
@@ -117,15 +125,51 @@ CassandraInputValidationConfiguration validationConfiguration(SidecarConfigurati
...
+    {
+        DriverConfiguration driverConfiguration = sidecarConfiguration.driverConfiguration();
+
+        ParameterizedClassConfiguration config = driverConfiguration.authProvider();
+        if (config == null)
+        {
+            // Fallback to the old one
+            Map<String, String> namedParameters = new HashMap<>();
+            namedParameters.put("username", driverConfiguration.username());
+            namedParameters.put("password", driverConfiguration.password());
```
frankgh, line 158: minor NIT:
```suggestion
            Map<String, String> namedParameters = Map<String, String> namedParameters = Map.of("username", driverConfiguration.username(),
                                                         "password", driverConfiguration.password());
```

server/src/test/java/org/apache/cassandra/sidecar/cluster/auth/ConfigProviderTest.java
```diff
@@ -0,0 +1,48 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+
+class ConfigProviderTest
+{
+    @Test
+    void testInterfaceGetters()
+    {
+        CqlAuthProvider provider = new ConfigProvider(Map.of(ConfigProvider.USERNAME_PARAM, "cassandra",
+                                                            ConfigProvider.PASSWORD_PARAM, "cassandra"));
```
frankgh: tiny NIT: formatting 
```suggestion
        CqlAuthProvider provider = new ConfigProvider(Map.of(ConfigProvider.USERNAME_PARAM, "cassandra",
                                                             ConfigProvider.PASSWORD_PARAM, "cassandra"));
```

server/src/test/java/org/apache/cassandra/sidecar/cluster/auth/FileProviderTest.java
```diff
@@ -0,0 +1,79 @@
...
+    {
+        Path usernamePath = tempDir.resolve("username");
+        Path passwordPath = tempDir.resolve("password");
+        Files.writeString(usernamePath, "cassandra-user\n");
+        Files.writeString(passwordPath, "cassandra-pass\n");
+
+        CqlAuthProvider provider = new FileProvider(Map.of(FileProvider.USERNAME_PATH_PARAM,
+                                                                       usernamePath.toString(),
+                                                                       FileProvider.PASSWORD_PATH_PARAM,
+                                                                       passwordPath.toString()));
```
frankgh: minor NIT: formatting


```suggestion
        CqlAuthProvider provider = new FileProvider(Map.of(FileProvider.USERNAME_PATH_PARAM, usernamePath.toString(),
                                                           FileProvider.PASSWORD_PATH_PARAM, passwordPath.toString()));
```

server/src/test/java/org/apache/cassandra/sidecar/cluster/auth/FileProviderTest.java
```diff
@@ -0,0 +1,79 @@
...
+    {
+        Path usernamePath = tempDir.resolve("username");
+        Path passwordPath = tempDir.resolve("password");
+        Files.writeString(usernamePath, " \n");
+        Files.writeString(passwordPath, "secret");
+
+        FileProvider provider = new FileProvider(Map.of(FileProvider.USERNAME_PATH_PARAM,
+                                                                       usernamePath.toString(),
+                                                                       FileProvider.PASSWORD_PATH_PARAM,
+                                                                       passwordPath.toString()));
```
frankgh: minor NIT:
```suggestion
        FileProvider provider = new FileProvider(Map.of(FileProvider.USERNAME_PATH_PARAM, usernamePath.toString(),
                                                        FileProvider.PASSWORD_PATH_PARAM, passwordPath.toString()));
```

frankgh (approved):
+1 looks good! I added some really minor comments

server/src/main/java/org/apache/cassandra/sidecar/cluster/CQLSessionProviderImpl.java
burmanm: This would cause NPE if the username/password is null, which is acceptable (and something the tests do). Map.of requires non-null for all values.

server/src/main/java/org/apache/cassandra/sidecar/modules/ConfigurationModule.java
burmanm: Same as above.

server/src/main/java/org/apache/cassandra/sidecar/cluster/CQLSessionProviderImpl.java
frankgh: good catch!

yifan-c:
The patch looks good. Can you rebase in order to trigger the github checks? The github actions was broken and has been fixed. 
In addition, please share the circleCI link (in JIRA) as well. Thank you!

frankgh:
Closed via https://github.com/apache/cassandra-sidecar/commit/b8ddad057c4db7b6ad061ba04323d9cc10622d96

PR #321, opened 2026-02-17
CASSSIDECAR-409 Add safety check to Live Migration data copy endpoint

nvharikrishna:

[CASSSIDECAR-409](https://issues.apache.org/jira/browse/CASSSIDECAR-409) Add safety check to Live Migration data copy endpoint

Summary of changes:
* Added JMX-based safety check for live migration data copy. Data copy task gets rejected if Sidecar is able to connect to the destination JMX port.
* Added PreCheck interface with default impl to allow customization of pre-check (executed before each iteration of file downloads).


server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
```diff
@@ -101,39 +107,85 @@ Future<LiveMigrationTask> createDataCopyTask(LiveMigrationDataCopyRequest reques
...
+     * Cassandra is running or not at the moment on the destination instance by checking if Sidecar
+     * was able to connect to the Cassandra instance's JMX port. It returns a failed future if Sidecar
+     * is able to connect to the JMX port of Cassandra.
+     *
+     * @param localInstance metadata for the local Cassandra instance
+     * @return Future that succeeds if Cassandra is not running, fails if it is running
+     */
+    private Future<Void> verifyCassandraNotRunning(InstanceMetadata localInstance)
+    {
+        return executorPools.internal().executeBlocking(() -> {
```
frankgh: this does not need to run on a blocking method. There's no IO happening here. The check is an in-memory check. Which brings the question about whether you want to do an online check here instead of relying on a cached value?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -124,6 +130,130 @@ public Future<OperationStatus> downloadFiles()
...
+        // Reached here means, source found and destination not found in gossip - safe to proceed
+        LOGGER.info("{} Gossip validation passed: destination {} not found in cluster gossip",
+                    logPrefix, destination);
+        return Future.succeededFuture();
+    }
+
+    private GossipInfoResponse.GossipInfo findGossipInfo(GossipInfoResponse gossipResponse,
+                                                         String instance) throws UnknownHostException
+    {
+        InstanceMetadata metadata = instancesMetadata.instanceFromHost(instance);
```
frankgh: is this `instance` managed by the local sidecar? `instancesMetadata` is only available for instances managed by the local Sidecar process.

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/LiveMigrationDataCopyRequest.java
```diff
@@ -71,8 +99,24 @@ public LiveMigrationDataCopyRequest(@JsonProperty("maxIterations") int maxIterat
...
         }
 
+        // Validate optional gossip fetch parameters if specified
+        if (gossipFetchBatchSize != null && gossipFetchBatchSize <= 0)
+        {
+            throw new IllegalArgumentException("Invalid gossipFetchBatchSize " + gossipFetchBatchSize
+                                               + ". It must be greater than zero when specified.");
+        }
+
+        if (gossipFetchMaxRetries != null && gossipFetchMaxRetries <= 0)
```
frankgh: should we also have some guardrails around an upperbound here? I'm thinking from a point of view of a malicious payload that can make the service retry for a long time?

frankgh (commented):
Looks good so far. I have a few minor comments.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
nvharikrishna: You are right, I took it out of 'executeBlocking'. Since gossip-based check is also present and it is only the first level of defense, I am fine with checking the cached value.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: Thanks for catching it! I think it changes most of the things in this pr. InstanceMetadata is needed for both destination (local) and source (remote). Unfortunately, I don't see a way to get either source instance metadata or cluster information (considering source and/or destination may be down briefly during migration), which are needed to fetch gossip info and perform validation. I will remove Gossipinfo-related checks. Probably, I will create a plug-in/placeholder so that if anyone wants to inject any custom check, they can do it.

nvharikrishna (commented):
@frankgh thanks for catching this: https://github.com/apache/cassandra-sidecar/pull/321/changes#r2898297905. It significantly changes the approach taken in this PR. I don't see a way to fetch cluster info or gossip reliably during migration without having the knowledge of non-local instance metadata. Going to remove the gossip check. But will add a placeholder/interface plug-in so that people can add their own checks if needed.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
```diff
@@ -101,39 +103,83 @@ Future<LiveMigrationTask> createDataCopyTask(LiveMigrationDataCopyRequest reques
...
+     * @param localInstance metadata for the local Cassandra instance
+     * @return Future that succeeds if Cassandra is not running, fails if it is running
+     */
+    private Future<Void> verifyCassandraNotRunning(InstanceMetadata localInstance)
+    {
+        try
+        {
+            CassandraAdapterDelegate delegate = localInstance.delegate();
+
+            if (delegate.isJmxUp())
```
frankgh: should we also check for CQL here?
```suggestion
            if (delegate.isJmxUp() || delegate.isNativeUp())
```

CHANGES.txt
```diff
@@ -1,5 +1,6 @@
 0.3.0
 -----
+ * Adding gossip based safety check to Live Migration data copy task endpoint (CASSSIDECAR-409)
```
frankgh: this should go under 0.4.0

frankgh (commented):
Looks good overall, I have one minor question.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
nvharikrishna: Makes sense, made the changes.

CHANGES.txt
nvharikrishna: done

frankgh (approved):
+1

CHANGES.txt
yifan-c: Please update the change entry and the PR title; they should be the same as the jira title

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
```diff
@@ -101,39 +103,83 @@ Future<LiveMigrationTask> createDataCopyTask(LiveMigrationDataCopyRequest reques
...
+                       if (!taskInMap.isCompleted())
+                       {
+                           // Accept new task if and only if the existing task has completed.
+                           return taskInMap;
+                       }
+                       else
+                       {
+                           return newTask;
+                       }
+                   }) == newTask;
```
yifan-c: nit: can you break it into 2 statements? It is easy to miss the `== newTask` portion.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
```diff
@@ -101,39 +103,83 @@ Future<LiveMigrationTask> createDataCopyTask(LiveMigrationDataCopyRequest reques
...
+
+                       if (!taskInMap.isCompleted())
+                       {
+                           // Accept new task if and only if the existing task has completed.
+                           return taskInMap;
+                       }
+                       else
+                       {
+                           return newTask;
+                       }
```
yifan-c: nit: can be easier to read with ternary operator

```suggestion
                       // Accept new task if and only if the existing taskInMap has completed.
                       return taskInMap.isCompleted() ? newTask : taskInMap;
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
```diff
@@ -101,39 +103,83 @@ Future<LiveMigrationTask> createDataCopyTask(LiveMigrationDataCopyRequest reques
...
+                   }) == newTask;
+
+                   if (!accepted)
+                   {
+                       return Future.failedFuture(
+                       new LiveMigrationDataCopyInProgressException("Another task is already under progress. Cannot accept new task."));
```
yifan-c, line 129: It is identical if you just throw the exception.

```suggestion
                       throw new LiveMigrationDataCopyInProgressException("Another task is already under progress. Cannot accept new task.");
```

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloadPreCheck.java
```diff
@@ -0,0 +1,105 @@
...
+ * allowing implementations to continuously validate that conditions remain safe throughout the
+ * multi-iteration copy process.</p>
+ *
+ * <p>A {@link #DEFAULT} no-op implementation is provided for cases where no pre-check is needed.
+ * Custom implementations can be bound via Guice to override the default behavior.</p>
+ *
+ * <p>Example use cases:</p>
+ * <ul>
+ *   <li>Gossip-based validation: verify the source node is present and the destination node
+ *       is absent from cluster gossip, preventing data copy to a node that has already joined</li>
```
yifan-c, line 41: The implementation does not exist

server/src/test/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManagerTest.java
```diff
@@ -324,7 +393,7 @@ private <T> void awaitForFuture(Future<T> future) throws InterruptedException
         CountDownLatch latch = new CountDownLatch(1);
         future.onComplete(res -> latch.countDown());
 
-        latch.await(100, TimeUnit.MILLISECONDS);
+        latch.await(2, TimeUnit.SECONDS);
```
yifan-c, line 396: Why the wait time is increased?

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
```diff
@@ -744,4 +768,51 @@ public FileAttributes(long size, long lastModifiedTime)
             this.lastModifiedTime = lastModifiedTime;
         }
     }
+
+    /**
+     * Implementation of {@link LiveMigrationFileDownloadPreCheck.PreCheckContext} that provides
+     * the downloader's context to pre-check implementations.
+     */
+    static class PreCheckContextImpl implements LiveMigrationFileDownloadPreCheck.PreCheckContext
```
yifan-c: make it `private static class`?

CHANGES.txt
nvharikrishna: done

server/src/test/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManagerTest.java
nvharikrishna: Observed that unit tests are failing rarely when there are JVM pauses (observed in my laptop). Thought it may turn flaky in CI environment, so increased the wait time.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloadPreCheck.java
nvharikrishna: This documentation is suggesting the kind of checks users can add (line: 35-38). But do not claim that they are implemented.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationFileDownloader.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
nvharikrishna: I wanted to return the future in both positive and negative scenarios because it seems simpler than throwing an exception, catching it elsewhere, and converting it to a failed future. I can change it if it makes a difference.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
nvharikrishna: done

server/src/main/java/org/apache/cassandra/sidecar/livemigration/DataCopyTaskManager.java
nvharikrishna: Move it to the front so that it can't be missed.

yifan-c (approved):

PR #323, opened 2026-02-20
CASSSIDECAR-411: RangeManager should be singleton in CDCModule

jyothsnakonisa:


bbotella (approved):
+1

Good catch!

yifan-c (approved):

yifan-c (commented):
Please fixed the test failures

CHANGES.txt
```diff
@@ -1,5 +1,6 @@
 0.3.0
 -----
+ * Rangemanager should be singleton in CDCModule (CASSSIDECAR-411)
```
jmckenzie-dev: Super nit: `RangeManager` to match casing on class name. 😅

server/src/test/java/org/apache/cassandra/sidecar/cluster/CassandraClientTokenRingProviderTest.java
```diff
@@ -282,12 +282,12 @@ private static TokenRange createMockTokenRange(Token start, Token end)
         }
     }
 
-    protected DnsResolver mockDnsResolver()
+    public static DnsResolver mockDnsResolver()
```
jmckenzie-dev, line 285: Why the increase in scope from private to public? I don't see any new consumers in the PR...

server/src/test/java/org/apache/cassandra/sidecar/cluster/CassandraClientTokenRingProviderTest.java
```diff
@@ -308,21 +308,21 @@ private InstancesMetadata mockInstancesMetadata()
...
-        InstanceMetadata instance3 = getMockInstanceMetaData(101000301, "local3", getMetadata());
+        InstanceMetadata instance1 = getMockInstanceMetaData(101000101, "localhost", getMetadata());
+        InstanceMetadata instance2 = getMockInstanceMetaData(101000201, "localhost2", getMetadata());
+        InstanceMetadata instance3 = getMockInstanceMetaData(101000301, "localhost3", getMetadata());
         when(instancesMetadata.instances()).thenReturn(List.of(instance1, instance2, instance3));
         return instancesMetadata;
     }
 
-    private Metadata getMetadata()
+    public static Metadata getMetadata()
```
jmckenzie-dev, line 318: Same as above - purpose of scope promotion? If intentional we should either annotated `@VisibleForTesting` or javadoc it w/@link so subsequent refactors pull this in / break / make noise in an obvious way.

jmckenzie-dev (approved):
nit on casing and question re: scoping; all things that can be resolved by author during commit, aren't a big deal, or we could follow up on a subsequent commit so we can get this change in.

server/src/test/java/org/apache/cassandra/sidecar/cluster/CassandraClientTokenRingProviderTest.java
jyothsnakonisa: I am reusing these helper methods in TestModule to fix failing tests.

PR #327, opened 2026-03-13
CASSSIDECAR-376: Update OperationalJob to support cluster-wide operations

brandondho:
### [CASSSIDECAR-376](https://issues.apache.org/jira/browse/CASSSIDECAR-376)

Updates the `OperationalJob` abstract class and related components to support cluster-wide node tracking for CEP-53. Single-node jobs now automatically track their execution lifecycle through node lists (`nodesPending`, `nodesExecuting`, `nodesSucceeded`, `nodesFailed`), while future cluster-wide jobs will manage these lists directly via protected setters.

### Changes

- `OperationalJob` gains a `nodeId` field and two-arg constructor. When `nodeId` is provided, node tracking lists are automatically managed through the job lifecycle in `execute()`. The single-arg constructor initializes empty lists for cluster-wide jobs to manage via protected setters.
- `OperationalJob` gains `startTime` and `lastUpdate` fields, automatically set during `execute()`
- `OperationalJobResponse` gains new fields (`startTime`, `nodesPending`, `nodesExecuting`, `nodesSucceeded`, `nodesFailed`, `lastUpdate`)
- `CassandraAdapterDelegate` gains `hostId()` method using JMX `getLocalHostId()` to fetch the Cassandra host UUID

### Future Work

This enables upcoming work on CEP-53:

- Implement job coordination for cluster-wide operations ([CASSSIDECAR-377](https://issues.apache.org/jira/browse/CASSSIDECAR-377))

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,6 +4,7 @@
 
 0.3.0
 -----
+ * Update OperationalJob class to support cluster-wide operations (CASSSIDECAR-376)
  * RangeManager should be singleton in CDCModule (CASSSIDECAR-411)
  * CDC: Add end-to-end CDC integration tests (CASSSIDECAR-308)
  * SchemaStorePublisherFactory should be Injectable in CachingSchemaStore (CASSSIDECAR-408)
--- a/client-common/src/main/java/org/apache/cassandra/sidecar/common/response/OperationalJobResponse.java
+++ b/client-common/src/main/java/org/apache/cassandra/sidecar/common/response/OperationalJobResponse.java
@@ -18,13 +18,20 @@
 
 package org.apache.cassandra.sidecar.common.response;
 
+import java.time.Instant;
+import java.util.List;
 import java.util.UUID;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import org.apache.cassandra.sidecar.common.DataObjectBuilder;
 import org.apache.cassandra.sidecar.common.data.OperationalJobStatus;
+import org.apache.cassandra.sidecar.common.utils.InstantIso8601Deserializer;
+import org.apache.cassandra.sidecar.common.utils.InstantIso8601Serializer;
 
 /**
  * Response structure of the operational jobs API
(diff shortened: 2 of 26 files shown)


brandondho:
CI: https://app.circleci.com/pipelines/circleci/TRR3wKvYSJgUgEYViPWJaY/FdxXdHgyZqcxr4N31cTUq?branch=CASSSIDECAR-376

pauloricardomg (approved):

CHANGES.txt
```diff
@@ -4,6 +4,7 @@
 
 0.3.0
 -----
+ * Update OperationalJob class to support cluster-wide operations (CASSSIDECAR-376)
```
frankgh: we should move this under 0.4.0 instead

server/src/main/java/org/apache/cassandra/sidecar/cluster/CassandraAdapterDelegate.java
```diff
@@ -520,6 +521,25 @@ public boolean isJmxUp()
         return nodeSettingsFromJmx != null;
     }
 
+    /**
+     * @return the Cassandra host UUID fetched via JMX
+     * @throws CassandraUnavailableException when JMX is not available
+     */
+    @NotNull
+    public UUID hostId() throws CassandraUnavailableException
```
frankgh: I think this belongs in the NodeSettings class instead, I think we should add it there instead

server/src/main/java/org/apache/cassandra/sidecar/handlers/ListOperationalJobsHandler.java
```diff
@@ -72,7 +70,11 @@ protected void handleInternal(RoutingContext context, HttpServerRequest httpRequ
         ListOperationalJobsResponse listResponse = new ListOperationalJobsResponse();
         jobManager.allInflightJobs()
                   .stream()
-                  .map(job -> new OperationalJobResponse(job.jobId(), RUNNING, job.name(), null))
+                  .map(job -> new OperationalJobResponse(job.jobId(), job.status(), job.name(), null,
```
frankgh: maybe introduce a builder for this constructor?

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/OperationalJobResponse.java
```diff
@@ -37,17 +38,35 @@ public class OperationalJobResponse
     private final OperationalJobStatus status;
     private final String operation;
     private final String reason;
+    private final String startTime;
```
frankgh: any reason why we use string for startTime and lastUpdate instead of a time data structure?

frankgh (commented):
Looks good in general. I have some suggestions. This commit captures the suggestions: https://github.com/frankgh/cassandra-sidecar/commit/054b99d28463d4e932a87e4ccd3199f79745c8a3

client-common/src/main/java/org/apache/cassandra/sidecar/common/response/OperationalJobResponse.java
frankgh: So I've been thinking about this more. This should not be a string, we should either have a native type (i.e. long) that represents the epoch in milliseconds. We can still have the String version of the start time in the OperationalJobResponse object, but it should be generated each time, and it should not be serialized in the JSON payload. Consumers should be able to  decide how to consume the value and display it

pauloricardomg: @frankgh did you have a chance to see e3e159edfa8693fc475157df089ba622f5ed1a2b ? This looks good to me as follow-up, if it looks good to you I can commit after CI looks good.

frankgh: yeah, looks good. I think the patch looks good. We just need to make sure CI is green

frankgh (approved):
+1 looks good to me! Thanks for the patch

PR #329, opened 2026-03-27
CASSSIDECAR-417 falling back to host equality test in CdcManager.getInstanceId

smiklosovic:


smiklosovic:
open to relax on logging, we can go with DEBUG level only.

jyothsnakonisa:
https://github.com/apache/cassandra-sidecar/pull/332 fixed this issue and this PR is no longer needed.

PR #332, opened 2026-03-30
CASSSIDECAR-417 : CdcManager.getInstanceId(instanceIp) returns -1 as it resolves ipAddress to null

jyothsnakonisa:


server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcManager.java
```diff
@@ -28,6 +28,8 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.apache.cassandra.sidecar.cluster.instance.InstanceMetadata;
```
skoppu22: Looks like this import is unused

skoppu22 (approved):

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcManager.java
```diff
@@ -207,20 +209,18 @@ public void stopConsumers()
...
-            String configuredIpAddress = instance.ipAddress();
-
-            // Option 1a: Normalize both to InetAddress and compare
-            if (resolveToSameAddress(instanceIp, configuredIpAddress))
-            {
-                return instance.id();
-            }
+            return instanceFetcher.instance(instanceIp).id();
+        }
+        catch (NoSuchCassandraInstanceException e)
```
skoppu22: I believe we need to catch broader exception like Exception or RuntimeException to ensure the same behaviour when other kinds of exceptions are thrown, for example NullPointerException from a corrupted metadata state

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcManager.java
```diff
@@ -207,20 +209,18 @@ public void stopConsumers()
...
+        try
         {
-            String configuredIpAddress = instance.ipAddress();
-
-            // Option 1a: Normalize both to InetAddress and compare
-            if (resolveToSameAddress(instanceIp, configuredIpAddress))
-            {
-                return instance.id();
-            }
+            return instanceFetcher.instance(instanceIp).id();
```
skoppu22, line 214: Looks like resolveToSameAddress is no longer called by any prod code after this change, and can be removed

skoppu22 (changes requested):
Minor changes needed

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcManager.java
jyothsnakonisa: Moved it to the test class.

skoppu22 (approved):

server/src/test/java/org/apache/cassandra/sidecar/cdc/CdcManagerTest.java
```diff
@@ -291,24 +281,92 @@ void testUnknownInstanceHandlesGracefully() throws IOException
...
     void testResolveToSameAddressFalse()
     {
-        String address1 = "127.0.0.1";
-        String address2 = "127.0.0.2";
-        assertThat(CdcManager.resolveToSameAddress(address1, address2)).isFalse();
+        assertThat(resolveToSameAddress("127.0.0.1", "127.0.0.2")).isFalse();
+    }
+
+    @Test
+    void testGetInstanceIdResolvesCorrectlyWhenIpAddressIsNull() throws IOException
```
jmckenzie-dev, line 306: We should comment this test to clarify its intent and how it differs from `testGetInstanceIdReturnsCorrectIdWhenIpAddressIsNull`. These 2 tests have very similar names and it's not clear how they differ at face value.

jmckenzie-dev (changes requested):

jmckenzie-dev (approved):

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcManager.java
```diff
@@ -207,34 +205,17 @@ public void stopConsumers()
...
-            return addr1.equals(addr2);
+            return instanceFetcher.instance(instanceIp).id();
         }
-        catch (UnknownHostException e)
+        catch (Exception e)
         {
-            LOGGER.warn("Could not resolve hostname: {}", e.getMessage());
-            return address1.equals(address2); // Fallback to string comparison
+            LOGGER.warn("Requested IP {} does not match with any instances", instanceIp);
+            return -1;
```
jmckenzie-dev: wdyt about, instead of using `-1` here as a magic number, we define a static constant `UNKNOWN_INSTANCE` or something we return when we fail to match and check against?

jyothsnakonisa: Yup agree, that looks better!

jyothsnakonisa:
CI : https://app.circleci.com/pipelines/gh/jyothsnakonisa/cassandra-sidecar/182/details

PR #333, opened 2026-04-08
CASSSIDECAR-422 SAI support in Sidecar

skoppu22:
Cassandra Analytics generates SSTable files, then invokes Sidecar’s uploadSSTables endpoint to stage them, followed by the importSSTables endpoint to move them into Cassandra’s data directory.

As part of adding SAI index support to Cassandra Analytics (see: https://issues.apache.org/jira/browse/CASSANALYTICS-31), the system will produce SAI index files in addition to SSTables. To support this, we need to update both the upload and import SSTables endpoints so they can handle SAI index files alongside SSTables. Will open a PR for SAI support in C* analytics after this PR merged in, because this PR is a pre-requisite for that


server/src/test/integration/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableImportHandlerIntegrationTest.java
```diff
@@ -201,4 +205,157 @@ private void populateTable(Session session, QualifiedTableName tableName, List<S
             session.execute(String.format("INSERT INTO %s (id) VALUES ('%s');", tableName, value));
         }
     }
+
+    @CassandraIntegrationTest
+    void testSSTableImportWithSaiIndexParams(VertxTestContext vertxTestContext)
```
frankgh: considering the amount of resources needed for CI , I am -1 on using these tests, we should migrate this test to `integration-tests` leveraging `SharedClusterSidecarIntegrationTestBase`

server/src/test/integration/org/apache/cassandra/sidecar/restore/jobdiscoverer/RestoreJobDiscovererSaiImportOptionsIntTest.java
```diff
@@ -0,0 +1,195 @@
...
+import static org.apache.cassandra.sidecar.restore.RestoreJobTestUtils.createJob;
+import static org.apache.cassandra.sidecar.restore.RestoreJobTestUtils.disableRestoreProcessor;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration tests verifying that SAI import options (failOnMissingIndex, validateIndexChecksum) are
+ * correctly persisted and propagated through the restore job pipeline: create job -> persist -> discover -> create ranges.
+ */
+@Tag("heavy")
+class RestoreJobDiscovererSaiImportOptionsIntTest extends IntegrationTestBase
```
frankgh: we shouldn't be introducing new tests to this framework as we are trying to deprecate it. Let's use the `integration-test` framework instead leveraging `SharedClusterSidecarIntegrationTestBase`

client-common/src/main/java/org/apache/cassandra/sidecar/common/data/SSTableImportOptions.java
```diff
@@ -18,21 +18,28 @@
...
+import java.util.LinkedHashMap;
 
 /**
  * Options for Cassandra import nodetool command. It is like properties.
  * Supports Json serialization and deserialization.
+ * <p>
+ * This class is used exclusively in the S3 restore job path ({@code CreateRestoreJobRequestPayload}).
  */
-public class SSTableImportOptions extends HashMap<String, String>
+public class SSTableImportOptions extends LinkedHashMap<String, String>
```
jyothsnakonisa, line 29: Is there a reason why this had to be changed to LinkedHashMap?

adapters/adapters-cassandra50/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra50/Cassandra50TableOperations.java
```diff
@@ -0,0 +1,71 @@
...
+ * parameters when importing SSTables.
+ */
+public class Cassandra50TableOperations extends CassandraTableOperations
+{
+    private final JmxClient jmxClient;
+
+    public Cassandra50TableOperations(JmxClient jmxClient)
+    {
+        super(jmxClient);
+        this.jmxClient = jmxClient;
```
jyothsnakonisa: Instead of having separate jmxclient field in this class, can u make the base class's field protected and use it here?

jyothsnakonisa (approved):
Looks good overall, left few questions

client-common/src/main/java/org/apache/cassandra/sidecar/common/data/SSTableImportOptions.java
skoppu22: The reason is deterministic JSON serialization order. SSTableImportOptions is serialized into 
  the importOptions field of CreateRestoreJobRequestPayload (S3 restore job path). With HashMap, the JSON key order depends on hash codes and is unpredictable; with LinkedHashMap, keys serialize in insertion order — matching the put() sequence in the constructor.

server/src/test/integration/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableImportHandlerIntegrationTest.java
skoppu22: Moved this to integration test using SharedClusterSidecarIntegrationTestBase

server/src/test/integration/org/apache/cassandra/sidecar/restore/jobdiscoverer/RestoreJobDiscovererSaiImportOptionsIntTest.java
skoppu22: Moved this to integration test using SharedClusterSidecarIntegrationTestBase

adapters/adapters-cassandra50/src/main/java/org/apache/cassandra/sidecar/adapters/cassandra50/Cassandra50TableOperations.java
skoppu22: Good catch, removed it

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraTableOperations.java
```diff
@@ -65,6 +65,30 @@ public List<String> importNewSSTables(@NotNull String keyspace,
...
+                                          boolean invalidateCaches,
+                                          boolean extendedVerify,
+                                          boolean copyData,
+                                          boolean failOnMissingIndex,
+                                          boolean validateIndexChecksum)
+    {
+        return importNewSSTables(keyspace, tableName, directory, resetLevel,
+                clearRepaired, verifySSTables, verifyTokens,
+                invalidateCaches, extendedVerify, copyData);
+    }
```
yifan-c, line 90: indentation

```suggestion
        return importNewSSTables(keyspace, tableName, directory, resetLevel,
                                 clearRepaired, verifySSTables, verifyTokens,
                                 invalidateCaches, extendedVerify, copyData);
```

server/src/main/java/org/apache/cassandra/sidecar/utils/SSTableImporter.java
```diff
@@ -248,8 +253,7 @@ private void drainImportQueue(ImportQueue queue)
                     successCount++;
                     LOGGER.info("Successfully imported SSTables with options={}, serviceTimeMillis={}",
                                 options, TimeUnit.NANOSECONDS.toMillis(serviceTimeNanos));
-                    promise.complete();
-                    cleanup(options);
+                    cleanup(options).onComplete(ar -> promise.complete());
```
yifan-c: This changes the behavior of the `SSTableImporter`. 
Previously, cleanup is a best-effort; the cleanup can fail and operators can recover manually. Now, if cleanup fails, import fails. 
I do not believe that is the desired behavior of the sstable importing.

server/src/testFixtures/java/org/apache/cassandra/sidecar/restore/RestoreJobTestUtils.java
yifan-c, line 1: Why is it moved from test scope to testFixture scope? The relocation leads to the textFixture depndencies update in `server/build.gradle`

skoppu22: RestoreJobTestUtils is now consumed by RestoreJobDiscovererSAIOptionsIntTest in integration-tests module as well existing tests in integration module. Hence moved to this to testFixtures to share across modules.

adapters/adapters-base/src/main/java/org/apache/cassandra/sidecar/adapters/base/CassandraTableOperations.java
skoppu22: Done

server/src/main/java/org/apache/cassandra/sidecar/utils/SSTableImporter.java
skoppu22: I was worried this will leave files behind, accumulate and occupy storage over the time. Reverted the change.

skoppu22:
Latest CI https://app.circleci.com/pipelines/github/skoppu22/cassandra-sidecar/58/workflows/29419afd-4b8a-440f-8f29-04f153297276

skoppu22:
Added entry to changes.txt

yifan-c (approved):

PR #335, opened 2026-04-09
ninja fix: update changes.txt for CASSSIDECAR-423

jyothsnakonisa:


frankgh:
@jyothsnakonisa there's no need to add an entry in CHANGES.txt for CI-only related changes.

PR #338, opened 2026-04-22
CASSSIDECAR-442 Add pre-test environment checks and aggregate jacoco report

rustyrazorblade:
Adds a checkEnvironment task (gradle/common/environmentChecks.gradle) that runs before all Test tasks and fails fast with actionable fix instructions if any of the following are not satisfied: /etc/hosts entries for localhost2 and localhost20, loopback aliases 127.0.0.2-20, dtest jars present, Java <= 17. All failures are reported together rather than stopping at the first.

Also adds jacocoRootReport to aggregate coverage across all subprojects.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -108,6 +108,8 @@ ext {
 
 }
 
+apply from: "${project.rootDir}/gradle/common/environmentChecks.gradle"
+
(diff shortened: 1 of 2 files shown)


yifan-c (approved):

yifan-c:
+1. Is there a jira ticket?

PR #339, opened 2026-04-23
CASSSIDECAR-373: Define storage provider interface and Cassandra implementation to support durable job tracking

andresbeckruiz:
## CASSSIDECAR-373

  Implements `CassandraStorageProvider`, the default Cassandra backed implementation of the `StorageProvider` interface for durable operational job state. 

  ## Changes

  - `StorageProvider` interface, which adds a provider-agnostic abstraction for persisting, querying, and coordinating operational jobs 
  - `OperationalJobRecord` is the data transfer object representing persisted job state
  - `OperationalJobConfiguration` provides configurable table TTL, and in the future can support remote Cassandra cluster connection
  - Three Cassandra table schemas:
    - `cluster_ops` — persists and tracks operational jobs (restarts, upgrades, etc.)
    - `cluster_ops_node_state` — tracks per-node status within an operation
    - `active_cluster_ops` — provides mutual exclusion of concurrent operations via LWT
  - Database accessors: (`ClusterOpsDatabaseAccessor`, `ClusterOpsNodeStateDatabaseAccessor`, `ActiveClusterOpsDatabaseAccessor`) 
  - `CassandraStorageProvider` composes the three accessors, acts as thin delegation layer
  
  - Integration tests for all three database accessors

  ## Future Work

  - Remote Cassandra cluster storage support 
  - Future [CEP-53](https://cwiki.apache.org/confluence/display/CASSANDRA/CEP-53%3A+Cassandra+Rolling+Restarts+via+Sidecar) work: 
      - CASSSIDECAR-374: Implement durable operational job tracker
      - CASSSIDECAR-375:  Add storage provider to enable JVM Distributed tests for cluster-wide operational jobs 
      - CASSSIDECAR-377: Implement job coordination for cluster-wide operations


server/src/main/java/org/apache/cassandra/sidecar/db/schema/ActiveClusterOpsSchema.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    protected String tableName()
+    {
+        return TABLE_NAME;
+    }
+
+    @Override
+    protected String createSchemaStatement()
+    {
+        return String.format("CREATE TABLE IF NOT EXISTS %s.%s (" +
```
isaacreath, line 67: Should we consider LCS for the compaction strategy on this table? May not be important given the (in-theory) low volume to this table. Just considering the challenges we face when using STCS with TTL tables and we can't use UCS as this should still work on Cassandra 4 clusters.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ActiveClusterOpsSchema.java
```diff
@@ -0,0 +1,113 @@
...
+                             "  cluster_name text," +
+                             "  operation_type text," +
+                             "  operation_id uuid," +
+                             "  PRIMARY KEY ((cluster_name), operation_type)" +
+                             ") WITH default_time_to_live = %s",
+                             keyspaceConfig.keyspace(), TABLE_NAME, tableTtl.toSeconds());
+    }
+
+    @Override
+    protected void prepareStatements(@NotNull Session session)
```
isaacreath, line 79: This class follows a different patten than the other schema classes that you implement below in that the below classes define a `CqlLiterals` inner class whereas this one just defines these as raw strings. Is there any reason why we aren't consistent across the schema classes in this PR? If not, I'd recommend consolidating on one of the two patterns.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ActiveClusterOpsSchema.java
```diff
@@ -0,0 +1,113 @@
...
+
+import com.datastax.driver.core.PreparedStatement;
+import com.datastax.driver.core.Session;
+import org.apache.cassandra.sidecar.common.server.utils.SecondBoundConfiguration;
+import org.apache.cassandra.sidecar.config.SchemaKeyspaceConfiguration;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Schema for the {@code active_cluster_ops} table, which tracks active operations
+ * and provides mutual exclusion via lightweight transactions (LWT).
```
isaacreath: Nit: We can improve this javadoc by explaining why this table exists and how it is used. The description from the CEP could be helpful here: https://cwiki.apache.org/confluence/display/CASSANDRA/CEP-53%3A+Cassandra+Rolling+Restarts+via+Sidecar#CEP53:CassandraRollingRestartsviaSidecar-sidecar_internal.active_cluster_ops

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ClusterOpsNodeStateSchema.java
```diff
@@ -0,0 +1,126 @@
...
+package org.apache.cassandra.sidecar.db.schema;
+
+import com.datastax.driver.core.PreparedStatement;
+import com.datastax.driver.core.Session;
+import org.apache.cassandra.sidecar.common.server.utils.SecondBoundConfiguration;
+import org.apache.cassandra.sidecar.config.SchemaKeyspaceConfiguration;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Schema for the {@code cluster_ops_node_state} table, which tracks node-level status within an operation.
```
isaacreath: Similarly, using the description from the CEP might be helpful: https://cwiki.apache.org/confluence/display/CASSANDRA/CEP-53%3A+Cassandra+Rolling+Restarts+via+Sidecar#CEP53:CassandraRollingRestartsviaSidecar-sidecar_internal.active_cluster_ops

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ClusterOpsSchema.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Override
+    protected String createSchemaStatement()
+    {
+        return String.format("CREATE TABLE IF NOT EXISTS %s.%s (" +
+                             "  cluster_name text," +
+                             "  operation_id timeuuid," +
+                             "  operation_type text," +
+                             "  status text," +
+                             "  node_execution_order frozen<list<frozen<list<text>>>>," +
```
isaacreath: Shouldn't this be a `frozen<list<forozen<list<uuid>>>>`? Since it stores the order of node ids in which the execution will run?

server/src/test/java/org/apache/cassandra/sidecar/db/ActiveClusterOpsDatabaseAccessorTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests for {@link ActiveClusterOpsDatabaseAccessor}
+ */
+class ActiveClusterOpsDatabaseAccessorTest
```
isaacreath: This test seems to mostly just be calling mocks. I'm not sure how much value it adds over writing an integration test for the database accessor. See similar testing strategy for `/api/v2/cassandra/settings`: https://github.com/apache/cassandra-sidecar/pull/264/changes#diff-a06cba1264721044cdd996db320d93d4297be64406f09bc3081ca67a1e52bf4f

server/src/test/java/org/apache/cassandra/sidecar/db/ClusterOpsDatabaseAccessorTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests for {@link ClusterOpsDatabaseAccessor}
+ */
+class ClusterOpsDatabaseAccessorTest
```
isaacreath: Same as the `ActiveClusterOpsDatabaseAccessorTest`, I think an integration test would be more appropriate than this.

server/src/test/java/org/apache/cassandra/sidecar/db/ClusterOpsNodeStateDatabaseAccessorTest.java
```diff
@@ -0,0 +1,119 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * Tests for {@link ClusterOpsNodeStateDatabaseAccessor}
+ */
+class ClusterOpsNodeStateDatabaseAccessorTest
```
isaacreath: Same as the `ActiveClusterOpsDatabaseAccessorTest`, I think an integration test would be more appropriate than this.

server/src/main/java/org/apache/cassandra/sidecar/db/ClusterOpsDatabaseAccessor.java
```diff
@@ -0,0 +1,126 @@
...
+    private OperationalJobRecord recordFromRow(Row row)
+    {
+        UUID operationId = row.getUUID("operation_id");
+        String operationType = row.getString("operation_type");
+        OperationalJobStatus status = OperationalJobStatus.valueOf(row.getString("status"));
+        List<List<String>> nodeExecutionOrder = row.get("node_execution_order", NODES_ORDER_TYPE);
+        Map<String, String> operationMetadata = row.getMap("operation_metadata", String.class, String.class);
+        return new OperationalJobRecord(operationId, operationType, status,
+                                        nodeExecutionOrder == null || nodeExecutionOrder.isEmpty() ? null : nodeExecutionOrder,
+                                        operationMetadata.isEmpty() ? null : operationMetadata);
```
isaacreath: I'm not the biggest fan of the ternay operators used to provide values for the constructor of `OperationalJobRecord`. It subtly makes the reader of this code need to think of many different states when this object is returned.

A few approaches we could consider here:

1) A builder for the `OperationalJobRecord`.
2) Move the logic into the constructor of the `OperationalJobRecord`
3) Extract this logic above. For example:

```
List<List<String>> nodeExecutionOrder = row.get("node_execution_order",NODES_ORDER_TYPE);
if (nodeExecutionOrder != null && nodeExecutionOrder.isEmpty()) 
{
  nodeExecutironOrder = null;
}
```

server/src/main/java/org/apache/cassandra/sidecar/db/ClusterOpsNodeStateDatabaseAccessor.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    public void updateNodeStatus(String clusterName, UUID operationId, UUID nodeId, OperationalJobStatus nodeStatus)
+    {
+        BoundStatement statement = tableSchema.insertNodeStatus()
+                                              .bind(clusterName, operationId, nodeId, nodeStatus.name());
+        execute(statement);
+    }
+
+    static final int BATCH_CHUNK_SIZE = 100;
```
isaacreath: Please put static properties above constructors (per Cassandra style guide: https://cassandra.apache.org/_/development/code_style.html)

server/src/main/java/org/apache/cassandra/sidecar/job/storage/CassandraStorageProvider.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    @Override
+    public void initialize()
+    {
+        // Schema initialization is handled by SidecarSchemaInitializer.
+        // TTL on all tables handles record pruning.
+        try
+        {
+            clusterName = sessionProvider.get().getCluster().getMetadata().getClusterName();
```
isaacreath: This creates the implicit link between the cluster which is storing the operations and the cluster on which the operation is executing.

I know that this PR has remote storage provider as out of scope, but making this something that we can set via constructor or via initialize would help make sure there is a pathway to store ops in a cluster separate from the one sidecar is connecting to.

isaacreath (changes requested):

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ActiveClusterOpsSchema.java
andresbeckruiz: Updated javadocs for all schemas

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ActiveClusterOpsSchema.java
andresbeckruiz: Added LCS as the compaction strategy for all tables.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ActiveClusterOpsSchema.java
andresbeckruiz: Updated to use `CqlLiterals` inner class

server/src/main/java/org/apache/cassandra/sidecar/db/ClusterOpsDatabaseAccessor.java
andresbeckruiz: Extracted ternary operators into clearer logic

server/src/main/java/org/apache/cassandra/sidecar/job/storage/CassandraStorageProvider.java
andresbeckruiz: Added `clusterName` argument to constructor

server/src/test/java/org/apache/cassandra/sidecar/db/ActiveClusterOpsDatabaseAccessorTest.java
andresbeckruiz: Replaced with ActiveClusterOpsDatabaseAccessorIntTest.java

server/src/test/java/org/apache/cassandra/sidecar/db/ClusterOpsDatabaseAccessorTest.java
andresbeckruiz: Replaced with ClusterOpsDatabaseAccessorIntTest.java

server/src/test/java/org/apache/cassandra/sidecar/db/ClusterOpsNodeStateDatabaseAccessorTest.java
andresbeckruiz: Replaced with ClusterOpsNodeStateDatabaseAccessorIntTest.java

andresbeckruiz:
CI: https://app.circleci.com/pipelines/github/andresbeckruiz/cassandra-sidecar/25/workflows/8e009e4b-f0b6-42d4-b5c7-511f5fef6faf

server/src/test/integration/org/apache/cassandra/sidecar/db/ActiveClusterOpsDatabaseAccessorIntTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+import org.apache.cassandra.sidecar.testing.IntegrationTestBase;
+import org.apache.cassandra.testing.CassandraIntegrationTest;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration tests for {@link ActiveClusterOpsDatabaseAccessor}
+ */
+class ActiveClusterOpsDatabaseAccessorIntTest extends IntegrationTestBase
```
isaacreath: Nit: Most other integration tests use the full name `IntegrationTest` instead of `IntTest`. I would recommend sticking with that convention. The only exception seeming to be the integration tests in the `db` package.

server/src/test/integration/org/apache/cassandra/sidecar/db/ActiveClusterOpsDatabaseAccessorIntTest.java
```diff
@@ -0,0 +1,94 @@
...
+{
+    @CassandraIntegrationTest
+    void testLwtOperations()
```
isaacreath: I know that the pattern of having a single large test like this is present in other integration test files (e.g., `RestoreJobDatabaseAccessorIntTest`). Is this strictly necessary? Or can we break this into many smaller test cases? 

If we can't (e.g., due to the integration testing framework needing to spin up a new cluster for each case), it may be nice to add a clear fail message, e.g. `getActiveOperations for a cluster should return null when no active operation exists but we found it did not` using the `withFailMessage` function.

server/src/main/java/org/apache/cassandra/sidecar/db/ClusterOpsNodeStateDatabaseAccessor.java
```diff
@@ -0,0 +1,111 @@
...
+import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.VisibleForTesting;
+
+/**
+ * Database accessor for the {@code cluster_ops_node_state} table.
+ */
+@Singleton
+public class ClusterOpsNodeStateDatabaseAccessor extends DatabaseAccessor<ClusterOpsNodeStateSchema>
+{
+    private static final int BATCH_CHUNK_SIZE = 100;
```
isaacreath: Should we consider making the BATCH_CHUNK_SIZE configurable through the constructor, with a default to this value? Not necessarily something we need to expose via `sidecar.yaml` today but would provide the pathway to doing so.

server/src/test/integration/org/apache/cassandra/sidecar/db/ActiveClusterOpsDatabaseAccessorIntTest.java
andresbeckruiz: In order to avoid spinning up a new cluster for each test case, added clear failure messages for each behavior tested

isaacreath (approved):
+1 (nb)

server/src/main/java/org/apache/cassandra/sidecar/job/storage/StorageProvider.java
```diff
@@ -0,0 +1,197 @@
...
+     * Implementations should throw {@link StorageProviderException} on write failure.
+     *
+     * @param jobId         the job identifier
+     * @param operationType the operation type (e.g. "restart")
+     * @param status        the new status
+     */
+    void updateJobStatus(UUID jobId, String operationType, OperationalJobStatus status);
+
+    /**
+     * Retrieve stored job records, up to the specified limit.
```
pauloricardomg: nit: Retrieve stored job records by descending time order

server/src/main/java/org/apache/cassandra/sidecar/db/ActiveClusterOpsDatabaseAccessor.java
```diff
@@ -0,0 +1,98 @@
...
+    public ActiveClusterOpsDatabaseAccessor(ActiveClusterOpsSchema schema, CQLSessionProvider sessionProvider)
+    {
+        super(schema, sessionProvider);
+    }
+
+    public boolean trySetActiveOperation(String clusterName, String operationType, UUID operationId)
+    {
+        BoundStatement statement = tableSchema.trySetActive().bind(clusterName, operationType, operationId);
+        statement.setConsistencyLevel(ConsistencyLevel.LOCAL_QUORUM);
+        statement.setSerialConsistencyLevel(ConsistencyLevel.LOCAL_SERIAL);
```
pauloricardomg, line 64: Shouldn't we be using global `SERIAL` to ensure the active operation will be consistent across all DCs?

server/src/main/java/org/apache/cassandra/sidecar/db/ActiveClusterOpsDatabaseAccessor.java
```diff
@@ -0,0 +1,98 @@
...
+            activeOps.put(row.getString("operation_type"), row.getUUID("operation_id"));
+        }
+        return activeOps;
+    }
+
+    public boolean clearActiveOperation(String clusterName, String operationType, UUID operationId)
+    {
+        BoundStatement statement = tableSchema.clearActive().bind(clusterName, operationType, operationId);
+        statement.setConsistencyLevel(ConsistencyLevel.LOCAL_QUORUM);
+        statement.setSerialConsistencyLevel(ConsistencyLevel.LOCAL_SERIAL);
```
pauloricardomg, line 99: Shouldn't we be using global SERIAL to ensure the active operation will be consistent across all DCs?

server/src/main/java/org/apache/cassandra/sidecar/db/ActiveClusterOpsDatabaseAccessor.java
andresbeckruiz: We are initially defaulting to `LOCAL_SERIAL` because `GLOBAL_SERIAL` for globally distributed clusters (eg: us-east, us-west, eu) would likely timeout. `LOCAL_SERIAL` can still guarantee that we only have one job ongoing per datacenter, assuming submitted nodes are restricted to that DC. 

Once configuration is added for the operational job framework, we should allow consistency level to be set as needed.

server/src/main/java/org/apache/cassandra/sidecar/db/ActiveClusterOpsDatabaseAccessor.java
andresbeckruiz: See https://github.com/apache/cassandra-sidecar/pull/339#discussion_r3243098147

server/src/main/java/org/apache/cassandra/sidecar/job/storage/StorageProvider.java
```diff
@@ -0,0 +1,198 @@
...
+    /**
+     * Update the status of an existing job.
+     * <p>
+     * Implementations should throw {@link StorageProviderException} on write failure.
+     *
+     * @param jobId         the job identifier
+     * @param operationType the operation type (e.g. "restart")
+     * @param status        the new status
+     */
+    void updateJobStatus(UUID jobId, String operationType, OperationalJobStatus status);
```
pauloricardomg: Is there any reason `operationType` is a String and not an enum ? this would be safer. We could have it with the currently supported operations supported by the operational jobs framework (even if they don't currently use the backend).

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ActiveClusterOpsSchema.java
```diff
@@ -0,0 +1,137 @@
...
+        return TABLE_NAME;
+    }
+
+    @Override
+    protected String createSchemaStatement()
+    {
+        return String.format("CREATE TABLE IF NOT EXISTS %s.%s (" +
+                             "  cluster_name text," +
+                             "  operation_type text," +
+                             "  operation_id uuid," +
```
pauloricardomg: why is this `uuid` and not `timeuuid` like in the `cluster_ops` table ?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ClusterOpsNodeStateSchema.java
```diff
@@ -0,0 +1,129 @@
...
+    {
+        return TABLE_NAME;
+    }
+
+    @Override
+    protected String createSchemaStatement()
+    {
+        return String.format("CREATE TABLE IF NOT EXISTS %s.%s (" +
+                             "  cluster_name text," +
+                             "  operation_id uuid," +
```
pauloricardomg: why is this `uuid` and not `timeuuid` like in the `cluster_ops` table ?

andresbeckruiz: Changed to `timeuuid` for consistency

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ActiveClusterOpsSchema.java
andresbeckruiz: Changed to `timeuuid` for consistency

server/src/main/java/org/apache/cassandra/sidecar/job/storage/StorageProvider.java
andresbeckruiz: Added `OperationType` enum

server/src/main/java/org/apache/cassandra/sidecar/db/ActiveClusterOpsDatabaseAccessor.java
isaacreath: If we're going for per-DC jobs, then it may make sense to add this into the schema as well so that we can get the active job by DC.

andresbeckruiz: Added datacenter to ActiveClusterOps schema in [55bdd43](https://github.com/apache/cassandra-sidecar/pull/339/commits/55bdd43c84f613c43558ee4a31091a410b3b78dd)

pauloricardomg (approved):

andresbeckruiz:
CI: https://app.circleci.com/pipelines/github/andresbeckruiz/cassandra-sidecar/69/workflows/edbd23a5-d194-4534-83f9-455270806a73

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/OperationalJobConfigurationImpl.java
```diff
@@ -0,0 +1,101 @@
...
+
+    protected OperationalJobConfigurationImpl(Builder builder)
+    {
+        this.tablesTtl = builder.tablesTtl;
+        validate();
+    }
+
+    private void validate()
+    {
+        if (tablesTtl.compareTo(MIN_TABLES_TTL) < 0)
```
sarankk, line 49: Why are we restricting `tablesTtl` with min value if we are allowing it to be configurable ?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ClusterOpsNodeStateSchema.java
```diff
@@ -0,0 +1,129 @@
...
+    protected String tableName()
+    {
+        return TABLE_NAME;
+    }
+
+    @Override
+    protected String createSchemaStatement()
+    {
+        return String.format("CREATE TABLE IF NOT EXISTS %s.%s (" +
+                             "  cluster_name text," +
```
sarankk, line 68: Can sidecar manage operations of other clusters? why do we have `cluster_name` in schema here?

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ClusterOpsSchema.java
```diff
@@ -0,0 +1,193 @@
...
+    protected String tableName()
+    {
+        return TABLE_NAME;
+    }
+
+    @Override
+    protected String createSchemaStatement()
+    {
+        return String.format("CREATE TABLE IF NOT EXISTS %s.%s (" +
+                             "  cluster_name text," +
```
sarankk, line 71: Same as above, can sidecar manage operations of other clusters?

client-common/src/main/java/org/apache/cassandra/sidecar/common/data/OperationType.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Enumerates the types of operational jobs that can be performed on a Cassandra cluster.
+ */
+public enum OperationType
+{
+    DECOMMISSION,
+    DRAIN,
+    MOVE,
+    REPAIR,
+    RESTART
```
sarankk: Nit: Don't see functionality for restart yet, could we add `RESTART` type later when functionality is added?

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/OperationalJobConfigurationImpl.java
```diff
@@ -0,0 +1,101 @@
...
+    @JsonProperty(value = "tables_ttl")
+    public SecondBoundConfiguration tablesTtl()
+    {
+        return tablesTtl;
+    }
+
+    @JsonProperty(value = "tables_ttl")
+    public void setTablesTtl(SecondBoundConfiguration tablesTtl)
+    {
+        this.tablesTtl = tablesTtl;
```
sarankk: Set method bypasses the validation done and is inconsistent. We can remove this, since it is not used

server/src/main/java/org/apache/cassandra/sidecar/db/ClusterOpsDatabaseAccessor.java
```diff
@@ -0,0 +1,174 @@
...
+                                         clusterName, jobId, operationType.name());
+        }
+        statement.setConsistencyLevel(ConsistencyLevel.LOCAL_QUORUM);
+        execute(statement);
+    }
+
+    private boolean shouldSetStartTime(String clusterName, UUID jobId)
+    {
+        OperationalJobRecord existing = findJob(clusterName, jobId);
+        return existing != null && existing.startTime() == null;
```
sarankk, line 136: `existing != null && existing.startTime() == null` check could cause start time to be updated by multiple sidecar instances at similar times, could we avoid multiple start time updates?

server/src/main/java/org/apache/cassandra/sidecar/db/ClusterOpsDatabaseAccessor.java
```diff
@@ -0,0 +1,174 @@
...
+
+    public void persistJob(String clusterName, OperationalJobRecord job)
+    {
+        Date lastUpdate = Date.from(Instant.now());
+        BoundStatement statement = tableSchema.insertJob()
+                                              .bind(clusterName,
+                                                    job.jobId(),
+                                                    job.operationType().name(),
+                                                    job.status().name(),
+                                                    lastUpdate,
```
sarankk, line 80: We are missing failure_reason and start_time in persist. Updated values in `OperationalJobRecord` will not be persisted

server/src/main/java/org/apache/cassandra/sidecar/job/storage/OperationalJobRecord.java
```diff
@@ -0,0 +1,208 @@
...
+     * @return the operation parameters, or null if not set
+     */
+    @Nullable
+    public Map<String, String> operationMetadata()
+    {
+        return operationMetadata;
+    }
+
+    @Override
+    public boolean equals(Object o)
```
sarankk: Why do we have `equals` and `hashCode` added for `OperationalJobRecord`? are we planning to have an in-memory copy of the jobs too in a Map? If not we should remove them

server/src/main/java/org/apache/cassandra/sidecar/job/storage/StorageProvider.java
```diff
@@ -0,0 +1,202 @@
...
+     */
+    @NotNull
+    List<OperationalJobRecord> findAllJobs(int limit);
+
+    // --- Active Operation Coordination ---
+
+    /**
+     * Set an operation as active if no other operation of the same type is currently active.
+     * <p>
+     * Implementations must provide compare-and-set (CAS) semantics to ensure only one active
```
sarankk, line 94: `ensure only one active operation of a given type runs at a time` should we restrict this based on operation type for only some operation types like repair? for other types like decommission, restart we might simultaneously decommission or restart few nodes?

sarankk (commented):
Thanks Andres, Overall looks good, left some comments.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/OperationalJobConfigurationImpl.java
andresbeckruiz: The goal is to provide a safety net to prevent accidental misconfiguration which could result in losing operational job state (eg: setting TTL to 1 day). [RestoreJobConfigurationImpl](https://github.com/apache/cassandra-sidecar/blob/trunk/server/src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java#L38-L39) employs a similar pattern of restricting minimum TTL for the restores job table

server/src/main/java/org/apache/cassandra/sidecar/db/ClusterOpsDatabaseAccessor.java
andresbeckruiz: `failure_reason` and `start_time` should not exist when a job is persisted for the first time, as the job has not failed or started yet.  Those fields are updated in `updateStatusWithStartTime` and `updateStatusWithFailure`.

server/src/main/java/org/apache/cassandra/sidecar/db/ClusterOpsDatabaseAccessor.java
andresbeckruiz: Yes, you are right that there could be a race between two separate Sidecar instances. The alternative would be using a LWT to guarantee one Sidecar sets the start time. Two Sidecars racing here would both set `start_time` to approximately the same value, so I opted for the simpler read-then-write to avoid adding Paxos overhead for an informational timestamp. That being said, I'm fine with switching to a LWT if that is preferred

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ClusterOpsNodeStateSchema.java
andresbeckruiz: `cluster_name` represents the cluster being opeated on in the operational job. It was added to allow a single Cassandra cluster to store operational job state for various clusters managed by separate Sidecar fleets. This also allows a separate Cassandra cluster from the one being managed by the Sidecar fleet to be used for state storage, a feature that was requested during the original design discussion of CEP-53. 

I've added a comment to clarify this in both schemas where it's used

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ClusterOpsSchema.java
andresbeckruiz: See reply above

server/src/main/java/org/apache/cassandra/sidecar/job/storage/StorageProvider.java
andresbeckruiz: The active operation lock ensures only one operation plan of a given type runs at a time, it does not restrict individual nodes. Within a single operation, multiple nodes can operate concurrently (with some restrictions-for example nodes should only restart concurrently as long as they don't replicate data with each other). This is controlled by `node_execution_order`.

Added a comment to the interface method to clarify this

server/src/main/java/org/apache/cassandra/sidecar/db/ClusterOpsDatabaseAccessor.java
sarankk: Got it, thanks for context, let's keep this. We can update if a more accurate timestamp is needed later.

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ClusterOpsNodeStateSchema.java
sarankk: @andresbeckruiz I see that through cluster ops and active ops database accessor we always write to the local table. If we want to provide the option to have a single cluster maintain ops data for multiple clusters, we would also need to update the schema to skip create statement steps optionally and add the ability for schema to use a session to connect to remote cluster for insert, update of job status. Is that something we are planning to add later on?

server/src/main/java/org/apache/cassandra/sidecar/db/ClusterOpsDatabaseAccessor.java
sarankk: If we use `persistJob` later on to update with updated `failure_reason`, we would miss the update silently. Currently I do not see uses of `persistJob` method. So will check in next PR how this method is used.

server/src/main/java/org/apache/cassandra/sidecar/job/storage/StorageProvider.java
sarankk: ah got it for a single `operation_type, operation_id` combination the lock holds. Makes sense 👍

sarankk (approved):
+1 Thanks Andres, Looks great to me. Approving since the pending comment does not require any code changes in this PR

server/src/main/java/org/apache/cassandra/sidecar/db/schema/ClusterOpsNodeStateSchema.java
andresbeckruiz: Good point, I think we can leave it out of this patch to limit the scope of this PR. I've created https://issues.apache.org/jira/browse/CASSSIDECAR-473 to track the work of enabling connection to a remote Cassandra cluster for durable state storage.

server/src/main/java/org/apache/cassandra/sidecar/db/ClusterOpsDatabaseAccessor.java
andresbeckruiz: Good point. Although `persistJob` will be used in the future for new jobs that shouldn't have `failure_reason` or `start_time` populated yet, if we wanted these values to not be null in the future we would silently miss the update. I've added `failure_reason` and `start_time` to the INSERT statement to be consistent with the full `OperationalJobRecord`.

sarankk (approved):
+1 Thanks, updated changes look good to me

PR #340, opened 2026-04-24
CASSSIDECAR-443: Support column types not parseable by Java 3.x driver

lukasz-antoniak:
Fixes [CASSSIDECAR-443](https://issues.apache.org/jira/browse/CASSSIDECAR-443).

### DIFF
--- a/integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraNodeOperationsIntegrationTest.java
+++ b/integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraNodeOperationsIntegrationTest.java
@@ -76,7 +76,7 @@ public class CassandraNodeOperationsIntegrationTest extends SharedClusterSidecar
         trustedClient().put(serverWrapper.serverPort, "localhost", ApiEndpointsV1.NODE_DRAIN_ROUTE)
                        .send());
 
-        assertThat(drainResponse.statusCode()).isEqualTo(OK.code());
+        assertThat(drainResponse.statusCode()).isIn(OK.code(), ACCEPTED.code());
 
         JsonObject responseBody = drainResponse.bodyAsJsonObject();
         assertThat(responseBody).isNotNull();
--- a/integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraVectorSchemaRouteIntegrationTest.java
+++ b/integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraVectorSchemaRouteIntegrationTest.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.routes;
+
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
+import io.vertx.core.http.HttpResponseExpectation;
+import org.apache.cassandra.sidecar.common.response.SchemaResponse;
+import org.apache.cassandra.sidecar.testing.QualifiedName;
+import org.apache.cassandra.sidecar.testing.SharedClusterSidecarIntegrationTestBase;
+import org.apache.cassandra.sidecar.utils.SimpleCassandraVersion;
+
+import static org.apache.cassandra.testing.utils.AssertionUtils.getBlocking;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assumptions.assumeThat;
+
+class CassandraVectorSchemaRouteIntegrationTest extends SharedClusterSidecarIntegrationTestBase
+{
+    protected static final int MIN_VERSION_WITH_VECTOR = 5;
+
+    @Override
+    protected void initializeSchemaForTest()
+    {
+        createTestKeyspace("test_keyspace", Map.of("replication_factor", 1));
+        createTestTable(new QualifiedName("test_keyspace", "int_table"),
+                        "CREATE TABLE IF NOT EXISTS %s (a int, b int, PRIMARY KEY (a))");
+        createTestTable(new QualifiedName("test_keyspace", "vector_table"),
+                        "CREATE TABLE IF NOT EXISTS %s (a int, b vector<float, 3>, PRIMARY KEY (a))");
+    }
+
+    @Override
+    protected void beforeClusterProvisioning()
+    {
+        assumeThat(SimpleCassandraVersion.create(testVersion.version()).major)
+        .as("Vector type is supported since Cassandra 5.0")
+        .isGreaterThanOrEqualTo(MIN_VERSION_WITH_VECTOR);
+    }
+
+    @Test
+    void testSchemaHandlerWithVectorTable()
+    {
+        String testRoute = "/api/v1/schema/keyspaces/test_keyspace";
+        SchemaResponse response = getBlocking(trustedClient()
+                                              .get(serverWrapper.serverPort, "localhost", testRoute)
+                                              .send()
+                                              .expecting(HttpResponseExpectation.SC_OK))
+                                  .bodyAsJson(SchemaResponse.class);
+        assertThat(response).isNotNull();
+        assertThat(response.keyspace()).isEqualTo("test_keyspace");
+        assertThat(response.schema()).contains("vector_table");
+    }
+}
--- a/server/src/main/java/org/apache/cassandra/sidecar/db/DriverUnsupportedSchemaCache.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/db/DriverUnsupportedSchemaCache.java
@@ -1,0 +1,293 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.db;
+
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.SortedMap;
+import java.util.TreeMap;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.KeyspaceMetadata;
+import com.datastax.driver.core.PreparedStatement;
+import com.datastax.driver.core.Row;
+import com.datastax.driver.core.Session;
+import com.datastax.driver.core.TableMetadata;
+import com.google.inject.Singleton;
+import io.vertx.core.Promise;
+import org.apache.cassandra.sidecar.common.server.CQLSessionProvider;
+import org.apache.cassandra.sidecar.common.server.data.Name;
+import org.apache.cassandra.sidecar.common.server.data.QualifiedTableName;
+import org.apache.cassandra.sidecar.common.server.utils.DurationSpec;
+import org.apache.cassandra.sidecar.common.utils.StringUtils;
+import org.apache.cassandra.sidecar.config.SidecarConfiguration;
+import org.apache.cassandra.sidecar.exceptions.CassandraUnavailableException;
+import org.apache.cassandra.sidecar.tasks.PeriodicTask;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.VisibleForTesting;
+
+/**
+ * The {@link DriverUnsupportedSchemaCache} class maintains cache of CQL schema for tables whose definition is not
+ * supported by driver natively. Java driver 3.x does not return {@link TableMetadata} for tables which schema
+ * could not be parsed. Since it does not currently support vector type, any table containing vector would not
+ * be visible. Cache is refreshed for the first time as soon as CQL connection is established or upon first lookup.
+ * Later, it is periodically refreshed according to configured schedule.
+ * TODO: Remove after upgrade to Java driver 4.x (CASSSIDECAR-421).
+ */
+@Singleton
+public class DriverUnsupportedSchemaCache implements PeriodicTask
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(DriverUnsupportedSchemaCache.class);
+    private static final String STATEMENT_DELIMITER = "\n\n";
+
+    private final SidecarConfiguration sidecarConfiguration;
+    private final CQLSessionProvider sessionProvider;
+    private final CQLSchemaAccessor schemaAccessor;
+
+    // cache contains only schemas unparseable by Java driver
+    // during every cache refresh, reference is being reassigned to achieve atomic swap
+    private volatile SortedMap<QualifiedTableName, String> schemaCache;
+    private volatile boolean initialized; // flag indicating whether cache has been populated at least once
+
+    private PreparedStatement tableListStatement;
+
+    public DriverUnsupportedSchemaCache(SidecarConfiguration sidecarConfiguration,
+                                        CQLSessionProvider sessionProvider)
+    {
+        this.sidecarConfiguration = sidecarConfiguration;
+        this.sessionProvider = sessionProvider;
+        this.schemaAccessor = new CQLSchemaAccessor(sessionProvider);
+        this.schemaCache = createCache();
+        this.initialized = false;
+    }
+
+    /**
+     * @return Schema for all tables across all keyspaces (only those not supported by Java driver).
+     */
+    @NotNull
+    public String getFullSchema()
+    {
+        // we cannot know whether schema was updated after last
+        // periodic refresh, so results may be stale
+        refreshIfUninitialized();
+        return getUnsupportedSchema(table -> true);
+    }
+
+    /**
+     * @return Schema for all tables within given keyspaces (only those not supported by Java driver).
+     */
+    @NotNull
+    public String getKeyspaceSchema(@NotNull Name keyspace)
+    {
+        // we cannot know whether schema was updated after last
+        // periodic refresh, so results may be stale
+        refreshIfUninitialized();
+        return getUnsupportedSchema(table -> keyspace.equals(table.getKeyspace()));
+    }
+
+    /**
+     * @return Schema for table not supported by Java driver's metadata, {@code null} otherwise.
+     */
+    @Nullable
+    public String getTableSchema(@NotNull Name keyspace, @NotNull Name table)
+    {
+        return getTableSchema(keyspace, table, true);
+    }
+
+    /**
+     * @param allowRefresh Flag indicating whether lookup of table schema via CQL query
+     *                     is allowed, when data not found in cache.
+     * @return Schema for table not supported by Java driver's metadata, {@code null} otherwise.
+     */
+    @Nullable
+    public String getTableSchema(@NotNull Name keyspace, @NotNull Name table, boolean allowRefresh)
+    {
+        refreshIfUninitialized();
+        QualifiedTableName name = new QualifiedTableName(keyspace, table);
+        String schema = schemaCache.get(name);
+        if (schema == null && allowRefresh)
+        {
+            // proactively fetch table schema, not to provide
+            // false-negative when table is not cached yet
+            // attention, method can block
+            schema = populateSchemaCache(schemaCache, name);
+        }
+        return schema;
+    }
+
+    @Override
+    public DurationSpec delay()
+    {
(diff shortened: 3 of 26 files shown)


server/src/main/java/org/apache/cassandra/sidecar/handlers/sstableuploads/SSTableUploadHandler.java
```diff
@@ -209,13 +214,17 @@ private Future<SSTableUploadRequestParam> validateKeyspaceAndTable(String host,
...
                        }
-
                        if (MetadataUtils.table(keyspaceMetadata, request.table()) == null)
                        {
-                           String message = String.format("Invalid table name '%s' supplied for keyspace '%s'",
-                                                          request.table(), request.keyspace());
-                           logger.error(message);
-                           return Future.failedFuture(wrapHttpException(HttpResponseStatus.BAD_REQUEST, message));
+                           // check for tables that are not parseable by Java driver's metadata feature
+                           boolean tableExists = driverUnsupportedSchemaCache.getTableSchema(request.keyspace(), request.table(), false) != null;
```
lukasz-antoniak: Do not refresh if table schema is not present in cache (prevent blocking due to CQL query).

server/src/main/java/org/apache/cassandra/sidecar/handlers/validations/ValidateTableExistenceHandler.java
```diff
@@ -97,7 +103,14 @@ protected void handleInternal(RoutingContext context,
             }
 
             TableMetadata tableMetadata = keyspaceMetadata.getTable(table);
-            if (tableMetadata == null)
+            boolean tableExists = tableMetadata != null;
+            if (!tableExists)
+            {
+                tableExists = driverUnsupportedSchemaCache.getTableSchema(new Name(keyspaceMetadata.getName()),
+                                                                          new Name(table),
+                                                                          false) != null;
```
lukasz-antoniak: Do not refresh if table schema is not present in cache (prevent blocking due to CQL query).

server/src/main/java/org/apache/cassandra/sidecar/db/DriverUnsupportedSchemaCache.java
```diff
@@ -0,0 +1,277 @@
...
+/**
+ * The {@link DriverUnsupportedSchemaCache} class maintains cache of CQL schema for tables whose definition is not
+ * supported by driver natively. Java driver 3.x does not return {@link TableMetadata} for tables which schema
+ * could not be parsed. Since it does not currently support vector type, any table containing vector would not
+ * be visible. Cache is refreshed for the first time as soon as CQL connection is established or upon first lookup.
+ * Later, it is periodically refreshed according to configured schedule.
+ * TODO: Remove after upgrade to Java driver 4.x (CASSSIDECAR-421).
+ */
+@Singleton
+public class DriverUnsupportedSchemaCache implements PeriodicTask
```
lukasz-antoniak, line 64: Could not come up with a better class name.

server/src/main/java/org/apache/cassandra/sidecar/handlers/KeyspaceSchemaHandler.java
```diff
@@ -94,7 +100,9 @@ private void handleWithMetadata(RoutingContext context, Name keyspace, Metadata
     {
         if (keyspace == null)
         {
-            SchemaResponse schemaResponse = new SchemaResponse(metadata.exportSchemaAsString());
+            String fullSchema = metadata.exportSchemaAsString();
+            fullSchema += driverUnsupportedSchemaCache.getFullSchema();
```
skoppu22: - If the first fullschema above doesn't end with newlines and the cache's schema is non-empty, the two schemas run together without whitespace separation.
- Also, we are doing this kind of concatenation in multiple places in this PR. Can we have a small utility function ensuring concatenation and whitespace separation etc handled properly and the same is called everywhere?

server/src/main/java/org/apache/cassandra/sidecar/handlers/validations/ValidateTableExistenceHandler.java
skoppu22: if tableMetadata is null and tableExists becomes true when driverUnsupportedSchemaCache found the table, this code still stores null in the RoutingContext instead of storing the actual metadata

lukasz-antoniak: That is true. I could not find anywhere any usage of this constant, so thought it does not harm. To construct any meaningful `TableMetadata`, we would need to parse the schema, so decided to leave it like this for now. Am I missing something?

server/src/main/java/org/apache/cassandra/sidecar/handlers/KeyspaceSchemaHandler.java
skoppu22: This is getting called in the event loop and the below call getFullSchema can invoke cache refresh code which runs under synchronized. Need to move this to worker pool using executeBlocking.

server/src/main/java/org/apache/cassandra/sidecar/db/DriverUnsupportedSchemaCache.java
```diff
@@ -0,0 +1,277 @@
...
+        }
+    }
+
+    private String populateSchemaCache(Map<QualifiedTableName, String> cache, QualifiedTableName table)
+    {
+        List<String> cqlSchema = schemaAccessor.getTableSchema(table.getKeyspace(), table.table());
+        if (cqlSchema != null)
+        {
+            String schema = String.join(STATEMENT_DELIMITER, cqlSchema);
+            cache.put(table, schema);
```
skoppu22, line 231: Instead of using passed cache, can we do this.schemaCache.put( ..) to ensure we are always inserting in the latest cache (if other thread just created a new cache during a refresh by any chance)

server/src/main/java/org/apache/cassandra/sidecar/db/DriverUnsupportedSchemaCache.java
```diff
@@ -0,0 +1,277 @@
...
+            else
+            {
+                schemaCache.clear();
+            }
+
+            initialized = true;
+        }
+        catch (CassandraUnavailableException ignored)
+        {
+            LOGGER.debug("Not yet connect to Cassandra cluster");
```
skoppu22: [minor] 'connected'

server/src/main/java/org/apache/cassandra/sidecar/handlers/KeyspaceSchemaHandler.java
lukasz-antoniak: `driverUnsupportedSchemaCache.getFullSchema()` will block only if we do not have schema initialised at least once. To prevent "first request" latency, we load the schema as soon as CQL connection is established (see `ConfigurationModule`). My purpose was to change as little code as possible due to this workaround. I had a version to do CQL lookup on non-event loop thread, but then I had to change complete `handleWithMetadata` method. What do you think?

server/src/main/java/org/apache/cassandra/sidecar/db/DriverUnsupportedSchemaCache.java
lukasz-antoniak: Not exactly, because we would not remove the tables that were dropped in the meantime. We swap the reference to `schemaCache` to make the swap atomic.

server/src/main/java/org/apache/cassandra/sidecar/db/DriverUnsupportedSchemaCache.java
lukasz-antoniak: That was copy-paste from `CachedLocalTokenRanges` :). Fixing in both places.

server/src/main/java/org/apache/cassandra/sidecar/db/DriverUnsupportedSchemaCache.java
```diff
@@ -0,0 +1,277 @@
...
+            {
+                LOGGER.debug("Tables not supported by Java driver metadata: {}", tables);
+                SortedMap<QualifiedTableName, String> newCache = createCache();
+                tables.forEach(table -> populateSchemaCache(newCache, table));
+                // replacing cache, because some tables might have been removed in the meanwhile
+                schemaCache = newCache;
+            }
+            else
+            {
+                schemaCache.clear();
```
skoppu22: Another thread may be iterating over schemaCache map in getUnsupportedSchema while we do schemaCache.clear() here, resulting in concurrent modification exception. Instead of clear, if we do schemaCache = createCache(), other thread can continue iterating older map.

lukasz-antoniak: Brilliant catch, thanks! I just learned that `synchronizedSortedMap` does not synchronize iterator-based traversals (JavaDoc).

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraVectorSchemaRouteIntegrationTest.java
```diff
@@ -0,0 +1,70 @@
...
+    {
+        String testRoute = "/api/v1/schema/keyspaces/test_keyspace";
+        SchemaResponse response = getBlocking(trustedClient()
+                                              .get(serverWrapper.serverPort, "localhost", testRoute)
+                                              .send()
+                                              .expecting(HttpResponseExpectation.SC_OK))
+                                  .bodyAsJson(SchemaResponse.class);
+        assertThat(response).isNotNull();
+        assertThat(response.keyspace()).isEqualTo("test_keyspace");
+        assertThat(response.schema()).contains("vector_table");
```
skoppu22, line 68: Nice. Can we have a integration test for SSTable upload path as well for vectors?

server/src/main/java/org/apache/cassandra/sidecar/handlers/validations/ValidateTableExistenceHandler.java
skoppu22: Perhaps we don't need to set table metadata in routing context when tableMetadata is null.

server/src/main/java/org/apache/cassandra/sidecar/modules/ConfigurationModule.java
```diff
@@ -126,6 +132,25 @@ CQLSessionProvider cqlSessionProvider(Vertx vertx,
...
 
+    @Provides
+    @Singleton
+    DriverUnsupportedSchemaCache driverUnsupportedSchemaCache(Vertx vertx,
+                                                              SidecarConfiguration sidecarConfiguration,
+                                                              CQLSessionProvider cqlSessionProvider)
+    {
+        DriverUnsupportedSchemaCache schemaCache = new DriverUnsupportedSchemaCache(sidecarConfiguration, cqlSessionProvider);
+        // trigger immediate cache population once Sidecar connects to Cassandra node
+        vertx.eventBus().localConsumer(ON_CASSANDRA_CQL_READY.address(), message -> schemaCache.refresh(true));
```
skoppu22: This will block event loop. Some other consumers of ON_CASSANDRA_CQL_READY are using EventBusUtils.onceLocalConsumer + PeriodicTaskExecutor.schedule() pattern to avoid blocking

server/src/main/java/org/apache/cassandra/sidecar/handlers/KeyspaceSchemaHandler.java
skoppu22: If server is down or CQL failed to connect during the first initialization, subsequent incoming call will run on the event loop blocking other incoming requests processing. Something like below might work. @frankgh is vertx expert, might suggest simpler option.

```
executorPools.service().executeBlocking(() -> driverUnsupportedSchemaCache.getFullSchema())
             .onSuccess(unsupportedSchema -> {
                 String fullSchema = metadata.exportSchemaAsString() + unsupportedSchema;
                 context.json(new SchemaResponse(fullSchema));
             });
```

lukasz-antoniak: I think I have applied an sensible solution. Please check.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraVectorSchemaRouteIntegrationTest.java
lukasz-antoniak: Added `SSTableVectorUploadHandlerIntegrationTest`.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraNodeOperationsIntegrationTest.java
```diff
@@ -76,7 +76,7 @@ void testNodeDrainOperationSuccess()
         trustedClient().put(serverWrapper.serverPort, "localhost", ApiEndpointsV1.NODE_DRAIN_ROUTE)
                        .send());
 
-        assertThat(drainResponse.statusCode()).isEqualTo(OK.code());
+        assertThat(drainResponse.statusCode()).isIn(OK.code(), ACCEPTED.code());
```
yifan-c: Why is it necessary to change the assertion?

server/src/test/integration/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableVectorUploadHandlerIntegrationTest.java
```diff
@@ -0,0 +1,228 @@
...
+import org.apache.cassandra.testing.CassandraIntegrationTest;
+import org.apache.cassandra.testing.IClusterExtension;
+import org.apache.cassandra.testing.utils.AssertionUtils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assumptions.assumeThat;
+
+/**
+ * Integration tests for {@link SSTableUploadHandler}, {@link SSTableImportHandler} and vector type
+ * TODO: Consider removing after upgrade to Java driver 4.x (CASSSIDECAR-421).
```
skoppu22, line 61: No need to remove this test after upgrading to 4.x driver I believe

server/src/main/java/org/apache/cassandra/sidecar/handlers/validations/ValidateTableExistenceHandler.java
```diff
@@ -97,12 +103,19 @@ protected void handleInternal(RoutingContext context,
...
+            }
+            if (!tableExists)
             {
                 String errMsg = "Table " + input.tableName() + " was not found for keyspace " + input.keyspace();
                 context.fail(wrapHttpException(HttpResponseStatus.NOT_FOUND, errMsg));
             }
-            else
+            else if (tableMetadata != null)
```
skoppu22: There is no else part below (i.e, no case handling tableMetadata = null) and context.next() or context.fail() not called in that case and handler will hang until timeout.

server/src/main/java/org/apache/cassandra/sidecar/modules/ConfigurationModule.java
```diff
@@ -126,6 +133,25 @@ CQLSessionProvider cqlSessionProvider(Vertx vertx,
...
 
+    @Provides
+    @Singleton
+    DriverUnsupportedSchemaCache driverUnsupportedSchemaCache(Vertx vertx,
+                                                              SidecarConfiguration sidecarConfiguration,
+                                                              CQLSessionProvider cqlSessionProvider)
+    {
+        DriverUnsupportedSchemaCache schemaCache = new DriverUnsupportedSchemaCache(sidecarConfiguration, cqlSessionProvider);
+        // trigger immediate cache population once Sidecar connects to Cassandra node
+        EventBusUtils.onceLocalConsumer(vertx.eventBus(), ON_CASSANDRA_CQL_READY.address(), ignored -> schemaCache.refresh(true));
```
skoppu22: Thanks for adding onceLocalConsumer. I believe the handler body still executes on the event loop. schemaCache.refresh(true) is synchronized and runs blocking CQL.

Every other ON_CASSANDRA_CQL_READY consumer in the codebase uses the pattern onceLocalConsumer(…, ignored -> executor.schedule(this)). Could we either register DriverUnsupportedSchemaCache as a PeriodicTask via executor.schedule(...), or wrap with vertx.executeBlocking(() -> schemaCache.refresh(true))?

server/src/main/java/org/apache/cassandra/sidecar/db/DriverUnsupportedSchemaCache.java
```diff
@@ -0,0 +1,287 @@
...
+
+    /**
+     * @param allowRefresh Flag indicating whether lookup of table schema via CQL query
+     *                     is allowed, when data not found in cache.
+     * @return Schema for table not supported by Java driver's metadata, {@code null} otherwise.
+     */
+    @Nullable
+    public String getTableSchema(@NotNull Name keyspace, @NotNull Name table, boolean allowRefresh)
+    {
+        refreshIfUninitialized();
```
yifan-c, line 131: Does it mean that the `allowRefresh: false` might not be honored if uninitialized?

server/src/main/java/org/apache/cassandra/sidecar/handlers/validations/ValidateTableExistenceHandler.java
yifan-c: If table is not initialized yet, it still may block, even through `false` is passed.

server/src/main/java/org/apache/cassandra/sidecar/handlers/sstableuploads/SSTableUploadHandler.java
yifan-c: Same. `getTableSchema` may block unconditionally, if not initialized.

server/src/main/java/org/apache/cassandra/sidecar/handlers/KeyspaceSchemaHandler.java
```diff
@@ -130,6 +144,27 @@ private Future<Metadata> metadata(String host)
...
+    /**
+     * Get CQL schema not parseable by Java driver (and therefore not present in {@link Metadata}).
+     *
+     * @param keyspace optional keyspace name, {@code null} includes schema for all keyspaces
+     * @return {@link Future} containing CQL schema
+     */
+    private Future<String> unsupportedSchema(Name keyspace)
+    {
+        return executorPools.service().executeBlocking(() -> {
+            // schema cache can block if it was not successfully initialized at least once
```
yifan-c, line 156: 👍

server/src/main/java/org/apache/cassandra/sidecar/db/DriverUnsupportedSchemaCache.java
```diff
@@ -0,0 +1,287 @@
...
+        }
+        catch (CassandraUnavailableException ignored)
+        {
+            LOGGER.debug("Not yet connected to Cassandra cluster");
+        }
+    }
+
+    private String populateSchemaCache(Map<QualifiedTableName, String> cache, QualifiedTableName table)
+    {
+        List<String> cqlSchema = schemaAccessor.getTableSchema(table.getKeyspace(), table.table());
```
yifan-c: getKeyspace and  table can return null value, but the `getTableSchema` method expects `@NotNull` values.

server/src/test/integration/org/apache/cassandra/sidecar/routes/sstableuploads/SSTableVectorUploadHandlerIntegrationTest.java
```diff
@@ -0,0 +1,228 @@
...
+            String testSchemaRoute = "/api/v1/cassandra/schema";
+            HttpResponse<Buffer> response = client.get(server.actualPort(), "127.0.0.1", testSchemaRoute).send()
+                                                  .toCompletionStage().toCompletableFuture().join();
+            assertThat(response.statusCode()).isEqualTo(HttpResponseStatus.OK.code());
+            assertThat(response.bodyAsString()).contains(tableName.tableName());
+        });
+
+        // create a snapshot called <tableName>-snapshot for tbl1
+        IClusterExtension<? extends IInstance> cluster = sidecarTestContext.cluster();
+        final String snapshotStdout = cluster.get(1).nodetoolResult("snapshot",
```
yifan-c: nit: drop the `final` to be consistent with the code style. 

(That said, I know it is easy to miss)

yifan-c (commented):
looks good overall.

server/src/main/java/org/apache/cassandra/sidecar/handlers/sstableuploads/SSTableUploadHandler.java
lukasz-antoniak: Fixed by running two futures in parallel, similar to `KeyspaceSchemaHandler`.

server/src/main/java/org/apache/cassandra/sidecar/modules/ConfigurationModule.java
lukasz-antoniak: `DriverUnsupportedSchemaCache` is already registered as periodic task and executed in fixed intervals. Tested it locally. I have wrapped execution with `vertx.executeBlocking(...)`. Thanks for pointing this one.

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/routes/CassandraNodeOperationsIntegrationTest.java
lukasz-antoniak: Below in the assertions, we check that job status is in created, running or succeeded state:
```
assertThat(responseBody.getString("jobStatus")).isIn(
OperationalJobStatus.CREATED.name(),
OperationalJobStatus.RUNNING.name(),
OperationalJobStatus.SUCCEEDED.name()
);
```
`ACCEPTED` maps to running state, and this test was flaky when I have executed complete test suite.
```
OperationalJobStatus status = job.status();
LOGGER.info("Job completion status={} jobId={}", status, job.jobId());
if (status.isCompleted())
{
    context.response().setStatusCode(HttpResponseStatus.OK.code());
}
else
{
    context.response().setStatusCode(HttpResponseStatus.ACCEPTED.code());
}
```

server/src/main/java/org/apache/cassandra/sidecar/db/DriverUnsupportedSchemaCache.java
lukasz-antoniak: Indeed, `allowRefresh` only controls if you allow to trigger a refresh on-demand if data is not present at current time. If the data was not ever initialized, we would always return `null`, so I am trying to prevent lots of false-negative answers.

lukasz-antoniak:
@skoppu22, @yifan-c, I think I addressed all comments, can you have another look?

yifan-c (approved):

skoppu22 (approved):

yifan-c:
This error is odd
```
[2](https://github.com/apache/cassandra-sidecar/actions/runs/25426292180/job/74582316487?pr=340#step:8:73)
* Exception is:
org.gradle.api.plugins.UnknownPluginException: Plugin [id: 'io.smallrye.openapi', version: '4.1.1'] was not found in any of the following sources:

- Gradle Core Plugins (plugin is not in 'org.gradle' namespace)
- Included Builds (No included builds contain this plugin)
- Plugin Repositories (could not resolve plugin artifact 'io.smallrye.openapi:io.smallrye.openapi.gradle.plugin:4.1.1')
  Searched in the following repositories:
    Gradle Central Plugin Repository
```

yifan-c:
Looks like it is just a glitch in github check. 
+1

PR #341, opened 2026-04-30
CASSSIDECAR-447: Scope all CDC dependencies exclusively to CdcModule

yifan-c:
Patch by Yifan Cai; Reviewed by TBD for CASSSIDECAR-447


server/src/main/java/org/apache/cassandra/sidecar/db/VirtualTablesDatabaseAccessor.java
yifan-c, line 1: The file is removed and the functionality (i.e. `isCdcOnRepairEnabled`) is merged into `CdcSystemViewsDatabaseAccessor` as they both access the settings vtable.

server/src/main/java/org/apache/cassandra/sidecar/modules/UtilitiesModule.java
```diff
@@ -88,6 +92,14 @@ SidecarVersionProvider sidecarVersionProvider()
         return new SidecarVersionProvider("/sidecar.version");
     }
 
+    @Provides
+    @Singleton
+    SidecarInstancesProvider sidecarInstancesProvider(InstancesMetadata instancesMetadata,
+                                                      ServiceConfiguration serviceConfiguration)
+    {
+        return new DynamicSidecarInstancesProvider(instancesMetadata, serviceConfiguration);
+    }
```
yifan-c, line 101: It was declared in the `CdcModule`, but should really be moved there as it is used by other components.

.github/workflows/ci.yml
yifan-c, line 1: I am going to patch separately. Do not review this file.

yifan-c: The change has been patched separately

server/src/main/java/org/apache/cassandra/sidecar/modules/ConfigurationModule.java
```diff
@@ -68,14 +69,30 @@
...
+    private final SidecarConfiguration config;
 
     /**
      * Constructs the Guice main module to run Cassandra Sidecar
      */
     public ConfigurationModule()
     {
         confPath = null;
+        config = null;
+    }
```
jyothsnakonisa: Can we get rid of this so that config/config path are not left unset by mistake. Removing this will give compiletime  error if config/path is not set correctly

jyothsnakonisa (approved):
Looks good overall, just minor comment.

jyothsnakonisa (approved):
Thanks Yifan! looks good to me

frankgh (approved):
+1 looks good to me

PR #343, opened 2026-05-06
CASSSIDECAR-451: Sidecar CDC should resolve real keyspace replication factors from the driver

Klose6:
Currently the analytics library uses ReplicationFactorSupplier.DEFAULT (SimpleStrategy/RF=1), which causes peer discovery to find only one replica per token range, it is better to add the support to resolve real keyspace replication factors from the driver.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 0.4.0
 -----
+ * Sidecar CDC should resolve real keyspace replication factors from the driver (CASSSIDECAR-451)
  * SAI support in Sidecar (CASSSIDECAR-422)
  * Update OperationalJob to support cluster-wide operations (CASSSIDECAR-376)
  * Support column types not parseable by Java 3.x driver (CASSSIDECAR-443)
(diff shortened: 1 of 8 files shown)


jberragan (commented):
Thanks for adding, not sure why this was neglected in the Sidecar implementation.

Klose6:
> Thanks for adding, not sure why this was neglected in the Sidecar implementation.

yeah, we found this during the cdc e2e testing.

Klose6:
@jberragan @bbotella could you take a look when you got a chance? thanks in advance!

bbotella:
Hi!! Really sorry for the delay!! I've been out traveling for a while, and just made it back. I'll be taking a look this week.

Klose6:
> Hi!! Really sorry for the delay!! I've been out traveling for a while, and just made it back. I'll be taking a look this week.

np, thank you so much!

jyothsnakonisa:
@Klose6  Thanks for the patch, I have this change as part of another PR https://github.com/apache/cassandra-sidecar/pull/347 Which is bringing in several changes for getting CDC working. Can we please wait for it to be merged? If there is anything else that you need to add in addition to that, we can work on getting those changes in.

Klose6:
> @Klose6 Thanks for the patch, I have this change as part of another PR #347 Which is bringing in several changes for getting CDC working. Can we please wait for it to be merged? If there is anything else that you need to add in addition to that, we can work on getting those changes in.

@jyothsnakonisa Sure I’ll wait for your PR to be merged first since it already contains my change. After that, I can rebase/update my branch accordingly if needed(but overall looks my changes is not needed after your PR). Thanks for coordinating this.
cc: @bbotella @jberragan

jyothsnakonisa:
@Klose6 I merged my PR to fix CDC issues yesterday, can u rebase and see if you need to add anything?

Klose6:
> @Klose6 I merged my PR to fix CDC issues yesterday, can u rebase and see if you need to add anything?

I just did a rebase and overall looks good, no need to add extra things, your PR uses the SchemaSupplier for the CDC enabled tables and my PR uses the InstanceMetadataFetcher, both will work but SchemaSupplier is simpler, I think we can just close this PR, thank you!

PR #344, opened 2026-05-06
CASSSIDECAR-453 cdc events routing

smiklosovic:


server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
```diff
@@ -136,7 +136,7 @@ public EventConsumer eventConsumer(CdcConfig conf)
                 instance.delegate().nodeSettings()).releaseVersion()
         ).getVersion();
         this.kafkaPublisher = KafkaPublisher.create(version,
-                                                    TopicSupplier.staticTopicSupplier(conf.kafkaTopic()),
```
smiklosovic, line 139: conf.kafkaTopic() is annotated with `@Nullable`. I do not know what we should do when it returns null. This was already the case before this patch. I do not see anything which would validate if kafka topic is null or not. Might be deferred to another ticket though.

jyothsnakonisa (approved):
Thanks Stefan, looks good to me. Please fix the check style

PR #346, opened 2026-05-11
CASSSIDECAR-454: Sidecar should wake up immediately on the instance r…

mansikhara:
## Summary                                                                                                                                                            
                                                                                                                                                                        
  UpdateRestoreJobHandler now immediately notifies the restore system after writing a phase signal (STAGE_READY or IMPORT_READY) to the DB, eliminating the 5–10 minute idle wait for the next RestoreJobDiscoverer polling cycle.                                                                                                            
                                                                                                                                                                        
  - **IMPORT_READY**: calls `RestoreJobManagerGroup.updateRestoreJob()` to propagate the new status to in-memory trackers, allowing `RestoreProcessor` to unblock staged
   ranges on its next 1-second tick.
  - **STAGE_READY**: calls `RestoreJobDiscoverer.processJobNow()` which discovers slices from the DB and submits them to `RestoreProcessor` immediately.                
                                                                                                                                                                        
  ## Design decisions                                                                                                                                                   
                                                                                                                                                                        
  - Notification is fire-and-forget on a worker thread (`executeBlocking`) — does not block the event loop or delay the HTTP 200 response.                              
  - Re-reads the full job from DB inside `notifyPhaseSignalMaybe` because `RestoreJobDatabaseAccessor.update()` returns a sparse object missing fields like keyspace and
   consistency level.                                                                                                                                                   
  - Both calls are additive and safe to miss — the DB write remains the durable source of truth, and the discovery loop still picks up the signal on its next cycle if
  the immediate notification fails.                                                                                                                                     
                                                            
  ## Test plan                                                                                                                                                          
                                                            
  - [x] `UpdateRestoreJobHandlerTest` — new tests verify IMPORT_READY triggers `updateRestoreJob` and STAGE_READY triggers `processJobNow`                              
  - [x] `RestoreJobDiscovererPhaseSignalIntTest` — new integration tests verify ranges are created immediately on STAGE_READY without explicit discovery, and
  IMPORT_READY does not create duplicates                                                                                                                               
  - [x] Existing `RestoreJobDiscovererNode*IntTest` tests pass (regression)


server/src/main/java/org/apache/cassandra/sidecar/handlers/restore/UpdateRestoreJobHandler.java
```diff
@@ -117,6 +125,12 @@ protected void handleInternal(RoutingContext context,
             }
 
             context.response().setStatusCode(HttpResponseStatus.OK.code()).end();
+            // Fire-and-forget on a worker thread — notifying the restore system should not
+            // block the event loop or delay the HTTP response.
+            executorPools.service().executeBlocking(() -> {
+                notifyPhaseSignalMaybe(job);
+                return null;
+            });
```
yifan-c: nit: it can be simplified to one line.

```suggestion
            executorPools.service().runBlocking(() -> notifyPhaseSignalMaybe(job));
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/restore/UpdateRestoreJobHandler.java
```diff
@@ -147,4 +161,38 @@ protected UpdateRestoreJobRequestPayload extractParamsOrThrow(RoutingContext con
...
+        if (job.status != RestoreJobStatus.IMPORT_READY && job.status != RestoreJobStatus.STAGE_READY)
+        {
+            return;
+        }
+
+        try
+        {
+            // The job returned from update() is sparse (only contains updated fields).
+            // Re-read the full job from DB to get all required fields (keyspace, consistency level, etc.).
+            RestoreJob fullJob = restoreJobDatabaseAccessor.find(job.jobId);
```
yifan-c: We should be able to skip this lookup. 
In the scope of the handler, it actually has the job object retrieved from database. See line#96, i.e. `        RoutingContextUtils.getAsFuture(context, SC_RESTORE_JOB)`. Note, the job object is before the update.

server/src/main/java/org/apache/cassandra/sidecar/handlers/restore/UpdateRestoreJobHandler.java
```diff
@@ -147,4 +161,38 @@ protected UpdateRestoreJobRequestPayload extractParamsOrThrow(RoutingContext con
...
+                return;
+            }
+
+            if (fullJob.status == RestoreJobStatus.IMPORT_READY)
+            {
+                restoreJobManagerGroup.updateRestoreJob(fullJob);
+            }
+            else if (fullJob.status == RestoreJobStatus.STAGE_READY)
+            {
+                restoreJobDiscoverer.processJobNow(fullJob);
```
yifan-c: I think for both `IMPORT_READY` and `STAGE_READY`, we can call `processJobNow`

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobManagerGroup.java
```diff
@@ -105,7 +105,7 @@ void removeJobInternal(RestoreJob restoreJob)
      *
      * @param restoreJob restore job to update
      */
-    void updateRestoreJob(RestoreJob restoreJob)
+    public void updateRestoreJob(RestoreJob restoreJob)
```
yifan-c: can potentially revert back to package-private, if `updateRestoreJob` does not need to be visible to the handler.

server/src/test/integration/org/apache/cassandra/sidecar/restore/jobdiscoverer/RestoreJobDiscovererPhaseSignalIntTest.java
```diff
@@ -0,0 +1,215 @@
...
+import static org.apache.cassandra.sidecar.restore.RestoreJobTestUtils.createJob;
+import static org.apache.cassandra.sidecar.restore.RestoreJobTestUtils.disableRestoreProcessor;
+import static org.apache.cassandra.testing.utils.AssertionUtils.loopAssert;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration tests verifying that phase signals (STAGE_READY, IMPORT_READY) sent via
+ * UpdateRestoreJobHandler immediately trigger processing without waiting for the discovery loop.
+ */
+class RestoreJobDiscovererPhaseSignalIntTest extends IntegrationTestBase
```
yifan-c: Please create the test in `integration-tests`, using `SharedClusterIntegrationTestBase`; as name suggests, the same in-jvm cassandra cluster is shared among test cases in the file.

`IntegrationTestBase` is deprecating, it consumes a lot of resources, as each test starts up and tears down in-jvm cluster.

yifan-c:
`CheckStyle` is failing. There is a gradle task to run the check at local quickly. 

```
./gradlew codeCheckTasks
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/restore/UpdateRestoreJobHandler.java
mansikhara: Done, thanks!

server/src/main/java/org/apache/cassandra/sidecar/handlers/restore/UpdateRestoreJobHandler.java
mansikhara: done!

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobManagerGroup.java
mansikhara: done!

server/src/main/java/org/apache/cassandra/sidecar/handlers/restore/UpdateRestoreJobHandler.java
mansikhara: done!

server/src/test/integration/org/apache/cassandra/sidecar/restore/jobdiscoverer/RestoreJobDiscovererPhaseSignalIntTest.java
mansikhara: done!

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/restore/RestoreJobDiscovererPhaseSignalIntTest.java
```diff
@@ -0,0 +1,187 @@
...
+/**
+ * Verifies that {@code STAGE_READY} / {@code IMPORT_READY} phase signals delivered via
+ * the {@code UpdateRestoreJobHandler} REST endpoint trigger immediate processing on the
+ * receiving Sidecar instance, without waiting for the next discovery loop cycle.
+ *
+ * <p>Both discovery loop delays are pinned to 1h via {@link #configurationOverrides()},
+ * so any successful observation of restore ranges within seconds can only be the work of
+ * the wake-up path added by CASSSIDECAR-454 — not the periodic discovery loop.
+ */
+class RestoreJobDiscovererPhaseSignalIntTest extends SharedClusterSidecarIntegrationTestBase
```
yifan-c, line 60: 👍

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/restore/RestoreJobDiscovererPhaseSignalIntTest.java
```diff
@@ -0,0 +1,187 @@
...
+                                                                  .withStatus(RestoreJobStatus.STAGE_READY)
+                                                                  .build());
+
+        loopAssert(10, 500, () -> assertThat(rangeAccessor.findAll(jobId, bucketId))
+                                  .describedAs("STAGE_READY should immediately create restore ranges via the wake-up path")
+                                  .isNotEmpty());
+    }
+
+    @Test
+    void testStageReadyCreatesCorrectTokenRanges()
```
yifan-c: This test seems to be largely the same with `testStageReadyImmediatelySubmitsSlices`, can we merge them and add the token value assertion in the `loopAssert` lambda?

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/restore/RestoreJobDiscovererPhaseSignalIntTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+        RestoreRangeDatabaseAccessor rangeAccessor =
+        serverWrapper.injector.getInstance(RestoreRangeDatabaseAccessor.class);
+        loopAssert(10, 500, () -> {
+            List<RestoreRange> ranges = rangeAccessor.findAll(jobId, bucketId);
+            assertThat(ranges).describedAs("STAGE_READY should produce trimmed ranges").isNotEmpty();
+            for (RestoreRange range : ranges)
+            {
+                assertThat(range.startToken()).isNotNull();
+                assertThat(range.endToken()).isNotNull();
```
yifan-c: Assert on the actual value, instead of not null. 
Start token should be 499 (as ranges have exclusive start end)
End token should be 1500.

server/src/main/java/org/apache/cassandra/sidecar/handlers/restore/UpdateRestoreJobHandler.java
```diff
@@ -87,36 +91,40 @@ protected void handleInternal(RoutingContext context,
...
+                                        metrics.tokenRefreshed.metric.update(1);
+                                    }
 
-            context.response().setStatusCode(HttpResponseStatus.OK.code()).end();
+                                    context.response().setStatusCode(HttpResponseStatus.OK.code()).end();
+                                    // Fire-and-forget on a worker thread — notifying the restore system should not
+                                    // block the event loop or delay the HTTP response.
+                                    executorPools.service()
+                                                 .runBlocking(() -> notifyPhaseSignalMaybe(existingJob, updatedJob.status));
+                                });
```
yifan-c: The refactoring should work. But, I'd like to suggest the change in `RestoreJobDatabaseAccessor` (see below). Basically, adding a new update override that takes `RestoreJob existingJob`. In the method, it returns an updated `existingJob` after successfully persisted to the database. Please give it a try and see if it looks cleaner. 

```java
    /**
     * Update fields in the restore job and persist
     *
     * @param payload fields to be updated
     * @param jobId job ID
     * @return the restore job object with only the updated fields
     * @throws DataObjectMappingException when secrets json cannot be serialized
     */
    public RestoreJob update(UpdateRestoreJobRequestPayload payload, UUID jobId)
    throws DataObjectMappingException
    {
        sidecarSchema.ensureInitialized();
        return update(payload, jobId, RestoreJob.builder());
    }

    /**
     * Update fields in the restore job and persist
     *
     * @param payload fields to be updated
     * @param existingJob existing restore job to be updated
     * @return the restore job object built from the existingJob and the updated fields
     * @throws DataObjectMappingException when secrets json cannot be serialized
     */
    public RestoreJob update(UpdateRestoreJobRequestPayload payload, RestoreJob existingJob)
    throws DataObjectMappingException
    {
        sidecarSchema.ensureInitialized();
        return update(payload, existingJob.jobId, existingJob.unbuild());
    }

    private RestoreJob update(UpdateRestoreJobRequestPayload payload, UUID jobId, RestoreJob.Builder updateBuilder)
    throws DataObjectMappingException
    {
        LocalDate createdAt = RestoreJob.toLocalDate(jobId);
        updateBuilder.createdAt(createdAt)
                     .jobId(jobId);

        RestoreJobSecrets secrets = payload.secrets();
        RestoreJobStatus status = payload.status();
        String jobAgent = payload.jobAgent();
        Date expireAt = payload.expireAtAsDate();
        Long sliceCount = payload.sliceCount();
        // all updates are going to the same partition. We use unlogged explicitly.
        // Cassandra internally combine those updates into the same mutation.
        BatchStatement batchStatement = new BatchStatement(BatchStatement.Type.UNLOGGED);
        ByteBuffer wrappedSecrets;
        if (secrets != null)
        {
            try
            {
                byte[] secretBytes = MAPPER.writeValueAsBytes(secrets);
                wrappedSecrets = ByteBuffer.wrap(secretBytes);
                batchStatement.add(tableSchema.updateBlobSecrets()
                                              .bind(createdAt, jobId, wrappedSecrets));
            }
            catch (JsonProcessingException e)
            {
                throw new DataObjectMappingException("Failed to serialize secrets", e);
            }
            updateBuilder.jobSecrets(secrets);
        }
        if (status != null)
        {
            batchStatement.add(tableSchema.updateStatus().bind(createdAt, jobId, status.name()));
            updateBuilder.jobStatus(status);
        }
        if (jobAgent != null)
        {
            batchStatement.add(tableSchema.updateJobAgent().bind(createdAt, jobId, jobAgent));
            updateBuilder.jobAgent(jobAgent);
        }
        if (expireAt != null)
        {
            batchStatement.add(tableSchema.updateExpireAt().bind(createdAt, jobId, expireAt));
            updateBuilder.expireAt(expireAt);
        }
        if (sliceCount != null)
        {
            batchStatement.add(tableSchema.updateSliceCount().bind(createdAt, jobId, sliceCount));
            updateBuilder.sliceCount(sliceCount);
        }

        execute(batchStatement);
        return updateBuilder.build();
    }
```

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/restore/RestoreJobDiscovererPhaseSignalIntTest.java
mansikhara: good catch! Thanks

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/restore/RestoreJobDiscovererPhaseSignalIntTest.java
mansikhara: done!

server/src/main/java/org/apache/cassandra/sidecar/handlers/restore/UpdateRestoreJobHandler.java
mansikhara: Done in the latest commit- handler is also cleaner now since it no longer has to rebuild the full job after the update.

server/src/main/java/org/apache/cassandra/sidecar/handlers/restore/UpdateRestoreJobHandler.java
```diff
@@ -87,36 +91,39 @@ protected void handleInternal(RoutingContext context,
...
-            if (job.secrets != null)
+            if (updatedJob.secrets != null)
             {
                 metrics.tokenRefreshed.metric.update(1);
             }
 
             context.response().setStatusCode(HttpResponseStatus.OK.code()).end();
+            // Fire-and-forget on a worker thread — notifying the restore system should not
+            // block the event loop or delay the HTTP response.
+            executorPools.service().runBlocking(() -> notifyPhaseSignalMaybe(updatedJob));
```
sarankk: Let's use `executorPools.internal()` instead? since this is internal handling and not client side facing 
```suggestion
            executorPools.internal().runBlocking(() -> notifyPhaseSignalMaybe(updatedJob));
```

server/src/main/java/org/apache/cassandra/sidecar/handlers/restore/UpdateRestoreJobHandler.java
```diff
@@ -147,4 +154,22 @@ protected UpdateRestoreJobRequestPayload extractParamsOrThrow(RoutingContext con
             throw wrapHttpException(HttpResponseStatus.BAD_REQUEST, "Invalid request payload", decodeException);
         }
     }
+
+    private void notifyPhaseSignalMaybe(RestoreJob updatedJob)
+    {
+        if (updatedJob.status != RestoreJobStatus.IMPORT_READY && updatedJob.status != RestoreJobStatus.STAGE_READY)
```
sarankk, line 171: Do we want to process`STAGED` status update similarly as well? @yifan-c I see you included this status as well

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -628,6 +628,22 @@ Map<Integer, Map<UUID, RestoreJobStatus>> jobsByDay()
...
+     * source of truth, and duplicate processing is deduplicated by existing idempotency checks.
+     *
+     * @param restoreJob the restore job to process immediately
+     */
+    public void processJobNow(RestoreJob restoreJob)
+    {
+        initLocalDatacenterMaybe();
+        RestoreJobManagerGroup restoreJobManagers = restoreJobManagerGroupSingleton.get();
+        restoreJobManagers.updateRestoreJob(restoreJob);
+        processSidecarManagedJobMaybe(restoreJob);
```
sarankk: fields within `JobIdsByDay` are updated from 2 different threads now. Also we seem to have `isExecuting` check to avoid concurrent executions even within `RestoreJobDiscoverer`. We should make `JobIdsByDay` thread safe. @yifan-c can confirm if this is a concern

sarankk (commented):
Thanks Mansi, looks good. Left some comments.

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
mansikhara: Yes, this is a great catch! I will work on the fix. Thanks.

server/src/main/java/org/apache/cassandra/sidecar/handlers/restore/UpdateRestoreJobHandler.java
yifan-c: I think it is to be addressed in https://issues.apache.org/jira/browse/CASSSIDECAR-463

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
yifan-c: Good catch. Instead of making `JobIdsByDay` thread-safe, a simple/clean way to address it is to wrap the lines in `processJobNow` with the same `isExecuting / finally` pattern used in `tryExecuteDiscovery`. The CAS operation on `isExecuting` provides a happens-before edge.

server/src/main/java/org/apache/cassandra/sidecar/handlers/restore/UpdateRestoreJobHandler.java
mansikhara: done!

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
mansikhara: done!

sarankk (approved):
+1 Thanks, Looks good to me

CHANGES.txt
```diff
@@ -8,6 +8,7 @@
  * Fix CDC message loss during Sidecar restarts (CASSSIDECAR-457)
  * Fix CDC resource leaks: thread leaks, Kafka resource cleanup, singleton SidecarCdcClient (CASSSIDECAR-456)
  * Fix breaking changes for Analytics 0.4.0 (CASSSIDECAR-455)
+ * Sidecar should wake up immediately on the instance receiving a phase signal instead of waiting for the discovery loop (CASSSIDECAR-454)
```
sarankk: Nit: Just noticed this item is not on top of the list, could you move latest change to top of the list?

mansikhara: done, thanks!

PR #347, opened 2026-05-12
Integrating Cassandra analytics 0.4.0 and CDC resource leaks and serialization fixes

jyothsnakonisa:


server/src/main/java/org/apache/cassandra/bridge/CassandraBridgeFactory.java
```diff
@@ -119,21 +116,8 @@ String jarResourceName(String... parts)
...
-                               }
-                               catch (MalformedURLException e)
-                               {
-                                   throw new RuntimeException(e);
-                               }
-                           }).toArray(URL[]::new);
-
         return AccessController.doPrivileged((PrivilegedAction<ClassLoader>) () ->
-                                                                             new PostDelegationClassLoader(urls, Thread.currentThread().getContextClassLoader()));
+                                                                             BaseCassandraBridgeFactory.buildClassLoader(resourceNames));
```
jmckenzie-dev: Can we left align w/1 tab here instead of having this aligned so far to the right?

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -132,8 +146,15 @@ private void configureSidecarServerEventListeners()
...
-                loadPublisher();
-                publishSchemas();
+                try
+                {
+                    loadPublisher();
+                    publishSchemas();
+                }
+                catch (Exception e)
+                {
+                    LOGGER.error("Failed to publish schemas to Kafka during initialization, CDC will still start", e);
```
jmckenzie-dev, line 160: Why do we continue to start even if our schema publishers fail? What was the thinking here? We should document the "why" here to explain this reasoning.

server/build.gradle
```diff
@@ -151,6 +151,7 @@ dependencies {
     implementation(group: "org.apache.cassandra", name: "cassandra-avro-converter_spark3_2.12", version: "${[project.analyticsVersion]}")
     implementation(group: "org.apache.cassandra", name: "cassandra-analytics-cdc_spark3_2.12", version: "${[project.analyticsVersion]}")
     implementation(group: "org.apache.cassandra", name: "cassandra-analytics-cdc-sidecar_spark3_2.12", version: "${[project.analyticsVersion]}")
+    compileOnly(group: "org.apache.cassandra", name: "cassandra-analytics-sidecar-client", version: "${[project.analyticsVersion]}")
```
jmckenzie-dev, line 157: We should comment on why this is here and `compileOnly` so it's clear for future maintainers. Not immediately obvious.

build.gradle
```diff
@@ -140,6 +140,10 @@ allprojects {
 
         // for dtest jar
         mavenLocal()
+
+        maven {
+            url "https://repository.apache.org/content/repositories/orgapachecassandra-1468/"
```
jmckenzie-dev: This needs to be removed before merge right?

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -91,13 +105,13 @@ public class CachingSchemaStore implements SchemaStore
...
         this.sidecarSchema = sidecarSchema;
         this.cqlToAvroSchemaConverter = cqlToAvroSchemaConverter;
-        this.avroSchemasCache.putAll(createSchemaCache(cassandraClusterSchemaMonitor.getCdcTables()));
-        AvroSchemas.registerLogicalTypes();
-        cassandraClusterSchemaMonitor.addSchemaChangeListener(this::onSchemaChanged);
         this.vertx = vertx;
         this.cdcConfig = cdcConfig;
         this.sidecarCdcStats = sidecarCdcStats;
         this.schemaStorePublisherFactory = schemaStorePublisherFactory;
+        this.avroSchemasCache.putAll(createSchemaCache(cassandraClusterSchemaMonitor.getCdcTables()));
```
jmckenzie-dev, line 116: Why the ordering change here? If there's a temporal or ordering based requirement, we should add a comment there about that so future maintainers don't regress.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -176,17 +209,26 @@ void onSchemaChanged()
...
-        loadPublisher();
-        publishSchemas();
+        try
+        {
+            loadPublisher();
+            publishSchemas();
+        }
+        catch (Exception e)
+        {
+            LOGGER.error("Failed to publish schemas to Kafka, CDC will still start", e);
```
jmckenzie-dev, line 228: Similar to publisher init above, why do we continue even if schema publishing fails? Let's document the reasoning and intent here. And if we're not 100% sure this is the right default, we should consider making this a configurable parameter for operators.

server/src/test/java/org/apache/cassandra/sidecar/cdc/CdcPublisherTests.java
```diff
@@ -111,204 +112,22 @@ void setUp()
...
             virtualTables,
             sidecarCdcStats,
-            avroSerializer,
-            rangeManager
-        );
-    }
-
-
-    @Test
-    void testSecretsProviderReturnsNullWhenSslDisabled()
```
jmckenzie-dev, line 133: I think we lost this test in the migration over to the new test file - was that deliberate?

server/src/main/java/org/apache/cassandra/sidecar/cdc/SidecarReplicationFactorSupplier.java
```diff
@@ -0,0 +1,73 @@
...
+    public ReplicationFactor getMaximalReplicationFactor()
+    {
+        String dc = cdcOptions.dc();
+        Set<CqlTable> tables = FutureUtils.get(schemaSupplier.getCdcEnabledTables());
+        return tables.stream()
+                     .map(CqlTable::replicationFactor)
+                     .filter(rf -> rf.getOptions().containsKey(dc))
+                     .max(Comparator.comparingInt(rf -> rf.getOptions().get(dc)))
+                     .orElseGet(() -> {
+                         LOGGER.warn("No CDC-enabled tables found for DC '{}'; falling back to RF=1 SimpleStrategy", dc);
```
jmckenzie-dev: Why do we _have_ this here? What's the cause for this fallback, what are the implications, etc. It's not clear to me why this wouldn't be a Bad Thing vs. failing out if we hit this point and can't get a replication factor for a table / keyspace.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcPublisher.java
jmckenzie-dev, line 212: The fact that we don't call `initialize` during the regular `run` and startup path is quite surprising and confusing. Why is this constructed this way? The `initialize` method looks like it actually only increments metrics and sets `isInitialized` to true. The flow there seems generally somewhat odd to me.

My naive assumption is we'd have some flow that is "initialize CDC publishing; when it's done, flag `isInitialized` as true." I'd expect both `run` and `restart` to lean on this shared code.

Right now, best I can tell we call `startConsumers` in `run` and then don't call it again. This is a problem because we call `stopConsumers` in our `CdcPublisher#stop` call, meaning post restart our consumers will be stopped and never restarted, at least from the context inside this class. We should either be responsible for all start/stop on consumers or none in this file.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
```diff
@@ -176,17 +209,26 @@ void onSchemaChanged()
...
+        }
         // Remove any old schema entries for deleted tables, this operation can be done in the end as this is
         // only for removing stale entries and no one is going to use these entries once the table is removed.
         // This doesn't have to be an atomic operation.
-        avroSchemasCache.keySet().retainAll(refreshedCdcTables.stream().map(cqlTable -> TableIdentifier.of(cqlTable.keyspace(), cqlTable.table())).collect(Collectors.toList()));
+        avroSchemasCache.keySet().retainAll(refreshedCdcTables.stream()
+                                                               .map(cqlTable -> TableIdentifier.of(cqlTable.keyspace(), cqlTable.table()))
+                                                               .collect(Collectors.toList()));
         vertx.eventBus().publish(ON_CDC_CACHE_WARMED_UP.address(), "Cdc cache warmed up");
```
jmckenzie-dev, line 236: We only publish the `ON_CDC_CACHE_WARMED_UP` method on `onSchemaChanged`. Does this mean initialization has a logical and temporal coupling on some reactive, config-based, event driven thing to take place to warm up the cache?

I'd expect we would initialize and be in a ready state out of the gate w/this published, not having to rely on a reactive pipeline of events to cascade through to trigger this.

integration-framework/src/main/java/org/apache/cassandra/sidecar/testing/TestCdcPublisher.java
```diff
@@ -49,30 +47,32 @@ public class TestCdcPublisher extends CdcPublisher
...
     {
-        super(vertx, sidecarConfiguration, executorPools, clusterConfigProvider,
-              schemaSupplier, sidecarInstancesProvider, clientConfig,
-              instanceMetadataFetcher, conf, databaseAccessor, cdcStats,
-              virtualTables, sidecarCdcStats, avroSerializer, rangeManagerProvider);
+        super(vertx, executorPools, clusterConfigProvider,
+              schemaSupplier, instanceMetadataFetcher, conf, databaseAccessor, cdcStats,
+              virtualTables, sidecarCdcStats, rangeManagerProvider,
+              cassandraBridgeFactory, sidecarCdcClientProvider,
+              null,
```
jmckenzie-dev: Passing these 2 as null today may work but it's a landmine for future modifications. Any changes in the future that try and access the 2 objects that this is passing in as null will cause things to NPE out in surprising ways in this test that'll cause trouble in the future. Can we mock in something here that'll satisfy the API of these 2 objects w/out introducing too much complexity to prevent that future regression?

server/src/main/java/org/apache/cassandra/sidecar/cdc/CdcManager.java
```diff
@@ -98,39 +96,40 @@ public CdcManager(EventConsumer eventConsumer,
...
         Map<String, Set<TokenRange>> ownedRanges = rangeManager.ownedTokenRanges();
         if (ownedRanges == null || ownedRanges.isEmpty())
         {
             throw new IllegalStateException("No owned token ranges right now, cql session may still be initializing.");
         }
 
-        // NEW: Deduplicate by (instanceId, tokenRange) to prevent duplicate consumers
-        Map<String, SidecarCdc> uniqueConsumers = new HashMap<>(ownedRanges.values().stream().mapToInt(Set::size).sum());
+        // Deduplicate by (instanceId, tokenRange) to prevent duplicate consumers
+        Map<String, CdcConsumerEntry> uniqueEntries = new HashMap<>(ownedRanges.values().stream().mapToInt(Set::size).sum());
```
jmckenzie-dev: I'd prefer if we named this something like `uniqueCDCStates` or something else that, via the name, informs _what_ it is. "uniqueEntries" is tautological and doesn't really give us much to go off.

jmckenzie-dev (changes requested):

build.gradle
jyothsnakonisa: Yes!

server/src/test/java/org/apache/cassandra/sidecar/cdc/CdcPublisherTests.java
jyothsnakonisa: None of the tests were removed. They got moved to other files, this one got moved to `SidecarClientSecretsProviderTests`, during reorganization of commits, it got moved to another commit.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
jyothsnakonisa: Added a comment explaining the reason

server/src/main/java/org/apache/cassandra/sidecar/cdc/SidecarReplicationFactorSupplier.java
jyothsnakonisa: Default implementation for this method in the base class fallbacks to RF1 and hence I am using that here, If you think we should fail in this scenario, I am open to that suggestion.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
jyothsnakonisa: This change is not part of this PR. This is my guess, CdcPublisher should only start if the CachingSchemaStore Is initialized to avoid reading commit logs before schema is available in-memory, otherwise we will run into errors. `ON_CDC_CACHE_WARMED_UP ` event is published when initialization of in -memory cache is complete to indicate that schemas are ready In memory and we can start processing commit logs for producing mutations.

server/build.gradle
jyothsnakonisa: Added a comment explaining the reason to add this

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
jyothsnakonisa: ON_CDC_CACHE_WARMED_UP event never gets called until there is a schema change if we don't proceed and CDC never starts. 

I agree with u that if we proceed when schema publishing fails, we will miss the update in schemastore and there might be issues during deserialization. WDYT of adding a retry when schema publishing fails? we can call ON_CDC_CACHE_WARMED_UP when it succeeds.

jmckenzie-dev: So 2 things:
1. Why do we not warm up the cache or process anything before there's a schema change? The flow of:
a. start process
b. be non-functional until
c. a schema change hits
d. THEN you warm up your cache and notify

Is surprising to me. I would expect we had 2 flows.
### Startup flow:
- Start process
- Preload / cache / get things ready w/what we know
- Notify about cache warmup

### Schema change flow
- Reload / recache things w/the new data
- Notify about cache warmup

Though writing it out there, "cache warmup" seems like the wrong name vs. "CACHE_READY" or something.

And 2: An exponential backoff on retry that terminates at N attempts sounds like a great idea for what to do when we can't publish schemas. Ultimately, if we can't publish schemas I don't _think_ we should startup CDC, since any subsequent schema changes will break things and by definition be some distance away from the source of the breakage.

server/src/test/java/org/apache/cassandra/sidecar/cdc/CdcPublisherTests.java
jmckenzie-dev: You sure? Looking at the diff I see:
```
    void testSecretsProviderWithSslEnabledNoKeystoreNoTruststore()
    void testSecretsProviderWithKeystoreOnly()
    void testSecretsProviderWithTruststoreOnly()
    void testSecretsProviderWithBothKeystoreAndTruststore()
    void testSecretsProviderUsesCorrectSslConfigKeys()
```
Don't see ReturnsNullWhenSslDisabled.

server/src/main/java/org/apache/cassandra/sidecar/cdc/SidecarReplicationFactorSupplier.java
jmckenzie-dev: To be honest I have no clue. 😀 That's why I asked why it was here because I have no frame of reference to know if this is the right thing to do or not.

If we try and get a maximal replication factor but straight up can't hit the system to pull it out, defaulting to the most permissive (RF=1) instead of the general default (RF=3) feels like a land-mine waiting to blow up in someone's face down the road doesn't it?

`getMinimalReplicationFactor` defaulting to 1 I could see. But `getMaximalReplicationFactor` having a fallback to 1 feels... :suspect: 

@yifan-c - any chance you have any context here?

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
jyothsnakonisa: When sidecar starts, the schema is initialized to empty string, when we first read actual schema that is detected as schema change.

Also, note that when we say cache warm up it is avro schemas in the CachingSchemaStore.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
jmckenzie-dev: Ah. This makes sense. Could you add a comment to the code explaining as much? 

Having the cache's warm-up signal being dependent on a schema change felt counterintuitive, since on regular startup you don't necessarily have a schema _**change**_ you have an initial schema _**load**_. This is probably just me being pedantic.

But if we're going through that same pathway to trigger a schema change any time a schema loads and just using that pathway for a single "if this then that" transitive trigger, that makes sense. "Hey, schema changed (whether it be load or change after load) - here's our downstream dependencies." kind of thing.

jyothsnakonisa: Added comment.

server/src/test/java/org/apache/cassandra/sidecar/cdc/CdcPublisherTests.java
jyothsnakonisa:
```
if (sslConfiguration == null || !sslConfiguration.enabled())
        {
            return null;
        }
```

This block got moved out of `SidecarClientSecretsProvider` and the secrets provider will not now return null. I think that is the reason why I removed the test.

server/src/main/java/org/apache/cassandra/sidecar/cdc/CachingSchemaStore.java
jyothsnakonisa: @jmckenzie-dev  I will address this in separate PR `CASSSIDECAR-466 CDC iterators are starting when publishing schema to schema store fails`

jmckenzie-dev (approved):
lgtm

jyothsnakonisa:
CI: https://app.circleci.com/pipelines/github/jyothsnakonisa?created_after=2026-05-05T00%3A00%3A00.000Z&created_preset=Last+2+weeks&selectedPipeline=gh%2Fjyothsnakonisa%2Fcassandra-sidecar%2F206&workflowId=5fccde62-7e3c-451e-b7a5-015127f9039e

jyothsnakonisa:
Pushed all 5 commits on to trunk to preserve commit history.

PR #348, opened 2026-05-14
CASSSIDECAR-461: Validate snapshot name during list snapshot

frankgh:
Patch by Francisco Guerrero; reviewed by TBD for CASSSIDECAR-461


server/src/main/java/org/apache/cassandra/sidecar/utils/CassandraInputValidator.java
```diff
@@ -57,7 +60,19 @@ public interface CassandraInputValidator
      * @throws NullPointerException    when the {@code snapshotName} is {@code null}
      * @throws CassandraInputException when the {@code snapshotName} contains invalid characters in the name
      */
-    String validateSnapshotName(@NotNull String snapshotName);
+    default String validateSnapshotName(@NotNull String snapshotName)
+    {
```
frankgh, line 65: moved implementation to default implementation in the interface. `RegexBasedCassandraInputValidator` didn't seem the appropriate place for a non-regex validation

server/src/main/java/org/apache/cassandra/sidecar/handlers/snapshots/ListSnapshotHandler.java
```diff
@@ -173,7 +173,7 @@ protected SnapshotRequestParam extractParamsOrThrow(RoutingContext context)
 
         return SnapshotRequestParam.builder()
                                    .qualifiedTableName(qualifiedTableName(context))
-                                   .snapshotName(context.pathParam("snapshot"))
+                                   .snapshotName(validator.validateSnapshotName(context.pathParam("snapshot")))
```
frankgh, line 176: I realized this might not be a real issue because we can't have `/` or `\0` (null character) in the path parameter. I tried coming up with a test, but the request fails before reaching the snapshot handler with a 404 (because the route is incorrect). I think we should still do the validation here in case in the future the snapshot name validation changes.

pauloricardomg (approved):

server/src/main/java/org/apache/cassandra/sidecar/handlers/snapshots/ListSnapshotHandler.java
yifan-c: try `%2e%2e/%2e%2e` (escaped encoding of `../..`) for the snapshot name?

frankgh: I did, it is still a 404

yifan-c: I found a case that relies on the validator to reject, 400 error instead of 404. The pathParam value is `..%2F..`, i.e. `../..`
Without the validator, the snapshot name is `../..` and it is passed to handler to process.

PR #349, opened 2026-05-15
CASSSIDECAR-460: Refactor OperationalJob to have data separate from execution logic

andresbeckruiz:
### Changes

  - `OperationalJobInfo` interface: A read-only view of job state providing accessors for job data without exposing execution logic 
  - `OperationalJob` implements `OperationalJobInfo`
  - Read paths updated: `OperationalJobTracker`, `OperationalJobManager`, handlers, and utilities now operate on `OperationalJobInfo` instead of `OperationalJob`
  - Job submission separated from execution: `OperationalJobManager.trySubmitJob()` decouples registration from execution, enabling the future `DurableOperationalJobTracker` to persist job state before execution begins
  - Unit tests for `OperationalJobInfo`
  
  ### Future Work

  - CASSSIDECAR-374: Implement durable operational job tracker

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
@@ -271,7 +271,7 @@ public abstract class OperationalJob implements Task<Void>, OperationalJobInfo
         if (fut.isComplete() && fut.failed() && fut.cause() != null)
         {
             String message = fut.cause().getMessage();
-            return message != null ? message : fut.cause().getClass().getName();
+            return message != null ? message : String.format("Encountered %s during job execution", fut.cause().getClass().getName());
         }
         return null;
     }
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
@@ -42,7 +42,7 @@ import org.jetbrains.annotations.Nullable;
 /**
  * An abstract class representing operational jobs that run on Cassandra
  */
-public abstract class OperationalJob implements Task<Void>
+public abstract class OperationalJob implements Task<Void>, OperationalJobInfo
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(OperationalJob.class);
 
@@ -216,6 +216,13 @@ public abstract class OperationalJob implements Task<Void>
      */
     public abstract boolean hasConflict(@NotNull List<OperationalJob> sameOperationJobs);
 
+    @Override
+    public String name()
+    {
+        String simpleName = this.getClass().getSimpleName();
+        return simpleName.isEmpty() ? this.getClass().getName() : simpleName;
+    }
+
     /**
      * Determines the status of the job. OperationalJob subclasses could choose to override the method.
      * <p>
@@ -254,6 +261,18 @@ public abstract class OperationalJob implements Task<Void>
     public Future<Void> asyncResult()
     {
         return executionPromise.future();
+    }
+
+    @Override
+    @Nullable
+    public String failureReason()
+    {
+        Future<Void> fut = asyncResult();
+        if (fut.isComplete() && fut.failed() && fut.cause() != null)
+        {
+            return fut.cause().getMessage();
+        }
+        return null;
     }
 
     /**
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
@@ -270,7 +270,8 @@ public abstract class OperationalJob implements Task<Void>, OperationalJobInfo
         Future<Void> fut = asyncResult();
         if (fut.isComplete() && fut.failed() && fut.cause() != null)
         {
-            return fut.cause().getMessage();
+            String message = fut.cause().getMessage();
+            return message != null ? message : fut.cause().getClass().getName();
         }
         return null;
     }
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
@@ -88,30 +88,37 @@ public abstract class OperationalJob implements Task<Void>, OperationalJobInfo
         this.nodesFailed = Collections.emptyList();
     }
 
+    /**
+     * {@inheritDoc}
+     */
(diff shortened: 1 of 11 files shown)


server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
```diff
@@ -100,11 +100,12 @@ public void trySubmitJob(OperationalJob job,
         {
             checkConflict(job);
 
-            // New job is submitted for all cases when we do not have a corresponding downstream job
-            jobTracker.computeIfAbsent(job.jobId(), jobId -> {
+            OperationalJob tracked = jobTracker.computeIfAbsent(job.jobId(), id -> job);
```
isaacreath: Why do we need to change the submission behavior? It appears that this accomplishes the same thing as the existing code?

andresbeckruiz: The goal was to separate storage of the `OperationalJob` from execution, but this can be done as part of https://issues.apache.org/jira/browse/CASSSIDECAR-374 once the `DurableOperationalJobTracker` is introduced. Reverted this change in [8188cfe](https://github.com/apache/cassandra-sidecar/pull/349/commits/8188cfe03829e996a8d11006caed180a4039c489)

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
```diff
@@ -73,9 +73,9 @@ public List<OperationalJob> allInflightJobs()
      * Fetch the job using its UUID
      *
      * @param jobId identifier of the job
-     * @return instance of the job or null
+     * @return instance of the job info or null
      */
-    public OperationalJob getJobIfExists(UUID jobId)
+    public OperationalJobInfo getJobIfExists(UUID jobId)
```
sarankk, line 78: `allInflightJobs()` still exposes all the details and only `ListOperationalJobsHandler` with read operational job permission calls it, do we want to update that as well to return only info related to job?

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
```diff
@@ -256,6 +263,18 @@ public Future<Void> asyncResult()
...
     }
 
+    @Override
+    @Nullable
+    public String failureReason()
+    {
+        Future<Void> fut = asyncResult();
+        if (fut.isComplete() && fut.failed() && fut.cause() != null)
+        {
+            return fut.cause().getMessage();
```
arjunashok: `getMessage()` can return null for exceptions created without a message (e.g. new RuntimeException()). The response would show a FAILED job with no reason.

server/src/test/java/org/apache/cassandra/sidecar/handlers/OperationalJobHandlerTest.java
```diff
@@ -184,6 +184,7 @@ public OperationalJobManager jobManager()
             when(failedMock.jobId()).thenReturn(failedUuid);
             when(failedMock.status()).thenReturn(OperationalJobStatus.FAILED);
             when(failedMock.asyncResult()).thenReturn(Future.failedFuture("Test failed"));
```
arjunashok: Nit. can be removed. `OperationalJobHandler` no longer calls asyncResult() on the job

sarankk (commented):
Thanks Andres, Lgtm, one minor comment

sarankk (approved):

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
andresbeckruiz: Good idea, addressed in [94c6f7c](https://github.com/apache/cassandra-sidecar/pull/349/commits/94c6f7c807b6c0c538d1bf186b1f8b57180c8457)

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
andresbeckruiz: Fixed to use class name if exception is created without a message in [94c6f7c](https://github.com/apache/cassandra-sidecar/pull/349/commits/94c6f7c807b6c0c538d1bf186b1f8b57180c8457)

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJob.java
```diff
@@ -256,6 +263,19 @@ public Future<Void> asyncResult()
...
 
+    @Override
+    @Nullable
+    public String failureReason()
+    {
+        Future<Void> fut = asyncResult();
+        if (fut.isComplete() && fut.failed() && fut.cause() != null)
+        {
+            String message = fut.cause().getMessage();
+            return message != null ? message : fut.cause().getClass().getName();
```
sarankk: Nit 
```suggestion
            return message != null ? message : String.format("Encountered %s during job execution", fut.cause().getClass().getName());
```

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobInfo.java
```diff
@@ -0,0 +1,68 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A read-only view of operational job data. Provides state accessors needed by handlers
+ * and utilities to inspect job status and report results, without exposing execution logic.
+ */
+public interface OperationalJobInfo
+{
+    UUID jobId();
```
frankgh, line 38: can we add javadocs to public interfaces?

frankgh (commented):
Looks good, but to maintain consistency with the repo, we should add javadocs to the newly introduced interface

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobInfo.java
andresbeckruiz: Added in [8f57576](https://github.com/apache/cassandra-sidecar/pull/349/commits/8f575766f3c0a3e33a2a8d443660422b961b084f)

frankgh (approved):
+1 thanks for the patch

andresbeckruiz:
CI: https://app.circleci.com/pipelines/github/andresbeckruiz/cassandra-sidecar/46/workflows/b2485d2a-1613-41ce-9b2e-5037c28e0352

PR #350, opened 2026-05-15
CASSSIDECAR-462: Split RestoreJobDiscoverer into a fast status-check loop and a slow slice-discovery loop

mansikhara:


server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobStatusChecker.java
```diff
@@ -0,0 +1,132 @@
...
+ * Fast status-check loop complement to {@link RestoreJobDiscoverer}. Runs on a short interval
+ * (see {@link RestoreJobConfiguration#jobDiscoveryStatusCheckInterval()}) and performs cheap
+ * point-reads of {@code job.status} for the in-flight jobs already known to the discoverer.
+ * When a status transition is detected it delegates to
+ * {@link RestoreJobDiscoverer#handleStatusTransition(RestoreJob)} so that sidecars react to
+ * phase signals without waiting for the slow full-scan loop. The slow loop remains the
+ * correctness and recovery guarantee (new-job discovery, expired-job aborts, etc.).
+ */
+@Singleton
+public class RestoreJobStatusChecker implements PeriodicTask
```
yifan-c: I can see that `RestoreJobStatusChecker` is introduced to have a clear separation of the status check logic. 
However, I am concerned about the `RestoreJobDiscoverer` dependency and the state sharing between public classes (adding public interfaces). 

I'd propose moving the checker as a non-static inner class in `RestoreJobDiscoverer`, so that the state can be shared more comfortably. It is also to reflect the relationship between checker and discoverer. Checker is not an independent component, but a scheduling belongs to the discoverer. The checker task can be deployed when deploying discoverer.

mansikhara: great suggestion! Done

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -87,6 +87,8 @@ public class RestoreJobDiscoverer implements PeriodicTask, RingTopologyChangeLis
     private final JobIdsByDay jobIdsByDay;
     private final RingTopologyRefresher ringTopologyRefresher;
     private final AtomicBoolean isExecuting = new AtomicBoolean(false);
+    private final Object stateLock = new Object();
```
skoppu22: Some references to jobIdsByDay in this class are protected by stateLock, but not all. There are multiple references accessing jobIdsByDay without lock. Need to move them inside the lock.

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
skoppu22: Now both slow and fast loop increment this. And fast loop never decrements it. slow loop thinks there are active jobs. hasInflightJobs() returns non zero and slow loop's delay thinks there are active jobs when really not.

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -219,37 +224,102 @@ private void executeInternal()
...
+            // resize to the earliestInDays with the minimum days as defined in jobDiscoveryMinimumRecencyDays
+            jobDiscoveryRecencyDays = Math.max(context.earliestInDays, restoreJobConfig.jobDiscoveryMinimumRecencyDays());
+            LOGGER.info("Exit job discovery. " +
+                        "inflightJobsCount={} " +
+                        "delay={} " +
+                        "jobDiscoveryRecencyDays={} " +
+                        "expiredJobs={} " +
+                        "abortedJobs={}",
+                        inflightJobsCount, delay(), jobDiscoveryRecencyDays, context.expiredJobs, context.abortedJobs);
+            metrics.activeJobs.metric.setValue(inflightJobsCount);
```
skoppu22: Fast loop doesn't report this metric. Our dashboard will not reflect effectiveness of fast job.

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -219,37 +224,102 @@ private void executeInternal()
     {
         Preconditions.checkState(periodicTaskExecutor != null, "Loop executor is not registered");
 
-        LOGGER.debug("Discovering restore jobs. " +
-                     "inflightJobsCount={} jobDiscoveryRecencyDays={}",
-                     inflightJobsCount, delay(), jobDiscoveryRecencyDays);
+        synchronized (stateLock)
```
skoppu22: This holds the lock for too long for fast loop to progress. I would suggest move inflightJobs() to inside JobIdsByDay and then synchronize JobIdsByDay methods. Then we don't need this stateLock. And also avoids incorrect inflightJobsCount problem.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
```diff
@@ -42,6 +42,8 @@ public class RestoreJobConfigurationImpl implements RestoreJobConfiguration
     MillisecondBoundConfiguration.parse("5m");
     private static final MillisecondBoundConfiguration DEFAULT_JOB_DISCOVERY_IDLE_LOOP_DELAY =
     MillisecondBoundConfiguration.parse("10m");
+    private static final MillisecondBoundConfiguration DEFAULT_JOB_DISCOVERY_STATUS_CHECK_INTERVAL =
```
skoppu22, line 48: Why default 1s? What is the reason behind choosing this for default?

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
skoppu22: For example,

We can mark every method of JobIdsByDay as synchronized, add below method to this class. Then we can remove class variables stateLock and inflightJobsCount. We can call jobIdsByDay.inflightJobIds().size() for inflight jobs count. 

```
        synchronized Set<UUID> inflightJobIds()
        {
            Set<UUID> result = new HashSet<>();
            for (Map<UUID, RestoreJobStatus> jobs : jobsByDay.values())
            {
                for (Map.Entry<UUID, RestoreJobStatus> entry : jobs.entrySet())
                {
                    if (!entry.getValue().isFinal())
                    {
                        result.add(entry.getKey());
                    }
                }
            }
            return result;
        }
```

server/src/test/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscovererTest.java
```diff
@@ -438,6 +438,193 @@ void testRestoreToLocalDatacenterOnlyJobIsOnHoldWhenLocalDatacenterIsUndetermine
...
+        when(mockJobAccessor.find(failingJobId)).thenThrow(new RuntimeException("db down")); // transient error
+
+        Promise<Void> promise = Promise.promise();
+        loop.statusCheckTask.execute(promise);
+
+        assertThat(promise.future().succeeded())
+        .describedAs("a missing or failing job should not abort the whole pass")
+        .isTrue();
+    }
+
```
skoppu22, line 628: Please add a test to verify race between slow and fast loop.

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
mansikhara: Thanks for the great feedback, Shailaja.
Resolved this by the larger refactor - JobIdsByDay is now self-synchronized (every method synchronized), so every access is uniformly protected and the outer stateLock is gone entirely.

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
mansikhara: Added `metrics.activeJobs.metric.setValue(jobIdsByDay.inflightJobIds().size())` at the end of handleStatusTransition so the gauge reflects fast-loop transitions immediately instead of waiting for the next slow-loop pass. Thanks!

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
mansikhara: This is a great catch!
Removed the `manual inflightJobsCount counter. hasInflightJobs()`, the discovery log, and metrics.activeJobs all compute the count on demand from `jobIdsByDay.inflightJobIds().size()`, so there's no derived counter that the two loops could let drift out of sync with the map. (Thread-safety is still provided every JobIdsByDay method is synchronized.)

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
mansikhara: Addressed the comment! Thanks

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
mansikhara: 1s caps the wake-up latency for peer Sidecars on a phase transition. The Sidecar that gets the HTTP phase signal reacts immediately via processJobNow; peers learn via DB point-read on this interval (vs. 5m/10m for the slow loop). Cost is N reads/sec where N is the typically small in-flight job count. Sub-second hits diminishing returns against Cassandra write to read visibility; multi-second creeps back toward slow-loop latency. Also added an inline comment at the constant.

server/src/test/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscovererTest.java
mansikhara: Done, thanks!

server/src/test/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscovererTest.java
```diff
@@ -625,6 +626,111 @@ void testStatusCheckTaskTolerantOfMissingOrFailingJobs()
...
 
+    // Hammers the slow discovery loop and the fast status-check task concurrently against
+    // shared JobIdsByDay state, with statuses flipping under their feet, to verify that
+    // per-method synchronization on JobIdsByDay is sufficient. Both loops mutate the map
+    // through processOneJob via different gates (isExecuting vs checkerExecuting), so they
+    // can interleave on every operation. The test asserts no exception leaks out of either
+    // loop and that the in-flight view converges to the latest committed state once both
+    // settle.
+    @Test
+    void testSlowAndFastLoopRaceDoesNotCorruptState() throws Exception
```
skoppu22, line 637: Good one

skoppu22 (approved):

integration-tests/src/integrationTest/org/apache/cassandra/sidecar/restore/RestoreJobStatusCheckerIntTest.java
```diff
@@ -0,0 +1,133 @@
...
+ * {@code StatusCheckTask}) reacts to a restore-job status transition within the configured fast-loop
+ * interval, without waiting for the slow discovery loop.
+ *
+ * <p>The test seeds an in-flight job, runs one discovery pass to populate the in-flight set,
+ * then flips the job's status directly in the {@code sidecar_internal} keyspace and asserts
+ * that the discoverer sees the terminal status (via {@link RestoreJobDiscoverer#inflightJobIds()})
+ * well within the slow-loop interval. The slow loop is configured to a much larger interval so
+ * a positive observation can only be the work of the fast status-check loop.
+ */
+class RestoreJobStatusCheckerIntTest extends SharedClusterIntegrationTestBase
```
yifan-c: nit: rename to `RestoreJobDiscovererStatusCheckIntTest` so that Intellij can correct associate the test with `RestoreJobDiscoverer` and vice versa.

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -87,8 +87,10 @@ public class RestoreJobDiscoverer implements PeriodicTask, RingTopologyChangeLis
     private final JobIdsByDay jobIdsByDay;
     private final RingTopologyRefresher ringTopologyRefresher;
     private final AtomicBoolean isExecuting = new AtomicBoolean(false);
-    private String localDatacenter = null;
-    private int inflightJobsCount = 0;
+    final StatusCheckTask statusCheckTask = new StatusCheckTask();
```
yifan-c: I think the reason of package-private statusCheckTask is solely for testing purpose. Can we instead have a getter with `@VisibleForTesting` and make this field `private final`?

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -242,14 +245,62 @@ private void executeInternal()
...
+    }
+
+    /**
+     * Applies the same transition handling as a full discovery pass, but only for a single freshly-read job.
+     * Called by the fast status-check loop when it detects a status change on a known in-flight job.
+     * A no-op when the job's status matches the last-known status cached during discovery.
+     */
+    void handleStatusTransition(RestoreJob currentJob)
+    {
+        initLocalDatacenterMaybe();
```
yifan-c: This method is called for every RestoreJob unnecessarily.

yifan-c (commented):
**High-level design suggestion**

The slow discovery loop is the **correctness guarantee** — it must always run on schedule. The fast status-check loop is a **best-effort latency optimization** — it is safe to skip any individual tick. All shared mutable state (`JobIdsByDay`) is intentionally not self-synchronized; a single external mutual-exclusion gate is the sole source of thread safety. The changes below enforce this by replacing two independent CAS flags (which gave both loops equal priority and allowed concurrent access to shared state) with a `ReentrantLock` whose call sites encode the asymmetry directly: `lock()` for the slow loop (blocks briefly, never skips), `tryLock()` for everything else (opportunistic).

**Concrete changes**

1. **Replace `AtomicBoolean isExecuting` with `ReentrantLock executionLock`.** Slow loop uses `lock()` (blocks until free, never skips); fast loop and `processJobNow` use `tryLock()` (skip if busy). This directly encodes the priority asymmetry: a shared CAS treats all callers equally, so a fast loop holding it when the slow loop's 5–10 min timer fires causes the slow loop to miss its entire cycle.

2. **Remove `checkerExecuting` from `StatusCheckTask`.** `ReentrantLock` is now the sole gate; a second independent flag re-introduces the two-gate concurrency problem.

3. **Remove `synchronized` from all `JobIdsByDay` methods.** They were only added because the two independent gates allowed concurrent access. With a single lock the original invariant — exactly one thread touches `JobIdsByDay` at a time — is fully restored.

4. **Restore `volatile int inflightJobsCount`.** `scheduleDecision()` is called by the timer without holding the lock. Reading `inflightJobIds()` there would be an unsynchronized read of `JobIdsByDay` once `synchronized` is removed. The original counter is written under the lock (visible via the lock's happens-before edge) and read as a plain volatile from `scheduleDecision()`. Also fixes the O(jobs)-allocation-per-tick regression for `hasInflightJobs()` and the gauge update.

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
```diff
@@ -163,6 +171,22 @@ public void setJobDiscoveryIdleLoopDelayMillis(long jobDiscoveryIdleLoopDelayMil
...
+     */
+    @Override
+    @JsonProperty(value = "job_discovery_status_check_interval")
+    public MillisecondBoundConfiguration jobDiscoveryStatusCheckInterval()
+    {
+        return jobDiscoveryStatusCheckInterval;
+    }
+
+    @JsonProperty(value = "job_discovery_status_check_interval")
+    public void setJobDiscoveryStatusCheckInterval(MillisecondBoundConfiguration jobDiscoveryStatusCheckInterval)
```
sarankk, line 185: Nit: This method is unused. Seems like we are following the pattern of having set methods, other set methods in class are deprecated.

yifan-c (approved):

server/src/main/java/org/apache/cassandra/sidecar/config/yaml/RestoreJobConfigurationImpl.java
mansikhara: Kept the setter — it's not actually unused, Jackson invokes it reflectively when deserializing sidecar.yaml via `SidecarConfigurationImpl.readYamlConfiguration`. It mirrors the shape of setJobDiscoveryActiveLoopDelay / setJobDiscoveryIdleLoopDelay right above. The @Deprecated setters in this class are the legacy *_millis aliases (e.g. `job_discovery_active_loop_delay_millis`) being phased out; `job_discovery_status_check_interval` is brand new with no legacy form, so there's nothing to deprecate. Happy to add a *_millis alias if you'd like one for parity, but otherwise leaving as-is.

sarankk (approved):

PR #351, opened 2026-05-15
CASSSIDECAR-424: Add ConfigurationProvider interfaces

pauloricardomg:
## Summary

Introduces the pluggable overlay storage abstraction as defined in CEP-62 for managing Cassandra configuration via Sidecar.

- **CassandraConfigurationOverlay**: Immutable model with version-agnostic `cassandraYaml` (JsonNode) and `extraJvmOpts`, with `updated()` centralizing merge logic for both components
- **ConfigurationOverlaySnapshot**: Wrapper with lazily computed cached SHA-256 content hash and last modified timestamp
- **ConfigurationProvider**: Pluggable storage interface with `getConfiguration` and `storeConfiguration` using hash-based optimistic concurrency
- **InMemoryConfigurationProvider**: Sample implementation with per-instance locking

## Test Plan

- CassandraConfigurationOverlayTest: object validation, yaml and jvm opts merge
- ConfigurationOverlaySnapshotTest: hash determinism, caching, format
- InMemoryConfigurationProviderTest: store/get, optimistic concurrency, instance isolation, concurrent access

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
@@ -178,6 +178,6 @@ public class CassandraConfigurationOverlay
         ObjectNode node = MAPPER.createObjectNode();
         node.set("cassandraYaml", cassandraYaml);
         node.set("extraJvmOpts", MAPPER.valueToTree(extraJvmOpts));
-        return node.toString();
+        return node.toPrettyString();
     }
 }
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
@@ -18,9 +18,8 @@
 
 package org.apache.cassandra.sidecar.configmanagement;
 
-import java.util.ArrayList;
 import java.util.Collections;
-import java.util.List;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Objects;
 
@@ -42,7 +41,8 @@ import org.jetbrains.annotations.Nullable;
  * the responsibility of the Configuration Manager.
  *
  * <p>The {@code extraJvmOpts} field contains JVM options that are appended to the Cassandra JVM startup
- * command. These are opaque strings not subject to schema validation.
+ * command. Each entry maps the full option flag (e.g. {@code -Dcassandra.jmx.local.port}) to its value
+ * (e.g. {@code 7199}). These are opaque strings not subject to schema validation.
  */
 public class CassandraConfigurationOverlay
 {
@@ -52,11 +52,11 @@ public class CassandraConfigurationOverlay
     private final JsonNode cassandraYaml;
 
     @NotNull
-    private final List<String> extraJvmOpts;
+    private final Map<String, String> extraJvmOpts;
 
     @JsonCreator
     public CassandraConfigurationOverlay(@JsonProperty("cassandraYaml") @Nullable JsonNode cassandraYaml,
-                                         @JsonProperty("extraJvmOpts") @Nullable List<String> extraJvmOpts)
+                                         @JsonProperty("extraJvmOpts") @Nullable Map<String, String> extraJvmOpts)
     {
         if (cassandraYaml == null)
         {
@@ -71,8 +71,8 @@ public class CassandraConfigurationOverlay
             this.cassandraYaml = cassandraYaml;
         }
         this.extraJvmOpts = extraJvmOpts != null
-                            ? Collections.unmodifiableList(new ArrayList<>(extraJvmOpts))
-                            : Collections.emptyList();
+                            ? Collections.unmodifiableMap(new LinkedHashMap<>(extraJvmOpts))
+                            : Collections.emptyMap();
     }
 
     /**
@@ -86,11 +86,11 @@ public class CassandraConfigurationOverlay
     }
 
     /**
-     * @return an unmodifiable list of extra JVM options
+     * @return an unmodifiable map of extra JVM options (option name to value)
      */
     @JsonProperty("extraJvmOpts")
     @NotNull
-    public List<String> extraJvmOpts()
+    public Map<String, String> extraJvmOpts()
     {
         return extraJvmOpts;
     }
@@ -98,17 +98,19 @@ public class CassandraConfigurationOverlay
     /**
      * Returns a new overlay with the given updates applied. The current instance is not modified.
      *
+     * <p>Both parameters follow the same semantics: a {@code null} value for a key removes that entry,
+     * a non-null value upserts it.
+     *
      * @param cassandraYamlUpdates field-level changes to cassandra.yaml: key = field name, value = new value.
      *                             A null or {@link com.fasterxml.jackson.databind.node.NullNode} value removes
      *                             the field. Pass {@code null} for no yaml changes.
-     * @param addJvmOpts           JVM options to append to the current list. Pass {@code null} for no additions.
-     * @param removeJvmOpts        JVM options to remove by value. Pass {@code null} for no removals.
+     * @param extraJvmOptsUpdates  JVM option changes: key = option name, value = new option value.
+     *                             A {@code null} value removes the option. Pass {@code null} for no changes.
      * @return a new overlay with the updates applied
      */
     @NotNull
     public CassandraConfigurationOverlay updated(@Nullable Map<String, JsonNode> cassandraYamlUpdates,
-                                                 @Nullable List<String> addJvmOpts,
-                                                 @Nullable List<String> removeJvmOpts)
+                                                 @Nullable Map<String, String> extraJvmOptsUpdates)
     {
         ObjectNode mergedYaml = ((ObjectNode) cassandraYaml).deepCopy();
         if (cassandraYamlUpdates != null)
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
```diff
@@ -0,0 +1,172 @@
...
+     *                             A null or {@link com.fasterxml.jackson.databind.node.NullNode} value removes
+     *                             the field. Pass {@code null} for no yaml changes.
+     * @param addJvmOpts           JVM options to append to the current list. Pass {@code null} for no additions.
+     * @param removeJvmOpts        JVM options to remove by value. Pass {@code null} for no removals.
+     * @return a new overlay with the updates applied
+     */
+    @NotNull
+    public CassandraConfigurationOverlay updated(@Nullable Map<String, JsonNode> cassandraYamlUpdates,
+                                                 @Nullable List<String> addJvmOpts,
+                                                 @Nullable List<String> removeJvmOpts)
```
isaacreath: Right now this function provides differing approaches for updating values between `cassandraYaml` and `extraJvmOpts`.  

For `cassandraYaml`, if I want to update an existing override, I pass the new value in through the `cassandraYamlUpdates` map.

For `extraJvmpOpts`, I need to add the existing JVM option to `removeJvmOpts` and the new value to `addJvmOpts`. 

Is there any way to offer a similar approach for both so that the semantics are consistent for callers?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * @return the cassandra.yaml overlay as a version-agnostic JSON object
+     */
+    @JsonProperty("cassandraYaml")
+    @NotNull
+    public JsonNode cassandraYaml()
+    {
+        return cassandraYaml;
```
isaacreath, line 71: It's technically possible to mutate `cassandraYaml` by accessing the value from here, casting it to a `ObjectNode`, then using `ObjectNode#put` to update a key. 

To preserve the immutable property of this object, we may want to consider returning a copy of the `cassandraYaml`.  That said, this could create a lot of unnecessary objects if the immutability properties aren't necessary.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @Override
+    public String toString()
+    {
+        ObjectNode node = MAPPER.createObjectNode();
+        node.put("hash", hash());
+        node.put("lastModified", lastModified.toString());
+        node.set("configuration", MAPPER.valueToTree(configuration));
+        return node.toString();
```
isaacreath: Nit: Do we want to consider using a pretty print string here? (i.e. `toPrettyString()`). Could make debugging easier.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationProvider.java
```diff
@@ -0,0 +1,66 @@
...
+public interface ConfigurationProvider
+{
+    /**
+     * Retrieve the configuration overlay for the given Cassandra instance.
+     *
+     * @param instance the Cassandra instance metadata
+     * @return the configuration overlay snapshot, or {@code null} if no overlay exists for the instance
+     */
+    @Nullable
+    ConfigurationOverlaySnapshot getConfiguration(InstanceMetadata instance);
```
isaacreath: Did you consider `getOverlay` / `storeOverlay` instead of `getConfiguration` / `storeConfiguration`? The use of overlay can indicate to the reader of callers of this interface that we are not getting a fully materialized config, but it may not be a big deal.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
pauloricardomg: Changed extraJvmOpts from List<String> to Map<String, String> to provide a consistent update API between cassandraYaml and extraJvmOpts. Both now follow the same semantics in the updated() method: pass a key with a non-null value to upsert, or a null value to remove. This eliminates the need for separate addJvmOpts/removeJvmOpts parameters. Keys use the full JVM flag format (e.g. -Dcassandra.jmx.local.port -> 7199). See commit 0f17ae72.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
pauloricardomg: Since this is an internal API, I opted to avoid defensive copies on every accessor call to minimize unnecessary object creation. I updated the constructor to deep-copy the input ObjectNode to prevent external mutation through the original reference, and updated the accessor Javadoc to mention that callers must not mutate the returned node (use updated() instead). This gives us immutability guarantees at construction time without the overhead of copying on every read. See commit e95958e1.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
pauloricardomg: Addressed on 7f8a81a9

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationProvider.java
pauloricardomg: good call, updated on f651dd81

server/src/test/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshotTest.java
```diff
@@ -0,0 +1,131 @@
...
+        ObjectNode yaml1 = MAPPER.createObjectNode();
+        yaml1.put("concurrent_reads", 32);
+        yaml1.put("memtable_flush_writers", 4);
+
+        ObjectNode yaml2 = MAPPER.createObjectNode();
+        yaml2.put("concurrent_reads", 32);
+        yaml2.put("memtable_flush_writers", 4);
+
+        CassandraConfigurationOverlay overlay1 = new CassandraConfigurationOverlay(yaml1, Arrays.asList("-Xmx4G"));
+        CassandraConfigurationOverlay overlay2 = new CassandraConfigurationOverlay(yaml2, Arrays.asList("-Xmx4G"));
```
isaacreath: Looks like CI is failing here. Needs an import? `import java.util.Arrays`

isaacreath:
+1 (nb)

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
```diff
@@ -0,0 +1,175 @@
...
+ * settings. It may contain settings from any Cassandra version supported by Sidecar (4.0, 4.1, 5.0, etc.).
+ * No version-specific validation is performed by this class; validation against a version-aware schema is
+ * the responsibility of the Configuration Manager.
+ *
+ * <p>The {@code extraJvmOpts} field contains JVM options that are appended to the Cassandra JVM startup
+ * command. These are opaque strings not subject to schema validation.
+ */
+public class CassandraConfigurationOverlay
+{
+    private static final ObjectMapper MAPPER = new ObjectMapper();
```
frankgh: ideally we should try to use the framework's JSON functionality to avoid allocating additional resources (for JSON processing) etc.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/InMemoryConfigurationProvider.java
```diff
@@ -0,0 +1,69 @@
...
+            }
+
+            if (current != null && (originalHash == null || !current.hash().equals(originalHash)))
+            {
+                return false;
+            }
+
+            overlays.put(instance.id(), newSnapshot);
+            return true;
+        }
```
frankgh: do we really need locks here? Isn't the concurrent hashmap already a DS that supports this natively?
```suggestion
        return overlays.compute(instance.id(), (k, current) -> {
            if (current == null && originalHash != null)
            {
                return null;
            }

            if (current != null && (originalHash == null || !current.hash().equals(originalHash)))
            {
                return current;
            }
            return newSnapshot;
        }) == newSnapshot;
```

frankgh (commented):
I think we should try to use vertx JSON framework here. I provided a sample of what this would look like here: https://github.com/apache/cassandra-sidecar/commit/23ba281f3344795ca2043fc3a4c967df1e04a9ba

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/InMemoryConfigurationProvider.java
pauloricardomg: The reason I used an explicit lock is to prevent contention within the CHM bucket but I guess this is premature optimization so I'm ok to remove and rely only on CHM.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
pauloricardomg: Good call, this is more fluent and simplify things.  The reason I used jackson is to perform json schema validation in the future but we can convert from vertx json to jackson vertx if needed in the future.

pauloricardomg:
@frankgh incorporated your feedback on 7ce2732, but also reverted a previous change to have extraJvmOptions as a map instead of a list (15673bd1141e74144589bccd0dadfe581b18252d) and added a new commit ad56510 to validate conflicting jvm boolean options are rejected during merge.

frankgh (approved):
+1 looks good. I think we need to rebase and ensure CI is green

pauloricardomg:
green CI: https://app.circleci.com/pipelines/circleci/KaZH6B8gKQFCZv6WTtM3yk/7N9zyX9UjY83jRQygcQ3Mf/49/workflows/7f8d241b-983a-46f8-8edd-0321d353c82c

PR #354, opened 2026-05-20
CASSSIDECAR-468 Various build fixes

rustyrazorblade:
This PR addresses several build / CI related issues.

* Removes the gradle setup github action blocked by ASF
* Adds standard OCI labels for container artifacts (repo, git sha, etc)
* Increases timeouts for failing tests (workaround only)
* Uses standard build layout for test xml and html artifacts, collected into single directory via a `doLast` block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,13 @@ To get up and running, create a temporary alias for every node except the first:
 
 Note that this does not persist across reboots, so you'll have to run it every time you restart.
 
+### File descriptor limit
+Integration tests open many sockets and kqueue selectors via Netty. The default macOS file descriptor limit (256) is too low and will cause `Too many open files` errors. Increase it before running tests:
+
+```
+ulimit -n 65536
+```
+
 Getting started: Running The Sidecar
 --------------------------------------
 
--- a/build.gradle
+++ b/build.gradle
@@ -306,6 +306,20 @@ subprojects {
     test {
         finalizedBy jacocoTestReport // report is always generated after tests run
     }
+
+    tasks.withType(Test).configureEach { testTask ->
+        reports {
+            junitXml.outputLocation.set(rootProject.file("build/test-results/${project.name}/${testTask.name}"))
+            html.outputLocation.set(rootProject.file("build/reports/tests/${project.name}/${testTask.name}"))
+        }
+        testLogging {
+            events "started", "passed", "skipped", "failed"
+            exceptionFormat "full"
+            showExceptions true
+            showCauses true
+            showStackTraces true
+        }
+    }
 }
 
 tasks.register('jacocoRootReport', JacocoReport) {
--- a/server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
@@ -23,12 +23,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import io.vertx.core.Vertx;
 import org.apache.cassandra.sidecar.config.SchemaKeyspaceConfiguration;
 import org.apache.cassandra.sidecar.config.SidecarConfiguration;
-import org.apache.cassandra.sidecar.utils.EventBusUtils;
-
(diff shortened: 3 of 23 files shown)


server/src/main/java/org/apache/cassandra/sidecar/db/schema/SidecarSchema.java
```diff
@@ -41,24 +37,22 @@ public class SidecarSchema implements TableSchemaFetcher
...
-                                            ignored -> isInitialized.set(true));
-        }
-        else
+        if (!this.schemaKeyspaceConfiguration.isEnabled())
         {
             LOGGER.info("Sidecar schema is disabled!");
         }
     }
 
+    public void markInitialized()
```
jmckenzie-dev: nit: this could be package private (at least in current usage form). Do we want this public? Given it mutates state I tend to err on the side of closing down as much as possible.

build.gradle
```diff
@@ -306,6 +304,34 @@ subprojects {
...
+        testLogging {
+            events "started", "passed", "skipped", "failed"
+            exceptionFormat "full"
+            showExceptions true
+            showCauses true
+            showStackTraces true
+        }
+        def proj = project
+        def rootProj = rootProject
+        doLast {
```
jmckenzie-dev: Claude's telling me (and I googled around and it seems correct) that `doLast` won't run on failing tasks. Looks like the canonical approach to getting things to the right location is something like:
```groovy
  tasks.withType(Test).configureEach { testTask ->
      reports {
          junitXml.outputLocation.set(rootProject.file("build/test-results/${project.name}/${testTask.name}"))
          html.outputLocation.set(rootProject.file("build/reports/tests/${project.name}/${testTask.name}"))
      }
      // ...testLogging config...
  }
```
That way the files will always route to where we want them and we don't have to resort to a `finalizedBy` or something to try and force the movement.

README.md
```diff
@@ -65,6 +65,13 @@ To get up and running, create a temporary alias for every node except the first:
 
 Note that this does not persist across reboots, so you'll have to run it every time you restart.
 
+### File descriptor limit
+Integration tests open many sockets and kqueue selectors via Netty. The default macOS file descriptor limit (256) is too low and will cause `Too many open files` errors. Increase it before running tests:
+
+```
+ulimit -n 65536
```
jmckenzie-dev, line 72: Any particular reason we don't just go unlimited or w/something like 256k or 1M?

server/src/test/integration/org/apache/cassandra/sidecar/db/RestoreJobDatabaseAccessorIntTest.java
```diff
@@ -47,7 +47,7 @@ class RestoreJobDatabaseAccessorIntTest extends IntegrationTestBase
     @CassandraIntegrationTest
     void testCrudOperations()
     {
-        waitForSchemaReady(10, TimeUnit.SECONDS);
+        waitForSchemaReady(30, TimeUnit.SECONDS);
```
jmckenzie-dev, line 50: Any reason we want to go to 30 seconds vs. say a couple minutes? I don't have a good sense for how bogged down the gh executor services get but since this is a deterministic latch wait on the schema showing up, seems like we could up the ceiling a bit. Flip side is: if this doesn't apply in 30 seconds you're probably hosed. 🤷🏻

jmckenzie-dev (changes requested):
Questions and nits; the only real thing I'm concerned about is `doLast` not running on task failure so files not getting moved. I think we have a solution for that in the recommendation though.

Great stuff - appreciate you helping stabilize CI!

build.gradle
rustyrazorblade: Ah, good call.  Fixing.

README.md
rustyrazorblade: We can, but I don't think we should... it can do some weird things to your system when you hit those numbers.  I was going to follow this up with a patch to fix the FD issue, it's pretty straightforward.

server/src/test/integration/org/apache/cassandra/sidecar/db/RestoreJobDatabaseAccessorIntTest.java
rustyrazorblade: I initially increased this because someone said it was a timeout problem, but then it wasn't.  I don't think we ever come even remotely close to this, it should be about a second or two.

jmckenzie-dev (approved):

PR #357, opened 2026-05-26
CASSSIDECAR-426: Add FileBasedConfigurationProvider for file-based overlay persistence

pauloricardomg:
Implement the default ConfigurationProvider that persists overlays as JSON files under per-instance directories ({configDir}/{instanceId}/config.json). Writes are atomic via temp file + rename. Hash-based optimistic concurrency control rejects stale updates. Add Jackson annotations and JavaTimeModule to ConfigurationOverlaySnapshot for direct JSON serialization.

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/FileBasedConfigurationProvider.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/FileBasedConfigurationProvider.java
@@ -1,0 +1,145 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.configmanagement;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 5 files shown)


server/src/main/java/org/apache/cassandra/sidecar/configmanagement/FileBasedConfigurationProvider.java
```diff
@@ -0,0 +1,139 @@
...
+            }
+
+            if (current != null && (originalHash == null || !current.hash().equals(originalHash)))
+            {
+                return current;
+            }
+
+            writeToDisk(instance, newSnapshot);
+            return newSnapshot;
+        }) == newSnapshot;
```
frankgh: Can we avoid as much as possible perform IO here?
```suggestion
        boolean storeOverlay = overlays.compute(instance.id(), (k, cached) -> {
            ConfigurationOverlaySnapshot current = cached != null ? cached : readFromDisk(instance);

            if (current == null && originalHash != null)
            {
                return null;
            }

            if (current != null && (originalHash == null || !current.hash().equals(originalHash)))
            {
                return current;
            }

            return newSnapshot;
        }) == newSnapshot;

        if (storeOverlay)
        {
            writeToDisk(instance, newSnapshot);
        }
        return storeOverlay;
```

frankgh (commented):
Looks good in general. CI is not happy yet though

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/FileBasedConfigurationProvider.java
pauloricardomg: Is this correct though? If there's 2 thread updating the config simultaneously this could cause lost updates.

ie. thread1 writes config `{concurrent_writes: 16}` while thread2 writes config `{concurrent_writes: 32}`, they could both succeed but only one of them would be written in the end.

we need to perform the write inside the lock to guarantee serializibility, unless I'm missing something

pauloricardomg: nevermind I see that we are caching the snapshot and updating that what would make the next update fail due to hash mismatch

pauloricardomg:
Addressed comments and findbugs error. 

CI is happy on circle: https://app.circleci.com/pipelines/circleci/KaZH6B8gKQFCZv6WTtM3yk/7N9zyX9UjY83jRQygcQ3Mf/55/workflows/3e448558-c834-48d0-b480-51fc6664d2f5

frankgh (approved):
+1 this PR needs rebasing

PR #358, opened 2026-05-29
CASSSIDECAR-427: Implement basic configuration retrieval in ConfigurationManager

pauloricardomg:
## Summary

Implement the basic GET configuration flow in the Configuration Manager. When handling retrieval requests, the manager fetches the current overlay from the `ConfigurationProvider`, merges it with the base template to compute the effective configuration, and returns the result with its SHA-256 hash and last modified timestamp.

### New classes

- **`ConfigurationManager`**: Retrieves effective configuration by loading the base template via `ConfigUtils.loadConfiguration` and merging it with the provider overlay via `ConfigurationOverlaySnapshot.overlay()`. Wraps provider failures in `ConfigurationManagerException`.
- **`ConfigurationManagerException`**: Unchecked exception wrapping provider failures with a descriptive message.
- **`ConfigUtils`**: Utilities for YAML loading (`loadYaml`), deep merge of configurations (`mergeConfigurations`), and loading a YAML file as a `ConfigurationOverlaySnapshot` (`loadConfiguration`).

### Changes to existing classes

- **`ConfigurationOverlaySnapshot`**: Adds an `overlay()` method that produces a new snapshot by deep-merging another snapshot on top, combining `cassandraYaml` recursively, merging `extraJvmOpts`, and using the max of both `lastModified` timestamps. Renames `overlay` field/accessor to `configuration` to better reflect its role when representing the effective (merged) configuration.

## Test plan

- [ ] `ConfigurationManagerTest`: Verifies effective configuration retrieval with no overlay, with overlay (merge semantics), and provider failure propagation
- [ ] `ConfigUtilsTest`: Verifies YAML loading, deep merge logic (overlay wins, new keys, nested objects, empty overlay), and `loadConfiguration` snapshot creation
- [ ] `ConfigurationOverlaySnapshotTest`: Verifies `overlay()` merge of yaml keys, JVM opts, max lastModified, and deep nested object merging
- [ ] `FileBasedConfigurationProviderTest`: Updated to use renamed `configuration()` accessor
- [ ] Run `./gradlew :server:test --tests "org.apache.cassandra.sidecar.configmanagement.*"` - all 50 tests pass

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -3,6 +3,7 @@
  * RPM package is created with wrong version (CASSSIDECAR-472)
  * Define storage provider interface and Cassandra implementation to support durable job tracking (CASSSIDECAR-373)
  * Validate snapshot name during list snapshot (CASSSIDECAR-461)
+ * Add basic configuration retrieval logic to ConfigurationManager (CASSSIDECAR-427)
 
 0.4.0
 -----
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,9 +1,9 @@
 0.5.0
 -----
+ * Add basic configuration retrieval logic to ConfigurationManager (CASSSIDECAR-427)
  * RPM package is created with wrong version (CASSSIDECAR-472)
  * Define storage provider interface and Cassandra implementation to support durable job tracking (CASSSIDECAR-373)
  * Validate snapshot name during list snapshot (CASSSIDECAR-461)
- * Add basic configuration retrieval logic to ConfigurationManager (CASSSIDECAR-427)
 
 0.4.0
 -----
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
@@ -127,6 +127,11 @@ public final class ConfigUtils
             Object overlayValue = field.getValue();
             Object baseValue = result.getValue(fieldName);
 
+            if (overlayValue == null)
+            {
+                continue;
+            }
+
             if (isJsonObject(baseValue) && isJsonObject(overlayValue))
             {
                 JsonObject merged = mergeConfigurations(
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
@@ -28,12 +28,12 @@ import java.util.Map;
 import java.util.Objects;
 
 import com.fasterxml.jackson.core.JsonParser;
-import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.json.jackson.DatabindCodec;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Utility methods for configuration operations: YAML loading and deep merge.
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
@@ -107,9 +107,8 @@ public final class ConfigUtils
 
     /**
      * Deep-merges the overlay onto the base configuration. For nested objects both base and overlay
-     * contain, fields are merged recursively. For all other node types (scalars, arrays),
-     * the overlay value replaces the base value. Null overlay values are skipped and the
-     * corresponding base value is preserved. The base node is not modified.
+     * contain, fields are merged recursively. For all other node types (scalars, arrays, nulls),
+     * the overlay value replaces the base value. The base node is not modified.
      *
      * <p>Overlays may introduce keys not present in the base configuration. Such keys are
      * added to the result as-is (scalars, arrays) or merged recursively (nested objects).
@@ -122,51 +121,8 @@ public final class ConfigUtils
     {
         Objects.requireNonNull(base, "base must not be null");
         Objects.requireNonNull(overlay, "overlay must not be null");
-        if (overlay.isEmpty())
-        {
-            return base.copy();
-        }
-
         JsonObject result = base.copy();
-        for (Map.Entry<String, Object> field : overlay)
-        {
-            String fieldName = field.getKey();
-            Object overlayValue = field.getValue();
-            Object baseValue = result.getValue(fieldName);
-
-            if (overlayValue == null)
-            {
-                continue;
-            }
-
-            if (isJsonObject(baseValue) && isJsonObject(overlayValue))
-            {
-                JsonObject merged = mergeConfigurations(
-                        asJsonObject(baseValue),
-                        asJsonObject(overlayValue));
-                result.put(fieldName, merged);
-            }
-            else
-            {
-                result.put(fieldName, overlayValue);
-            }
-        }
+        result.mergeIn(overlay, true);
         return result;
-    }
-
-    @SuppressWarnings("unchecked")
-    private static boolean isJsonObject(Object value)
-    {
-        return value instanceof JsonObject || value instanceof Map;
-    }
-
-    @SuppressWarnings("unchecked")
-    private static JsonObject asJsonObject(Object value)
-    {
-        if (value instanceof JsonObject)
-        {
-            return (JsonObject) value;
-        }
-        return new JsonObject((Map<String, Object>) value);
     }
 }
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
@@ -29,6 +29,7 @@ import java.util.Objects;
 
 import com.fasterxml.jackson.core.JsonParser;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
 import io.vertx.core.json.JsonObject;
@@ -58,7 +59,16 @@ public final class ConfigUtils
      * @param yamlPath path to the YAML configuration file, or {@code null} for an empty snapshot
      * @return a snapshot representing the file contents
      */
-    public static ConfigurationOverlaySnapshot loadConfiguration(Path yamlPath)
+    /**
+     * Loads configuration from the given YAML path, returning the cached snapshot if the file
+     * has not been modified since it was last read.
+     *
+     * @param yamlPath the path to the YAML file, or {@code null} for an empty snapshot
+     * @param cached   a previously loaded snapshot to reuse if the file is unchanged, or {@code null}
+     * @return the cached snapshot if still valid, or a freshly loaded snapshot
+     */
+    public static ConfigurationOverlaySnapshot loadConfiguration(@Nullable Path yamlPath,
+                                                                  @Nullable ConfigurationOverlaySnapshot cached)
     {
         if (yamlPath == null)
         {
@@ -67,6 +77,10 @@ public final class ConfigUtils
         try
         {
             Instant lastModifiedBefore = Files.getLastModifiedTime(yamlPath).toInstant();
+            if (cached != null && cached.lastModified().equals(lastModifiedBefore))
+            {
+                return cached;
+            }
             JsonObject yaml = loadYaml(yamlPath);
             Instant lastModifiedAfter = Files.getLastModifiedTime(yamlPath).toInstant();
             if (!lastModifiedBefore.equals(lastModifiedAfter))
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
@@ -49,17 +49,6 @@ public final class ConfigUtils
     }
 
     /**
-     * Loads a YAML configuration file and returns it as a {@link ConfigurationOverlaySnapshot}.
-     * The snapshot's {@code lastModified} is set to the file's modification time and
-     * {@code extraJvmOpts} is empty.
-     *
-     * <p>If {@code yamlPath} is {@code null}, returns an {@linkplain ConfigurationOverlaySnapshot#emptySnapshot()
-     * empty snapshot}.
-     *
-     * @param yamlPath path to the YAML configuration file, or {@code null} for an empty snapshot
-     * @return a snapshot representing the file contents
-     */
-    /**
      * Loads configuration from the given YAML path, returning the cached snapshot if the file
      * has not been modified since it was last read.
      *
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
@@ -101,6 +101,9 @@ public final class ConfigUtils
      * contain, fields are merged recursively. For all other node types (scalars, arrays, nulls),
      * the overlay value replaces the base value. The base node is not modified.
      *
+     * <p>Overlays may introduce keys not present in the base configuration. Such keys are
+     * added to the result as-is (scalars, arrays) or merged recursively (nested objects).
+     *
      * @param base    the base configuration tree
      * @param overlay the overlay tree whose values take precedence
      * @return a new tree with the merged result
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
@@ -25,8 +25,10 @@ import java.nio.file.Path;
 import java.time.Instant;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Objects;
 
 import com.fasterxml.jackson.core.JsonParser;
+import org.jetbrains.annotations.NotNull;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import com.fasterxml.jackson.dataformat.yaml.YAMLGenerator;
 import io.vertx.core.json.JsonObject;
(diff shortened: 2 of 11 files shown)


server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
```diff
@@ -21,34 +21,64 @@
...
+     * @return a new snapshot with the merged configuration and the max of both lastModified timestamps
+     */
+    @NotNull
+    public ConfigurationOverlaySnapshot overlay(@NotNull ConfigurationOverlaySnapshot other)
+    {
+        JsonObject mergedYaml = ConfigUtils.mergeConfigurations(configuration.cassandraYaml(),
+                                                                other.configuration().cassandraYaml());
+
+        Map<String, String> mergedOpts = new LinkedHashMap<>(configuration.extraJvmOpts());
+        mergedOpts.putAll(other.configuration().extraJvmOpts());
```
isaacreath: This will add new keys for any key which is not present in the `mergedOpts` map. Is that the behavior we want to expose? Or do we want to limit overlays to known keys in the base template?

Either way worth explicitly documenting in the javadoc.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
```diff
@@ -21,34 +21,64 @@
...
+
+        Instant mergedLastModified = lastModified.isAfter(other.lastModified)
```
isaacreath, line 101: Since the merge produces a new `ConfigurationOverlaySnapshot`, shouldn't `lastModified` be `Instant.now()`? If not, what are the intended semantics of `lastModified`?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
```diff
@@ -0,0 +1,157 @@
...
+            return overlay.copy();
+        }
+        if (overlay == null || overlay.isEmpty())
+        {
+            return base.copy();
+        }
+
+        JsonObject result = base.copy();
+        JsonObject overlayCopy = overlay.copy();
+        for (Map.Entry<String, Object> field : overlayCopy)
```
isaacreath: Similar question as above about the intended semantics when a overlay contains keys not in the base configuration.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
isaacreath: The javadoc in https://github.com/apache/cassandra-sidecar/pull/358/changes#diff-e190c28aea1ad1a802b4cbcdadc180917c653b3b9aac34b0eb23eb4c680e277bR48-R58 helps clarify this a bit.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
```diff
@@ -0,0 +1,157 @@
...
+            if (isJsonObject(baseValue) && isJsonObject(overlayValue))
+            {
+                JsonObject merged = mergeConfigurations(
+                        asJsonObject(baseValue),
+                        asJsonObject(overlayValue));
+                result.put(fieldName, merged);
+            }
+            else
+            {
+                result.put(fieldName, overlayValue);
```
skoppu22: Looks like if base value for a key is not null and overlay value for the same is null, this line still picks overlay value null. Is this expected? We are not doing this for JsonObjects as line 114 above picks base copy. But we are doing this differently for individual keys.

skoppu22: My assumption is overlay contains only overridden keys, while base contains all. For example,

If base is
a: 'a'
b: 'b'
c: 'c'

Then if overlay is
b:'b2'

Does it not mean admins want to only override b, while picking base value for others? In that case, the line 136 missing this case right?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
```diff
@@ -0,0 +1,157 @@
...
+    {
+        if (yamlPath == null)
+        {
+            return ConfigurationOverlaySnapshot.emptySnapshot();
+        }
+        JsonObject yaml = loadYaml(yamlPath);
+        Instant lastModified;
+        try
+        {
+            lastModified = Files.getLastModifiedTime(yamlPath).toInstant();
```
skoppu22: There is no synchronization to avoid race condition between loading yaml at line 65 and retrieving last modified time at line 69. There is a possibility of associating different modified time for older yaml if the file was modified in between.

As we are reading from disk, 
- either we need to read both contents and timestamp in a single call if such an API exists
- or do optimistic synchronization
t1 = read modified time
data = read contents
t2 = read modified time 
if t2 != t1, file was modified and repeat the action.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
skoppu22: Don't we need to call validateNoConflictingBooleanOpts on these merged JVM opts?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
```diff
@@ -84,29 +114,43 @@ public CassandraConfigurationOverlay overlay()
     public JsonObject toJson()
     {
         return new JsonObject()
+               .put("hash", hash())
```
skoppu22, line 146: fromJson doesn't have "hash", while toJson having it

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
```diff
@@ -0,0 +1,77 @@
...
+     * with the overlay from the {@link ConfigurationProvider}.
+     *
+     * @param instance the Cassandra instance metadata
+     * @return a snapshot of the effective configuration with its SHA-256 hash and last modified timestamp
+     * @throws ConfigurationManagerException if the provider fails to retrieve the overlay
+     */
+    @NotNull
+    public ConfigurationOverlaySnapshot getEffectiveConfiguration(InstanceMetadata instance)
+    {
+        ConfigurationOverlaySnapshot baseSnapshot = ConfigUtils.loadConfiguration(baseTemplatePath);
```
skoppu22: Do we plan to cache base snapshot in upcoming PRs, as it doesn't change frequently, and we are involving IO every time here.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
```diff
@@ -0,0 +1,157 @@
...
+        {
+            return overlay.copy();
+        }
+        if (overlay == null || overlay.isEmpty())
+        {
+            return base.copy();
+        }
+
+        JsonObject result = base.copy();
+        JsonObject overlayCopy = overlay.copy();
```
skoppu22: Why do we need to make a copy of overlay when we are not modifying it?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
pauloricardomg: Yes, that's expected. hash is computed from the configuration contents and cached — it's not stored state. toJson() includes it for observability (e.g., logging, API  responses), but fromJson() correctly skips it since it will be recomputed on demand from the deserialized configuration.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
pauloricardomg: overlays can add keys not present in the base template, added it to javadoc on 7c81cc8a

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
pauloricardomg: Just overlaying the configuration over a base template does not produce a new configuration. last modified is the maximum modification timestamp of the base template and the overlay. lastModified is only instant.now when updating the overlay or the base template by modifying the file on disk. Added clarification to javadoc on 99072c35

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
pauloricardomg: clarified on https://github.com/apache/cassandra-sidecar/commit/7c81cc8ac80ccc8282820e06cc2dc44787fe8664

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
pauloricardomg: Add null handling on ad29f2c2 and also added test case to ensure overlay will be correct for that case.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
pauloricardomg: good catch, added optimistic concurrency on 031b999f

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
pauloricardomg: Makes sense, updated 6ec34b80 to prefer base if there's a conflicting boolean jvm opt on overlay.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
pauloricardomg: Added base template caching on b0125eb4

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
pauloricardomg: not needed, removed on f62ab37a

pauloricardomg:
@isaacreath @skoppu22 thanks for the helfpul comments, addressed them

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
```diff
@@ -0,0 +1,182 @@
...
+{
+    static final YAMLFactory YAML_FACTORY = new YAMLFactory()
+            .disable(YAMLGenerator.Feature.WRITE_DOC_START_MARKER);
+
+    private ConfigUtils()
+    {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
```
skoppu22: Looks like older JavaDoc left over here

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
```diff
@@ -21,34 +21,80 @@
...
+        JsonObject mergedYaml = ConfigUtils.mergeConfigurations(configuration.cassandraYaml(),
+                                                                other.configuration().cassandraYaml());
+
+        Map<String, String> mergedOpts = new LinkedHashMap<>(configuration.extraJvmOpts());
+        for (Map.Entry<String, String> entry : other.configuration().extraJvmOpts().entrySet())
+        {
+            String key = entry.getKey();
+            if (CassandraConfigurationOverlay.hasConflictingBooleanOpt(mergedOpts, key))
+            {
+                continue;
```
skoppu22, line 96: Are we preferring base option if there is a conflict? Either we need to log a warning or throw an error. And javadoc says 'merges', doesn't mention that we are picking base option here on conflict. It can be confusing for the users, better to callout whatever the approach we pick here.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
pauloricardomg: Fixed on 23a3ed8be43f67f52a96d6f958d9db7f4ef700fb

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
pauloricardomg: Normally it shouldn't be possible to add conflicting option, since we validate that when updating configs. However in case the remote provider adds a conflicting option it's better to ignore and log a warning then fail, otherwise this can prevent Cassandra from starting. I logged a warning and updated javadoc on f95be70

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+        catch (IOException e)
+        {
+            throw new UncheckedIOException("Failed to load YAML from " + yamlPath, e);
+        }
+    }
+
+    /**
+     * Deep-merges the overlay onto the base configuration. For nested objects both base and overlay
+     * contain, fields are merged recursively. For all other node types (scalars, arrays, nulls),
```
skoppu22: [Minor] 'nulls' can be removed

skoppu22 (approved):

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
pauloricardomg: Good catch, fixed on c9d81a5. Thanks for the review 🙏

isaacreath:
+1 (nb).

isaacreath (approved):

CHANGES.txt
```diff
@@ -3,6 +3,7 @@
  * RPM package is created with wrong version (CASSSIDECAR-472)
  * Define storage provider interface and Cassandra implementation to support durable job tracking (CASSSIDECAR-373)
  * Validate snapshot name during list snapshot (CASSSIDECAR-461)
+ * Add basic configuration retrieval logic to ConfigurationManager (CASSSIDECAR-427)
```
frankgh: can we move this to the top of the list?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
```diff
@@ -0,0 +1,87 @@
...
+    private final Path baseTemplatePath;
+
+    @Nullable
+    private volatile ConfigurationOverlaySnapshot cachedBaseSnapshot;
+
+    /**
+     * @param provider         the configuration provider for fetching overlays
+     * @param baseTemplatePath path to the base cassandra.yaml template, or {@code null} for an empty base
+     */
+    public ConfigurationManager(ConfigurationProvider provider, @Nullable Path baseTemplatePath)
```
frankgh, line 44: is this only used in tests right now?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
```diff
@@ -0,0 +1,172 @@
...
+                        asJsonObject(baseValue),
+                        asJsonObject(overlayValue));
+                result.put(fieldName, merged);
+            }
+            else
+            {
+                result.put(fieldName, overlayValue);
+            }
+        }
+        return result;
```
frankgh, line 126: Can't we use `mergeIn` instead?
```suggestion
        JsonObject result = base.copy();
        result.mergeIn(overlay, true /*deep merge*/);
        return result;
```

server/src/test/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtilsTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+        JsonObject result = ConfigUtils.mergeConfigurations(base, overlay);
+
+        assertThat(result.getString("a")).isEqualTo("a");
+        assertThat(result.getString("b")).isEqualTo("b2");
+        assertThat(result.getString("c")).isEqualTo("c");
+    }
+
+    @Test
+    void testMergeNullOverlayValueKeepsBaseValue()
```
frankgh: I'm trying to understand whether this is the correct behavior. Why do we not honor the null values from the overlay?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
```diff
@@ -84,29 +143,43 @@ public CassandraConfigurationOverlay overlay()
...
+                CassandraConfigurationOverlay.fromJson(json.getJsonObject("configuration"));
+        return new ConfigurationOverlaySnapshot(lastModified, configuration);
     }
 
     private String computeHash()
     {
         try
         {
-            byte[] bytes = overlay.toJson().toBuffer().getBytes();
+            byte[] bytes = configuration.toJson().toBuffer().getBytes();
```
frankgh, line 182: I'm curious why are we using a sha-256 for the hash? can't we use `configuration.hashCode()` ?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
pauloricardomg: yes, the idea is to add only the business logic now and the API will be introduced in [CASSSIDECAR-435](https://issues.apache.org/jira/browse/CASSSIDECAR-435)

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationOverlaySnapshot.java
pauloricardomg: this allows the hash to be computed externally independent from the server by hashing the json

CHANGES.txt
pauloricardomg: done on df9e305b1c997b3fce237de9c21978d2912881ff

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtils.java
pauloricardomg: we couldn't before due to the null-skipping behavior but I incorporated and removed null-skipping on b66b235a1b61f670e6432256d0cfe615228a583a

server/src/test/java/org/apache/cassandra/sidecar/configmanagement/ConfigUtilsTest.java
pauloricardomg: you're right, I updated here to override null from overlay b66b235a1b61f670e6432256d0cfe615228a583a

frankgh (approved):
+1

PR #359, opened 2026-06-10
CASSSIDECAR-374: Implement durable operational job tracker

andresbeckruiz:
## [CASSSIDECAR-374](https://issues.apache.org/jira/browse/CASSSIDECAR-374)

Original PR made against [CASSSIDECAR-373](https://github.com/apache/cassandra-sidecar/pull/339) branch with review comments: https://github.com/andresbeckruiz/cassandra-sidecar/pull/3. 

  ### Changes

  - `DurableOperationalJobTracker`: Implementation of `OperationalJobTracker` that persists job state via `StorageProvider`. Live local jobs are cached in a local `ConcurrentHashMap`,  completed jobs are served from storage
  - `OperationalJobManager`: Separate job tracking from execution — the job is registered in the tracker (and persisted) before execution begins, ensuring durable trackers always have a record before the job runs
  - Unit tests for `DurableOperationalJobTracker` and persist-before-execute verification in `OperationalJobManagerTest`
  - Integration test for `DurableOperationalJobTracker` with Cassandra-backed `StorageProvider`

  ### Future Work

  - [CASSSIDECAR-378](https://issues.apache.org/jira/browse/CASSSIDECAR-378): Add support for creation and coordination of local Sidecar Jobs

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
@@ -32,6 +32,7 @@ import org.slf4j.LoggerFactory;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import org.apache.cassandra.sidecar.common.data.OperationalJobStatus;
+import org.apache.cassandra.sidecar.concurrent.TaskExecutorPool;
 import org.apache.cassandra.sidecar.config.ServiceConfiguration;
 import org.apache.cassandra.sidecar.job.storage.OperationalJobRecord;
 import org.apache.cassandra.sidecar.job.storage.StorageProvider;
@@ -54,13 +55,16 @@ public class DurableOperationalJobTracker implements OperationalJobTracker
 
     private final ConcurrentHashMap<UUID, OperationalJob> liveJobs;
     private final StorageProvider storageProvider;
+    private final TaskExecutorPool executor;
 
     @Inject
     public DurableOperationalJobTracker(ServiceConfiguration serviceConfiguration,
-                                        StorageProvider storageProvider)
+                                        StorageProvider storageProvider,
+                                        TaskExecutorPool executor)
     {
         this.liveJobs = new ConcurrentHashMap<>(serviceConfiguration.operationalJobTrackerSize());
         this.storageProvider = storageProvider;
+        this.executor = executor;
     }
 
     /**
@@ -121,31 +125,28 @@ public class DurableOperationalJobTracker implements OperationalJobTracker
      */
     private void updateTerminalStatus(OperationalJob job)
     {
-        OperationalJobStatus terminalStatus = job.status();
-        String failureReason = job.failureReason();
-        for (int attempt = 1; attempt <= MAX_STATUS_UPDATE_ATTEMPTS; attempt++)
+        updateTerminalStatus(job, 1);
+    }
+
+    private void updateTerminalStatus(OperationalJob job, int attempt)
+    {
+        try
         {
-            try
+            storageProvider.updateJobStatus(job.jobId(), job.operationType(), job.status(), job.failureReason());
+        }
+        catch (RuntimeException e)
+        {
+            LOGGER.warn("Failed to update terminal status for job {} (attempt {}/{}). error={}",
+                        job.jobId(), attempt, MAX_STATUS_UPDATE_ATTEMPTS, e.getMessage());
+            if (attempt < MAX_STATUS_UPDATE_ATTEMPTS)
             {
-                storageProvider.updateJobStatus(job.jobId(), job.operationType(), terminalStatus, failureReason);
-                return;
+                executor.setTimer(RETRY_DELAY_MS * attempt,
+                                  id -> updateTerminalStatus(job, attempt + 1));
             }
-            catch (RuntimeException e)
+            else
             {
-                LOGGER.warn("Failed to update terminal status for job {} (attempt {}/{}). error={}",
-                            job.jobId(), attempt, MAX_STATUS_UPDATE_ATTEMPTS, e.getMessage());
-                if (attempt < MAX_STATUS_UPDATE_ATTEMPTS)
-                {
-                    try
-                    {
-                        Thread.sleep(RETRY_DELAY_MS * attempt);
-                    }
-                    catch (InterruptedException ie)
-                    {
-                        Thread.currentThread().interrupt();
-                        return;
-                    }
-                }
+                LOGGER.error("Exhausted retries when updating terminal status for job {}. " +
+                             "Manual intervention may be required to correct job metadata.", job.jobId(), e);
             }
         }
     }
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
@@ -80,8 +80,8 @@ public class DurableOperationalJobTracker implements OperationalJobTracker
             storageProvider.persistJob(OperationalJobRecord.fromOperationalJob(job));
 
             job.asyncResult().onComplete(ar -> {
-                liveJobs.remove(job.jobId());
                 updateTerminalStatus(job);
+                liveJobs.remove(job.jobId());
             });
 
             return job;
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
@@ -96,9 +96,11 @@ public class DurableOperationalJobTracker implements OperationalJobTracker
                 updateTerminalStatus(job);
                 liveJobs.remove(job.jobId());
             })).onFailure(e -> {
-                liveJobs.remove(jobId, job);
-                LOGGER.error("Failed to persist job {} to storage. Job will not be tracked durably.",
+                // The persist failed, but the job is already executing on a separate executor. Keep it in
+                // liveJobs so in-process status queries and conflict detection still see it.
+                LOGGER.error("Failed to persist job {} to storage. Job will be tracked in-memory only.",
                              jobId, e);
+                job.asyncResult().onComplete(ar -> liveJobs.remove(job.jobId()));
             });
         }
 
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
@@ -47,6 +47,16 @@ import org.jetbrains.annotations.Nullable;
  * {@link OperationalJob} references for the current process, since executing jobs
  * with Vert.x promises cannot be reconstituted from storage. Once a job completes,
  * it is removed from the local map, and subsequent lookups are served from storage.
+ *
+ * <p>Storage currently receives only two writes per job: the initial
+ * {@link OperationalJobStatus#CREATED CREATED} record on submission and the terminal status on
+ * completion. As a result, if the sidecar restarts mid-operation, or if the
+ * terminal-status update exhausts its retries (see {@link #updateTerminalStatus}), the persisted
+ * record can remain stuck at {@code CREATED} even though the operation has since progressed or
+ * finished. There is currently no marker distinguishing a record that is genuinely still
+ * {@code CREATED} from one whose true state was simply never recorded; adding such a marker together
+ * with a reconciliation sweep (leveraging {@link StorageProvider#findAllJobs(int)}) is tracked as
+ * follow-up work.
  */
 @Singleton
 public class DurableOperationalJobTracker implements OperationalJobTracker
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -53,6 +54,7 @@ public class DurableOperationalJobTracker implements OperationalJobTracker
     private static final Logger LOGGER = LoggerFactory.getLogger(DurableOperationalJobTracker.class);
     private static final int MAX_STATUS_UPDATE_ATTEMPTS = 3;
     private static final long RETRY_DELAY_MS = 100;
+    private static final long RETRY_JITTER_MS = 100;
 
     private final ConcurrentHashMap<UUID, OperationalJob> liveJobs;
     private final StorageProvider storageProvider;
@@ -218,8 +220,10 @@ public class DurableOperationalJobTracker implements OperationalJobTracker
                         job.jobId(), attempt, MAX_STATUS_UPDATE_ATTEMPTS, e.getMessage());
             if (attempt < MAX_STATUS_UPDATE_ATTEMPTS)
             {
-                executor.setTimer(RETRY_DELAY_MS * attempt,
-                                  id -> updateTerminalStatus(job, attempt + 1));
+                // Add jitter so that concurrent sidecar processes retrying against the same transient
+                // Cassandra blip do not all retry in lockstep
+                long delay = RETRY_DELAY_MS * attempt + ThreadLocalRandom.current().nextLong(RETRY_JITTER_MS);
+                executor.setTimer(delay, id -> updateTerminalStatus(job, attempt + 1));
             }
             else
             {
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
@@ -51,7 +51,7 @@ import org.jetbrains.annotations.Nullable;
 public class DurableOperationalJobTracker implements OperationalJobTracker
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(DurableOperationalJobTracker.class);
-    private static final int MAX_STATUS_UPDATE_ATTEMPTS = 2;
+    private static final int MAX_STATUS_UPDATE_ATTEMPTS = 3;
     private static final long RETRY_DELAY_MS = 100;
 
     private final ConcurrentHashMap<UUID, OperationalJob> liveJobs;
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
@@ -90,15 +90,13 @@ public class DurableOperationalJobTracker implements OperationalJobTracker
             executor.executeBlocking(() -> {
                 storageProvider.persistJob(OperationalJobRecord.fromOperationalJob(job));
                 return null;
-            }).onFailure(e -> {
+            }).onSuccess(v -> job.asyncResult().onComplete(ar -> {
+                updateTerminalStatus(job);
+                liveJobs.remove(job.jobId());
+            })).onFailure(e -> {
                 liveJobs.remove(jobId, job);
                 LOGGER.error("Failed to persist job {} to storage. Job will not be tracked durably.",
                              jobId, e);
-            });
-
-            job.asyncResult().onComplete(ar -> {
-                updateTerminalStatus(job);
-                liveJobs.remove(job.jobId());
             });
         }
 
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
@@ -74,18 +74,35 @@ public class DurableOperationalJobTracker implements OperationalJobTracker
     @Override
     public OperationalJob computeIfAbsent(UUID jobId, Function<UUID, OperationalJob> mappingFunction)
     {
-        return liveJobs.computeIfAbsent(jobId, id -> {
-            OperationalJob job = mappingFunction.apply(id);
+        if (!storageProvider.isAvailable())
+        {
+            throw new IllegalStateException("Storage provider is not available");
+        }
 
-            storageProvider.persistJob(OperationalJobRecord.fromOperationalJob(job));
+        boolean[] created = {false};
+        OperationalJob job = liveJobs.computeIfAbsent(jobId, id -> {
+            created[0] = true;
+            return mappingFunction.apply(id);
+        });
+
+        if (created[0])
+        {
+            executor.executeBlocking(() -> {
+                storageProvider.persistJob(OperationalJobRecord.fromOperationalJob(job));
+                return null;
+            }).onFailure(e -> {
+                liveJobs.remove(jobId, job);
+                LOGGER.error("Failed to persist job {} to storage. Job will not be tracked durably.",
+                             jobId, e);
+            });
 
             job.asyncResult().onComplete(ar -> {
                 updateTerminalStatus(job);
                 liveJobs.remove(job.jobId());
             });
+        }
 
-            return job;
-        });
+        return job;
     }
 
     @Nullable
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
@@ -38,6 +38,7 @@ import org.apache.cassandra.sidecar.concurrent.TaskExecutorPool;
 import org.apache.cassandra.sidecar.config.ServiceConfiguration;
 import org.apache.cassandra.sidecar.job.storage.OperationalJobRecord;
 import org.apache.cassandra.sidecar.job.storage.StorageProvider;
+import org.apache.cassandra.sidecar.utils.InvocationTrackingFunction;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
@@ -92,13 +93,11 @@ public class DurableOperationalJobTracker implements OperationalJobTracker
             throw new IllegalStateException("Storage provider is not available");
         }
 
-        boolean[] created = {false};
-        OperationalJob job = liveJobs.computeIfAbsent(jobId, id -> {
-            created[0] = true;
-            return mappingFunction.apply(id);
-        });
+        InvocationTrackingFunction<UUID, OperationalJob> mappingFunctionTracker =
+        new InvocationTrackingFunction<>(mappingFunction);
+        OperationalJob job = liveJobs.computeIfAbsent(jobId, mappingFunctionTracker);
 
-        if (created[0])
+        if (mappingFunctionTracker.wasInvoked())
         {
             persistInitialRecord(job, 1);
         }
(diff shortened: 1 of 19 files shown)


server/src/main/java/org/apache/cassandra/sidecar/job/storage/OperationalJobRecord.java
```diff
@@ -173,6 +179,70 @@ public Map<String, String> operationMetadata()
         return operationMetadata;
     }
 
+    @Override
+    @Nullable
+    public UUID nodeId()
+    {
+        return null;
```
pauloricardomg, line 176: Why did we not include this field in the cluster ops schema?

server/src/main/java/org/apache/cassandra/sidecar/job/storage/OperationalJobRecord.java
```diff
@@ -173,6 +179,70 @@ public Map<String, String> operationMetadata()
...
+    public List<UUID> nodesPending()
+    {
+        return Collections.emptyList();
+    }
+
+    @Override
+    @NotNull
+    public List<UUID> nodesExecuting()
+    {
+        return Collections.emptyList();
```
pauloricardomg: Is there any reason why we're not populating these nodes* fields from the `cluster_ops_node_state` table ?

server/src/main/java/org/apache/cassandra/sidecar/job/storage/OperationalJobRecord.java
pauloricardomg: Also why is this field needed, I don't see it being used anywhere.

server/src/test/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTrackerTest.java
```diff
@@ -0,0 +1,332 @@
...
+        CountDownLatch latch = new CountDownLatch(1);
+
+        tracker.computeIfAbsent(jobId, id -> job);
+        executorPool.executeBlocking(job::execute);
+
+        job.asyncResult().onComplete(ar -> latch.countDown());
+        assertThat(latch.await(5, TimeUnit.SECONDS)).isTrue();
+
+        loopAssert(2, () -> {
+            verify(storageProvider).updateJobStatus(eq(jobId), eq(OperationType.DECOMMISSION), eq(SUCCEEDED), isNull());
```
pauloricardomg: this assertion is failing

server/src/test/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTrackerTest.java
```diff
@@ -0,0 +1,332 @@
...
+        CountDownLatch latch = new CountDownLatch(1);
+
+        tracker.computeIfAbsent(jobId, id -> job);
+        executorPool.executeBlocking(job::execute);
+
+        job.asyncResult().onComplete(ar -> latch.countDown());
+        assertThat(latch.await(5, TimeUnit.SECONDS)).isTrue();
+
+        loopAssert(2, () -> {
+            verify(storageProvider).updateJobStatus(eq(jobId), eq(OperationType.DECOMMISSION), eq(FAILED), eq("test failure"));
```
pauloricardomg: this assertion is failing

server/src/test/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTrackerTest.java
```diff
@@ -0,0 +1,332 @@
...
+
+        CountDownLatch latch = new CountDownLatch(1);
+        tracker.computeIfAbsent(jobId, id -> job);
+        executorPool.executeBlocking(job::execute);
+
+        job.asyncResult().onComplete(ar -> latch.countDown());
+        assertThat(latch.await(5, TimeUnit.SECONDS)).isTrue();
+
+        loopAssert(2, () -> {
+            verify(storageProvider, times(2)).updateJobStatus(eq(jobId), eq(OperationType.DECOMMISSION), eq(SUCCEEDED), isNull());
```
pauloricardomg: this assertion is failing

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public OperationalJob computeIfAbsent(UUID jobId, Function<UUID, OperationalJob> mappingFunction)
+    {
+        return liveJobs.computeIfAbsent(jobId, id -> {
+            OperationalJob job = mappingFunction.apply(id);
+
+            storageProvider.persistJob(OperationalJobRecord.fromOperationalJob(job));
```
pauloricardomg: Can we do this outside the CHM lock ? it's a bad practice to perform I/O inside the lock

server/src/main/java/org/apache/cassandra/sidecar/job/storage/OperationalJobRecord.java
```diff
@@ -173,6 +179,70 @@ public Map<String, String> operationMetadata()
...
+    @Nullable
+    public UUID nodeId()
+    {
+        return null;
+    }
+
+    @Override
+    public String name()
+    {
+        return operationType.name();
```
pauloricardomg: Live jobs return lowercase from name():
```java
  // NodeDecommissionJob
  private static final String OPERATION = "decommission";
  public String name() { return OPERATION; }
```

  OperationalJobRecord (from this patch) returns the Java enum's .name():
```java
  // OperationalJobRecord.java
  public String name() {
      return operationType.name();  // → "DECOMMISSION" (Java enum names are uppercase)
  }
```
  The handler builds the API response like this:
```java
  // ListOperationalJobsHandler.java:77
  .operation(job.name())
```

  So a client polling a job sees:
  1. While the job is live: "operation": "decommission"
  2. After the job completes and is evicted from the local map (subsequent calls served from storage): "operation": "DECOMMISSION"

  The same job changes its operation field from lowercase to uppercase just because the backing object switched from a live
  NodeDecommissionJob to an OperationalJobRecord. That's a breaking change for any client doing string comparison on the operation field.

server/src/main/java/org/apache/cassandra/sidecar/job/storage/OperationalJobRecord.java
```diff
@@ -173,6 +179,70 @@ public Map<String, String> operationMetadata()
...
+        return status == OperationalJobStatus.RUNNING;
+    }
+
+    /**
+     * Creates an {@link OperationalJobRecord} from a live {@link OperationalJob}.
+     *
+     * @param job the operational job to convert
+     * @return a new record capturing the job's current state
+     */
+    public static OperationalJobRecord fromOperationalJob(OperationalJob job)
```
pauloricardomg, line 231: When a job completes, the DurableOperationalJobTracker evicts it from the live map. After that, get(jobId) returns
  an OperationalJobRecord from storage. Let's trace what data is available at each stage:

```
  While the job is live (e.g., NodeDecommissionJob):
  nodeId()        → UUID of the node being decommissioned
  startTime()     → when execution began
  nodesPending()  → [nodeId] initially, then [] after execution starts
  nodesExecuting()→ [nodeId] while running
  nodesSucceeded()→ [nodeId] after success
  nodesFailed()   → [nodeId] after failure
```

  The factory that persists the job only captures 3 fields:
```java
  // OperationalJobRecord.fromOperationalJob(job)
  return new OperationalJobRecord(job.jobId(), job.operationType(), job.status());
  //                               ^^^^^^^^     ^^^^^^^^^^^^^^^     ^^^^^^^^^^
  //                               that's it — nodeId, startTime, node lists are all discarded
```

  After completion, get() returns the record with stubbed implementations:
```
  nodeId()         → null            (always)
  startTime()      → null            (constructor sets it to null)
  nodesPending()   → emptyList       (always)
  nodesExecuting() → emptyList       (always)
  nodesSucceeded() → emptyList       (always)
  nodesFailed()    → emptyList       (always)
```

  So an API client observing a decommission job would see:
```
  DURING execution:                    AFTER completion (from storage):
  ─────────────────────────────────    ─────────────────────────────────
  {                                    {
    "jobId": "abc-123",                  "jobId": "abc-123",
    "operation": "decommission",         "operation": "DECOMMISSION",
    "jobStatus": "RUNNING",              "jobStatus": "SUCCEEDED",
    "startTime": "2026-06-12T...",       "startTime": null,
    "nodesExecuting": ["node-uuid"],     "nodesExecuting": [],
    "nodesPending": [],                  "nodesPending": [],
    "nodesSucceeded": [],                "nodesSucceeded": [],
    "nodesFailed": []                    "nodesFailed": []
  }                                    }
```

  All the provenance — which node was decommissioned, when it started, whether it succeeded per-node — is lost once the job leaves the
  local map. The record in Cassandra only has (jobId, operationType, status).

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
```diff
@@ -0,0 +1,153 @@
...
+
+            job.asyncResult().onComplete(ar -> {
+                liveJobs.remove(job.jobId());
```
pauloricardomg: liveJobs removal should come after updateTerminalStatus to avoid `RUNNING → CREATED → SUCCEEDED.` when polling status between removal and update

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
pauloricardomg: what happens if `!storageProvider.isAvailable` ?

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
```diff
@@ -0,0 +1,153 @@
...
+    private final StorageProvider storageProvider;
+    private final TaskExecutorPool executor;
+
+    @Inject
+    public DurableOperationalJobTracker(ServiceConfiguration serviceConfiguration,
+                                        StorageProvider storageProvider,
+                                        TaskExecutorPool executor)
+    {
+        this.liveJobs = new ConcurrentHashMap<>(serviceConfiguration.operationalJobTrackerSize());
+        this.storageProvider = storageProvider;
```
pauloricardomg, line 81: who is expected to initialize the storage provider ? can this class assume it's already initialized or do we want to initialize here?

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
```diff
@@ -0,0 +1,153 @@
...
+    private final ConcurrentHashMap<UUID, OperationalJob> liveJobs;
+    private final StorageProvider storageProvider;
+    private final TaskExecutorPool executor;
+
+    @Inject
+    public DurableOperationalJobTracker(ServiceConfiguration serviceConfiguration,
+                                        StorageProvider storageProvider,
+                                        TaskExecutorPool executor)
+    {
+        this.liveJobs = new ConcurrentHashMap<>(serviceConfiguration.operationalJobTrackerSize());
```
pauloricardomg, line 80: Unlike InMemoryOperationalJobTracker which has removeEldestEntry, the durable tracker has no eviction. If
  asyncResult() never completes (e.g., executeBlocking fails to submit due to full pool), the entry stays in the map forever. Is this expected?

pauloricardomg (commented):
I noticed the durable job tracker is never initialized, do we plan to do this in a follow-up PR ? Is the idea that the operator will choose whether to use the durable or the in memory job tracker ?

server/src/test/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTrackerTest.java
andresbeckruiz: Fixed

server/src/test/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTrackerTest.java
andresbeckruiz: Fixed

server/src/test/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTrackerTest.java
andresbeckruiz: Fixed

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
skoppu22: There are 2 problems here
- Performing a IO in computeIfAbsent which runs under a lock
- This computeIfAbsent is called by trySubmitJob which is called by handlers i.e, blocks the vert.x event loop

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
```diff
@@ -0,0 +1,153 @@
...
+    {
+        try
+        {
+            storageProvider.updateJobStatus(job.jobId(), job.operationType(), job.status(), job.failureReason());
+        }
+        catch (RuntimeException e)
+        {
+            LOGGER.warn("Failed to update terminal status for job {} (attempt {}/{}). error={}",
+                        job.jobId(), attempt, MAX_STATUS_UPDATE_ATTEMPTS, e.getMessage());
+            if (attempt < MAX_STATUS_UPDATE_ATTEMPTS)
```
skoppu22: - This method is called from above method with attempt=1, if 1st attempt fails, we are doing only one more retry. Should we increase value of MAX_STATUS_UPDATE_ATTEMPTS ?
- Also, if retries also failed to update, we will have stale entries in CREATED stated. We need a periodic thread running once in a while and removing stale entries (this can be deferred to later as not a blocker).

server/src/main/java/org/apache/cassandra/sidecar/job/storage/OperationalJobRecord.java
andresbeckruiz: `nodeId` was not added to the cluster ops schema because for cluster-wide records, there's no single associated node. Per-node state is tracked separately in the `cluster_ops_node_state` table, and the nodes included in a job are recorded in the `node_execution_order` column. 

`nodeId` is part of the `OperationalJobInfo` interface, which was designed to capture the common read-only fields from `OperationalJob`. `nodeId` already existed for single-node jobs (drain, decommission, move). Since `OperationalJobRecord` now implements this interface, it returns null since cluster-wide records aren't associated with a single node.

server/src/main/java/org/apache/cassandra/sidecar/job/storage/OperationalJobRecord.java
andresbeckruiz: Populated these fields in [d8d047e](https://github.com/apache/cassandra-sidecar/pull/359/commits/d8d047e49479acdd16748a5a7f3a3ca322c7e9af)

server/src/main/java/org/apache/cassandra/sidecar/job/storage/OperationalJobRecord.java
andresbeckruiz: Good catch. Fixed in [791b750](https://github.com/apache/cassandra-sidecar/pull/359/commits/791b750cde9b9db77695c7a8f7ea0be692c64da8)

server/src/main/java/org/apache/cassandra/sidecar/job/storage/OperationalJobRecord.java
andresbeckruiz: The record in Cassandra is intended to start with this information. The nodes list information concern is addressed in [d8d047e](https://github.com/apache/cassandra-sidecar/pull/359/commits/d8d047e49479acdd16748a5a7f3a3ca322c7e9af). 

For start time, the `OperationalJobCoordinator` introduced in CASSSIDECAR-377 will be responsible for setting this when calling `updateJobStatus` to set the status to RUNNING. See: https://github.com/apache/cassandra-sidecar/blob/c5782978480d259add3c8968d5bb0518a3dd3e50/server/src/main/java/org/apache/cassandra/sidecar/db/ClusterOpsDatabaseAccessor.java#L117-L122

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
andresbeckruiz: Addressed in [adee1db](https://github.com/apache/cassandra-sidecar/pull/359/commits/adee1db8ac1fbea721935df6a45d74e3f67dedbf)

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
andresbeckruiz: Added guard in [464f03e](https://github.com/apache/cassandra-sidecar/pull/359/commits/464f03e82124a4a74bc835b995f25f06856bbc6b)

andresbeckruiz: Addressed in [464f03e](https://github.com/apache/cassandra-sidecar/pull/359/commits/464f03e82124a4a74bc835b995f25f06856bbc6b). Moved the job persistence to outside of the CHM lock, and used `executeBlocking` for the I/O operation. 

There is a possible issue where we return to the client a successful job submission but the job persistence later fails. This should be addressed in a follow up patch to change the return type of this method to `Future<OperationalJob> `and chain the return to the client on the job persistence. The other option is to move job persistence to `OperationalJob.execute,` but this couples jobs with the storage layer. 

I have left this out of this patch to limit scope creep.

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
andresbeckruiz: Good catch. If executeBlocking fails to submit due to a full pool, the entry will stay in the map forever. As mentioned in the comment below, I will defer a cleanup job for stale entries to a follow up PR, unless you think it should be addressed here.

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
andresbeckruiz: Increased MAX_STATUS_UPDATE_ATTEMPTS to 3 in [d8a6bf9](https://github.com/apache/cassandra-sidecar/pull/359/commits/d8a6bf9c5fe86537e83f979d10ba71af4f0f38b7). 

Agreed that there should be a periodic thread to remove stale entries in a later patch.

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
andresbeckruiz: The `StorageProvider` will be initialized externally. `DurableOperationalJobTracker` assumes it's already initialized and guards against unavailability with the `isAvailable()` check added in [464f03e](https://github.com/apache/cassandra-sidecar/pull/359/commits/464f03e82124a4a74bc835b995f25f06856bbc6b). 

The idea is that an operator will choose between the durable or in-memory tracker via configuration. I'm deferring these edits to `sidecar.yaml` to a follow up PR once the config is actually being used by a cluster-wide operation.

andresbeckruiz:
CI: https://app.circleci.com/pipelines/github/andresbeckruiz/cassandra-sidecar/97/workflows/644be457-446b-475d-be14-1ea9b9953089

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
```diff
@@ -0,0 +1,233 @@
...
+            executor.executeBlocking(() -> {
+                storageProvider.persistJob(OperationalJobRecord.fromOperationalJob(job));
+                return null;
+            }).onFailure(e -> {
+                liveJobs.remove(jobId, job);
+                LOGGER.error("Failed to persist job {} to storage. Job will not be tracked durably.",
+                             jobId, e);
+            });
+
+            job.asyncResult().onComplete(ar -> {
```
skoppu22: I believe we need to chain this code into above executeBlocking's onSuccess to ensure the this runs only after state persisted.

pauloricardomg (approved):

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
andresbeckruiz: Good catch. Chained the completion handler into the persist's onSuccess so we only register it after the `CREATED` record is persisted, otherwise a failed persist would still call `updateTerminalStatus` on a record that was never persisted. Also added a test covering that case in [373e367](https://github.com/apache/cassandra-sidecar/pull/359/commits/373e3670fba86cc68ad75df4c5c95b6b0100700c)

server/src/test/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTrackerTest.java
```diff
@@ -0,0 +1,455 @@
...
+        doThrow(new StorageProviderException("persist failed")).when(storageProvider).persistJob(any());
+
+        CountDownLatch completed = new CountDownLatch(1);
+        tracker.computeIfAbsent(jobId, id -> job);
+        executorPool.executeBlocking(job::execute);
+        job.asyncResult().onComplete(ar -> completed.countDown());
+
+        assertThat(completed.await(5, TimeUnit.SECONDS)).isTrue();
+        // The tracker registers its completion handler before this test does, so a stale
+        // updateJobStatus would already have run once the job has completed.
```
skoppu22: [minor] looks like this comment needs to be updated, as we expect updateJobStatus to not run

server/src/test/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTrackerTest.java
```diff
@@ -0,0 +1,455 @@
...
+            .when(storageProvider).updateJobStatus(eq(jobId), eq(OperationType.DRAIN), eq(SUCCEEDED), isNull());
+
+        CountDownLatch latch = new CountDownLatch(1);
+        tracker.computeIfAbsent(jobId, id -> job);
+        executorPool.executeBlocking(job::execute);
+
+        job.asyncResult().onComplete(ar -> latch.countDown());
+        assertThat(latch.await(5, TimeUnit.SECONDS)).isTrue();
+
+        loopAssert(2, () -> {
```
skoppu22, line 481: Is this supposed to be 3 as we updated MAX_STATUS_UPDATE_ATTEMPTS to 3?

skoppu22 (approved):

server/src/test/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTrackerTest.java
andresbeckruiz: Addressed in [b59c9d9](https://github.com/apache/cassandra-sidecar/pull/359/commits/b59c9d946578b5a6410cd25afc48bce6597b0758)

server/src/test/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTrackerTest.java
andresbeckruiz: Addressed in [b59c9d9](https://github.com/apache/cassandra-sidecar/pull/359/commits/b59c9d946578b5a6410cd25afc48bce6597b0758)

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
```diff
@@ -0,0 +1,231 @@
...
+        boolean[] created = {false};
+        OperationalJob job = liveJobs.computeIfAbsent(jobId, id -> {
+            created[0] = true;
+            return mappingFunction.apply(id);
+        });
+
+        if (created[0])
+        {
+            executor.executeBlocking(() -> {
+                storageProvider.persistJob(OperationalJobRecord.fromOperationalJob(job));
```
arjunashok: Seems like this initial persist gets zero retries and gets removed from `liveJobs`

Since `OperationalJobManager.trySubmitJob` already kicked off `job.execute()` on a separate executor before this callback runs, the job keeps running against Cassandra while the tracker has already forgotten it.

Subsequently, `get(jobId)` returns null, and `inflightJobsByOperation()` no longer sees it, and could let a second decommission/move/repair job through while the first is still executing underneath. 

Should this keep the job in liveJobs (serving it in a durability-degraded state) rather than dropping it entirely on persist failure?

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
```diff
@@ -0,0 +1,231 @@
...
+        updateTerminalStatus(job, 1);
+    }
+
+    private void updateTerminalStatus(OperationalJob job, int attempt)
+    {
+        try
+        {
+            storageProvider.updateJobStatus(job.jobId(), job.operationType(), job.status(), job.failureReason());
+        }
+        catch (RuntimeException e)
```
arjunashok, line 252: Now that retries are bumped to 3,  when all  exhaust, the storage record is left at the state it last successfully wrote (typically CREATED). There's no flag distinguishing 'this record's status is stale/unreliable' from 'this job genuinely hasn't started.' 

A caller polling `GET /job/{id}` sees "not started yet" for a job that finished, which can either hang waiting for a terminal state that will never come, or time out and resubmit the same operation (which is not safe for decommission/move). It also blocks the periodic reconciliation sweep we want to add as we would need a way to tell "genuinely still CREATED" from "we know this finished but failed to record it".

Couple of options: (a) a metadata field on OperationalJobRecord marking it durability-degraded, checked only by the reconciliation sweep (which can come later), or (b) a new OperationalJobStatus value (e.g. UNKNOWN) visible through the same field every existing caller already reads.

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
arjunashok: For 3 attempts it's probably fine as-is, but consider adding jitter (see `ClusterLeaseClaimTask.initialDelay()` for existing precedent in this codebase) so that concurrent sidecar processes retrying against the same transient Cassandra blip don't all retry in lockstep.

arjunashok:
Noticed `trySetActiveOperation/clearActiveOperation/getActiveOperation` on `StorageProvider` aren't called anywhere yet. 

Conflict checks (checkConflict/inflightJobsByOperation) are still scoped to one sidecar's in-memory liveJobs, not the cross-cluster CAS lock the interface describes. Is wiring that up deferred to CASSSIDECAR-378, or expected in this PR?

arjunashok (commented):
Seems like only two states reach storage: `CREATED` state on submit and the terminal status on completion. `RUNNING` and per-node progress (updateNodeStatus/updateNodeStatuses) are not persisted, and there's no startup/periodic reconciliation (findAllJobs is currently only called from tests). 

So a sidecar restart mid-decommission/move/repair leaves the record stuck at `CREATED` (same symptom as the retry-exhaustion case in my comment), just more common. 

Worth calling out explicitly as a known limitation for long-running jobs  (RestoreJobDiscoverer already does this for restore jobs, if useful as reference).

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
andresbeckruiz: Good idea, addressed in [696b576](https://github.com/apache/cassandra-sidecar/pull/359/commits/696b5760f77bd24f8eb88443ba3901d1e865e1bf) and added a test to ensure job is maintained in liveJobs until completion, regardless of persistJob success.

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
andresbeckruiz: This is a good point.  However, a caller wouldn't see `CREATED` for an actual running job in the first place: `get()` serves the live job from `liveJobs` first, so an in-progress job reports `RUNNING`, not the storage record. We only fall back to the stored CREATED after the job has truly finished (since `updateTerminalStatus` runs only on completion), or after a restart, when `liveJobs` is empty.

Your comment above still holds true if Sidecar restarts-- we could resubmit an operation even if it's running and says "CREATED" in storage. However, this is not a regression, since the current in memory tracker returns `null` from `get()` once the job leaves memory, so it's equally capable of making a client think that the job has not started or does not exist and then resubmit.

Given there is no regression, I would probably defer choosing between a) and b) until a PR is made for a reconciliation sweep later on that will actually use it. If this sounds good, I can make a JIRA documenting this follow up. I have also added a Javadoc documenting the limitation and follow up in [b238bd4](https://github.com/apache/cassandra-sidecar/pull/359/commits/b238bd41a68ea650ad30f98d83630ba523f49057)

arjunashok: Agreed on deferring the design choice until the reconciliation sweep PR. The Javadoc covers it well. Could you file that JIRA(unless one already exists) and link it here (and maybe in the class doc) so it's tracked? Thanks.

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
arjunashok: Thanks. The initial persistJob call still gets exactly one attempt, no retry so a transient storage blip at creation time could downgrade that job to in-memory-only tracking for its  lifetime

Worth giving the initial persist the same (3x + jitter)short retry treatment, or is that intentionally out of scope here?

frankgh (approved):
+1 I have some small NITs that I have captured here: https://github.com/apache/cassandra-sidecar/commit/632b3fad59a4a871ba0cb206ae58de19c9c25327 if you are interested. But functionally the patch looks good.

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
andresbeckruiz: Added retries in [9945171](https://github.com/apache/cassandra-sidecar/pull/359/commits/994517168096c333462116fb7000d0ed03d8b9e5)

server/src/main/java/org/apache/cassandra/sidecar/job/DurableOperationalJobTracker.java
andresbeckruiz: JIRA can be found [here](https://issues.apache.org/jira/browse/CASSSIDECAR-482?jql=labels%20%3D%20CEP-53), and added to class Javadoc as well in [bb7592b](https://github.com/apache/cassandra-sidecar/pull/359/commits/bb7592b6af2a33b1f092e51942354914c0d56fa7)

PR #360, opened 2026-06-10
CASSSIDECAR-377: Implement job coordination for cluster-wide operations

andresbeckruiz:
## [CASSSIDECAR-377](https://issues.apache.org/jira/browse/CASSSIDECAR-377)

Original PR made against [CASSSIDECAR-373](https://github.com/apache/cassandra-sidecar/pull/339) branch with review comments: https://github.com/andresbeckruiz/cassandra-sidecar/pull/2. 

  ### Changes

  - `OperationalJobCoordinator` interface 
  - `StorageOperationalJobCoordinator`: Implementation that delegates to `StorageProvider`'s compare and set based methods to acquire a lock for an active operation
  - `OperationalJob.operationType()`: New abstract method returning `OperationType` enum, implemented by all concrete jobs 
  - `OperationalJobManager` integration: If a job requires coordination, the manager acquires the active lock via the coordinator before execution; throws `OperationalJobConflictException` if rejected
  - Unit tests for `StorageOperationalJobCoordinator` and coordinator integration in `OperationalJobManagerTest`

 This ticket only covers the job activation path when a job is submitted to be executed immediately. Clearing locks after completion and status update activation will be implemented in later tickets (CASSSIDECAR-378, CASSSIDECAR-379).

  ### Future Work

  - [CASSSIDECAR-378](https://issues.apache.org/jira/browse/CASSSIDECAR-378): Add support for creation and coordination of local Sidecar Jobs
  - [CASSSIDECAR-379](https://issues.apache.org/jira/browse/CASSSIDECAR-379): Add PATCH operational-jobs/{id} API to update status of existing operational jobs

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
@@ -29,9 +29,11 @@ import org.slf4j.LoggerFactory;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import org.apache.cassandra.sidecar.common.server.utils.DurationSpec;
+import org.apache.cassandra.sidecar.common.utils.Preconditions;
 import org.apache.cassandra.sidecar.concurrent.ExecutorPools;
 import org.apache.cassandra.sidecar.concurrent.TaskExecutorPool;
 import org.apache.cassandra.sidecar.exceptions.OperationalJobConflictException;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * An abstraction of the management and tracking of long-running jobs running on the sidecar.
@@ -41,18 +43,33 @@ public class OperationalJobManager
 {
     protected final Logger logger = LoggerFactory.getLogger(this.getClass());
     private final OperationalJobTracker jobTracker;
-
+    @Nullable
+    private final OperationalJobCoordinator coordinator;
     private final TaskExecutorPool internalExecutorPool;
 
     /**
-     * Creates a manager instance with a default sized job-tracker.
+     * Creates a manager instance without a coordinator.
      *
      * @param jobTracker the tracker for the operational jobs
      */
-    @Inject
     public OperationalJobManager(OperationalJobTracker jobTracker, ExecutorPools executorPools)
     {
+        this(jobTracker, null, executorPools);
+    }
+
+    /**
+     * Creates a manager instance with a coordinator for cluster-wide operation mutual exclusion.
+     *
+     * @param jobTracker  the tracker for the operational jobs
+     * @param coordinator the coordinator for cluster-wide operations, or {@code null} if not needed
+     */
+    @Inject
+    public OperationalJobManager(OperationalJobTracker jobTracker,
+                                 @Nullable OperationalJobCoordinator coordinator,
+                                 ExecutorPools executorPools)
+    {
         this.jobTracker = jobTracker;
+        this.coordinator = coordinator;
         this.internalExecutorPool = executorPools.internal();
     }
 
@@ -104,6 +121,7 @@ public class OperationalJobManager
             OperationalJob tracked = jobTracker.computeIfAbsent(job.jobId(), jobId -> job);
             if (tracked == job)
             {
+                checkCoordination(job);
                 internalExecutorPool.executeBlocking(job::execute);
             }
         }
@@ -130,6 +148,28 @@ public class OperationalJobManager
         if (job.hasConflict(sameOperationJobs))
         {
             throw new OperationalJobConflictException("The same operational job is already running on Cassandra. operationName='" + job.name() + '\'');
+        }
+    }
+
+    /**
+     * For jobs that require cluster-wide coordination, attempts to acquire the active operation lock
+     * via the coordinator. Throws a conflict if another operation is already active.
+     *
+     * @param job instance of the job to coordinate
+     * @throws OperationalJobConflictException when the coordinator cannot activate the operation
+     */
+    private void checkCoordination(OperationalJob job) throws OperationalJobConflictException
+    {
+        if (job.requiresCoordination())
+        {
+            Preconditions.checkState(coordinator != null,
+                                     "Job requires coordination but no OperationalJobCoordinator is configured");
+            boolean activated = coordinator.trySetActive(job.operationType(), job.jobId());
+            if (!activated)
+            {
+                throw new OperationalJobConflictException("An active operation already exists. operationType='"
+                                                         + job.operationType() + '\'');
+            }
         }
     }
 }
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
@@ -121,7 +121,7 @@ public class OperationalJobManager
             OperationalJob tracked = jobTracker.computeIfAbsent(job.jobId(), jobId -> job);
             if (tracked == job)
             {
-                checkCoordination(job);
+                tryCoordination(job);
                 internalExecutorPool.executeBlocking(job::execute);
             }
         }
@@ -158,7 +158,7 @@ public class OperationalJobManager
      * @param job instance of the job to coordinate
      * @throws OperationalJobConflictException when the coordinator cannot activate the operation
      */
-    private void checkCoordination(OperationalJob job) throws OperationalJobConflictException
+    private void tryCoordination(OperationalJob job) throws OperationalJobConflictException
     {
         if (job.requiresCoordination())
         {
--- a/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
@@ -127,7 +127,12 @@ public class OperationalJobManager
         {
             if (ar.succeeded() && Boolean.TRUE.equals(ar.result()))
             {
-                job.asyncResult().onComplete(result -> releaseActiveOperationLock(job));
+                // Distributed jobs that finish on other nodes opt out of auto-release; the orchestration
+                // layer clears the lock once all nodes reach a terminal state.
+                if (job.releasesOnCompletion())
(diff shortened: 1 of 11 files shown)


server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
```diff
@@ -101,10 +118,12 @@ public void trySubmitJob(OperationalJob job,
             checkConflict(job);
 
             // New job is submitted for all cases when we do not have a corresponding downstream job
-            jobTracker.computeIfAbsent(job.jobId(), jobId -> {
+            OperationalJob tracked = jobTracker.computeIfAbsent(job.jobId(), jobId -> job);
```
pauloricardomg: do we need to remove from jobTracker if there's a coordination conflict ? Or perhaps should we just add to jobTracker in case the tryCoordination was successful, so tryCoordination should come before jobtracker.computeIfAbsent ?

skoppu22: This leaves stale job entry in CREATED state if the coordination fails.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
```diff
@@ -131,4 +150,26 @@ private void checkConflict(OperationalJob job) throws OperationalJobConflictExce
...
+     * via the coordinator. Throws a conflict if another operation is already active.
+     *
+     * @param job instance of the job to coordinate
+     * @throws OperationalJobConflictException when the coordinator cannot activate the operation
+     */
+    private void tryCoordination(OperationalJob job) throws OperationalJobConflictException
+    {
+        if (job.requiresCoordination())
+        {
+            Preconditions.checkState(coordinator != null,
```
skoppu22: Preconditions throws illegal state exception, which is not handled in the caller

server/src/test/java/org/apache/cassandra/sidecar/job/OperationalJobManagerTest.java
```diff
@@ -166,4 +179,93 @@ protected Future<Void> executeInternal() throws OperationalJobException
...
+
+        OperationalJob job = createCoordinatedJob(UUIDs.timeBased());
+        BiConsumer<OperationalJob, OperationalJobConflictException> onComplete = (j, ex) -> {
+            assertThat(ex).isInstanceOf(OperationalJobConflictException.class);
+            assertThat(ex.getMessage()).contains("An active operation already exists");
+            latch.countDown();
+        };
+
+        manager.trySubmitJob(job, onComplete, executorPool.service(), SecondBoundConfiguration.parse("5s"));
+        assertThat(latch.await(10, TimeUnit.SECONDS)).isTrue();
```
skoppu22, line 254: Also verify that tracker doesn't have entry for this job after conflict detected.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
```diff
@@ -131,4 +150,26 @@ private void checkConflict(OperationalJob job) throws OperationalJobConflictExce
...
+                                     "Job requires coordination but no OperationalJobCoordinator is configured");
+            boolean activated = coordinator.trySetActive(job.operationType(), job.jobId());
```
skoppu22: I am afraid to merge this PR without calling clearActive. If we delay adding that in a subsequent PR and if someone enables requiresCoordination() for any job meanwhile, then no more further operations will be allowed. I would recommend implementing calling clearActive in the same PR.

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
isaacreath: Would it make sense to mark the job as `FAILED` if coordination fails? And set the `failureReason` to indicate that the job was not able to start due to coordination failure?

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
andresbeckruiz: No more unhandled `IllegalStateException`, addressed in [3661c4a](https://github.com/apache/cassandra-sidecar/pull/360/commits/3661c4aaaedac3205088f1aa2d60d909357ae234)

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
andresbeckruiz: Coordination now happens before tracking and jobs that fail coordination are now marked as `FAILED`, addressed in [3661c4a](https://github.com/apache/cassandra-sidecar/pull/360/commits/3661c4aaaedac3205088f1aa2d60d909357ae234)

server/src/test/java/org/apache/cassandra/sidecar/job/OperationalJobManagerTest.java
andresbeckruiz: Jobs are now marked as `FAILED` if coordination fails, so no conflict will happen as `inflightJobsByOperation` does not return completed jobs. Added

`assertThat(tracker.inflightJobsByOperation(job.name())).doesNotContain(job);
`

in [3661c4a](https://github.com/apache/cassandra-sidecar/pull/360/commits/3661c4aaaedac3205088f1aa2d60d909357ae234) to confirm

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
andresbeckruiz: Added in [097a9af](https://github.com/apache/cassandra-sidecar/pull/360/commits/097a9af44a5b85b2c4de5f945492176310878904)

andresbeckruiz:
CI: https://app.circleci.com/pipelines/github/andresbeckruiz/cassandra-sidecar/101/workflows/d27734e4-047e-4299-9029-ed0c9624184c

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
```diff
@@ -41,18 +43,33 @@ public class OperationalJobManager
 {
     protected final Logger logger = LoggerFactory.getLogger(this.getClass());
     private final OperationalJobTracker jobTracker;
-
+    @Nullable
+    private final OperationalJobCoordinator coordinator;
```
isaacreath: The nullable `OperationalJobCoordinator` feels like a bit of a smell to me. It seems like the goal here is to support backwards compatibility for the `OperationalJobManager` for users of the `OperationalJobManager` to not have to include coordination (e.g., for existing decommission jobs), am I right about that?

If so, I'd suggest that instead of having a null `OperationalJobCoordinator`, we have a second implementation, `DisabledOperationalJobCoordinator` or something similar, that fails all coordination tasks with an error saying that coordination is not currently supported by this sidecar instance (and maybe how to configure coordination if the user wants to have coordinated jobs).

This way we make the coordination enabled / disabled logic part of the behavior of the coordinator itself, as opposed to having a null dependency here. It can also clean up all accesses to the `OperationalJobCoordinator` as we no longer have to worry about it being null.

isaacreath (changes requested):

pauloricardomg (approved):

skoppu22 (approved):

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
andresbeckruiz: Good idea. Replaced the `@Nullable OperationalJobCoordinator` with a `DisabledOperationalJobCoordinator` in [405a93d](https://github.com/apache/cassandra-sidecar/pull/360/commits/405a93dad1e653cf6bf53cdc31eccfa6fb3b19e3)

To clarify, backwards compatibility for existing uncoordinated jobs (eg: decommission) is already handled by `requiresCoordination() == false`, so they short circuit before touching the coordinator. Therefore, the `DisabledOperationalJobCoordinator` only matters for a job requiring coordination submitted to an instance without coordination support. 

Also updated this patch to bind the `OperationalJobCoordinator` to the `DisabledOperationalJobCoordinator`. Configuration to enable this class (as well as durable tracking) will be added in CASSSIDECAR-378.

isaacreath (approved):
+1 (nb)

server/src/main/java/org/apache/cassandra/sidecar/job/OperationalJobManager.java
```diff
@@ -99,19 +107,60 @@ public void trySubmitJob(OperationalJob job,
...
+        }
+
+        // Acquiring the active operation lock might perform blocking storage I/O so it must
+        // run off the event loop
+        acquireActiveOperationLock(job)
+        .onComplete(ar ->
+        {
+            if (ar.succeeded() && Boolean.TRUE.equals(ar.result()))
+            {
+                job.asyncResult().onComplete(result -> releaseActiveOperationLock(job));
```
arjunashok: `requiresCoordination()` is a clean job-level flag the manager just reads. 

The release side isn't symmetric though, as `trySubmitJob` always auto-releases on local completion, with no way for a job to opt out. That's right for today's single-node jobs, but doesn't match clearActive's javadoc (orchestration-layer clearing for jobs that finish elsewhere). 

Rather than just documenting that gap, could we add a parallel hook for release? 
e.g. releasesOnCompletion(), default true;  the manager just reads it instead of special-casing job types, so a future distributed job can skip the automatic release without any change to `OperationalJobManager`.

andresbeckruiz: Good catch-- addressed in [5d5afc6](https://github.com/apache/cassandra-sidecar/pull/360/commits/5d5afc6f40244c8937050fcfa14f8978cb8aff3d)

arjunashok (approved):

frankgh (approved):
+1

andresbeckruiz:
Rebased to prepare for merge ^

PR #365, opened 2026-06-17
CASSSIDECAR-475: Restore job fails with "Keyspace does not exist" for keyspaces created with quoted mixed-case names

bianca-stanciu29:
Cassandra CQL allows keyspace names to be defined as quoted identifiers, which
preserves their exact casing (e.g. CREATE KEYSPACE “MyKeyspace”).

The DataStax Java Driver (com.datastax.driver.core.Metadata) stores such keyspaces with their case-preserved name internally. When querying the Driver’s Metadata API, the argument must also be quoted for the Driver to perform a case-sensitive lookup. Without quoting, the Driver treats the argument as a case-insensitive identifier, normalizes it to
lowercase, and fails to find the keyspace.

A restore job targeting a keyspace created with a quoted mixed-case name (e.g.
“MyKeyspace”) fails with “Keyspace does not exist” during SSTable range validation,
even though the keyspace is present in the cluster.


server/src/main/java/org/apache/cassandra/sidecar/cluster/locator/CachedLocalTokenRanges.java
```diff
@@ -104,7 +104,7 @@ public Map<Integer, Set<TokenRange>> localTokenRanges(String keyspace, boolean f
             return Collections.emptyMap();
         }
 
-        if (metadata.getKeyspace(keyspace) == null)
+        if (metadata.getKeyspace(Metadata.quoteIfNecessary(keyspace)) == null)
```
skoppu22: We need to check for keyspace == null before this, otherwise quoteIfNecessary may access null string

frankgh: keyspace should not be null here ever

server/src/main/java/org/apache/cassandra/sidecar/coordination/CassandraClientTokenRingProvider.java
```diff
@@ -83,7 +83,7 @@ public Map<Integer, Set<TokenRange>> localTokenRanges(String keyspace, boolean f
     {
         checkAndReloadReloadCaches();
         Metadata metadata = instancesMetadata.instances().get(0).delegate().metadata();
-        if (keyspace == null || metadata.getKeyspace(keyspace) == null)
+        if (keyspace == null || metadata.getKeyspace(Metadata.quoteIfNecessary(keyspace)) == null)
```
skoppu22: Using quoteIfNecessary may cause regression when a mixed letter keyspace name used without quotes, because quoteIfNecessary decides quotes only based on the chars existing in the keyspace string.  Using MetadataUtils.keyspace would avoid this I believe.

Keyspace created as | Cassandra internal name | Sidecar keyspace_name | Pre-PR getKeyspace | Post-PR getKeyspace(quoteIfNecessary)
-- | -- | -- | -- | --
"MyKeyspace" (quoted) | MyKeyspace | MyKeyspace | folds→mykeyspace → null (bug) | "MyKeyspace"→MyKeyspace → found (fixed)
MyKeyspace (unquoted) | mykeyspace | MyKeyspace | folds→mykeyspace → found | "MyKeyspace"→MyKeyspace → null (REGRESSION)
myks (lowercase) | myks | myks | found | found

skoppu22: Same is applicable for other changes in this PR

server/src/main/java/org/apache/cassandra/sidecar/cluster/locator/CachedLocalTokenRanges.java
bianca-stanciu29: Fixed, moved to MetadataUtils.keyspace() which handles the null keyspace case internally before any lookup.

yifan-c (changes requested):
The fix is good in general. But I do not think it covers all code paths. 

For example, updating the integration test `RestoreJobDiscovererPhaseSignalIntTest` to use mixed case names and it fails. 

```
QualifiedName USER_KEYSPACE_TABLE = new QualifiedName("Restore_phase_signal_ks", "T", true, true);
```

CHANGES.txt
```diff
@@ -8,6 +8,7 @@
 
 0.4.0
 -----
+ * Fix restore job failing with "Keyspace does not exist" for quoted mixed-case keyspace names (CASSSIDECAR-475)
```
bianca-stanciu29, line 11: Hey @skoppu22, @frankgh, @yifan-c  just wanted to let you know I've addressed all the feedback!                                       
                                                                                                                                         
For the null check and regression concerns raised by @skoppu22 and @frankgh, I replaced all the raw `getKeyspace` / `quoteIfNecessary` calls with a `MetadataUtils.keyspace()` helper that does a two-step lookup: tries the raw name first, then falls back to the quoted  version. This way it handles both regular lowercase keyspaces and mixed-case quoted ones without any regression.                     
                                                                                                                                         
  For @yifan-c 's point about missing code paths, you were right, `ValidateTableExistenceHandler` wasn't covered. Fixed that too, and  updated `RestoreJobDiscovererPhaseSignalIntTest` to use the mixed-case name exactly as you suggested in your comment.
                                                                                                                                         
  Everything is in commit `f7f73187`. Thanks for the thorough review, let me know if anything else needs attention!

yifan-c (approved):
Thanks for addressing the comments!

frankgh (approved):
+1 thanks for the fix

CHANGES.txt
bianca-stanciu29: Thanks @yifan-c @frankgh for approving! The only red check is Integration tests lightweight (Java 17, C* 5.0), it's a pre-existing timing flake in RepairIntegrationTest.repairTest() (asserts HTTP 202, but the tiny 10-row test repair can finish inside the 5s wait window and return 200 instead, unrelated to this change). Could someone re-run that job (or merge once it's green)? I don't have write access to do either myself. Thanks!

PR #367, opened 2026-06-22
Tighten path containment in Live Migration file route

nvharikrishna:

[CASSSIDECAR-479](https://issues.apache.org/jira/browse/CASSSIDECAR-479) Tightens the path check on the Live Migration file route so that symlinks inside a data directory cannot point to files outside the configured directory.

Earlier, the route only checked the URL text for .. patterns. That check does not catch symlinks, so a symlink within the data dir could still resolve to a file elsewhere on disk. The route now also resolves the path with toRealPath() and confirms the resolved path stays inside the configured base directory.

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileResolveHandler.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileResolveHandler.java
@@ -18,13 +18,13 @@
 
 package org.apache.cassandra.sidecar.handlers.livemigration;
 
-import java.net.URLDecoder;
-import java.nio.charset.StandardCharsets;
+import java.io.IOException;
+import java.net.URI;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.nio.file.PathMatcher;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -42,7 +42,6 @@ import io.vertx.core.http.HttpServerRequest;
 import io.vertx.core.net.SocketAddress;
 import io.vertx.ext.auth.authorization.Authorization;
 import io.vertx.ext.web.RoutingContext;
-import io.vertx.ext.web.handler.HttpException;
 import org.apache.cassandra.sidecar.acl.authorization.BasicPermissions;
 import org.apache.cassandra.sidecar.cluster.instance.InstanceMetadata;
 import org.apache.cassandra.sidecar.concurrent.ExecutorPools;
@@ -52,6 +51,7 @@ import org.apache.cassandra.sidecar.handlers.AbstractHandler;
 import org.apache.cassandra.sidecar.handlers.AccessProtected;
 import org.apache.cassandra.sidecar.handlers.FileStreamHandler;
 import org.apache.cassandra.sidecar.livemigration.LiveMigrationInstanceMetadataUtil;
+import org.apache.cassandra.sidecar.livemigration.LiveMigrationInstanceMetadataUtil.ResolvedPath;
 import org.apache.cassandra.sidecar.utils.CassandraInputValidator;
 import org.apache.cassandra.sidecar.utils.InstanceMetadataFetcher;
 import org.jetbrains.annotations.NotNull;
@@ -60,6 +60,7 @@ import org.jetbrains.annotations.Nullable;
 import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.DIR_INDEX_PARAM;
 import static org.apache.cassandra.sidecar.common.ApiEndpointsV1.DIR_TYPE_PARAM;
 import static org.apache.cassandra.sidecar.livemigration.LiveMigrationPlaceholderUtil.replacePlaceholder;
+import static org.apache.cassandra.sidecar.utils.HttpExceptions.wrapHttpException;
 
 /**
  * Handler that resolves and validates file paths for live migration operations.
@@ -94,7 +95,7 @@ public class LiveMigrationFileResolveHandler extends AbstractHandler<Void> imple
         String dirType = context.pathParam(DIR_TYPE_PARAM);
         if (null == dirType || dirType.isEmpty() || null == LiveMigrationDirType.find(dirType))
         {
-            throw new HttpException(HttpResponseStatus.BAD_REQUEST.code(), "Invalid directory type: " + dirType);
+            throw wrapHttpException(HttpResponseStatus.BAD_REQUEST, "Invalid directory type: " + dirType);
         }
 
         String dirIndex = context.pathParam(DIR_INDEX_PARAM);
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
```diff
@@ -58,6 +63,14 @@ public class LiveMigrationInstanceMetadataUtil
 {
     private static final Logger LOGGER = LoggerFactory.getLogger(LiveMigrationInstanceMetadataUtil.class);
 
+    /**
+     * Caches the canonical (symlinks resolved) form of each base directory keyed by its lexical
+     * path. Base directories come from {@link InstanceMetadata}, which is fixed at startup, so the
+     * canonical form is stable for the process lifetime. The set of distinct base directories is
+     * bounded by configuration (~5–10 per instance), so no eviction is required.
+     */
+    private static final ConcurrentMap<Path, Path> BASE_DIR_CANONICAL = new ConcurrentHashMap<>();
```
yifan-c: I know that the `LiveMigrationInstanceMetadataUtil` is a static helper already. But the static map `BASE_DIR_CANONICAL` catches my attention. I have 2 suggestions.
1. It is common in this project to inject util via guice. Can we do the same, instead of an unmanaged global object out side of guice?
2. To harden the assumption on the size, can we set a size limit and emit warning if exceeding certain size (say 10) when inserting into the map.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileResolveHandler.java
```diff
@@ -123,65 +124,91 @@ protected void handleInternal(RoutingContext rc,
...
         try
         {
-            localFile = LiveMigrationInstanceMetadataUtil.localPath(normalizedPath, instanceMeta).toString();
+            resolved = LiveMigrationInstanceMetadataUtil.resolveLexically(normalizedPath, instanceMeta);
         }
         catch (IllegalArgumentException e)
         {
-            LOGGER.warn("Invalid path", e);
-            rc.response().setStatusCode(HttpResponseStatus.NOT_FOUND.code()).end();
+            rc.fail(wrapHttpException(HttpResponseStatus.BAD_REQUEST, e.getMessage(), e));
```
yifan-c: Is it expected? 404 -> 400

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
```diff
@@ -299,13 +336,68 @@ public static Path localPath(@NotNull String fileUrl,
...
+         *                                  other than missing file
+         * @throws IllegalArgumentException if the canonical path escapes the base directory
+         */
+        public void verifyContainment() throws IOException
+        {
+            if (!Files.exists(resolvedPath))
+            {
+                throw new NoSuchFileException(resolvedPath.toString());
+            }
+            Path canonical = resolvedPath.toRealPath();
```
yifan-c: looks like `toRealPath` already throws `NoSuchFileException` when the file does not exist. The above check is redundant. 

> @throws  IOException – if the file does not exist or an I/O error occurs

nvharikrishna: > @throws IOException – if the file does not exist or an I/O error occurs

Documentation says IOException, and implementation may throw just IOException and not NoSuchFileException. So, checking for file existence and throwing `NoSuchFileException` explicitly.

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileResolveHandler.java
nvharikrishna: Thanks for pointing it out. There is some ambiguity w.r.t usage of IllegalArgumentException. Changed it to BAD_REQUEST as `resolveLexically` throws it for malformed URL. This usage is overlapping with file/dir that don’t exist. Introduced new exception to bring clear separation between malformed URLs and well constructed URLs, but file/directory doesn't exist so that existing behavior is preserved.

server/src/main/java/org/apache/cassandra/sidecar/livemigration/LiveMigrationInstanceMetadataUtil.java
nvharikrishna: This class is a utility class and all its methods are static, so Guice cannot be used. Compared to all other work done during downloading, I think the save by this cache is very minimal. For simplicity, I can remove the cache entirely.

yifan-c: Removing the cache sounds good. 

I am ok with the the utility class w/o managing its own static/global resources. Just to share the info re: guice + utility class, it is possible. Those are the utility/helper classes using Guice in the codebase.                                 
  - CacheFactory — factory for building caches                                                                                                                  
  - DigestVerifierFactory — factory for DigestVerifier instances                                                                                                
  - FileStreamer — streams files to HTTP responses                                                                                                              
  - InstanceMetadataFetcher — retrieves instance info from instanceId/hostname                                                                                  
  - SSTableImporter — performs SSTable imports into a Cassandra instance                                                                                        
  - SSTableUploader — handles SSTable uploads                                                                                                                   
  - SSTableUploadsPathBuilder — builds paths to the SSTable uploads staging dir                                                                                 
  - SidecarClientProvider — Guice Provider<SidecarClient> for the singleton client                                                                                                                                                                               
  - snapshots.SnapshotPathBuilder — builds/validates snapshot paths on a host                                                                                   
  - restore.RestoreJobUtil — restore job helper utilities                                                                                                       
  - concurrent.ExecutorPools — provides shared executor/thread pools

yifan-c (approved):

server/src/main/java/org/apache/cassandra/sidecar/handlers/livemigration/LiveMigrationFileResolveHandler.java
```diff
@@ -123,65 +125,99 @@ protected void handleInternal(RoutingContext rc,
...
+            return;
         }
         catch (IllegalArgumentException e)
         {
-            LOGGER.warn("Invalid path", e);
-            rc.response().setStatusCode(HttpResponseStatus.NOT_FOUND.code()).end();
+            // URL must have been malformed, report it as bad request
+            rc.fail(wrapHttpException(HttpResponseStatus.BAD_REQUEST, e.getMessage(), e));
             return;
         }
```
frankgh: NIT, we can simplify this a bit:


```suggestion
        ResolvedPath resolved = LiveMigrationInstanceMetadataUtil.resolveLexically(normalizedPath, instanceMeta);
```

and then handle with the override:

```
    @Override
    protected void processFailure(Throwable cause, RoutingContext context, String host, SocketAddress remoteAddress, Void request)
    {
        if (cause instanceof UnknownMigrationPrefixException)
        {
            context.fail(wrapHttpException(HttpResponseStatus.NOT_FOUND, cause.getMessage(), cause));
        }
        else
        {
            super.processFailure(cause, context, host, remoteAddress, request);
        }
    }
```

The `IllegalArgumentException` is already handled correctly in `org.apache.cassandra.sidecar.handlers.AbstractHandler#determineHttpException`

frankgh (approved):
+1 Looks good. I have one minor suggestion

PR #368, opened 2026-06-22
CASSSIDECAR-480 CCM demo script

nvharikrishna:

[CASSSIDECAR-480](https://issues.apache.org/jira/browse/CASSSIDECAR-480) Added documentation to try live migration feature using CCM.


frankgh (approved):
+1 great write up! Very detailed. I followed the steps on a clean machine and it worked fine. Since the cluster was pretty empty the process was quick. I think we can have some tlp-stress load going to the local ccm cluster to simulate real traffic during the migration

nvharikrishna:
@frankgh Thanks for the review!

>  I think we can have some tlp-stress load going to the local ccm cluster to simulate real traffic during the migration

Thought it would be a little heavy to talk about one more tool during the demo/hands-on script. Pushed a change to insert rows in a for loop to generate data. Does it sound good?

frankgh:
> Does it sound good?

That sounds reasonable

frankgh (approved):
+1

PR #369, opened 2026-06-29
CASSSIDECAR-429: Add RFC 6902-inspired JSON Patch support to ConfigurationManager

pauloricardomg:
## Summary

Implements patch operations for the Configuration Manager using RFC 6902-inspired semantics. Paths reference the effective configuration structure (base template + overlay merged) but mutations target the overlay only. Supported operations: `add`, `remove`, `replace`, `test`. `move` and `copy` are intentionally excluded as they introduce ambiguous semantics in the two-tier configuration model.

## Configuration Model

The configuration uses a two-tier model:

- **Base template**: the stock `cassandra.yaml` provided by the operator, read-only via this API.
- **Overlay**: a sparse set of values explicitly set via patch operations, stored by the `ConfigurationProvider`.
- **Effective config**: the result of deep-merging base template + overlay. Overlay values take precedence. This is what the GET API returns and what Cassandra uses at startup.

```
Base Template          Overlay              Effective Config
+-----------------+    +-------------+      +-------------------+
| cluster: "prod" |    | reads: 128  |  =>  | cluster: "prod"   |
| reads: 32       | +  | flush: 8    |      | reads: 128        |
| flush: 4        |    +-------------+      | flush: 8          |
+-----------------+                         +-------------------+
```

Patch operations target paths in the **effective config** (what the client sees from GET) but only mutate the **overlay**. The base template is never modified by this API.

### Patch Operations

| Operation | Behavior | Precondition |
|-----------|----------|--------------|
| `add` | Upsert value in overlay | Parent must exist in effective config (RFC 6902) |
| `remove` | Remove key from overlay; effective config falls back to template | Key must exist in overlay (not just template) |
| `replace` | Set value in overlay | Key must exist in effective config |
| `test` | Assert effective config value matches expected; no mutation | Path must exist in effective config |

### Copy-Siblings Strategy (nested paths)

When a patch targets a nested path within a top-level `cassandraYaml` key, the **entire top-level key value** is copied from the effective config into the overlay before applying the leaf change. This ensures the overlay is always self-contained at the top-level key granularity, preventing orphan fragments if the base template later removes the parent structure.

**Example:**

Template:
```json
{"memtable": {"configurations": {"trie": {"class_name": "TrieMemtable", "max_shard_count": 4}, "skiplist": {"class_name": "SkipListMemtable"}}}}
```

Patch: `replace /configuration/cassandraYaml/memtable/configurations/trie/max_shard_count` with value `8`

Resulting overlay (entire `memtable` key stored, not just the leaf):
```json
{"memtable": {"configurations": {"trie": {"class_name": "TrieMemtable", "max_shard_count": 8}, "skiplist": {"class_name": "SkipListMemtable"}}}}
```

The trade-off is that sibling fields are "pinned" in the overlay. If the base template later modifies `skiplist`, the overlay's copy takes precedence. Operators can `remove` the sibling path from the overlay to re-sync with the template.

### Main Classes

| Class | Responsibility |
|-------|---------------|
| `ConfigurationPatchOperation` | Value object: `Op` enum (ADD, REMOVE, REPLACE, TEST) + path + value |
| `ConfigurationPatchValidator` | Validates ops before application: path format, value presence, duplicate detection, segment parsing |
| `ConfigurationPatchApplier` | Applies validated ops: precondition checks (phase 1), then mutations (phase 2) for atomicity |
| `ConfigurationPatchException` | Thrown on validation/application failures with the failing operation attached |
| `ConfigurationConflictException` | Thrown on hash mismatch (optimistic concurrency) |
| `ConfigurationManager.patchConfiguration` | Orchestrates: validate ops, check hash, apply patch, CAS via provider |

### Validation Flow

1. **Structural validation** (`ConfigurationPatchValidator`): path prefix, segment parsing, value presence, duplicate mutation paths
2. **Hash check** (`ConfigurationManager`): caller's `expectedHash` vs current effective config hash
3. **Precondition checks** (`ConfigurationPatchApplier` phase 1): per-op checks against effective config and overlay
4. **Mutations** (`ConfigurationPatchApplier` phase 2): only if all preconditions pass

### Additional Validations

- **JVM option key format**: must start with `-D`, `-X`, or `-XX:`
- **Conflicting boolean JVM opts**: `-XX:+Foo` and `-XX:-Foo` cannot coexist (remove first, then add)
- **Nested paths into array values**: rejected with clear error (array index traversal not supported)
- **`extraJvmOpts` paths must be flat**: no nested segments allowed

### Test Overview

| Test Class | Focus |
|-----------|-------|
| `ConfigurationPatchValidatorTest` | Path parsing, value presence, duplicates, error cases |
| `ConfigurationPatchApplierTest` | All 4 ops at top-level and nested, copy-siblings, arrays, JVM validation, atomicity |
| `ConfigurationManagerTest` | End-to-end through the manager: hash conflicts, provider failures, concurrency |

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
@@ -39,6 +39,9 @@ import org.jetbrains.annotations.Nullable;
  * <p>The {@code extraJvmOpts} field contains JVM options that are appended to the Cassandra JVM startup
  * command. Each entry maps the full option flag (e.g. {@code -Dcassandra.jmx.local.port}) to its value
  * (e.g. {@code 7199}). These are opaque strings not subject to schema validation.
+ *
+ * <p>This class is immutable. Mutations are performed by {@link ConfigurationPatchApplier} which
+ * constructs a new instance with the desired changes.
  */
 public class CassandraConfigurationOverlay
 {
@@ -114,59 +117,6 @@ public class CassandraConfigurationOverlay
         return new CassandraConfigurationOverlay(cassandraYaml, extraJvmOpts);
     }
 
-    /**
-     * Returns a new overlay with the given updates applied. The current instance is not modified.
-     *
-     * <p>Both parameters follow the same semantics: a {@code null} value for a key removes that entry,
-     * a non-null value upserts it.
-     *
-     * @param cassandraYamlUpdates field-level changes to cassandra.yaml: key = field name, value = new value.
-     *                             A {@code null} value removes the field. Pass {@code null} for no yaml changes.
-     * @param extraJvmOptsUpdates  JVM option changes: key = option name, value = new option value.
-     *                             A {@code null} value removes the option. Pass {@code null} for no changes.
-     * @return a new overlay with the updates applied
-     */
-    @NotNull
-    public CassandraConfigurationOverlay updated(@Nullable Map<String, Object> cassandraYamlUpdates,
-                                                 @Nullable Map<String, String> extraJvmOptsUpdates)
-    {
-        JsonObject mergedYaml = cassandraYaml.copy();
-        if (cassandraYamlUpdates != null)
-        {
-            for (Map.Entry<String, Object> entry : cassandraYamlUpdates.entrySet())
-            {
-                if (entry.getValue() == null)
-                {
-                    mergedYaml.remove(entry.getKey());
-                }
-                else
-                {
-                    mergedYaml.put(entry.getKey(), entry.getValue());
-                }
-            }
-        }
-
-        LinkedHashMap<String, String> mergedOpts = new LinkedHashMap<>(extraJvmOpts);
-        if (extraJvmOptsUpdates != null)
-        {
-            for (Map.Entry<String, String> entry : extraJvmOptsUpdates.entrySet())
-            {
-                if (entry.getValue() == null)
-                {
-                    mergedOpts.remove(entry.getKey());
-                }
-                else
-                {
-                    mergedOpts.put(entry.getKey(), entry.getValue());
-                }
-            }
-        }
-
-        validateNoConflictingBooleanOpts(mergedOpts);
-
-        return new CassandraConfigurationOverlay(mergedYaml, mergedOpts);
-    }
-
     static boolean hasConflictingBooleanOpt(Map<String, String> existing, String key)
     {
         String conflicting = conflictingBooleanOpt(key);
@@ -184,19 +134,6 @@ public class CassandraConfigurationOverlay
             return "-XX:+" + key.substring(5);
         }
         return null;
-    }
-
-    private static void validateNoConflictingBooleanOpts(Map<String, String> opts)
-    {
-        for (String key : opts.keySet())
-        {
-            if (hasConflictingBooleanOpt(opts, key))
-            {
-                String option = key.substring(5);
-                throw new IllegalArgumentException(
-                    "Conflicting boolean JVM options: -XX:+" + option + " and -XX:-" + option);
-            }
-        }
     }
 
     @Override
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
@@ -62,7 +62,7 @@ public class CassandraConfigurationOverlay
 
     /**
      * Returns the cassandra.yaml overlay as a version-agnostic JSON object. Callers must not mutate the
-     * returned object; use {@link #updated} to produce a new overlay with changes applied.
+     * returned object; use {@link ConfigurationPatchApplier} to produce a new overlay with changes applied.
      *
      * @return the cassandra.yaml overlay as a version-agnostic JSON object
      */
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
@@ -100,7 +100,9 @@ public class ConfigurationManager
      * For nested paths within a top-level {@code cassandraYaml} key,
      * the entire top-level key value is copied from the effective config into the overlay before
      * applying the leaf change (copy-siblings strategy). This ensures the overlay is always
-     * self-contained at the top-level key granularity.
+     * self-contained at the top-level key granularity. As a consequence, editing a nested leaf pins
+     * its sibling leaves against later base-template drift, and removing an overlaid leaf reverts it to
+     * the current base value; see {@link ConfigurationPatchApplier} for details.
      *
      * <p>All operations are validated atomically: either all succeed or none are applied.
      *
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
@@ -145,7 +145,7 @@ public class ConfigurationManager
                                                                  ? currentOverlay.configuration()
                                                                  : new CassandraConfigurationOverlay(null, null);
             CassandraConfigurationOverlay updatedOverlay =
-                    ConfigurationPatchApplier.apply(parsedOps, effectiveConfig.configuration(), currentOverlayConfig);
+                    ConfigurationPatchApplier.apply(parsedOps, baseSnapshot.configuration(), currentOverlayConfig);
             ConfigurationOverlaySnapshot newOverlaySnapshot = new ConfigurationOverlaySnapshot(Instant.now(),
                                                                                                updatedOverlay);
 
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
@@ -28,7 +28,11 @@ import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
 /**
- * Manages configuration of Cassandra instances via Sidecar
+ * Manages configuration of Cassandra instances via Sidecar.
+ *
+ * <p>The <em>effective configuration</em> is the result of merging the base cassandra.yaml template
+ * with the instance overlay retrieved from the {@link ConfigurationProvider}. It is what the instance
+ * actually sees; the overlay is only the delta persisted on top of the base template.
  */
 public class ConfigurationManager
 {
@@ -50,8 +54,7 @@ public class ConfigurationManager
     }
 
     /**
-     * Computes the effective configuration for the given instance by merging the base template
-     * with the overlay from the {@link ConfigurationProvider}.
+     * Computes the effective configuration for the given instance.
      *
      * @param instance the Cassandra instance metadata
      * @return a snapshot of the effective configuration with its SHA-256 hash and last modified timestamp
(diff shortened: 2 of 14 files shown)


server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
```diff
@@ -84,4 +86,94 @@ private ConfigurationOverlaySnapshot getBaseSnapshot()
...
+            {
+                throw new ConfigurationConflictException(expectedHash, effectiveConfig.hash());
+            }
+
+            // 3. Apply patch operations to the overlay (precondition checks + mutations)
+            CassandraConfigurationOverlay currentOverlayConfig = currentOverlay != null
+                                                                 ? currentOverlay.configuration()
+                                                                 : new CassandraConfigurationOverlay(null, null);
+            CassandraConfigurationOverlay updatedOverlay =
+                    ConfigurationPatchApplier.apply(parsedOps, effectiveConfig.configuration(), currentOverlayConfig);
```
skoppu22: If incoming patch request has a conflicting jvm option against base yaml, this ConfigurationPatchApplier.apply only checks against overlay and doesn't detect conflict. storeOverlay call below stores it. Then baseSnapshot.overlay call below detects conflict with the base, prints warning and returns config without the conflicting option to the caller. The same thing happens in subsequent calls as well.
Also, returned config should be same as stored config, which is not true in this case.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationPatchValidator.java
```diff
@@ -0,0 +1,283 @@
...
+    static final String CASSANDRA_YAML_PREFIX = PATH_PREFIX + CASSANDRA_YAML_SECTION + "/";
+    static final String EXTRA_JVM_OPTS_PREFIX = PATH_PREFIX + EXTRA_JVM_OPTS_SECTION + "/";
+
+    // Allows: -Dproperty.name, -Xmx, -Xss, -XX:+Flag, -XX:-Flag, -XX:Flag
+    // Rejects: -javaagent, -agentpath, -agentlib, keys with = or shell metacharacters
+    static final Pattern JVM_OPT_KEY_PATTERN = Pattern.compile(
+            "^-(D[a-zA-Z][a-zA-Z0-9._-]*|X[a-z][a-zA-Z0-9]*|XX:[+-]?[a-zA-Z][a-zA-Z0-9_]*)$");
+
+    // XX flags that accept arbitrary commands as values
+    static final Set<String> BLOCKED_XX_FLAGS = Set.of(
```
skoppu22: There are more options which can take paths etc e.g. -XX:ErrorFile, -XX:HeapDumpPath, -XX:LogFile, -Xloggc, and -D-style system properties. All absolute paths are allowed. Good to have allowlist instead of blocklist to be specific about which options are allowed and which paths are allowed.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationPatchApplier.java
```diff
@@ -0,0 +1,451 @@
...
+            return new JsonObject((Map<String, Object>) value).copy();
+        }
+        return new JsonObject();
+    }
+
+    private static boolean valueEquals(@NotNull Object actual, @Nullable Object expected)
+    {
+        if (actual instanceof JsonObject && expected instanceof Map)
+        {
+            return actual.equals(new JsonObject((Map<String, Object>) expected));
```
skoppu22: As we anyhow converting `expected` to JsonObject, can we expect the caller to pass both actual and expected as JsonObject, so we don't have to worry if passed params are not exactly matching instanceof expectations?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
skoppu22: Good to have test case for the same

skoppu22:
Well structured code !

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationManager.java
pauloricardomg: Good catch! Addressed in 17fb8896.
apply() now checks boolean JVM opt conflicts against the effective config (base + overlay), not just the overlay, and rejects the patch before storeOverlay is called. Stored and returned configs no longer diverge, and the warn-and-drop path in baseSnapshot.overlay() is no longer hit.

Tests: ConfigurationManagerTest.testPatchConflictingBooleanJvmOptRejected (rejected, nothing stored) and testPatchReturnedConfigMatchesStoredConfig (returned == re-read).

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationPatchValidator.java
pauloricardomg: Addressed in 5c676938 by expanding the blocklist to cover path-bearing/file-writing options (-XX:ErrorFile, -XX:HeapDumpPath, -XX:LogFile, -XX:FlightRecorderOptions, -XX:StartFlightRecording, -Xloggc, -Xlog, -Xbootclasspath) alongside the existing command-executing -XX:OnError/-XX:OnOutOfMemoryError.

I honestly don't know the full set of valid JVM configs, and an allowlist would mean enumerating every legitimate Cassandra JVM option (all the -Xms/-Xmx/-Xss/-XX GC and heap tuning flags plus the whole -Dcassandra.* / -Dcom.sun.management.* / -Djava.* system-property surface), which would make the list massive and hard to keep current. This isn't a security issue in practice: the patch endpoint requires admin credentials, so we're just removing the obviously insecure options and can add more as we find them. If you feel strongly about an allowlist, I'm happy to switch to one.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationPatchApplier.java
pauloricardomg: Addressed in d973ec40. The patch handler builds each operation's value via JsonObject.getValue() (from context.body().asJsonArray()), which returns Vert.x JSON types - JsonObject for objects, JsonArray for arrays, or scalars - the same types resolveValue returns for the effective config. So both sides are already comparable and the TEST precondition is now a plain Objects.equals with no instanceof/conversion. Added coverage for object- and array-valued TEST operations.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
skoppu22: [nit] stale reference 'updated'

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationPatchValidator.java
```diff
@@ -0,0 +1,295 @@
...
+            "-XX:LogFile",
+            "-XX:FlightRecorderOptions",
+            "-XX:StartFlightRecording",
+            "-Xloggc",
+            "-Xlog",
+            "-Xbootclasspath");
+
+    // Allows: alphanumeric, dots, colons, slashes, @, +, commas, spaces, hyphens (max 512 chars)
+    // Rejects: shell metacharacters (;|&$`), quotes, newlines, and other control characters
+    static final Pattern JVM_OPT_VALUE_PATTERN = Pattern.compile("^[a-zA-Z0-9._:/@+, -]{0,512}$");
```
skoppu22: Is allowing space intentional? Just thinking does this allow injection of non-allowed keys or value by any chance?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationPatchApplier.java
```diff
@@ -0,0 +1,491 @@
...
+        {
+            Object child = parent.getValue(segments.get(i));
+            JsonObject childObj = asJsonObject(child);
+            if (childObj == null)
+            {
+                return;
+            }
+            parent = childObj;
+        }
+        parent.remove(segments.get(segments.size() - 1));
```
skoppu22, line 449: [nit] Don't we need to remove parent if it is empty after only child removal?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationPatchApplier.java
```diff
@@ -0,0 +1,491 @@
...
+                checkNestedPathTraversable(effectiveYaml, parsed);
+                Object actualValue = resolveValue(effectiveYaml, parsed.topLevelKey(), parsed.nestedSegments());
+                if (actualValue == null)
+                {
+                    throw new ConfigurationPatchException(
+                            "Test failed: path does not exist in effective config: '" + op.path() + "'", op);
+                }
+                // Both sides are Vert.x JSON types: resolveValue returns JsonObject/JsonArray/scalars,
+                // and op.value() is read from the request via JsonObject.getValue (see
+                // ConfigurationPatchHandler), which wraps objects/arrays the same way.
```
skoppu22: Is ConfigurationPatchHandler coming in upcoming PRs, I couldn't locate it ?

skoppu22 (approved):

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationPatchApplier.java
pauloricardomg: Per RFC 6902, `remove` deletes only the targeted member and never prunes a now-empty parent - the client must remove the parent in a separate op. For example, removing the only overridden key under `server_encryption_options`:

{ "op": "remove", "path": "/cassandraYaml/server_encryption_options/optional" }

leaves `server_encryption_options: {}` behind; the client has to follow up with `{ "op": "remove", "path": "/cassandraYaml/server_encryption_options" }` to drop it. So this is intentional/spec-compliant, not a bug - the empty parent is only removed if explicitly requested.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/CassandraConfigurationOverlay.java
pauloricardomg: Good catch! addressed on 62c0ebc

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationPatchValidator.java
pauloricardomg: Good catch - space is no longer allowed on d818e07 since it would break cassandra startup script (the Cassandra launcher word-splits unquoted values during eval).

I also updated the value pattern to permit the JSON characters `" { } [ ]`. That is needed for JSON-valued options like `-Dcassandra.settings.default_compaction={"class_name":"SizeTieredCompactionStrategy","parameters":{...}}`.

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/ConfigurationPatchApplier.java
pauloricardomg: This will come in upcoming PR for [CASSSIDECAR-436](https://issues.apache.org/jira/browse/CASSSIDECAR-436): https://github.com/pauloricardomg/cassandra-sidecar/pull/7

I updated the comment to remove this reference on 38b8780

pauloricardomg:
Thanks for the review @skoppu22 !

frankgh (approved):
+1

QFREEE:
The copy-siblings strategy makes sense to me, and I think the trade-off is reasonable.

One thing that may be worth considering is making this behavior visible to API consumers. A nested patch appears narrow, but it can copy and pin sibling values in the overlay. Could the PATCH response, API documentation, or logs indicate that the containing top-level section was copied into the overlay?

This may fit better in the CASSSIDECAR-436 endpoint PR, but having that visibility would make later base-template drift easier to understand and troubleshoot.

pauloricardomg:
> This may fit better in the [CASSSIDECAR-436](https://issues.apache.org/jira/browse/CASSSIDECAR-436) endpoint PR, but having that visibility would make later base-template drift easier to understand and troubleshoot.

will add this to CASSSIDECAR-436

frankgh (approved):
+1

PR #371, opened 2026-07-19
CASSSIDECAR-484: Fix the spurious oldest segment age in CdcRawDirectorySpaceCleaner

Klose6:
CdcRawDirectorySpaceCleaner re-reads segment metadata via File.lastModified() after the initial listFiles() scan. If Cassandra (or any other actor) reclaims a segment in the interim, the File.lastModified() will return 0, and nowInMillis - 0 collapses to wall-clock milliseconds.

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
@@ -22,6 +22,7 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
@@ -419,9 +420,10 @@ public class CdcRawDirectorySpaceCleaner implements PeriodicTask
         private final File indexFile;
         private final long segmentId;
         private final long len;
-        // Snapshot lastModified at construction so a concurrent delete (e.g. Cassandra
-        // reclaiming cdc_raw) between the directory scan and later reads does not turn
-        // the value into 0L, which would make (now - lastModified) explode to wall-clock time.
+        // Snapshot size and lastModified from a single readAttributes call so a concurrent
+        // change between separate stat fields cannot produce an inconsistent pair (e.g. len > 0
+        // with lastModified == 0). A later delete before the cleanup loop is still handled by
+        // the lastModified > 0 guards.
         private final long lastModified;
 
         CdcRawSegmentFile(File logFile)
@@ -429,9 +431,22 @@ public class CdcRawDirectorySpaceCleaner implements PeriodicTask
             this.file = logFile;
             final String name = logFile.getName();
             this.segmentId = parseSegmentId(name);
-            this.len = logFile.length();
-            this.lastModified = logFile.lastModified();
             this.indexFile = CdcUtil.getIdxFile(logFile);
+            long size = 0L;
+            long mtime = 0L;
+            try
+            {
+                BasicFileAttributes attrs = Files.readAttributes(logFile.toPath(), BasicFileAttributes.class);
+                size = attrs.size();
+                mtime = attrs.lastModifiedTime().toMillis();
+            }
+            catch (IOException e)
+            {
+                // File removed or unreadable between listFiles and construction; len/mtime
+                // stay 0 and the lastModified > 0 guards skip metric emission and deletion.
+            }
+            this.len = size;
+            this.lastModified = mtime;
         }
 
         public boolean exists()
--- a/server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
@@ -232,8 +232,13 @@ public class CdcRawDirectorySpaceCleaner implements PeriodicTask
         Collections.sort(segmentFiles);
         long nowInMillis = timeProvider.currentTimeMillis();
 
-        // track the age of the oldest commit log segment to give indication of the time-window buffer available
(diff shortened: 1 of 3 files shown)


server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -398,13 +419,18 @@ protected static class CdcRawSegmentFile implements Comparable<CdcRawSegmentFile
...
+        // the value into 0L, which would make (now - lastModified) explode to wall-clock time.
+        private final long lastModified;
 
         CdcRawSegmentFile(File logFile)
         {
             this.file = logFile;
             final String name = logFile.getName();
             this.segmentId = parseSegmentId(name);
             this.len = logFile.length();
+            this.lastModified = logFile.lastModified();
```
jyothsnakonisa: There could be a race condition between two assignments.

You could use Files.readAttributes here to get both of them atomically.

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
jyothsnakonisa, line 216: "segment might already be gone" is represented by lastModified == 0. There are two if conditions that you are adding on the lastModified == 0 condition to fix the bug. This works but what do you think about this alternative?

In the listing segment files pipeline, why don't you filter out segment files which are no longer present, so that you don't need to have if conditions later on?
```suggestion
.map(CdcRawSegmentFile::createSegmentFile)
.filter(Objects::nonNull)
.filter(
```

jyothsnakonisa: CdcRawSegmentFile.createSegmentFile could be a static method to create segment files by reading file attributes atomically to avoid race conditions. Could be something like the following.


```
static CdcRawSegmentFile createSegmentFile(File logFile)
        {
            try
            {
                BasicFileAttributes attrs = Files.readAttributes(logFile.toPath(), BasicFileAttributes.class);
                return new CdcRawSegmentFile(logFile, attrs.size(), attrs.lastModifiedTime().toMillis());
            }
            catch (IOException e)
            {
                // most commonly NoSuchFileException: segment was reclaimed concurrently
                LOGGER.warn("Skipping cdc segment that disappeared before its attributes could be read path={}",
                            logFile, e);
                return null;
            }
        }
```

jyothsnakonisa (commented):
Thanks for the patch @Klose6 left some minor comments

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
Klose6: Hi, thanks for the review! Here the issue we found is sometime the oldestSegmentAge metrics emitted is a very large and unrealistic value(like~50 years) which indicates the segmentFiles.get(0).lastModified()) is 0, this is caused by the segment file was deleted between the listing and deletion, and add the filter in the listing should not help here because the segment can still be gone later before this deletion, so we would like to add the check to fix the spurious value in the metrics here.

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
Klose6: sure, let me try to use the Files.readAttributes() to assign all the values based on that snapshot.

Klose6: Updated to use Files.readAttributes() for the len and modifiedTime, thanks!

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
jyothsnakonisa: Thanks for the reply — but I think this is already handled now that `CdcRawSegmentFile` reads attributes atomically (from my other comment). Nothing reads the file again after that — the age check and delete loop both use the saved value, not a fresh read. So a segment disappearing *after* listing can't cause `lastModified == 0` anymore.

The only way `lastModified` is still `0` is if reading attributes fails during listing itself — which is exactly what my filter suggestion catches.

And if a segment gets deleted after listing but before we delete it, that's already safe — `Files.deleteIfExists()` just no-ops.

So instead of the three `lastModified > 0` checks, we could filter out unreadable segments at listing time. That also avoids `0` meaning two things — "file is gone" vs. a genuinely real epoch timestamp, which right now would never get cleaned up.

jyothsnakonisa: One more thing worth calling out: filtering also removes the need for the `directorySizeBytes -= length` special-casing in the delete loop. Right now, when `lastModified == 0`, the loop assumes the file's already gone and discounts its size from the local budget using the stale cached length — but if that ever turns out to be a false positive (a file with a real epoch mtime that's still on disk), the loop would wrongly believe space was freed and could stop cleaning up too early.

With filtering, that branch goes away entirely. A file that fails `readAttributes()` just isn't in the list, so it's never counted in the first place. And a file with a genuinely real epoch mtime stays in the list and gets treated like any other old segment — normal age/delete logic applies, no special-casing needed.

server/src/main/java/org/apache/cassandra/sidecar/tasks/CdcRawDirectorySpaceCleaner.java
```diff
@@ -430,7 +471,7 @@ public long indexLength()
 
         public long lastModified()
         {
-            return file.lastModified();
+            return lastModified;
```
jyothsnakonisa, line 474: `indexLength()` still calls `File.length()` live, unlike `length()`/`lastModified()` which are now frozen at construction. Probably fine since idx files aren't reclaimed by Cassandra the same way — just flagging for consistency.

PR #373, opened 2026-08-04
adjustments to clarify the README - CASSSIDECAR-488

aar0np:
As per [CASSSIDECAR-488](https://issues.apache.org/jira/browse/CASSSIDECAR-488):

- Revising CIP-1 to CEP1.
- Adding Cassandra version qualifiers to be 4.0 or greater.
- Adding instructions on how to download and run the binaries.
- Additional instructions for defining environment variables.
- Clarifying instructions for running the build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ The latest version of the Sidecar can be found at: https://dlcdn.apache.org/cass
 Build Prerequisites
 -------------------
 We depend on the Cassandra in-jvm dtest framework for testing. 
-Because these jars are not published, you must manually build the dtest jars before you can build the project.
+Because these jars are not published, you must manually build the dtest jars before you can build the project. This can be done using  `build-dtest-jars.sh` from the `scripts/` directory.
 
 The build script supports two parameters:
 - `REPO` - the Cassandra git repository to use for the source files. This is helpful if you need to test with a fork of the Cassandra codebase.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,33 +30,51 @@ We use the [Sidecar for Apache Cassandra JIRA](https://issues.apache.org/jira/pr
...
 Build Prerequisites
 -------------------
 We depend on the Cassandra in-jvm dtest framework for testing. 
 Because these jars are not published, you must manually build the dtest jars before you can build the project.
 
-```shell
-./scripts/build-dtest-jars.sh
-```
-
 The build script supports two parameters:
```
yifan-c, line 51: after moving this script to the bottom of the section, "The build script" can read confusing.

README.md
yifan-c, line 84: This is specifically required for macOS only.

README.md
```diff
@@ -30,33 +30,51 @@ We use the [Sidecar for Apache Cassandra JIRA](https://issues.apache.org/jira/pr
...
-ip address is 127.0.0.x, where x is the node id. 
+```shell
+./scripts/build-dtest-jars.sh
+```
+
+### macOS network aliases
+
+For multi-node in-jvm dtests, network aliases will need to be setup for each Cassandra node. The tests assume each node's ip address is 127.0.0.x, where x is the node id. 
 
 For example if you populated your cluster with 3 nodes, create interfaces for 127.0.0.2 and 127.0.0.3 (the first node of course uses 127.0.0.1).
```
yifan-c: Those 2 paragraphs are common and should not be under `macOS network aliases`

README.md
aar0np: Got it. Adding an extra sentence to clarify that.

README.md
aar0np: That "macOS network aliases" header is still there, it's just a little higher up.

README.md
aar0np: Oh...I see what you mean. Let me move those two paragraphs ABOVE the "macOS network aliases" header like it was before.

aar0np:
Ok, corrections made.

PR #374, opened 2026-08-04
CASSSIDECAR-465: Improve HTTP 429 handling across sidecar server and …

arjunashok:
…client

### Changes
- The sidecar's upload concurrency limiter returns 429, but `BasicRetryPolicy` had no case for it. As a result, clients failed  after one attempt instead of backing off.
- Adds 429 handling to `BasicRetryPolicy` (honors `Retry-After`, falls back to configured backoff); fixes a prerequisite bug in `RequestExecutor.schedule()` that fired retries immediately instead of waiting (this was silently breaking the existing 503 backoff too, independent of 429); adds a `Retry-After` header to the server's 429 response, via config: `retry_after_seconds`.
- `retryAfterSeconds()` is a `default` interface method for backward compatibility

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 0.5.0
 -----
+ * Improve HTTP 429 handling across sidecar server and client (CASSSIDECAR-465)
  * Add RFC 6902-inspired JSON Patch support to ConfigurationManager (CASSSIDECAR-429)
  * Implement job coordination for cluster-wide operations (CASSSIDECAR-377)
  * Fix dead/dropped CDC lifecycle metrics and duplicate SidecarCdcStats interface (CASSSIDECAR-489)
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,6 @@
 0.5.0
 -----
- * Improve HTTP 429 handling across sidecar server and client (CASSSIDECAR-465)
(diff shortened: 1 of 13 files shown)


client/src/main/java/org/apache/cassandra/sidecar/client/RequestExecutor.java
```diff
@@ -372,6 +372,9 @@ protected void schedule(long delayMillis, Runnable runnable)
         {
             singleThreadExecutorService.schedule(runnable, delayMillis, TimeUnit.MILLISECONDS);
         }
-        runnable.run();
+        else
+        {
+            runnable.run();
+        }
```
yifan-c, line 378: wow. This is another bug that is more serious than 429 handling in client.

CHANGES.txt
```diff
@@ -1,5 +1,6 @@
 0.5.0
 -----
+ * Improve HTTP 429 handling across sidecar server and client (CASSSIDECAR-465)
```
yifan-c: The patch include 2 fixes. I am fine with it. Just need to reflect it correctly in the change log. 

```suggestion
 * Support retrying on HTTP 429 responses and fix duplicate immediate execution in RequestExecutor retry scheduling (CASSSIDECAR-465)
```

yifan-c (approved):
The fixes look good to me

jyothsnakonisa (approved):
LGTM

PR #376, opened 2026-08-12
CASSSIDECAR-463: Add fast_forward_enabled flag to enable eager pipeli…

mansikhara:
…ning of staging and importing in coordinated writes


CHANGES.txt
```diff
@@ -13,6 +13,7 @@
 
 0.4.0
 -----
+ * Add fast_forward_enabled flag to enable eager pipelining of staging and importing in coordinated writes (CASSSIDECAR-463)
```
sarankk: Lets add this under `0.5.0`, `0.4.0` release is already out

NEWS.txt
```diff
@@ -21,6 +21,9 @@ New features
 
 Upgrading
 ---------
+    - In CASSSIDECAR-463, a new internal table (restore_job_v7) is created to track the fastForwardEnabled
```
sarankk: Similarly here, lets add under `0.5`

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -653,6 +681,18 @@ boolean shouldLogJob(RestoreJob job)
             return oldStatus == null || job.status == RestoreJobStatus.CREATED || oldStatus != job.status;
         }
 
+        /**
+         * Look up the status of the job seen from the prior run, without recording the current one.
+         *
+         * @return the status of the job seen from the prior run, or null if the job has not been seen
+         */
+        @Nullable
+        RestoreJobStatus lastSeenStatus(RestoreJob job)
```
sarankk: Can we use `getKnownStatus()` instead, functionality seems similar

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreJob.java
```diff
@@ -216,20 +223,77 @@ public boolean hasExpired(long referenceTimestampMillis)
...
+     * staging phase could never conclude once importing has started.
+     *
+     * @return the set of range statuses that satisfy the success criteria
+     */
+    public Set<RestoreRangeStatus> satisfyingRangeStatuses()
+    {
+        RestoreRangeStatus expected = expectedNextRangeStatus();
+        if (fastForwardEnabled && expected == RestoreRangeStatus.STAGED)
+        {
+            return EnumSet.of(RestoreRangeStatus.STAGED, RestoreRangeStatus.SUCCEEDED);
```
sarankk: Shall we return immutable Sets here?

sarankk (approved):
+1 Looks good to me, minor comments

CHANGES.txt
mansikhara: Done, thanks!

NEWS.txt
mansikhara: Done!

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
mansikhara: done!

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreJob.java
mansikhara: Done!

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CreateRestoreJobRequestPayload.java
```diff
@@ -109,6 +111,9 @@ public static Builder builder(RestoreJobSecrets secrets, long expireAtInMillis)
      * @param consistencyLevel    consistency level a job should satisfy
      * @param localDatacenter     the local datacenter name; required if using local consistency level and localDatacenterOnly is specified
      * @param localDatacenterOnly whether the job should restore to the specified local datacenter only
+     * @param fastForwardEnabled  whether Sidecar should eagerly pipeline the staging and the importing phases,
+     *                            i.e. stage while the job is in {@link RestoreJobStatus#CREATED} status and import
+     *                            once the job is in {@link RestoreJobStatus#STAGED} status
      */
     @JsonCreator
     public CreateRestoreJobRequestPayload(@JsonProperty(JOB_ID) UUID jobId,
```
yifan-c, line 122: Please add `@JsonIgnoreProperties(ignoreUnknown = true)` to the class. 

The annotation lets a newer client talk to an older server safely, by having the server ignore fields it doesn't recognize. 
It should have been added sooner. Adding it now won't help the scenario where a client sends a payload with `JOB_FAST_FORWARD_ENABLED` to an old server that doesn't know about it yet, but it should help with future field additions.

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreJob.java
```diff
@@ -216,20 +223,77 @@ public boolean hasExpired(long referenceTimestampMillis)
...
+     */
+    public Set<RestoreRangeStatus> satisfyingRangeStatuses()
+    {
+        RestoreRangeStatus expected = expectedNextRangeStatus();
+        if (fastForwardEnabled && expected == RestoreRangeStatus.STAGED)
+        {
+            return Sets.immutableEnumSet(RestoreRangeStatus.STAGED, RestoreRangeStatus.SUCCEEDED);
+        }
+        return Sets.immutableEnumSet(expected);
+    }
```
yifan-c: The if-block does not make sense to me. 

When fast-forward is enabled, if the expected next range status is STAGED, it means the job status is not yet STAGED (i.e. consistency checker has not passed the staging-phase), then we should not expect range status to also include SUCCEEDED. 

I need to put more thought in the day time.

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreJob.java
```diff
@@ -216,20 +223,77 @@ public boolean hasExpired(long referenceTimestampMillis)
...
-        Preconditions.checkArgument(status != RestoreJobStatus.CREATED,
+        // Ranges of a fast forward job are already staging while the job is in CREATED status, so their progress can
+        // be examined. For all other jobs, no range exists yet in CREATED status.
+        Preconditions.checkArgument(fastForwardEnabled || status != RestoreJobStatus.CREATED,
                                     "Cannot check progress for restore job in CREATED status. jobId: " + jobId);
 
-        return status == RestoreJobStatus.STAGE_READY || status == RestoreJobStatus.STAGED
+        // While the job is still staging, or has just staged, the ranges are expected to reach STAGED.
+        // Note that shouldStageNow() covers the CREATED status of a fast forward job.
+        return shouldStageNow() || status == RestoreJobStatus.STAGED
```
yifan-c: I think the condition should be `shouldStageNow() || (status == RestoreJobStatus.STAGED && !fastForwardEnabled)`. 

Continue mapping job status STAGED to range status STAGED is in the normal flow only, where it waits for the external controller to advance the job status to IMPORT_READY to start checking for SUCCEEDED. When the job status is still STAGED, the consistency checker continues check for range status STAGED. 
It is changed with fast-forward, when job status is STAGED, it already expects SUCCEEDED as it starts to import right away.

yifan-c (commented):
not a full review – submitting the comments so far.

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreJob.java
yifan-c: Putting together a table  for the range status determination with and without fast-forward

| fast-forward? | current _JOB_ status | expected _RANGE_ status |
| -- | -- | -- |
| FALSE | STAGE_READY | STAGED | 
| FALSE | STAGED | STAGED (waiting for external coordinator to signal IMPORT_READY) | 
| FALSE | IMPORT_READY | SUCCEEDED |
| TRUE | CREATED | STAGED |
| TRUE | STAGE_READY | STAGED |
| TRUE | STAGED | SUCEEDED (sidecar should start import in this mode) |
| TRUE | IMPORT_READY | SUCEEDED |

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreJob.java
yifan-c: I think the comment: https://github.com/apache/cassandra-sidecar/pull/376/changes#r3802617419 is what actually needed, instead of this method. There is no need to expand the result into a set. Call-sites can call `expectedNextRangeStatus` directly.

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -465,8 +474,13 @@ private void processSidecarManagedJobMaybe(RestoreJob job)
         if (shouldFindSlicesAndSubmit(job))
         {
             findSlicesAndSubmit(job);
-            // Mark the flag. It prevents finding slices (which is expensive) until the flag is unset.
-            jobIdsByDay.markSlicesDiscovered(job);
+            if (!isEagerStagingPhase(job))
```
yifan-c: The condition can be simplified to the following. 

```suggestion
            if (job.status != RestoreJobStatus.CREATED)
```

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -507,10 +521,23 @@ private boolean isLocalDatacenterExcluded(RestoreJob job)
...
     {
-        return (job.status == RestoreJobStatus.STAGE_READY || job.status == RestoreJobStatus.IMPORT_READY)
+        // A fast forward job stages its slices while it is still in CREATED status. The slices are uploaded
+        // progressively, so the discovery is repeated on every run, i.e. it does not consult the discovered flag.
+        if (isEagerStagingPhase(job))
+        {
+            return true;
+        }
+
+        return (job.status == RestoreJobStatus.STAGE_READY || job.shouldImportNow())
```
yifan-c: Another simplification, together with https://github.com/apache/cassandra-sidecar/pull/376/changes#r3809029703
```java
    private boolean shouldFindSlicesAndSubmit(RestoreJob job)
    {
        return (job.shouldStageNow() || job.shouldImportNow())
               && !jobIdsByDay.isSliceDiscovered(job);
    }
```

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
```diff
@@ -507,10 +521,23 @@ private boolean isLocalDatacenterExcluded(RestoreJob job)
...
                && !jobIdsByDay.isSliceDiscovered(job);
     }
 
+    // Whether the job is a fast forward job that has started staging ahead of the STAGE_READY signal
+    private boolean isEagerStagingPhase(RestoreJob job)
+    {
+        return job.fastForwardEnabled && job.status == RestoreJobStatus.CREATED;
+    }
+
```
yifan-c: The method can be removed

client-common/src/main/java/org/apache/cassandra/sidecar/common/request/data/CreateRestoreJobRequestPayload.java
mansikhara: Added, thanks. Agreed on the reasoning- it doesn't help the new-client → old-server case for `fastForwardEnabled` itself, but it makes the next field addition safe, and it matches what most of the response payloads in client-common already do.

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreJob.java
mansikhara: Makes sense, thanks for the great catch. Fixed

server/src/main/java/org/apache/cassandra/sidecar/db/RestoreJob.java
mansikhara: Agreed and fixed, thanks!

One follow-on this exposed: RestoreJobConsistencyChecker gated its force-a-discovery-run on `status == IMPORT_READY`, which misses the fast-forward case now that the import phase begins at STAGED. Changed it to `restoreJob.shouldImportNow()` and renamed the flag to firstCheckOfImportPhase. Non-fast-forward behavior is unchanged.

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
mansikhara: Done!

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
mansikhara: Removed!

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreJobDiscoverer.java
mansikhara: Done, thanks!

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreRangeTask.java
```diff
@@ -193,8 +192,8 @@ else if (job.status == RestoreJobStatus.IMPORT_READY)
                 }
                 else
                 {
-                    String msg = "Unexpected restore job status. Expected only STAGE_READY or IMPORT_READY when " +
-                                 "processing active slices. Found status: " + job.statusWithOptionalDescription();
+                    String msg = "Unexpected restore job status. Expected the job to be ready to stage or to import " +
+                                 "when processing active slices. Found status: " + job.statusWithOptionalDescription();
```
yifan-c: nit: how about also appending "and fast-forward: enabled/disabled"?

yifan-c (approved):

server/src/main/java/org/apache/cassandra/sidecar/restore/RestoreRangeTask.java
mansikhara: Done!

PR #378, opened 2026-08-18
CASSSIDECAR-434: Add failure policy handling for ConfigurationProvider unavailability

pauloricardomg:
## Overview

`ConfigurationManager` reads and writes cassandra.yaml overlays through a pluggable
`ConfigurationProvider`. Remote implementations can be
unreachable, and today any provider exception surfaces as a generic
`ConfigurationManagerException` - HTTP 500, with no way to keep serving configuration
during an outage.

This adds `FailurePolicyWrapper`, a decorator that mirrors successful provider results
to a local file cache and applies a configurable policy when the provider is down:

- `CACHED_READ_WRITE` - reads and writes fall back to the cache
- `CACHED_READ_ONLY` - reads fall back to the cache, writes are rejected
- `FAIL` - every operation throws

Failures are classified by `isUnavailable()`. Domain errors and provider bugs
(`ConfigurationManagerException`, `IllegalArgumentException`, `NullPointerException`,
`IllegalStateException`, `UnsupportedOperationException`) propagate unchanged so
clients still get the right error; everything else - IOException, socket and timeout
failures, provider client exceptions - is treated as an outage and hits the policy.

The cache is a `FileBasedConfigurationProvider` writing `cached_overlay.json`. Wrapping a
`FileBasedConfigurationProvider` is a no-op - a local provider cannot be unreachable
in the network sense, and double-caching it would serve no purpose.

`FailurePolicy` is threaded through the constructor but not yet read from sidecar.yaml
or wired via DI. That plumbing is a follow-up.

## Known limitation

Writes accepted by the cache under `CACHED_READ_WRITE` are local-only. Once the
provider recovers, the next read returns the delegate's value and overwrites the cache,
so writes made during the outage are lost. There is no reconciliation. This is
documented on the enum constant.

## Review guide

Start with `FailurePolicy` for the three behaviors, then `FailurePolicyWrapper`. The
parts worth attention:

- **`isUnavailable()`** - exception classification is by exclusion, so an unrecognized provider
  client exception is considered unavailability and uses the cache rather than failing.
- **`getOverlay()` null handling** - a `null` from the delegate provider means the overlay was
  deleted upstream, and the wrapper caches an empty snapshot rather than leaving the
  stale entry, so a deleted overlay is not resurrected during a later outage. A `null`
  from the *cache* means nothing was ever cached, and throws instead.
- **`updateCache()`** - best-effort. It hash-compares to skip redundant disk writes and
  swallows every cache exception with a log, because a cache failure must not fail an
  operation the delegate already committed. Writes only refresh the cache when
  `storeOverlay` returned `true`.

`FileBasedConfigurationProvider` is a mechanical change: the filename constant became a
constructor parameter with `overlay.json` as the default.

## Testing

`FailurePolicyWrapperTest` exercises each policy against a delegate that is healthy,
down, and down with nothing cached, plus the cases that are easy to get wrong: `wrap()`
bypassing file-based providers, no cache update on CAS conflict, deleted overlays not
resurrected from cache, and domain rejections propagating instead of falling back.

`ConfigurationManagerTest` adds four end-to-end tests for read fallback, `FAIL`, and
patching under both cached policies. Existing tests moved to the new constructor with
`FailurePolicy.FAIL`, which preserves their prior behavior.

### DIFF
--- a/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/FailurePolicyWrapper.java
+++ b/server/src/main/java/org/apache/cassandra/sidecar/configmanagement/FailurePolicyWrapper.java
@@ -1,0 +1,231 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.sidecar.configmanagement;
+
+import java.nio.file.Path;
+import java.util.Objects;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.cassandra.sidecar.cluster.instance.InstanceMetadata;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A decorator around {@link ConfigurationProvider} that applies a configurable
+ * {@link FailurePolicy} when the downstream provider is unavailable.
+ *
+ * <p>On successful calls to the delegate, the result is cached locally via a
+ * {@link FileBasedConfigurationProvider} writing to {@code cached_overlay.json}.
+ * When the delegate fails, behavior is governed by the configured failure policy.
+ *
+ * <p>Only failures indicating the delegate could not be reached activate the failure policy; any
+ * other failure is propagated to the caller unchanged. See {@link #isUnavailable(Throwable)}.
+ */
+public class FailurePolicyWrapper implements ConfigurationProvider
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(FailurePolicyWrapper.class);
+    static final String CACHED_OVERLAY_FILE_NAME = "cached_overlay.json";
+
+    private final ConfigurationProvider delegate;
+    private final FileBasedConfigurationProvider cache;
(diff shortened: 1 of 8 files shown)


server/src/main/java/org/apache/cassandra/sidecar/configmanagement/FailurePolicyWrapper.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+        catch (Exception e)
+        {
+            return handleReadFailure(instance, e);
+        }
+
+        // A null result means the overlay was deleted upstream. Cache an empty overlay (rather
+        // than leaving the stale entry) so a deleted overlay is not resurrected from cache during
+        // a later outage. An empty overlay merges to a no-op, so it is equivalent to no overlay.
+        updateCache(instance, result != null ? result : ConfigurationOverlaySnapshot.emptySnapshot());
```
skoppu22, line 99: Can we skip updating the cache when policy set to FAIL as Cache won't be read in that case?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/FailurePolicyWrapper.java
```diff
@@ -0,0 +1,231 @@
...
+     * Best-effort update of the local cache after a successful delegate operation.
+     * Cache failures (e.g. disk full, permission errors, corrupt cache file) are logged
+     * and swallowed so they do not fail an operation the delegate already completed.
+     */
+    private void updateCache(InstanceMetadata instance, ConfigurationOverlaySnapshot snapshot)
+    {
+        try
+        {
+            ConfigurationOverlaySnapshot cached = cache.getOverlay(instance);
+            String cachedHash = cached != null ? cached.hash() : null;
```
skoppu22, line 214: Can we maintain cachedHash in memory to avoid reading from Cache file for every read?

server/src/main/java/org/apache/cassandra/sidecar/configmanagement/FailurePolicyWrapper.java
```diff
@@ -0,0 +1,231 @@
...
+            // Skip the disk write when the cache already holds the same content.
+            if (snapshot.hash().equals(cachedHash))
```
skoppu22, line 216: There is a possibility of race condition between threads and leaving Cache stale

 Thread A: delegate.getOverlay() → v1  
 Thread B: delegate.getOverlay() → v2  
 Thread B: updateCache(v2)  →  cache now holds v2 
 Thread A: updateCache(v1) -> v1 != v2, writes v1

